PR #71, opened 2026-08-04
Add first-class OpenCode documentation and support

ainetx:
## Summary by CodeRabbit

* **New Features**
  * Added opt-in support for OpenCode v1.18.4 as an additional agent host.
  * Generates OpenCode-native subagents while preserving user-managed files and configurations.
  * Added collision detection, stale-file reconciliation, preview reporting, and read-only integration inspection.
  * Reports partial results when existing files cannot be safely managed.

* **Documentation**
  * Updated architecture, product, setup, and integration guidance with OpenCode compatibility, limitations, and usage instructions.
  * Clarified that local CI runs exclude jobs requiring GitHub-hosted context or secrets.

* **Tests**
  * Added compatibility fixtures and comprehensive coverage for generation, ownership, collisions, and partial outcomes.


skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5515,302 @@ def _process_markdown_subagents(
...
+        raise ValueError(
+            f"Output path '{out_path}' escapes project root '{project_root}' — "
+            "path traversal is not allowed"
+        ) from exc
+
+    rel_path = _safe_relpath(canonical, root_resolved)
+    install_marker = project_root / ".opencode" / ".cf-studio-installed"
+    if canonical.exists():
+        existing = _read_existing_text_file(canonical)
+        if rel_path in unowned_outputs or not _is_opencode_owned_subagent(
```
ainetx: ### Unowned-collision exclusion is permanent and never re-validated once the sentinel is restored

**Severity:** Major

**Problem**
Ownership of a `.opencode/agents/cf-*.md` file is proven by two independent signals: the install sentinel `.opencode/.cf-studio-installed` existing, and the file's content containing `_GENERATED_MARKER`. If the sentinel happens to be missing on a given run (deleted by accident, wiped along with other `.opencode/` state, or simply never created yet on a machine that already has the generated `cf-*.md` files present), `_is_opencode_owned_subagent` returns False even though the file's content marker proves it actually is a Studio-generated file. In `_write_opencode_subagent_or_preserve` (and identically in `_reconcile_opencode_subagents`) that negative result causes the path to be recorded into `.opencode/.cf-studio-unowned-outputs.json`. On every subsequent run, the very first check is `if rel_path in unowned_outputs or not _is_opencode_owned_subagent(...)` — the `rel_path in unowned_outputs` branch short-circuits and is checked independently of the live ownership signal, so even after the sentinel is restored and the file content still carries the generated marker, the file is permanently treated as an unowned collision. `_forget_opencode_unowned_output` is only ever called from the success path when Studio recreates that exact path from scratch (file absent) — never just because ownership can now be proven again.

**How to reproduce**
1. `cf agents generate --opencode` runs successfully, creating `.opencode/.cf-studio-installed` and `.opencode/agents/cf-foo.md` (with `_GENERATED_MARKER`).
2. Something removes just the sentinel file (accidental `rm`, a clean of `.opencode/` state, a fresh checkout on another machine/CI where the sentinel isn't persisted but the generated `.md` file is).
3. `cf agents generate --opencode` runs again: `sentinel_existed_before_run` is False, so `_is_opencode_owned_subagent` returns False for `cf-foo.md` even though it still contains the generator marker; the path is recorded as unowned, and the sentinel is recreated at the end of this same run.
4. Every future run now has the sentinel present and the file still carries the generated marker, but the file remains permanently "preserved"/excluded because `rel_path in unowned_outputs` is checked first and never cleared.

**Expected behavior**
Ownership should be re-derivable: once both proof signals (sentinel present + content marker present) hold again, the file should be reclaimed (removed from the unowned-outputs record and managed normally), not permanently frozen by a stale exclusion recorded during a transient sentinel outage.

**Actual behavior**
The unowned-outputs record is checked unconditionally before the live ownership signals and is only ever cleared when Studio recreates the exact same path from a non-existent state — never when the live ownership proof becomes valid again for an existing file.

```text
run1: sentinel missing -> ownership check fails -> path recorded in unowned_outputs.json -> sentinel recreated
run2..N: sentinel present + marker present (real ownership provable)
         but check order is: `rel_path in unowned_outputs` (true, short-circuits) OR NOT is_owned
         -> file permanently treated as unowned collision, never rebuilt/cleaned again
```

**Impact**
A single run with a transiently missing sentinel permanently disables Studio's ability to rebuild, update, or clean up that generated subagent file — a one-way, irreversible loss of ownership for a genuinely Studio-owned file.

**Suggested correction**
In `_write_opencode_subagent_or_preserve` and `_reconcile_opencode_subagents`, only treat `rel_path in unowned_outputs` as authoritative when the live ownership check also fails; when the live signals do prove ownership (sentinel present and marker present) for a path already in `unowned_outputs`, call `_forget_opencode_unowned_output` and manage the file normally instead of unconditionally preserving it.

**How to verify**
Add a test that: creates a real Studio-owned file, deletes only the sentinel, runs generation once (path gets recorded as unowned), recreates the sentinel, runs generation again, and asserts the file is now recognized as owned and rebuilt rather than permanently preserved.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5515,302 @@ def _process_markdown_subagents(
...
+    _forget_opencode_unowned_output(
+        rel_path,
+        project_root,
+        unowned_outputs,
+        subagents_result,
+        dry_run,
+    )
+
+
+def _is_opencode_owned_subagent(
```
ainetx, line 5872: ### Ownership proof for OpenCode files relies on a marker shared with every other agent adapter, not a per-file/per-agent Studio marker

**Severity:** Major

**Problem**
`_is_opencode_owned_subagent()` proves a `.opencode/agents/cf-*.md` file is Studio-owned by checking only (1) the directory sentinel `.opencode/.cf-studio-installed` and (2) that the file's content contains `_GENERATED_MARKER` — the literal string `<!-- Generated by cf agents -- do not edit -->`. That exact same constant is also embedded by the Claude, Cursor, Windsurf and Copilot subagent templates elsewhere in this file. It is not scoped to OpenCode, to the specific agent name, or to the file's expected content — a single generic literal reused identically across every adapter. Once the OpenCode directory sentinel exists, any `.opencode/agents/cf-*.md` file that happens to contain this exact substring will be judged "owned" by `_write_opencode_subagent_or_preserve()` and `_reconcile_opencode_subagents()`, causing Studio to silently overwrite it on the next generate, or delete it outright if its name is no longer in `desired_names`.

**How to reproduce**
1. `cfs generate-agents --opencode` runs once, creating `.opencode/.cf-studio-installed` and `.opencode/agents/cf-codegen.md`.
2. A user later manually places their own notes/snippet at `.opencode/agents/cf-mynotes.md`, which happens to contain the literal string `<!-- Generated by cf agents -- do not edit -->` (e.g. copied from a `.claude/agents/*.md` file they were referencing, or from public Studio documentation/source).
3. On the next `cfs generate-agents --opencode` run, since `cf-mynotes` is not in `desired_names`, `_reconcile_opencode_subagents` treats it as a stale-but-owned file (sentinel present + marker substring present) and unlinks it — the user's file is deleted without any preservation check ever flagging a real collision.

**Expected behavior**
Ownership proof should be specific enough that only a file Studio itself generated for that exact name/agent can pass — e.g. embed the per-file agent name/hash in the marker line (OpenCode's own frontmatter already carries `name:`), or verify the full generated template shape/checksum rather than a substring match against a marker string reused verbatim by unrelated adapters.

**Actual behavior**
The "per-file Studio marker" is the same global literal used by all other agent templates, and ownership is decided by a simple substring containment check against that shared literal plus the directory sentinel — nothing ties the marker to OpenCode specifically or to the individual file's name/content.

```text
_GENERATED_MARKER (one global literal)
  used by: claude template, cursor template, windsurf template, copilot template, opencode template
  _is_opencode_owned_subagent(path) == sentinel.exists() AND marker in path.read_text()
  -> any file containing the marker substring is "OpenCode-owned", regardless of origin
```

**Impact**
A user-owned, non-Studio-generated file under `.opencode/agents/cf-*.md` can be silently overwritten or permanently deleted on a routine generate/rebuild run, violating the PR's core safety invariant that Studio must never touch a file it cannot specifically prove it produced.

**Suggested correction**
Make the per-file marker OpenCode/name-specific (e.g. `<!-- Generated by cf agents (opencode:{name}) -- do not edit -->` or include a content hash of the expected template) and verify it matches the specific file being checked, instead of testing for containment of a marker literal shared across every other agent adapter in the codebase.

**How to verify**
Add a test mirroring the existing collision/read-only test, but seed the collision file with the exact `_GENERATED_MARKER` string (as another adapter's template would produce) while giving it content that was never rendered by the OpenCode template; assert Studio still treats it as unowned/collision rather than adopting and later overwriting or deleting it.

tests/test_cli_agents_e2e.py
```diff
@@ -64,7 +64,77 @@ def _bootstrap_generator_project(root: Path) -> None:
...
+            collision = root / ".opencode" / "agents" / "cf-codegen.md"
+            collision.parent.mkdir(parents=True)
+            collision_content = "# User-owned OpenCode agent\n"
+            collision.write_text(collision_content, encoding="utf-8")
+
+            exit_code, stdout, _stderr = _run_main(
+                [
+                    "--json",
+                    "generate-agents",
+                    "--opencode",
```
ainetx, line 108: ### generate-agents write path never tested with canonical `--agent opencode` form

**Severity:** Minor

**Problem**
The PR's claimed behavior includes equivalence between `--agent opencode` and the `--opencode` convenience flag. The only e2e test that exercises `generate-agents` for OpenCode (`test_generate_agents_opencode_collision_is_nonfatal_and_generates_remaining_owned_outputs`) uses `--opencode` exclusively. `--agent opencode` is exercised only against the read-only `agents` status command, never against the write path (`generate-agents`). The two argparse branches (`--opencode` flag vs. `--agent` value) are simple enough that this is a low-risk gap, but it means the write-path equivalence claim is not directly verified end-to-end.

**How to reproduce**
Search `tests/test_cli_agents_e2e.py` for `--agent opencode` and `--opencode`; observe the write-path (`generate-agents`) test only uses `--opencode`.

**Expected behavior**
An e2e or coverage test invoking `generate-agents --agent opencode` (in addition to `--opencode`) and asserting equivalent output/state to the existing `--opencode` test.

**Actual behavior**
Only `--opencode` is exercised on the write path; `--agent opencode` write-path equivalence is untested.

```text
generate-agents --opencode        -> tested (e2e)
generate-agents --agent opencode  -> untested for generate-agents (only tested for read-only `agents`)
```

**Impact**
A future refactor of `_resolve_requested_agents` or the `--agent` argparse wiring that breaks `--agent opencode` specifically for `generate-agents` (while leaving `--opencode` and the `agents` command working) would not be caught by this test suite.

**Suggested correction**
Duplicate (or parametrize) the `generate-agents --opencode` e2e test to also run with `--agent opencode` and assert identical results.

**How to verify**
Add the test, confirm it passes today, and confirm it would fail if `--agent opencode` handling were broken independently of `--opencode`.

tests/test_subagent_registration.py
```diff
@@ -545,6 +588,238 @@ def test_openai_generates_per_agent_toml_files(self):
...
+                ],
+            )
+            unowned_record = json.loads(
+                (root / ".opencode" / ".cf-studio-unowned-outputs.json").read_text(
+                    encoding="utf-8"
+                )
+            )
+            self.assertIn(".opencode/agents/cf-codegen.md", unowned_record["paths"])
+
+    def test_opencode_preserves_marked_modified_collision_across_runs(self):
```
ainetx, line 742: ### No test of a legitimate (non-collision) second OpenCode run being recognized as owned

**Severity:** Major

**Problem**
Every OpenCode test that calls `_run_agents` twice (`test_opencode_preserves_marked_modified_collision_across_runs` and `test_opencode_preserves_marked_modified_stale_agent_across_runs`) seeds the second run with a file that is already unowned/collision on the first run (no real `.cf-studio-installed` sentinel exists before the first run, so `_is_opencode_owned_subagent` returns False from the start). Both tests therefore only prove that an already-unowned path stays unowned/excluded across reruns. There is no test — unit or e2e — that runs OpenCode generation twice starting from a clean project (as the existing `test_idempotent_second_run_no_updates` does for `claude`) to prove the positive ownership-detection branch: a file legitimately created by Studio on run 1 (real sentinel + real `_GENERATED_MARKER`) is correctly recognized as owned on run 2 and rebuilt/left unchanged with `status: PASS`, rather than being misclassified as an unproven collision.

**How to reproduce**
Read the two multi-run OpenCode tests in `tests/test_subagent_registration.py`; observe both start from a pre-seeded, already-unowned file rather than a clean project with a real prior Studio-generated output.

**Expected behavior**
A test that calls generation twice on a clean tree (mirroring `test_idempotent_second_run_no_updates`), asserting the second run reports `status == "PASS"`, no `preserved`/collision outputs, and that changing a registry description causes the marker-owned file to be rebuilt with new content on a later run.

**Actual behavior**
No such test exists; only the negative (already-unowned-stays-unowned) branch is covered for multi-run behavior.

```text
covered:    unowned -> generate -> still unowned (collision preserved)
uncovered:  owned   -> generate -> still owned (rebuilt / PASS, not misclassified as collision)
```

**Impact**
A regression that flips the ownership check (e.g. inverts the `sentinel_existed` boolean, or a bug that makes `_is_opencode_owned_subagent` always return False for existing files) would make every legitimate second `generate-agents --opencode` run treat its own previously-generated files as unowned collisions — permanently disabling regeneration and silently degrading every run to PARTIAL — yet no test in this PR would catch it.

**Suggested correction**
Add an OpenCode idempotency test that runs generation twice from a clean project and asserts the second run is fully owned (`status: PASS`, no partial/collision), exercising the true-ownership branch of `_is_opencode_owned_subagent` and the overwrite path of `_write_opencode_subagent_or_preserve`.

**How to verify**
Run the added test against the current implementation and confirm it passes; then mutate `_is_opencode_owned_subagent`'s sentinel/content check (e.g. always return False) and confirm the new test fails while the existing suite would not have caught it.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5797,16 +6313,19 @@ def _build_agents_arg_parser(
...
     parser = argparse.ArgumentParser(prog=prog, description=description)
+    # @cpt-begin:cpt-studio-flow-agent-integration-generate-opencode:p1:inst-opencode-explicit-select
     agent_group = parser.add_mutually_exclusive_group(required=False)
     agent_group.add_argument(
         "--agent",
         default=None,
         help=(
-            "Agent/IDE key (e.g., windsurf, cursor, claude, copilot, openai). "
+            "Agent/IDE key (e.g., windsurf, cursor, claude, copilot, openai, opencode). "
             "Omit to target all supported agents."
```
ainetx: ### CLI help text doesn't disclose that OpenCode is excluded from the default agent set

**Severity:** Major

**Problem**
The `--agent` flag's help string lists `opencode` as one of its example values in the same sentence that says omitting the flag targets "all supported agents" — but omitting `--agent` actually excludes OpenCode from the default set (`_DEFAULT_RECOGNIZED_AGENTS`). This opt-in/exclusion rule is documented only in an internal code comment, never in the CLI's own `--help` output.

**How to reproduce**
1. Run `cfs generate-agents --help` (or `cfs agents --help`).
2. Read the `--agent` help text: `"Agent/IDE key (e.g., windsurf, cursor, claude, copilot, openai, opencode). Omit to target all supported agents."`
3. Run `cfs generate-agents` with no `--agent` flag and observe that no `.opencode/` files are touched.

**Expected behavior**
The help text should state that OpenCode is opt-in only and excluded from the default set, e.g. "Omit to target the default agent set (excludes OpenCode — use --agent opencode or --opencode to opt in)."

**Actual behavior**
The help text implies omitting `--agent` includes every listed agent, including `opencode`, which is incorrect.

```text
--agent opencode  -> included
(no --agent flag) -> _resolve_requested_agents() -> _DEFAULT_RECOGNIZED_AGENTS (opencode excluded)
                      but --help says "Omit to target all supported agents"
```

**Impact**
Users relying only on `--help` may believe OpenCode is generated by default, or have no idea from the CLI itself that OpenCode requires explicit opt-in.

**Suggested correction**
Update the `--agent` help string to state the default-set exclusion explicitly, and/or add a short parser `epilog` summarizing the OpenCode opt-in model.

**How to verify**
Re-run `--help` and confirm the printed text states the default-exclusion behavior, not only a source comment.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -6237,6 +6826,17 @@ def _add_legacy_preview_counts(preview: Dict[str, Any], legacy_preview: Dict[str
...
         preview["delete"] += len(section_result.get("deleted", []))
+    if legacy_preview.get("agent") != "opencode":
+        return
+    preview["preserved"] += sum(
+        1
+        for output in legacy_preview.get("subagents", {}).get("outputs", [])
+        if (
+            isinstance(output, dict)
+            and output.get("action") == "preserved"
+            and output.get("reason") == "opencode_unowned_collision"
```
ainetx: ### Preview "preserved" count omits the stale-unowned reconciliation reason, understating the preview summary

**Severity:** Major

**Problem**
The preview/dry-run "preserved" counter only counts OpenCode outputs whose `reason == "opencode_unowned_collision"`. Stale-file reconciliation preserves orphaned unowned files with a different reason, `"opencode_stale_unowned"`, which this filter silently excludes from the count.

**How to reproduce**
1. Create a stale `.opencode/agents/cf-*.md` file that is no longer referenced by any kit agent and is not Studio-owned (no marker/sentinel match).
2. Run `cfs generate-agents --opencode --dry-run`.
3. Observe the preview's `preserved` count.

**Expected behavior**
The preview's `preserved` counter reflects every preserved output, including stale-unowned preservations, so a dry run doesn't misreport "no changes needed" when a preserved stale file actually exists.

**Actual behavior**
Only `opencode_unowned_collision` outputs are counted; `opencode_stale_unowned` outputs are invisible in the aggregate preview count (though still present in the raw per-agent outputs list).

```text
preview["preserved"] += count(outputs where reason == "opencode_unowned_collision")
                                                    ^ misses "opencode_stale_unowned"
```

**Impact**
A dry run whose only pending action is a stale, preserved collision can misreport the run as fully in sync, hiding a real preserved-file condition from the user.

**Suggested correction**
Broaden the reason filter to include `"opencode_stale_unowned"` (and arguably `"opencode_stale_delete_failed"`), or drop the reason filter and count every `action == "preserved"` output in the OpenCode subagents section.

**How to verify**
Add a test asserting `preview["preserved"] >= 1` when only an `opencode_stale_unowned` preserved output exists.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+) -> None:
+    """Write only marker-owned OpenCode agent files; preserve collisions."""
+    root_resolved = project_root.resolve()
+    canonical = out_path.resolve()
+    try:
+        canonical.relative_to(root_resolved)
+    except ValueError as exc:
+        raise ValueError(
+            f"Output path '{out_path}' escapes project root '{project_root}' — "
+            "path traversal is not allowed"
```
ainetx, line 5825: ### Path-traversal detection raises an uncaught ValueError that crashes the whole generate-agents command

**Severity:** Critical

**Problem**
When an OpenCode subagent's resolved output path escapes the project root, the write helper raises a bare `ValueError`. No caller in the chain up to `cmd_generate_agents` wraps this call in a `try/except`, so the exception propagates uncaught and aborts the entire CLI invocation — including every other agent being processed in the same run — instead of yielding a per-agent error result.

**How to reproduce**
1. Arrange for a kit agent's OpenCode output path to resolve outside `project_root` (e.g. via a misconfigured/attacker-influenced kit `agents.toml` entry or output directory).
2. Run `cfs generate-agents --opencode` (or a default run that includes OpenCode processing).
3. Observe the process crashes with an unhandled traceback rather than reporting a structured per-agent error.

**Expected behavior**
A path-traversal detection should be isolated as a per-agent error appended to the command's structured result (mirroring how this same module already handles the identical traversal case for other, non-OpenCode output-path resolution call sites), letting the rest of the batch complete and the command exit with a controlled non-zero/PARTIAL status.

**Actual behavior**
The `ValueError` propagates uncaught through the whole call chain and crashes the CLI process for every agent in the run, not just the offending one.

```text
_write_opencode_subagent_or_preserve()
  -> raise ValueError("... path traversal is not allowed")
_process_opencode_subagents() [no try/except]
  -> _process_subagents() [no try/except]
    -> _process_single_agent() [no try/except]
      -> cmd_generate_agents() [no try/except]
         => uncaught crash, whole command aborts
```

**Impact**
A single misconfigured or malicious kit agent entry can take down the entire `generate-agents` invocation for every target agent, not just OpenCode.

**Suggested correction**
Wrap the call to the write helper in the OpenCode subagent-processing loop in a `try/except ValueError` that appends the error to the command's structured errors list and continues, matching the existing pattern used at the other traversal-detection call sites in this file.

**How to verify**
Add a test that forces the traversal branch and asserts the command returns a structured error result instead of raising.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+    dry_run: bool,
+    sentinel_existed_before_run: bool,
+) -> None:
+    """Generate only cf-namespaced, marker-owned OpenCode subagents."""
+    output_dir = (project_root / ".opencode" / "agents").resolve()
+    unowned_outputs = _load_opencode_unowned_outputs(project_root)
+    desired_names: Set[str] = set()
+    for kit_agent in kit_agents:
+        name = kit_agent["name"]
+        if not name.startswith("cf-"):
```
ainetx, line 6004: ### Kit agents without a "cf-" prefix are silently dropped from OpenCode generation with zero diagnostic output

**Severity:** Major

**Problem**
The OpenCode subagent loop filters out any kit agent whose name doesn't start with `cf-` via a plain `continue`, with no warning, error, or output-list entry. Kit agent names are not required to start with `cf-` by the agent-entry validator (it only rejects unsafe path-like names), so a legitimately-named third-party/kit-authored agent (e.g. `my-helper`) silently never appears under `.opencode/agents/` — while the very next branch, for a missing target path, does record a diagnostic via the missing-target-marking helper.

**How to reproduce**
1. Register a kit whose `agents.toml` defines an agent entry with a name that doesn't start with `cf-` (e.g. `my-helper`) and a valid prompt target.
2. Run `cfs generate-agents --opencode`.
3. Compare against `cfs generate-agents --claude` (or another non-OpenCode target) for the same kit.

**Expected behavior**
Either OpenCode support is documented as requiring the `cf-` prefix and the run surfaces a warning/skip entry explaining the omission, or the agent is generated like it is for every other target.

**Actual behavior**
The agent is silently absent from `.opencode/agents/` with no trace in the command's outputs or errors, while it is correctly generated for every other agent target.

```text
for kit_agent in kit_agents:
    if not name.startswith("cf-"):
        continue                      # <- no diagnostic recorded here
    if not target_agent_rel:
        _mark_missing_subagent_target(name, subagents_result)  # <- diagnostic recorded here
```

**Impact**
Silent, target-specific data loss: a kit author's custom (non-`cf-`) agent quietly never gets OpenCode support, with no signal pointing at the cause.

**Suggested correction**
Record a skip entry (or call the equivalent of a warning function) when filtering out a non-`cf-` name, matching the observability level of the missing-target-path case.

**How to verify**
Add a test with a kit agent named without the `cf-` prefix; assert the OpenCode result contains a visible skip/warning entry referencing that agent name, not silence.

tests/fixtures/opencode/v1.18.4/compatibility.json
```diff
@@ -0,0 +1,32 @@
...
+  "skills": {
+    "project_agent_compatible_path": ".agents/skills/{name}/SKILL.md",
+    "adapter_behavior": "consume_without_generating_or_overwriting"
+  },
+  "agents": {
+    "project_path": ".opencode/agents/{name}.md",
+    "name_prefix": "cf-",
+    "required_frontmatter": [
+      "description",
+      "mode: subagent"
```
ainetx, line 30: ### compatibility.json fixture is validated circularly against the generator's own output, not against OpenCode's real schema

**Severity:** Major

**Problem**
`compatibility.json`'s `required_frontmatter` lists `["description", "mode: subagent"]`, but no production code loads or validates against this fixture — it is only read by a test helper that checks each listed string is a substring of text produced by the generator template, which itself unconditionally emits the literal line `mode: subagent`. The check is therefore circular: it validates the generator's hard-coded output against a description of that same hard-coded output, not against an independently-modeled schema. No failing/edge-case input (e.g. missing `description`, a different `mode` value) is ever exercised through the real command path.

**How to reproduce**
1. Read `tests/fixtures/opencode/v1.18.4/compatibility.json`'s `required_frontmatter` field.
2. Read the test that consumes it: it does `for required_field in compatibility["agents"]["required_frontmatter"]: self.assertIn(required_field, content)` against text produced by the same generator template being tested.
3. Note there is no OpenCode-side loader/validator in production code that parses this fixture at all.

**Expected behavior**
The fixture should model the field name / type / required-ness of OpenCode's actual documented frontmatter schema (per OpenCode's own docs, only `description` is required; `mode` is optional), and be exercised with both compliant and deliberately non-compliant inputs to prove the check can actually fail.

**Actual behavior**
The fixture mirrors the generator's own fixed output and is checked only against that same output — it cannot catch a real schema drift or a generator regression that changes the emitted frontmatter shape, because the assertion would just as easily "pass" against whatever the generator happens to emit.

**Impact**
No real regression protection against OpenCode schema changes or against Studio emitting non-compliant frontmatter; the test would not fail even if the fixture and the generator diverged from the true external contract together.

**Suggested correction**
Rebuild the fixture as a proper field table (required vs. optional, with types) sourced independently from OpenCode's documented schema, and add a test that feeds a deliberately non-conforming rendered file through a schema-aware assertion to prove failing inputs are distinguishable.

**How to verify**
Confirm the rebuilt test fails when the generator's hard-coded `mode: subagent` line is removed or changed, proving the check is tied to the schema rather than to the generator's own fixed output.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+            _warn_agents(message)
+            _remember_opencode_unowned_output(
+                rel_path,
+                project_root,
+                unowned_outputs,
+                subagents_result,
+                dry_run,
+            )
+            return
+    _write_or_skip(canonical, content, subagents_result, project_root, dry_run)
```
ainetx: ### File-write failures are not isolated per item, so one OSError aborts the entire generate-agents batch

**Severity:** Major

**Problem**
The shared write helper used by the OpenCode subagent-generation loop (and by other agent targets) calls `write_text(...)` directly with no `try/except OSError` around it — only the read side is guarded. A filesystem failure (permission denied, disk full, path removed mid-run) on a single file's write propagates uncaught out of the whole per-agent loop, aborting the entire `generate-agents` run instead of being isolated as a per-item error.

**How to reproduce**
1. Make the OpenCode agents output directory (or one target file within it) unwritable.
2. Run `cfs generate-agents --opencode` (or a default run covering multiple agents, of which OpenCode is one).
3. Observe the command crashes with an unhandled traceback rather than reporting a per-item error and continuing with the remaining files/agents.

**Expected behavior**
A write failure for one file should be recorded as a per-item error result, and processing should continue for the remaining kit agents and target agent types.

**Actual behavior**
The `OSError` from `write_text` is unhandled and terminates the whole command, discarding whatever results had already been collected for other items in the same run.

```text
canonical.write_text(content, encoding="utf-8")   # no try/except
```

**Impact**
A single transient or environmental write failure (disk full, permission drift, antivirus lock) turns what should be a partial-success report into a hard crash, losing all already-collected results for the rest of the batch.

**Suggested correction**
Wrap the write calls in the shared write helper in `try/except OSError`, append a per-item error to the result's errors/outputs, and continue the loop rather than propagating.

**How to verify**
Add a test that makes one target file's write raise `OSError` (e.g. via monkeypatching) among several agents in the batch, and assert the command still completes with a partial/error status for that one item while the rest succeed.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -7220,6 +7828,19 @@ def _human_agents_list(  # pylint: disable=too-many-branches
...
+            generated = r.get("generated", [])
+            collisions = r.get("collision", [])
+            if generated:
+                any_files = True
+                ui.step(f"opencode: {len(generated)} native agent file(s) installed")
+                for path in generated:
+                    ui.substep(f"  {path}")
+            else:
+                ui.step("opencode: no native agent files")
+            for collision in collisions:
```
ainetx: ### `agents --opencode` human output drops the collision reason and never shows install-marker state

**Severity:** Minor

**Problem**
The read-only `agents` command's human-readable renderer prints a collision's file path, but never its `reason` field, and never surfaces whether the OpenCode install-marker sentinel is present — both of which are available in the JSON output for the same command.

**How to reproduce**
1. Run `cfs agents --opencode` (or `--agent opencode`) in a project with an existing collision.
2. Compare the human-readable output to the `--json` output for the same invocation.

**Expected behavior**
The human-readable text should convey the same information as the JSON payload: the collision's reason (why it wasn't touched) and whether the install marker is present.

**Actual behavior**
Only the collision's path is printed; the `reason` and `sentinel` fields are silently omitted from human-mode output.

```text
for collision in collisions:
    ui.warn(f"  collision: {collision.get('path', '?')}")   # reason never read
# sentinel state never rendered anywhere in this function
```

**Impact**
Lower severity than a total information loss (the collision's existence and path are still visible), but a user on the default human-readable path cannot tell *why* a file is flagged as a collision or whether the install marker exists without falling back to `--json`.

**Suggested correction**
Print the collision reason alongside the path (e.g. `collision: <path> (<reason>)`), and add a line reporting the install-marker/sentinel state.

**How to verify**
Add a human-mode (non-JSON) assertion capturing stdout for `agents --opencode` with a collision and checking both the reason and the sentinel state appear in the text.

tests/test_cli_agents_e2e.py
```diff
@@ -64,7 +64,120 @@ def _bootstrap_generator_project(root: Path) -> None:
...
+mode = "readonly"
+""",
+        encoding="utf-8",
+    )
+    (agents / "cf-codegen.md").write_text("Code generation instructions.\n", encoding="utf-8")
+    (agents / "cf-pr-review.md").write_text("PR review instructions.\n", encoding="utf-8")
+
+
 class TestCLIAgentsE2E(unittest.TestCase):
+    def test_generate_agents_opencode_collision_is_nonfatal_and_generates_remaining_owned_outputs(self):
```
ainetx, line 90: ### No CLI-level end-to-end test covers the plain success path for `generate-agents --opencode`

**Severity:** Minor

**Problem**
Every OpenCode `generate-agents` end-to-end test that drives the real CLI entry point (argv → `main()` → exit code) sets up a pre-existing collision and asserts a `PARTIAL` status. The plain success path — non-empty kit-agent scope, zero collisions, brand-new files written — is only covered by a test that calls the internal processing function directly, bypassing argument parsing, `main()`, and JSON/exit-code emission entirely.

**How to reproduce**
1. Search the CLI end-to-end test file for OpenCode `generate-agents` tests: both existing ones set up a collision file before running.
2. Search the unit test file for the success-path OpenCode generation test: it calls the internal per-agent processing function directly, not through the CLI entry point.

**Expected behavior**
At least one CLI-level (real argv/exit-code) test should cover `generate-agents --opencode` with a clean scope and no collisions, asserting the success exit code and status through the same path used in production.

**Actual behavior**
The success/no-collision combination for the real CLI entry point is untested; only the collision/partial combination is verified end-to-end.

**Impact**
A regression in argument parsing, JSON serialization, or exit-code computation specific to the plain-success OpenCode combination could slip through even though the underlying generation logic itself is unit-tested.

**Suggested correction**
Add an end-to-end test mirroring the existing collision test but without a pre-existing collision file, asserting a successful exit code and status through the real CLI entry point.

**How to verify**
Run the new test; it should pass against the current implementation (this is a coverage gap, not a functional bug) and would catch a future regression in the untested combination.

tests/test_cmd_generate_agents_v2.py
```diff
@@ -810,9 +810,9 @@ def test_confirmation_counts_include_legacy_workflows(self):
         confirm_args = []
         original_confirm = None
 
-        def _spy_confirm(args, preview_create, preview_update, preview_delete=0):
+        def _spy_confirm(args, preview_create, preview_update, preview_delete=0, **_kwargs):
```
ainetx, line 819: ### No test in this file verifies OpenCode bypasses manifest-v2 translation

**Severity:** Minor

**Problem**
This file's only change in the PR is a one-line signature compatibility fix to tolerate a new keyword argument on the confirmation spy helper. No test verifying that OpenCode selection bypasses manifest-v2 translation was added here — that specific coverage exists only in a different test file.

**How to reproduce**
Search this file for any reference to "opencode": only the incidental signature-compatibility change is present.

**Expected behavior**
If this file is the expected home for v2-pipeline behavioral tests, it should contain (or reference) the assertion that OpenCode is excluded from manifest-v2 agent/skill generation calls.

**Actual behavior**
The OpenCode-bypasses-v2-translation assertion lives entirely in a different test module, not in this one.

**Impact**
Low — the behavior is genuinely covered elsewhere, so this is an organizational/discoverability gap rather than a functional gap. A reviewer or future maintainer scanning this file specifically for v2-pipeline/OpenCode interaction coverage will find none.

**Suggested correction**
Either move or duplicate the OpenCode-bypass assertion into this file to match its apparent scope, or note explicitly (e.g. in a comment) that this coverage is intentionally maintained elsewhere.

**How to verify**
Confirm the OpenCode-bypass test exists and passes in its current location, and that this file's scope/naming convention is documented if the split is intentional.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+        # rather than trusting the recorded exclusion unconditionally.
+        if is_owned and rel_path in unowned_outputs:
+            _forget_opencode_unowned_output(
+                rel_path, project_root, unowned_outputs, subagents_result, dry_run
+            )
+        if not is_owned:
+            message = f"preserved unowned OpenCode agent collision: {rel_path}"
+            subagents_result["partial"] = True
+            subagents_result["outputs"].append(
+                {"path": rel_path, "action": "preserved", "reason": "opencode_unowned_collision"}
```
ainetx: ### generate-agents human output never names which OpenCode file was preserved or why

**Severity:** Major

**Problem**
When `generate-agents --opencode` finishes with a preserved collision, the human-readable (non-JSON) output only prints the generic category label for the partial reason (e.g. "preserved_outputs") per agent — never the actual file path or the specific reason. The per-file detail (`path`, `reason`) already exists in the collected result but is dropped before rendering; only `--json` exposes it.

**How to reproduce**
1. Run `cfs generate-agents --opencode --yes` (non-JSON) in a project with a pre-existing collision.
2. Compare the console output to the `--json` output for the same run.

**Expected behavior**
The human-readable report should name the specific file(s) preserved due to a collision and why, mirroring what `--json` already exposes.

**Actual behavior**
The footer only prints the joined category names for a partial result, never the underlying per-file path/reason detail that was already collected:

```text
ui.warn(f"  partial reason for {agent_name}: {category_text}")
# category_text is just the category name (e.g. "preserved_outputs"),
# never the actual preserved-file path or reason string
```

**Impact**
Operators using the default human-readable CLI output see "something is partial" with no actionable detail about which file collided or why generation left it untouched — undermining the usability of the collision-preservation feature outside JSON/automation contexts.

**Suggested correction**
Expand each preserved-output entry in the footer (or add an OpenCode-aware branch to the file-actions renderer) to print the specific path and reason, not just the category label.

**How to verify**
Add a non-JSON e2e test for `generate-agents --opencode --yes` with a pre-existing collision and assert the captured stdout contains the collided file's path and reason text, not just a category name.

architecture/DECOMPOSITION.md
```diff
@@ -286,7 +286,7 @@ Studio DESIGN is decomposed into features organized around architectural layers
 
 - **Design Constraints Covered**:
 
-  None
+  - `p1` - `cpt-studio-constraint-opencode-owned-output`
```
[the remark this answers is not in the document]
ainetx: Fixed in efb8116 — Feature 2.5's Scope, Requirements Covered, and Data sections now include OpenCode (`cpt-studio-fr-core-opencode`, `.opencode/*` paths).

architecture/DESIGN.md
```diff
@@ -763,14 +789,15 @@ Bridges the gap between Studio's unified skill system and the diverse file forma
...
 - Normalize generated agent output names so new files use a `cf-` prefix or a dedicated Studio-owned path; migrate legacy non-prefixed generated storytelling files only when ownership checks prove they are generated, and preserve user-edited files
-- Support `--agent` flag for single-agent regeneration
+- Support `--agent` flag for single-agent regeneration; OpenCode is selected only by canonical `--agent opencode` or convenience `--opencode`, never by the initial default target set
 
 ##### Responsibility boundaries
 
-Does NOT maintain agent-specific state. Does NOT define SKILL extension content — collects from kit files. Does NOT persist agent selection in config.
+Does NOT maintain agent-specific state. Does NOT define SKILL extension content — collects from kit files. Does NOT persist agent selection in config. In the OpenCode rollout it does NOT manage root `AGENTS.md`, user `opencode.json`, OpenCode installation or execution, runtime model/provider defaults, or user-owned `.opencode` content.
```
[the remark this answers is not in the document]
ainetx: Fixed in efb8116 — DESIGN.md's OpenAI entry now reads `.agents/skills/` + `.codex/agents/`.

tests/test_cmd_generate_agents_v2.py
```diff
@@ -810,7 +810,7 @@ def test_confirmation_counts_include_legacy_workflows(self):
...
             confirm_args.append((preview_create, preview_update, preview_delete))
             return True  # proceed
```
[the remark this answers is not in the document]
ainetx: Fixed in efb8116 — `_spy_confirm` now returns `"PROCEED"`; the test reaches `_run_v2_pipeline` and was strengthened with an outcome assertion (return code + written proxy file) so a pipeline-skip regression is caught.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in efb8116 — live ownership is now re-checked before consulting `unowned_outputs`; when both proof signals hold again, `_forget_opencode_unowned_output` is called and the file is reclaimed instead of staying permanently excluded. Covered by `test_opencode_reclaims_ownership_after_sentinel_restored`.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in efb8116 (name check) and tightened in f7c6c7f (exact-line match instead of substring, per your follow-up finding below) — ownership now requires the file's own `name:` frontmatter line to match its filename, not just the shared `_GENERATED_MARKER` literal. Covered by `test_opencode_generic_shared_marker_alone_does_not_prove_ownership` and `test_owned_subagent_name_match_is_exact_not_substring`.

tests/test_cli_agents_e2e.py
ainetx: Fixed in efb8116 — added `test_generate_agents_agent_opencode_form_matches_convenience_flag`, exercising `generate-agents --agent opencode` on the write path with the same assertions as the existing `--opencode` test.

tests/test_subagent_registration.py
ainetx: Fixed in efb8116 — added `test_opencode_idempotent_second_run_recognizes_real_ownership`: a clean two-run project now asserts the second run is `status: PASS` with no preserved/collision outputs, exercising the true-ownership branch.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+        paths.clear()
+        paths.update(current)
+        return True
+    finally:
+        if lock_fd is not None:
+            try:
+                _fcntl.flock(lock_fd, _fcntl.LOCK_UN)
+            except OSError:
+                pass
+            os.close(lock_fd)
```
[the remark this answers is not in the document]
ainetx: Fixed in f7c6c7f — the Windows fallback now uses an O_CREAT|O_EXCL sentinel lock with a bounded retry and stale-lock recovery, matching the POSIX fcntl path's serialization instead of proceeding unlocked. Covered by `test_save_falls_back_to_sentinel_lock_when_fcntl_unavailable` and `test_save_windows_fallback_refuses_to_proceed_when_lock_held`.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+    """Persist OpenCode ownership exclusions without replacing user state.
+
+    Serializes concurrent writers with an advisory lock and re-reads the
+    on-disk record under that lock, applying only the caller's single
+    `add`/`remove` delta rather than blindly overwriting with `paths` — this
+    prevents one process's write from silently discarding another
+    concurrent process's exclusion (lost-update).
+    """
+    record_path = _opencode_unowned_outputs_path(project_root)
+    lock_path = record_path.with_name(record_path.name + ".lock")
```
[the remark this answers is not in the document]
ainetx: Fixed in f7c6c7f — registered `.opencode/.cf-studio-unowned-outputs.json.lock` as a managed output in `_collect_marker_and_configured_outputs` and added it to `.gitignore`.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5527,399 @@ def _process_markdown_subagents(
...
+    Requiring the file's own frontmatter `name:` line to match the filename
+    this specific path is expected to carry ties the proof to this file's
+    identity, not just to a generic substring.
+    """
+    return bool(
+        (install_marker.is_file() if sentinel_existed is None else sentinel_existed)
+        and content is not None
+        and _GENERATED_MARKER in content
+        and f"name: {expected_name}" in content
+    )
```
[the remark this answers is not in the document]
ainetx: Fixed in f7c6c7f — applied your suggested diff (exact-line match via `content.splitlines()` instead of substring containment). Covered by `test_owned_subagent_name_match_is_exact_not_substring`.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — `--agent`'s help text now states omitting it targets the default set, which excludes `opencode`.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — the preview now counts every preserved OpenCode output regardless of reason, not only `opencode_unowned_collision`.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — the path-traversal check in `_write_opencode_subagent_or_preserve` now appends a per-item error and returns instead of raising; verified none of the other 6 call sites of `_write_or_skip` in this file wrap it either (this bug wasn't OpenCode-specific, but the fix here is scoped to the OpenCode call path per this PR).

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — a non-`cf-`-prefixed kit agent now records a visible `skipped`/`opencode_non_cf_prefix` output entry with a warning instead of a silent `continue`.

tests/fixtures/opencode/v1.18.4/compatibility.json
ainetx: Fixed in bbfac4c — split the fixture into `required_frontmatter` (OpenCode's own requirement: `description` only) vs. `studio_emitted_frontmatter` (this generator's own choice: `mode: subagent`), and added `test_opencode_required_frontmatter_check_is_schema_aware_not_circular` proving the check distinguishes compliant from non-compliant input.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — the same wrap that isolates the traversal ValueError (see the reply above on the sibling finding) also catches `OSError` from the `_write_or_skip` call and records it as a per-item error instead of propagating.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — human-mode output now prints the collision reason alongside the path, plus a sentinel-present line, matching what `--json` already exposes.

tests/test_cli_agents_e2e.py
ainetx: Fixed in bbfac4c — added `test_generate_agents_opencode_clean_scope_succeeds_through_cli_entry_point`, covering the zero-collision success path through the real CLI entry point.

tests/test_cmd_generate_agents_v2.py
ainetx: Addressed in bbfac4c — added a docstring note pointing to `test_agents_coverage.py::test_v2_pipeline_bypasses_opencode_manifest_translation` as the intentional home for that coverage, per your suggested lighter option.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in bbfac4c — the partial-reason footer now also prints each preserved output's path/reason and each collected error message, not just the category label.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5725,6 +6317,63 @@ def _process_single_agent(
...
+    subagents_result = _process_subagents(
+        "opencode",
+        project_root,
+        studio_root,
+        cfg,
+        cfg_path,
+        dry_run,
+        opencode_sentinel_existed_before_run=sentinel_existed_before_run,
+    )
+    if not dry_run and not subagents_result.get("ownership_recording_failed"):
```
ainetx: ### First-run recording failure leaves install marker unwritten, misclassifying Studio files as unowned on the next run

**Severity:** Major

**Problem**
`_write_install_marker` is gated on `not subagents_result.get("ownership_recording_failed")` (line 6359). When `_save_opencode_unowned_outputs` fails during the first-ever Studio OpenCode run — for example, on a disk-full or temp-file creation error — it sets `ownership_recording_failed = True`. The marker write is skipped. Files may have been successfully generated in that run, but `.opencode/.cf-studio-installed` is never created. On the next run, `sentinel_existed_before_run = install_marker.is_file()` evaluates to `False`. Every call to `_is_opencode_owned_subagent(content, marker, stem, sentinel_existed=False)` short-circuits to `False` regardless of the file's marker content, so all Studio-generated files from run 1 are treated as ownership-unproven collisions and routed to `_preserve_opencode_collision`.

**How to reproduce**
1. Start with a project that has no `.opencode/.cf-studio-installed` marker (first-ever run).
2. Patch `tempfile.mkstemp` to raise `OSError` so `_save_opencode_unowned_outputs` fails after agents have been written to disk.
3. Observe: agents are present in `.opencode/agents/`, install marker is absent, `partial=True`.
4. Remove the patch and run `cfs generate-agents --opencode` again with normal I/O.
5. Observe: all previously-generated `cf-*.md` files are preserved as collisions rather than recognized as owned and regenerated.

**Expected behavior**
Studio should recognize previously-generated files as owned on run 2 even if the collision-record write failed on run 1. The install marker should be written independently of whether the unowned-outputs record-write succeeded.

**Actual behavior**
All Studio-generated files from run 1 appear as unowned collisions on run 2 (and every subsequent run) until a fully-successful run finally writes the marker. The result reports `partial=True` and preserves every Studio-generated file instead of overwriting it.

```text
run 1: agents written ──► record-write fails ──► ownership_recording_failed=True
                                                          │
                                                          ▼
                                               marker NOT written (gated on flag)
                                                          │
run 2: sentinel_existed=False ◄────────────────────────────
         │
         ▼
   _is_opencode_owned_subagent returns False for ALL files
         │
         ▼
   _preserve_opencode_collision() for every file ──► partial=True, stuck state
```

**Impact**
Any first-run I/O failure creates a persistent state where Studio cannot update its own generated files. The core regeneration scenario fails silently on every subsequent run until the underlying I/O condition resolves and a full clean run completes.

**Suggested correction**
Decouple the install-marker write from the unowned-record write outcome. Write the marker even when `ownership_recording_failed=True`, since the marker only needs to prove that Studio ran and generated files — not that the collision record is complete. The collision record's incompleteness is already surfaced via `partial=True` and the `ownership_recording_failed` flag.

**How to verify**
1. Patch `tempfile.mkstemp` to raise `OSError` during a first-run `cfs generate-agents --opencode`.
2. Confirm agents written but marker absent.
3. Remove the patch and run again.
4. Assert the result has `partial=False`, no `preserved` entries for Studio-generated `cf-*.md` files, and files are regenerated cleanly.

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5473,13 +5534,429 @@ def _process_markdown_subagents(
...
+            # process rather than blocking forever or proceeding unlocked.
+            deadline = time.monotonic() + 10.0
+            while lock_fd is None:
+                try:
+                    lock_fd = os.open(str(lock_path), os.O_CREAT | os.O_EXCL | os.O_WRONLY)
+                except FileExistsError:
+                    if time.monotonic() > deadline:
+                        try:
+                            if time.time() - lock_path.stat().st_mtime > 10.0:
+                                lock_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
ainetx: Fixed in 946d289 — removed the age-based auto-clear entirely. On timeout the Windows fallback now refuses to proceed and warns, leaving the sentinel in place, instead of deleting it (which could steal an active-but-slow writer's lock). Covered by `test_sentinel_lock_does_not_delete_active_writers_lock_on_timeout`.

skills/studio/scripts/studio/commands/agents.py
ainetx: Fixed in 946d289 — the install marker write is no longer gated on `ownership_recording_failed`; it's written whenever `not dry_run`, independent of whether the collision-record write succeeded. Verified the fix by temporarily reverting it locally and confirming the new regression test (`test_opencode_marker_written_despite_first_run_record_write_failure`) fails without the fix and passes with it.

ainetx:
Fixed in 854f97e — addressed two new SonarCloud findings surfaced on the previous commit (946d289):

- **python:S1192** (`.opencode` literal duplicated 3×, line 5572): introduced `_OPENCODE_AGENTS_DIRNAME = ".opencode"` and used it at all three call sites.
- **python:S3776** (`_human_agents_list` cognitive complexity 29 > 15, line 7964): extracted the OpenCode-row and per-agent-row rendering into `_human_agents_list_render_opencode`/`_human_agents_list_render_row`.

Also found and fixed an unrelated latent bug while re-running the CI-equivalent verification gate locally: `generate_report()`'s `_rel()` helper in `utils/coverage.py` compared an *unresolved* `project_root` against a *resolved* file path, so `is_relative_to()` was always `False` whenever the project root sits behind a symlink (e.g. macOS `/var` → `/private/var`), silently falling back to absolute paths as "files" dict keys instead of project-relative ones. Fixed by resolving both sides, matching `spec_coverage.py`'s own `_rel_path()` convention. Covered by `test_generate_report_files_key_survives_a_symlinked_project_root` (confirmed red without the fix, green with it).

Full verification gate (pytest -n6 + coverage ≥90%, pylint, cfs validate, spec-coverage ≥0.46 granularity) and all 24 CI jobs are green on 854f97e.

Issue #72, opened 2026-08-12
Add decision/outcome telemetry, a workflow eval-harness, and advisory semantic coverage (advisory, non-gating)


SanjeevSolanki:
comment lines are stripped/ignored by GitHub markdown.) -->

## Summary

Studio's deterministic **structural** assurance layer (`cfs validate`, `spec-coverage`, CPT markers, enforced gates) is strong and differentiated. The layer that's currently hard to observe and test is the **reasoning half** — what the agent decided during a run, and whether produced code is *semantically* faithful to its spec, not just marker-covered.

I'd like to propose three small, **advisory (non-gating)** additions that make that half observable, testable, and measurable — without changing any existing exit code or gate behavior. Happy to implement them as separate PRs.

## Motivation

- **Decisions aren't logged.** Telemetry currently records command + git identity — there's no record of what happened *inside* a run (intent, candidate options, model/tier, validator outcome, human accept/reject), so agent behavior can't be analyzed empirically.
- **Workflows have no regression signal.** Nothing runs a workflow against fixture scenarios and scores whether it followed its own discipline (stop points, reviewer steps). Behavior can regress silently.
- **Coverage rewards markers, not correctness.** `spec-coverage` scores `@cpt` marker density — a file can read 100% covered while implementing the wrong behavior. No advisory signal for semantic fidelity.

## What I'd add — three separate, advisory features

**1. Decision/outcome telemetry (local-only, opt-out).** A structured local JSONL log per decision: intent · candidate set · model/tier · validator pass/fail · findings count · human accept/reject. No network egress; fail-safe (logging can never raise into the caller).

**2. Workflow eval-harness (deterministic + optional advisory judge).** Runs a workflow against fixture scenarios, reports a **compliance % per workflow** with regression detection. Structural scorer is deterministic (no model calls); any LLM-judge layer is advisory, separated, and evaluated against a labeled gold set with reported accuracy/consistency.

**3. Advisory semantic coverage.** An advisory layer *alongside* structural `spec-coverage` judging whether a code block actually implements its linked requirement (covered / partial / wrong), flagging high-density/low-fidelity files. Never changes the structural number or gate result.

## Design principles

- **Advisory never gates** — none alter an exit code or block a build; the deterministic layer stays authoritative.
- **Deterministic where promised; measured where not** — structural scorers reproducible; LLM-judge components report their own accuracy and run-to-run variance (no silent non-determinism).
- **Local & private by default** — telemetry is local-only and opt-out.

[...]

PR #80, opened 2026-08-17
Instrument the command dispatcher + validators with decision/outcome telemetry

SanjeevSolanki:
Follow-up to #79 (the decision-log writer): this wires the writer into the engine so telemetry is actually produced. Part of the telemetry direction in #72.

## What
- A single **fail-open dispatch wrapper** around the CLI's `handler(rest)` records one `invocation` event per command — command, exit code, duration, and a safe `args_shape` (counts only, never raw argv) — for every command, on both normal and exception exits, without changing any command's behaviour or exit code.
- **Correlation:** a `contextvars` decision id, established once per run and scoped to it, so events recorded inside a command inherit the run's id. The engine mints the id; honouring an inherited upstream id (proxy→engine) is a follow-up.
- **Validators:** `validate` / `spec-coverage` / `validate-kits` / `self-check` record their final verdict as a `validation` event via the run's id. `record()` is fail-safe (never raises into the caller), so these are unguarded one-liners.

## Guardrails
- Fail-open, not fail-silent — a wrapper failure logs one line and never affects the command; no new network destinations; local-only + opt-out honoured (unchanged from #79); valid schema-1 events (no schema change).

## Tests / gates
- New `tests/test_cli_telemetry.py`: emission, correlation, fail-open (record/setup raising → exit code unchanged), crash logging, and id scoping. Telemetry redirected to a throwaway path in tests so it doesn't pollute file-snapshot assertions.
- Green: full suite, per-file ≥90% + diff 100% coverage, pylint, vulture, `cfs validate`, `spec-coverage`.

## Summary by CodeRabbit

* **New Features**
  * Added command execution reporting with duration, command details, argument shape, exit status, and correlated decision IDs.
  * Added final status reporting for self-checks, specification coverage, validation, and kit validation.

* **Reliability**
  * Reporting failures no longer interrupt command execution.
  * Decision context is cleared after each command run.

* **Tests**
  * Added coverage for reporting, correlation, cleanup, failure handling, and command exceptions.

### DIFF
--- a/skills/studio/scripts/studio/cli.py
+++ b/skills/studio/scripts/studio/cli.py
@@ -342,15 +342,12 @@ def _main_impl(argv_list: List[str]) -> int:
         return _report_unknown_command(cmd)
     # @cpt-end:cpt-studio-algo-core-infra-route-command:p1:inst-if-no-handler
     # @cpt-begin:cpt-studio-algo-core-infra-route-command:p1:inst-execute-handler
-    # Telemetry is fail-open: it must never change command behaviour or exit code
-    # (fail-open, not fail-silent). Establish one decision id for the run so events
-    # recorded inside the handler (e.g. validation) correlate to this invocation.
-    decision_id = ""
-    try:
-        decision_id = decision_log.new_decision_id()
-        decision_log.set_current_decision_id(decision_id)
-    except Exception as exc:  # pylint: disable=broad-except
-        logging.getLogger("studio").debug("decision-log setup failed: %s", exc)
+    # Telemetry never changes command behaviour: new_decision_id / set_current /
+    # record_invocation are all non-raising (record() is fail-safe and surfaces a real
+    # write failure itself), so no guard is needed here. One decision id per run lets
+    # events recorded inside the handler (e.g. validation) correlate to this invocation.
+    decision_id = decision_log.new_decision_id()
+    decision_log.set_current_decision_id(decision_id)
     started = perf_counter()
     handler_result = 1
     try:
@@ -364,25 +361,17 @@ def _main_impl(argv_list: List[str]) -> int:
             handler_result = 0 if exc.code is None else 1
         raise
     finally:
-        # Record on every exit — normal, SystemExit, or a raising handler.
-        try:
-            decision_log.record_invocation(
-                cmd,
-                exit_code=handler_result,
-                duration_ms=int((perf_counter() - started) * 1000),
-                args_shape={"argc": len(rest),
-                            "flags": sum(1 for arg in rest if arg.startswith("-"))},
-                decision_id=decision_id,
-            )
-        except Exception as exc:  # pylint: disable=broad-except
-            # Fail-open, not fail-silent: surface one line via the logger (→ stderr).
-            logging.getLogger("studio").warning(
-                "decision-log invocation record failed (ignored): %s", exc)
-        # Always scope the id to this run, even if recording raised.
-        try:
-            decision_log.set_current_decision_id("")
-        except Exception as exc:  # pylint: disable=broad-except
-            logging.getLogger("studio").debug("decision-log reset failed (ignored): %s", exc)
+        # Record on every exit — normal, SystemExit, or a raising handler — then scope
+        # the id to this run so a later event can't inherit it.
+        decision_log.record_invocation(
+            cmd,
+            exit_code=handler_result,
+            duration_ms=int((perf_counter() - started) * 1000),
+            args_shape={"argc": len(rest),
+                        "flags": sum(1 for arg in rest if arg.startswith("-"))},
+            decision_id=decision_id,
+        )
+        decision_log.set_current_decision_id("")
     # @cpt-end:cpt-studio-algo-core-infra-route-command:p1:inst-execute-handler
 
 
--- a/skills/studio/scripts/studio/cli.py
+++ b/skills/studio/scripts/studio/cli.py
@@ -14,7 +14,10 @@ IMPORTANT: This module MUST NOT contain business logic.
 import sys
 import logging
 from pathlib import Path
+from time import perf_counter
 from typing import Callable, List, Optional
+
+from .utils import decision_log
 
 _CLI_STDERR_HANDLER_NAME = "studio-cli-stderr"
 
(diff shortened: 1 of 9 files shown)


skills/studio/scripts/studio/cli.py
```diff
@@ -339,8 +342,37 @@ def _main_impl(argv_list: List[str]) -> int:
...
+    try:
+        decision_id = decision_log.new_decision_id()
+        decision_log.set_current_decision_id(decision_id)
+    except Exception as exc:  # pylint: disable=broad-except
+        logging.getLogger("studio").debug("decision-log setup failed: %s", exc)
+    started = perf_counter()
+    handler_result = 1
+    try:
+        handler_result = handler(rest)
+        return handler_result
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed — the wrapper now catches `SystemExit`, records its real code (0 for `--help`, 2 for arg errors; `None`→0, message→1) and re-raises it unchanged. Added a parametrized regression test `test_systemexit_records_its_real_code` covering all four cases.

skills/studio/scripts/studio/cli.py
```diff
@@ -339,8 +342,37 @@ def _main_impl(argv_list: List[str]) -> int:
...
+                duration_ms=int((perf_counter() - started) * 1000),
+                args_shape={"argc": len(rest),
+                            "flags": sum(1 for arg in rest if arg.startswith("-"))},
+                decision_id=decision_id,
+            )
+            decision_log.set_current_decision_id("")
+        except Exception as exc:  # pylint: disable=broad-except
+            # Fail-open, not fail-silent: surface one line via the logger (→ stderr).
+            logging.getLogger("studio").warning(
+                "decision-log invocation record failed (ignored): %s", exc)
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed — the id reset is now its own guarded step in the `finally`, so it always runs even if `record_invocation` raises. Covered by `test_wrapper_fail_open_when_record_raises` (record raises) and `test_decision_id_is_scoped_to_the_run`.

SanjeevSolanki:
@coderabbitai review

skills/studio/scripts/studio/cli.py
```diff
@@ -339,8 +342,47 @@ def _main_impl(argv_list: List[str]) -> int:
...
+        if isinstance(exc.code, int):
+            handler_result = exc.code
+        else:
+            handler_result = 0 if exc.code is None else 1
+        raise
+    finally:
+        # Record on every exit — normal, SystemExit, or a raising handler.
+        try:
+            decision_log.record_invocation(
+                cmd,
```
ainetx: **Severity: MAJOR**

**What was found**

Real decision-log write failures are silently ignored. `record()` catches filesystem and serialization exceptions and returns `False`; 
`record_invocation()` propagates that value, but the dispatcher discards it and warns only when an exception escapes. Therefore, an unwritablelog can drop an invocation event without the visible warning promised by the PR’s “fail-open, not fail-silent” guardrail.

**How to verify**

Add a regression test that enables decision logging and makes `_append_locked()` raise `OSError`. Verify that:

1. the command preserves its original exit code and behavior;
2. no invocation event is written;
3. exactly one visible warning is emitted;
4. explicit opt-out and no-project cases remain quiet.

**How to fix**

Distinguish successful writes, intentional no-ops, and actual failures—for example, with `WRITTEN`, `SKIPPED`, and `FAILED` results—and have 
the dispatcher warn only for `FAILED`. Alternatively, emit the warning inside `record()` when it catches a real write exception. Do not warn 
for expected disabled/no-project cases.

ainetx: **Severity: MAJOR**

**What was found**

Real decision-log write failures are silently ignored. `record()` catches filesystem and serialization exceptions and 
returns `False`; `record_invocation()` propagates that value, but the dispatcher discards it and warns only when an 
exception escapes. Therefore, an unwritable log can drop an invocation event without the visible warning promised by the 
PR’s “fail-open, not fail-silent” guardrail.

**How to verify**

Add a regression test that enables decision logging and makes `_append_locked()` raise `OSError`. Verify that:

1. the command preserves its original exit code and behavior;
2. no invocation event is written;
3. exactly one visible warning is emitted;
4. explicit opt-out and no-project cases remain quiet.

**How to fix**

Distinguish successful writes, intentional no-ops, and actual failures—for example, with `WRITTEN`, `SKIPPED`, and 
`FAILED` results—and have the dispatcher warn only for `FAILED`. Alternatively, emit the warning inside `record()` when itcatches a real write exception. Do not warn for expected disabled/no-project cases.

ainetx (changes requested):
## Deep Review

**Verdict: Changes requested**

### Severity: MAJOR

**What was found**

Real decision-log write failures can be silently ignored. \`record()\` catches filesystem and serialization exceptions andreturns \`False\`; \`record_invocation()\` propagates that value, but the dispatcher discards it and warns only when an 
exception escapes. This can drop invocation events without the visible warning promised by the PR guardrail: “fail-open, 
not fail-silent”.

Inline finding: https://github.com/constructorfabric/studio/pull/80#discussion_r3795152412

**How to verify**

Enable decision logging and make \`_append_locked()\` raise \`OSError\`. Confirm that:

1. the command preserves its original exit code and behavior;
2. no invocation event is written;
3. exactly one visible warning is emitted;
4. explicit opt-out and no-project cases remain quiet.

**How to fix**

Distinguish successful writes, intentional no-ops, and actual failures—for example, using \`WRITTEN\`, \`SKIPPED\`, and 
\`FAILED\` outcomes—and warn only for \`FAILED\`. Alternatively, emit the warning where \`record()\` catches a real writeexception. Expected disabled/no-project cases should remain quiet.

skills/studio/scripts/studio/cli.py
SanjeevSolanki: Good catch — you're right, this was fail-*silent* for the case that mattered. `record()` swallowed write failures and returned `False`, so the dispatcher's guard (which only fired on a *raised* exception) never triggered for a genuinely unwritable log.

Fixed per your option (b) in f19e011b:

- **The warning now lives inside `record()`.** On the first real write/serialization failure it emits exactly one `WARNING`, then latches telemetry off for the rest of the run — so it neither spams a line per event nor keeps retrying a target we already know is unwritable. The "off for this run" wording is now literally true.
- **The exception is redacted before logging.** An `OSError` carries the absolute log path (`$HOME` included), so I run it through the same `_redact` used for the startup notice — a failure prints `~/…/decisions.jsonl`, not the username.
- **The dispatch wrapper is now guard-free.** Id-minting, the contextvar set, and `record_invocation` are all non-raising, so there's nothing left for it to catch.

Regression tests for your four checks: (1) exit code preserved and (2) no event written when `_append_locked` raises `OSError` (through `cli.main()`); (3) exactly one visible warning and the write is attempted only once (latched); (4) explicit opt-out and no-project stay silent. Also asserts `$HOME` is absent from the warning.

SanjeevSolanki: Duplicate of the thread just below — same finding, addressed in f19e011b: the warning now fires once inside `record()` (then latches telemetry off), and the exception is redacted. Full detail and the regression tests are in the other thread. 👍

Issue #89, opened 2026-08-17
validate/spec-coverage: a green gate does not prove the spec is implemented — close the empty-scan and denominator gaps


Oleg67:
## The gap

`cfs validate` and `cfs spec-coverage` are the two commands a team wires as required CI checks, so
their exit codes carry the whole weight of "this spec is implemented". Three states currently
produce a green exit while nothing was actually verified:

**1. No codebase registered → the coverage gate can never fail.** With `codebase = []` and no
`[[systems.autodetect.codebase]]` block, `spec-coverage` returns:

```json
{"status":"PASS",
 "summary":{"total_files":0,"covered_files":0,"coverage_pct":0.0,"granularity_score":0.0},
 "message":"No codebase files found in registry"}
```

0.0 % against a required 80 and 0.0 granularity against a required 0.7 — both missed by the maximum
possible margin, reported as PASS, exit 0. It does not self-correct when code lands: adding a fully
marked source file leaves `total_files: 0` and the same message. A team can wire this as a required
check on day one and it will never once fail.

**2. A registered-but-empty tree passes both gates.** A FEATURE with three checked `to_code = true`
IDs and zero code files gives `cfs validate` exit 0 "All checks passed" — while the same run prints
`Code files: 0` and `Code coverage: 0/3`. Add one *unmarked* `.py` file and the identical artifact
state produces 16 errors and exit 2. The checks are correct; they are unreachable when nothing was
scanned.

**3. Marker enforcement is conditioned on that same non-empty tree.** An ID marked `to_code = true`
and checked `[x]`, with every CDSL instruction checked, no `@cpt` markers and no code, validates
clean:

```json
{"status":"PASS","error_count":0,"warning_count":0,
 "code_files_scanned":0,"to_code_ids_total":1,"code_ids_found":0}
```

`to_code_ids_total: 1` against `code_ids_found: 0` sit in the same JSON object and produce neither
an error nor a warning.

**4. A single scope marker claims its whole file, so the coverage number is unstable.** Removing one
scope-marker comment line from a 342-line file, with no other change, moved coverage
**62.4 % → 9.71 %** and granularity **0.2953 → 0.8187**. The two metrics move in opposite
directions under the same edit, which makes a coverage figure hard to reason about — and means
structural files that cannot host an instruction block (re-export modules, exception declarations,
`Protocol` ports) silently depress the denominator.

## The change

- **Registration**: `cfs init` registers a detected codebase, or warns explicitly when it cannot —
[...]

PR #90, opened 2026-08-17
Add a regression corpus for empty-scan verdicts in validate and spec-coverage

Oleg67:
**What.** Adds `tests/test_enforcement_empty_scan.py` — a regression corpus covering the repository
states in which `cfs validate` and `cfs spec-coverage` report success without having assessed
anything, plus the neighbouring states that must keep passing so a later change to either command
cannot be too broad. No production code is touched.

**Why.** These states are easy to reach and currently indistinguishable from a clean run:

- A codebase registry with no entry makes `spec-coverage` return
  `{"status": "PASS", "summary": {"total_files": 0, "coverage_pct": 0.0}}` and exit 0 — against any
  thresholds. It does not self-correct when marked code lands.
- A `FULL` codebase entry that resolves to zero files passes both commands, while `validate` prints
  `Code files: 0` and `Code coverage: 0/3` in the same run.
- A checked `to_code="true"` ID with no markers and no code validates clean, with
  `to_code_ids_total: 1` and `code_ids_found: 0` in the same JSON object and neither error nor
  warning.

The corpus states each of these as a test so the behaviour is pinned and any change to it is
visible.

**How it is organised.** Three groups:

1. **Reachability proofs — green today.** `cross_validate_code()` called with *no* code files already
   emits one coverage error per unmet `to_code` ID. Worth having as a test in its own right: it shows
   the checks work on empty input, so the difference lives in the caller-side path, not in the
   validation logic.
2. **Regression pins — green today.** A marked implementation stays clean; one unmarked file still
   fails; a repository that claims nothing is owed nothing; an unassessable report without thresholds
   keeps exit 0. These are the cases a stricter verdict must not break — in particular the
   spec-first repository (PRD and ADR written, no implementation yet) which should not be failed.
3. **Three `xfail(strict=True)` cases** stating the verdict this corpus argues for where the commands
   currently answer differently: an empty scan reported as `PASS`; requested thresholds not reached
   at all when nothing was scanned; and an unregistered codebase being indistinguishable from an
   entry that resolves to zero files (both return `_empty_coverage_result()`, so the report cannot
   state its own denominator).

The markers are **strict**, so an unexpected pass fails the suite. If any of these is settled the
other way, or the behaviour changes for another reason, the corpus reports it rather than going
quietly green.

**The rule the three proposed cases share.** "Cannot assess" is not "PASS", and the exit code should
follow whether the caller asked for a guarantee: if thresholds were passed, or a checked
`to_code="true"` ID exists, and that cannot be assessed → non-zero exit; if only a report was asked
for → exit 0, but say plainly that nothing was measured. Group 2 exists to keep that from becoming
over-strict.

**Gates.** Full suite green on the rebased state (4,417 passed, 3 skipped, 18 xfailed). `pylint`,
`vulture-ci` and `spec-coverage` do not scope `tests/`, and no artifact changed, so `cfs validate` is
unaffected. No new dependencies; no CPT tracing required (tests are not engine code).

---

Context: this is the test-side companion to #89.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression coverage for empty and documentation-only scans.
  * Verified consistent enforcement outcomes across marked, unmarked, and populated codebases.
  * Confirmed report-only empty scans retain a zero exit status.
  * Added checks for deterministic results and appropriate handling of unassessable or incorrectly reported scans.

### DIFF
--- a/tests/test_enforcement_empty_scan.py
+++ b/tests/test_enforcement_empty_scan.py
@@ -1,0 +1,242 @@
+"""Enforcement reachability: an empty scan must not read as compliance.
+
+Each test names the repository state it builds and the verdict that state
+produces. The rule this corpus argues for is that "cannot assess" is not
+"PASS", and that the exit code should follow whether the caller asked for a
+guarantee:
+
+* a guarantee was requested (thresholds passed, or a checked ``to_code="true"``
+  ID exists) and it cannot be assessed -> non-zero exit;
+* only a report was requested -> exit 0, but the output says plainly that
+  nothing was measured.
+
+Most tests here pin behaviour that is already correct, so they are ordinary
+regression cover. The three marked ``xfail(strict=True)`` state the verdict this
+corpus proposes for cases the commands currently answer differently; the marker
+records that today's behaviour differs, and its reason names the code that
+decides. Because the marker is strict, an unexpected pass fails the suite -- so
+if a case is settled the other way, or the behaviour changes, the corpus says so
+instead of going quietly green.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from io import StringIO
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "skills" / "studio" / "scripts"))
+
+from studio.commands.spec_coverage import cmd_spec_coverage
+from studio.utils.artifacts_meta import ArtifactsMeta, CodebaseEntry, Kit, SystemNode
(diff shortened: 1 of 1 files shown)


Oleg67:
Superseded by #93, closing in favour of it.

This PR documented three cases as `xfail(strict)` because no fix existed yet. #93 fixes all three and carries the same tests as plain assertions, so the change and its tests land together rather than as two PRs whose correctness depends on merge order — if this one had merged second, its strict markers would have turned into `XPASS(strict)` failures on `main`.

The strict markers did their job on the way out: run against #93's branch, all three flipped to `XPASS(strict)` and failed, which is what told me one change closed all three.

PR #93, opened 2026-08-18
feat(spec-coverage): report applicability and honour demanded thresholds on an empty scope

Oleg67:
Part of the issue #89.

**What.** `cfs spec-coverage` now reports whether there was anything to assess, and honours a threshold the caller actually asked for even when the scope turns out to be empty.

**Why.** A scan with nothing in scope returned a bare `PASS` at exit 0 regardless of the request:

```json
{"status":"PASS","summary":{"total_files":0,"coverage_pct":0.0,"granularity_score":0.0},
 "message":"No codebase files found in registry"}
```

Two consequences:

- **A demanded guarantee was silently dropped.** `--min-coverage 90` over zero files reported PASS and exited 0 — the maximum possible miss, reported as success. A required CI check configured that way can never fail, and does not start failing when marked code lands.
- **An empty result was indistinguishable from a covered one.** Callers could not tell "nothing was measured" from "everything measured passed" — and an unregistered codebase produced byte-identical output to a registered entry whose path resolves to no files, which are two different mistakes with two different fixes.

**How.** The report answers three questions separately instead of overloading `status`:

| Question | Answer |
|---|---|
| Did anything fail? | `status` — an empty scope with no thresholds requested fails nothing, so it stays `PASS` |
| Was there anything to assess? | `applicable: false`, with a message naming the registered-entry count and distinguishing *unregistered* from *resolved to zero files* |
| Was a guarantee demanded that cannot be given? | the exit code — non-zero, with one `threshold_failures` entry per requested threshold, only when one was requested |

Keeping `status: PASS` for the no-threshold case is deliberate: the check *ran* and correctly found nothing to cover, so it is a completed assessment with an empty result, not a skipped one — and a spec-first repository with documents but no implementation yet should not be failed. `applicable` is what makes that honest rather than merely quiet. The field name is open to your preference.

**⚠️ Behaviour change for anyone pinning exit codes:** passing a `--min-*` threshold over an empty scope now exits 2 instead of 0. Runs that request no thresholds are unaffected, and no existing test needed changing — including `test_no_codebase_files`, which still asserts `status == "PASS"`.

**Tests.** Adds `tests/test_enforcement_empty_scan.py` (11 tests): the three cases above, plus regression cover that a stricter verdict must not break — a marked implementation stays clean, one unmarked file still fails, a repository that claims nothing is owed nothing, and an empty report without thresholds keeps exit 0. Also pins that `cross_validate_code()` already reports every unmet `to_code` ID when given no code files, which matters for the follow-up in #89.

**Gates.** `make test` 4432 passed / 4 skipped / 15 xfailed · `pylint` clean · `vulture-ci` clean · `cfs validate` 216/216, 0 errors 0 warnings · `make spec-coverage` all thresholds met, with the changed file at 87% / granularity 0.90 (instruction-block traced, no whole-file scope marker). No new dependencies; stdlib only.

Supersedes #90, which documented these three cases as `xfail` before a fix existed — closing that in favour of this, so the tests and the change land together.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved empty spec-coverage results with clearer context for unregistered codebases and registered entries containing no files.
  * Empty scans now indicate when coverage checks are not applicable and provide threshold-specific failure details.
  * Corrected exit statuses for enforced and non-enforced empty scans.
  * Improved recursive counting for selected registered codebases.
  * Coverage thresholds now pass when results meet the specified value.
  * Updated human-readable and JSON reports with clearer empty-state messaging and warnings.

* **Tests**
  * Added regression coverage for empty, documentation-only, populated, deterministic, and threshold-enforced scan scenarios.

### DIFF
--- a/architecture/features/spec-coverage.md
+++ b/architecture/features/spec-coverage.md
@@ -68,8 +68,8 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - User runs `cfs spec-coverage --min-file-granularity 0.3` → same as above, exit code 2 if any covered file granularity below threshold
 
 **Error Scenarios**:
-- No codebase entries registered → ERROR with hint to configure artifacts.toml
-- No code files found → report with 0% coverage
+- No codebase entries registered → report with `applicable: false` and a hint to configure artifacts.toml; exit code 2 only if a positive threshold was demanded
+- No code files found → report with `applicable: false` naming how many registered entries resolved to no files, and 0% coverage
 
 **Steps**:
 1. [x] - `p1` - User invokes `cfs spec-coverage [--min-coverage N] [--min-file-coverage N] [--min-granularity N] [--min-file-granularity N] [--verbose]` - `inst-user-spec-coverage`
@@ -90,7 +90,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - [x] - `p1` - Validate selected `--system` values and build unknown-system failure payloads - `inst-validate-systems`
 - [x] - `p1` - Filter ignored and out-of-root files before scanning - `inst-filter-ignored-files`
 - [x] - `p1` - Build empty coverage result when registry yields no code files - `inst-empty-report`
-- [x] - `p1` - Count codebase entries registered by the selected systems, so an unregistered registry is distinguishable from one that resolves to no files - `inst-count-registered-entries`
+- [x] - `p1` - Count codebase entries registered by the selected systems and recurse into child systems, including the default all-systems selection, so an unregistered registry is distinguishable from one that resolves to no files - `inst-count-registered-entries`
 - [x] - `p1` - Detect which threshold flags demand an enforceable guarantee, ignoring non-positive values that any scope satisfies - `inst-detect-requested-thresholds`
 - [x] - `p1` - Apply per-report threshold checks and accumulate failure messages - `inst-apply-thresholds`
 - [x] - `p1` - Resolve paths relative to project root for human-readable output - `inst-rel-path`
@@ -217,7 +217,7 @@ The system **MUST** calculate instruction density per covered file: `min(1.0, bl
 
 - [x] `p1` - **ID**: `cpt-studio-dod-spec-coverage-report`
 
-The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. The report **MUST** state whether there was anything to assess, so a scope that yielded no code files is distinguishable from a fully covered one. Exit code 0 when above thresholds, 2 when below or when a positive threshold was demanded over a scope that could not be assessed.
+The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. The report **MUST** state whether there was anything to assess, so a scope that yielded no code files is distinguishable from a fully covered one. Exit code 0 when at or above thresholds, 2 when below or when a positive threshold was demanded over a scope that could not be assessed.
 
 **Implements**:
 - `cpt-studio-flow-spec-coverage-report`
@@ -255,4 +255,4 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
 - [x] Scanning completes in ≤ 5 seconds for typical repositories
 - [x] An empty scope reports `applicable: false` and says why, on both the JSON and human surfaces
 - [x] A positive `--min-*` threshold over an empty scope exits 2; a non-positive one exits 0
-- [x] All output is valid JSON to stdout with exit codes 0/1/2
+- [x] JSON mode emits valid JSON to stdout and human mode emits formatted text, with exit codes 0/1/2 on both
--- a/architecture/features/spec-coverage.md
+++ b/architecture/features/spec-coverage.md
@@ -79,7 +79,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 5. [x] - `p1` - Calculate coverage metrics using `cpt-studio-algo-spec-coverage-metrics` - `inst-calc-metrics`
 6. [x] - `p1` - Calculate granularity scores using `cpt-studio-algo-spec-coverage-granularity` - `inst-calc-granularity`
 7. [x] - `p1` - Generate report using `cpt-studio-algo-spec-coverage-report` - `inst-gen-report`
-8. [x] - `p1` - **IF** any threshold flag set AND metric below threshold → exit code 2 - `inst-if-threshold`
+8. [x] - `p1` - **IF** any positive threshold flag set AND (metric below threshold OR nothing was assessed) → exit code 2 - `inst-if-threshold`
 9. [x] - `p1` - **RETURN** JSON report (summary, per-file stats, uncovered files) - `inst-return-report`
 
 **Supporting**:
@@ -90,6 +90,8 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - [x] - `p1` - Validate selected `--system` values and build unknown-system failure payloads - `inst-validate-systems`
 - [x] - `p1` - Filter ignored and out-of-root files before scanning - `inst-filter-ignored-files`
 - [x] - `p1` - Build empty coverage result when registry yields no code files - `inst-empty-report`
+- [x] - `p1` - Count codebase entries registered by the selected systems, so an unregistered registry is distinguishable from one that resolves to no files - `inst-count-registered-entries`
+- [x] - `p1` - Detect which threshold flags demand an enforceable guarantee, ignoring non-positive values that any scope satisfies - `inst-detect-requested-thresholds`
 - [x] - `p1` - Apply per-report threshold checks and accumulate failure messages - `inst-apply-thresholds`
 - [x] - `p1` - Resolve paths relative to project root for human-readable output - `inst-rel-path`
 - [x] - `p1` - Route JSON report to file or terminal UI - `inst-output-report`
@@ -215,7 +217,7 @@ The system **MUST** calculate instruction density per covered file: `min(1.0, bl
 
 - [x] `p1` - **ID**: `cpt-studio-dod-spec-coverage-report`
 
-The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. Exit code 0 when above thresholds, 2 when below.
+The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. The report **MUST** state whether there was anything to assess, so a scope that yielded no code files is distinguishable from a fully covered one. Exit code 0 when above thresholds, 2 when below or when a positive threshold was demanded over a scope that could not be assessed.
 
 **Implements**:
 - `cpt-studio-flow-spec-coverage-report`
@@ -251,4 +253,6 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
 - [x] `--verbose` flag includes per-file marker details in report
 - [x] Report format mirrors `coverage.py` JSON structure (summary + per-file)
 - [x] Scanning completes in ≤ 5 seconds for typical repositories
+- [x] An empty scope reports `applicable: false` and says why, on both the JSON and human surfaces
+- [x] A positive `--min-*` threshold over an empty scope exits 2; a non-positive one exits 0
 - [x] All output is valid JSON to stdout with exit codes 0/1/2
--- a/skills/studio/scripts/studio/commands/spec_coverage.py
+++ b/skills/studio/scripts/studio/commands/spec_coverage.py
@@ -190,10 +190,17 @@ def _count_selected_codebase_entries(meta, system_slugs: set[str] | None) -> int
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
     total = 0
 
+    def count_subtree(node: object) -> int:
+        """Entries registered by ``node`` and every descendant of it."""
+        return len(getattr(node, "codebase", None) or []) + sum(
+            count_subtree(child) for child in getattr(node, "children", [])
+        )
+
     def visit(node: object) -> None:
+        """Add the subtree of ``node`` once it is in scope, mirroring file collection."""
         nonlocal total
         if system_slugs is None or getattr(node, "slug", "") in system_slugs:
-            total += len(getattr(node, "codebase", None) or [])
+            total += count_subtree(node)
             return
         for child in getattr(node, "children", []):
             visit(child)
--- a/skills/studio/scripts/studio/commands/spec_coverage.py
+++ b/skills/studio/scripts/studio/commands/spec_coverage.py
@@ -41,25 +41,26 @@ def _build_spec_coverage_parser() -> argparse.ArgumentParser:
         "--min-coverage",
         type=float,
         default=None,
-        help="Minimum coverage percentage (0-100). Exit 2 if below.",
+        help="Minimum coverage percentage (0-100). Exit 2 if below, or if nothing was assessed.",
     )
     parser.add_argument(
         "--min-file-coverage",
         type=float,
         default=None,
-        help="Minimum per-file coverage percentage (0-100). Exit 2 if any file is below.",
+        help="Minimum per-file coverage percentage (0-100). Exit 2 if any file is below, or if nothing was assessed.",
     )
     parser.add_argument(
         "--min-granularity",
         type=float,
         default=None,
-        help="Minimum granularity score (0-1). Exit 2 if below.",
+        help="Minimum granularity score (0-1). Exit 2 if below, or if nothing was assessed.",
     )
     parser.add_argument(
         "--min-file-granularity",
         type=float,
         default=None,
-        help="Minimum per-file granularity score (0-1). Exit 2 if any covered file is below.",
+        help="Minimum per-file granularity score (0-1). Exit 2 if any covered file is below, "
+             "or if nothing was assessed.",
     )
     parser.add_argument(
         "--system",
@@ -187,33 +188,33 @@ def _count_selected_codebase_entries(meta, system_slugs: set[str] | None) -> int
     Distinguishes "nothing is registered" from "what is registered resolves to
     no files" -- two different mistakes that otherwise produce identical output.
     """
-    # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
-    total = 0
-
+    # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-count-registered-entries
     def count_subtree(node: object) -> int:
         """Entries registered by ``node`` and every descendant of it."""
         return len(getattr(node, "codebase", None) or []) + sum(
             count_subtree(child) for child in getattr(node, "children", [])
         )
 
-    def visit(node: object) -> None:
-        """Add the subtree of ``node`` once it is in scope, mirroring file collection."""
-        nonlocal total
+    def visit(node: object) -> int:
+        """Count ``node``'s subtree once it is in scope, mirroring file collection."""
         if system_slugs is None or getattr(node, "slug", "") in system_slugs:
-            total += count_subtree(node)
-            return
-        for child in getattr(node, "children", []):
-            visit(child)
+            return count_subtree(node)
+        return sum(visit(child) for child in getattr(node, "children", []))
 
-    for system_node in meta.systems:
-        visit(system_node)
-    return total
-    # @cpt-end:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
+    return sum(visit(system_node) for system_node in meta.systems)
+    # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-count-registered-entries
 
 
 def _requested_thresholds(args) -> List[str]:
-    """Names of the thresholds the caller asked to be enforced."""
-    # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
+    """Names of the thresholds whose satisfaction the caller actually demanded.
+
+    A non-positive threshold is met by any scope, empty or not, so it demands no
+    guarantee and is not counted as one. Without that, ``--min-coverage 0`` would
+    fail an empty scope while passing a populated one sitting at 0.0% -- and the
+    exit code has to keep answering a single question: was a guarantee demanded
+    that cannot be given?
+    """
+    # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-detect-requested-thresholds
     requested = []
     for flag, attr in (
         ("--min-coverage", "min_coverage"),
@@ -221,10 +222,11 @@ def _requested_thresholds(args) -> List[str]:
         ("--min-granularity", "min_granularity"),
         ("--min-file-granularity", "min_file_granularity"),
     ):
-        if getattr(args, attr, None) is not None:
+        value = getattr(args, attr, None)
+        if value is not None and value > 0:
             requested.append(flag)
     return requested
-    # @cpt-end:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
+    # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-detect-requested-thresholds
 
 
 def _empty_coverage_result(registered_entries: int = 0, requested_thresholds=None) -> dict:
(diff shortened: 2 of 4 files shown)


skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -181,18 +181,80 @@ def _filter_ignored_files(code_files_to_scan: List[Path], project_root: Path, me
...
+    # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
+    total = 0
+
+    def visit(node: object) -> None:
+        nonlocal total
+        if system_slugs is None or getattr(node, "slug", "") in system_slugs:
+            total += len(getattr(node, "codebase", None) or [])
+            return
+        for child in getattr(node, "children", []):
+            visit(child)
```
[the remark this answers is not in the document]
Oleg67: Confirmed and fixed in c770b228 — thanks, this was a real defect.

Two notes on the scope:

**It also affects the default run, not just an explicit `--system`.** With no selector, `system_slugs is None`, so the first branch matches every root node immediately and returns before reaching `children` — descendants went uncounted there too. That is the more commonly exercised path, so I widened the fix and the tests accordingly.

**The committable suggestion as posted does not parse** — the `+` between `len(...)` and `sum(...)` is missing, so it would be a syntax error if applied. The prose diff above it is correct; I implemented that shape rather than the suggestion block.

The fix mirrors `_collect_codebase_files`' recursion so the reported denominator describes exactly the subtree the scan walked:

```python
    def count_subtree(node: object) -> int:
        """Entries registered by ``node`` and every descendant of it."""
        return len(getattr(node, "codebase", None) or []) + sum(
            count_subtree(child) for child in getattr(node, "children", [])
        )
```

Covered by two new tests — a parent whose only registered entry lives on a child and resolves to zero files, asserted for the default run and for `--system` on the parent. Both were red before the change and green after; the nested-system case had no fixture in the corpus before, which is why it slipped through.

Gates re-run on the amended state: full suite **4,434 passed, 4 skipped, 15 xfailed** on Python 3.11; `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` all thresholds met; `validate` 0 errors / 0 warnings, code coverage 216/216; `vulture-ci` clean; `pylint` clean on the two configured targets.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -181,18 +181,87 @@ def _filter_ignored_files(code_files_to_scan: List[Path], project_root: Path, me
...
+
+    def count_subtree(node: object) -> int:
+        """Entries registered by ``node`` and every descendant of it."""
+        return len(getattr(node, "codebase", None) or []) + sum(
+            count_subtree(child) for child in getattr(node, "children", [])
+        )
+
+    def visit(node: object) -> None:
+        """Add the subtree of ``node`` once it is in scope, mirroring file collection."""
+        nonlocal total
```
ainetx: **[MINOR]** Stylistic inconsistency: this closure accumulates via `nonlocal total` (a scalar), while the two structurally identical sibling tree-walkers in this file (`_collect_system_slugs`, `_collect_selected_system_files`) mutate a container passed by reference instead. No functional impact — flagged for consistency only.

tests/test_enforcement_empty_scan.py
```diff
@@ -0,0 +1,306 @@
...
+
+        assert report["status"] == "PASS"
+        assert report.get("applicable") is False
+
+    def test_requested_thresholds_cannot_be_satisfied_by_an_empty_scan(self):
+        """0.0% against a required 90 is the maximum possible miss, not a pass."""
+        with TemporaryDirectory() as directory:
+            ctx = _context(Path(directory), codebase=[])
+
+            code, _ = _run_spec_coverage(ctx, ["--min-coverage", "90"])
```
ainetx: **[MINOR]** Only `--min-coverage` is tested over an empty scope here (and only for `code != 0`, not the actual failure-message content). `--min-file-coverage`, `--min-granularity`, and `--min-file-granularity` are never exercised over an empty scope anywhere in this file. A per-flag or message-content regression in the other three thresholds would pass the whole suite undetected.

ainetx (commented):
Automated Deep Review of this PR (34 checks across 6 phases, each independently re-verified by a second agent before being reported).

2 of 5 originally-flagged MAJOR issues are already fixed by the latest commit (`c770b228`) — the recursion bug and its missing regression test — nice fix, confirmed by re-reading the new code and tests.

Remaining findings unresolved as of `c770b228`. Two are inline below (within this PR's diff); the rest touch files this PR doesn't modify, so GitHub won't let me anchor them to a line — listing them here instead:

**[MAJOR, MEDIUM confidence] Pre-existing latent defect surfaced by this PR** — `skills/studio/scripts/studio/utils/artifacts_meta.py`, `ArtifactsMeta.from_dict` (~line 597): a malformed `systems` field (wrong type, or non-dict list entries) silently coerces to `systems = []` with no error recorded, indistinguishable from a legitimately empty registry. Not part of this PR's diff, but this PR's new reliance on the entry count makes the gap newly user-visible. Suggested fix: record a warning/error when `raw_systems` isn't a valid list-of-dicts.

**[MAJOR, CONFIRMED] Human-readable output never updated** — `skills/studio/scripts/studio/commands/spec_coverage.py`, `_human_spec_coverage` (~line 507, pre-existing, untouched by this PR): never reads the new `applicable` field, so a developer running `spec-coverage` **without** `--json` against an empty/unregistered scope still sees generic `Files 0/0 covered ... All thresholds met.` — reproducing the exact ambiguity this PR set out to fix, just on the most common interactive surface. Suggested fix: check `data.get("applicable")` there and render `data.get("message")` when `False`.

**[MAJOR, CONFIRMED] New behavior never exercised through real CLI/CI** — `.github/workflows/ci.yml` (~line 142): `make spec-coverage` always targets the real, populated `studio` system, never an empty scope; the repo's only real-CLI-entry-point test file (`tests/test_cli_validation_e2e.py`) is in the same situation. A wiring-level regression in this PR's new path would ship undetected by anything in this repo's CI. Suggested fix: add one e2e test bootstrapping a system with `codebase=[]` through the real CLI, asserting on `applicable`/exit code.

**[MINOR] Feature spec doc stale** — `architecture/features/spec-coverage.md` (~line 71, untouched by this PR): still frames exit code 2 purely as "metric below threshold"; no mention of the new empty-scope+requested-threshold path or the `applicable` field.

**[MINOR] CLI `--help` text stale** — `skills/studio/scripts/studio/commands/spec_coverage.py`, the four `--min-*` argparse help strings (~line 44, untouched by this PR): still say "Exit 2 if below" only; don't mention the new empty-scope trigger.

**[MINOR] Threshold values unvalidated** — `skills/studio/scripts/studio/commands/spec_coverage.py`, `_check_min_coverage` and its three siblings (~line 274-334, untouched by this PR): no range validation on `--min-*` values; a negative value silently makes the check a permanent no-op (functionally equivalent to omitting the flag, so real-world impact is low, but still worth an argparse-level rejection).

Oleg67:
Thanks — this was a genuinely useful review. I verified every finding against the code rather than taking the labels, and reproduced the confirmed ones. Addressed in `15cf21cf`.

## The human-output finding is correct, and it was the important one

Reproduced by running the command path in human mode. Unregistered and registered-but-resolves-to-nothing produced **byte-identical** output:

```
  Spec Coverage
    Files: 0/0 covered
    Coverage: 0.0%
    Granularity: 0.0000

  ✓ All thresholds met.
```

You were right that this reproduces the exact ambiguity the change set out to remove, on the surface most people use. Now:

```
  ⚠ No codebase entries are registered, so no code files were scanned
```
```
  ⚠ No code files found: 1 registered codebase entry resolved to 0 files
```

`_human_spec_coverage` renders the message and no longer emits a verdict it did not reach.

One refinement: the **threshold** path already worked — it rendered `⚠ cannot assess --min-coverage: …` and `✗ Threshold check failed.` correctly. Only the no-threshold path was blind, which is precisely the path the design keeps at `PASS`/exit 0 and where the applicability flag is therefore the *only* signal. So the defect was narrower than stated but more consequential, because that path had no other way to communicate.

## Two things the review didn't catch — one of them mine

**1. I had introduced a false positive.** Measured:

```
--min-coverage 0    → exit 2, FAIL
--min-coverage -5   → exit 2, FAIL
```

A non-positive threshold is met by any scope; a populated repository sitting at 0.0% passes `--min-coverage 0`, so an empty one has to agree or the flag means two different things depending on what happens to be registered. All four flags behaved this way. Fixed — only a positive threshold now counts as a demanded guarantee. Your **threshold-values-unvalidated** finding is what led me to it, so that MINOR paid for itself even though I've resolved it differently (see below).

**2. My CPT tracing was dishonest and I'd rather say so than have it found later.** I had traced the entry counter and the threshold-request check under `inst-state-uncovered`, whose declared transition is *"FROM NOT_RUN TO UNCOVERED WHEN no CDSL markers found in any code file"*. Reading the registry is not that. It cleared `spec-coverage` without declaring the new logic. Both now have declared instructions of their own (`inst-count-registered-entries`, `inst-detect-requested-thresholds`), and step 8 — which said exit 2 was `metric below threshold` only — now describes the second trigger. So your "feature spec doc stale" MINOR was really two problems, and the second one was mine.

## Disposition of the rest

| Finding | Action |
|---|---|
| Human output never updated | **Fixed** — formatter + tests |
| Feature spec doc stale | **Fixed** — step 8, DoD, two acceptance criteria, plus the tracing correction above |
| CLI `--help` stale | **Fixed** — my change is what made it wrong, so it belongs here |
| Threshold values unvalidated | **Resolved differently** — see below |
| New behavior never exercised through real CLI/CI | **Partly addressed**, partly disputed — see below |
| `from_dict` malformed `systems` | **Confirmed, deferring** — see below |
| `nonlocal` style | **Fixed** — replied inline |
| Only `--min-coverage` tested | **Fixed** — replied inline |

**Threshold range validation** — I've declined argparse-level rejection and fixed the semantics instead. Rejecting negatives at parse time is a behaviour change for existing callers that is out of scope for this PR, and the `> 0` rule removes the only place the missing validation now bites. Happy to add the argparse guard if you'd rather have it, but as a separate change.

**Real-CLI e2e** — half agreed. The corpus does drive `cmd_spec_coverage(argv)`, so argv parsing and exit-code plumbing are exercised; and `_human_spec_coverage` already had a unit-test class in `tests/test_ui_human_mode.py`. So the gap was a missing *case*, not missing infrastructure — I've added it in both places, including an end-to-end human-mode test that opts out of the autouse JSON-mode fixture in `conftest.py`, since without that opt-out every JSON assertion could hold while the human surface stayed broken. I don't think a bootstrapped-project e2e adds much beyond that, but say the word and I'll add one.

**`from_dict`** — confirmed as a real pre-existing defect, and I agree this PR makes it newly consequential: my message now asserts *"no codebase entries are registered"* as a fact, which a silently-dropped malformed `systems` field would make false. The same silent-coercion pattern also applies to `kits` a few lines above. It belongs in its own PR against a file this one doesn't touch, so I've left it — let me know if you'd prefer it folded in.

## Test quality

Every production change in this PR is mutation-tested — reverting it turns the relevant tests red:

| Reverted change | Result |
|---|---|
| Human formatter ignores `applicable` | **5 failed** |
| Any non-`None` threshold counts as a demand | **1 failed** |
| Entry counter stops at the matched node | **2 failed** |

Gates on the final state, Python 3.11: full suite **4,442 passed, 4 skipped, 15 xfailed**; `validate` 0 errors / 0 warnings, code coverage 216/216; `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` all thresholds met; `vulture-ci` clean; `pylint` clean.

## Separate finding, surfaced while verifying yours

Checking whether the gate actually validates my two new instruction declarations, I found that the CDSL marker parser **silently truncates an instruction slug at the first character outside `[a-z0-9-]`** instead of rejecting it. Measured, changing only the code-side marker:

| Code marker | `cfs validate` |
|---|---|
| `inst-count-registered-entries` (matches the artifact) | 0 errors |
| `inst-zzz-unrelated` | 2 errors — `code-inst-missing` + `code-inst-orphan` |
| `inst-count-registered-entrie` (truncated) | 2 errors |
| `inst-count-registered-entries-x` | 2 errors |
| **`inst-count-registered-entriesX`** | **0 errors** |

`codebase.py:31` and `:36` end in `:inst-(?P<inst>[a-z0-9-]+)` with no terminator, so the match succeeds and the trailing `X` is discarded — the marker traces as a *different* instruction than the one written in the file. So `inst-fooBar` silently traces as `inst-foo`, `inst-my_thing` as `inst-my`, and two markers `inst-fooX`/`inst-fooY` collapse onto the same key. The `(?P<id>cpt-[a-z0-9][a-z0-9-]+)` group has the same shape.

Entirely pre-existing and unrelated to this PR's diff, so I haven't touched it. Both directions of instruction checking do work correctly otherwise, which I verified separately — so my new declarations *are* gate-enforced. Want me to file this as its own issue?

skills/studio/scripts/studio/commands/spec_coverage.py
Oleg67: Fair, and the underlying reason the styles differed is that mine accumulated a scalar — you can't mutate an `int` through a closure, which is why it needed `nonlocal` where `_collect_system_slugs` and `_collect_selected_system_files` just mutate their captured `set`/`list`.

Rather than allocate a container to sum, I've dropped the accumulator entirely so both helpers return values:

```python
    def visit(node: object) -> int:
        """Count ``node``'s subtree once it is in scope, mirroring file collection."""
        if system_slugs is None or getattr(node, "slug", "") in system_slugs:
            return count_subtree(node)
        return sum(visit(child) for child in getattr(node, "children", []))

    return sum(visit(system_node) for system_node in meta.systems)
```

No `nonlocal`, and shorter than the version you commented on. It does still read differently from the two siblings, but for a counting walk I think returning is clearer than seeding a mutable box — happy to match the sibling shape instead if you'd rather the file stay uniform.

tests/test_enforcement_empty_scan.py
Oleg67: Agreed on both halves — this was my gap, and it hid a real bug.

All four flags are now exercised over an empty scope, and the failure *text* is asserted rather than just `code != 0`:

```python
        for flag, value in (
            ("--min-coverage", "90"),
            ("--min-file-coverage", "60"),
            ("--min-granularity", "0.46"),
            ("--min-file-granularity", "0.3"),
        ):
            ...
            assert code == 2, flag
            assert report["status"] == "FAIL", flag
            assert report["applicable"] is False, flag
            assert report["threshold_failures"] == [
                f"cannot assess {flag}: 0 files from 0 registered codebase entries"
            ], flag
```

Widening it to all four is also what exposed the false positive described in the summary comment: every flag was failing an empty scope even at a threshold of `0`, which any scope satisfies. A per-flag test would have caught that from the start.

There is now a companion test pinning the non-positive case in both directions, across all four flags.

architecture/features/spec-coverage.md
```diff
@@ -90,6 +90,8 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - [x] - `p1` - Validate selected `--system` values and build unknown-system failure payloads - `inst-validate-systems`
 - [x] - `p1` - Filter ignored and out-of-root files before scanning - `inst-filter-ignored-files`
 - [x] - `p1` - Build empty coverage result when registry yields no code files - `inst-empty-report`
+- [x] - `p1` - Count codebase entries registered by the selected systems, so an unregistered registry is distinguishable from one that resolves to no files - `inst-count-registered-entries`
+- [x] - `p1` - Detect which threshold flags demand an enforceable guarantee, ignoring non-positive values that any scope satisfies - `inst-detect-requested-thresholds`
```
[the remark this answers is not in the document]
Oleg67: Valid — the spec sentence didn't state the scope even though the code walks the subtree. Fixed in `18d7817e` by matching the wording already used two lines above for `inst-collect-codebase-files` ("and recurse into child systems"), since both walk the same tree:

> Count codebase entries registered by the selected systems **and recurse into child systems, including the default all-systems selection**, so an unregistered registry is distinguishable from one that resolves to no files

For the record, the code was already correct here — the recursion bug you flagged on the previous head was fixed in `c770b228`, and the default `system_slugs is None` case was fixed at the same time (it had the same defect, which the original report didn't cover). Two tests pin both paths and go red if either regresses. This change is the spec catching up to that, which is worth having: the whole point of the document is that it describes the code.

architecture/features/spec-coverage.md
```diff
@@ -215,7 +217,7 @@ The system **MUST** calculate instruction density per covered file: `min(1.0, bl
 
 - [x] `p1` - **ID**: `cpt-studio-dod-spec-coverage-report`
 
-The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. Exit code 0 when above thresholds, 2 when below.
+The system **MUST** output a JSON report with: summary (total files, covered files, coverage %, granularity score), per-file statistics (path, total lines, covered lines, coverage %, granularity, uncovered line ranges), and list of completely uncovered files. The report format **MUST** mirror `coverage.py` JSON output structure. The report **MUST** state whether there was anything to assess, so a scope that yielded no code files is distinguishable from a fully covered one. Exit code 0 when above thresholds, 2 when below or when a positive threshold was demanded over a scope that could not be assessed.
```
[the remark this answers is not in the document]
Oleg67: Correct at the boundary. `_check_min_coverage` and its three siblings return "no failure" when `metric >= threshold`, so a metric exactly equal to its threshold passes. Changed to "at or above thresholds" in `18d7817e`.

architecture/features/spec-coverage.md
```diff
@@ -251,4 +253,6 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
 - [x] `--verbose` flag includes per-file marker details in report
 - [x] Report format mirrors `coverage.py` JSON structure (summary + per-file)
 - [x] Scanning completes in ≤ 5 seconds for typical repositories
+- [x] An empty scope reports `applicable: false` and says why, on both the JSON and human surfaces
```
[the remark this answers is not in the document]
Oleg67: Agreed, and this contradiction is one I introduced — the criterion I added references the human surface, sitting directly above a pre-existing one asserting all output is JSON. Both can't be true, and the JSON-only claim was already inaccurate since human mode is the *default* (JSON requires the global `--json` flag).

Fixed in `18d7817e`:

> JSON mode emits valid JSON to stdout and human mode emits formatted text, with exit codes 0/1/2 on both

architecture/features/spec-coverage.md
```diff
@@ -251,4 +253,6 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
...
+- [x] A positive `--min-*` threshold over an empty scope exits 2; a non-positive one exits 0
```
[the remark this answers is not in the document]
Oleg67: This is the most useful of the five, because it isn't a wording slip — it's a spec/code divergence that predates this PR.

The Error Scenarios section said:

> - No codebase entries registered → ERROR with hint to configure artifacts.toml

The command has **never** done that. Before this branch it returned `{"status": "PASS", ..., "message": "No codebase files found in registry"}` and exit 0; it now returns `PASS` with `applicable: false`, or exit 2 when a positive threshold was demanded. So the document has been promising an error path that no version of the code implemented — which is a small instance of exactly the problem this PR is about: a spec or gate asserting a guarantee that nothing backs.

Corrected in `18d7817e` to the contract that actually ships, and the neighbouring scenario was updated too since it describes the same code path:

> - No codebase entries registered → report with `applicable: false` and a hint to configure artifacts.toml; exit code 2 only if a positive threshold was demanded
> - No code files found → report with `applicable: false` naming how many registered entries resolved to no files, and 0% coverage

I've kept the hint about configuring `artifacts.toml`, since that part of the original scenario was good advice and is still what a user needs to do.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -41,25 +41,26 @@ def _build_spec_coverage_parser() -> argparse.ArgumentParser:
         "--min-coverage",
         type=float,
         default=None,
-        help="Minimum coverage percentage (0-100). Exit 2 if below.",
+        help="Minimum coverage percentage (0-100). Exit 2 if below, or if nothing was assessed.",
     )
```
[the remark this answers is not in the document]
Oleg67: Right, and this was a stale-by-one-commit problem: I wrote that help text before restricting the trigger to positive thresholds, then didn't revisit it. All four now read, e.g.:

```
Minimum coverage percentage (0-100). Exit 2 if below; a positive value also
exits 2 if nothing was assessed.
```

This matches both the code (`_requested_thresholds` skips non-positive values) and the tests that pin `--min-coverage 0` and `-5` to exit 0 on an empty scope. Fixed in `18d7817e`.

tests/test_enforcement_empty_scan.py
```diff
@@ -0,0 +1,428 @@
...
+                ctx = _context(Path(directory), codebase=[])
+
+                code, report = _run_spec_coverage(ctx, [flag, value])
+
+            assert code == 2, flag
+            assert report["status"] == "FAIL", flag
+            assert report["applicable"] is False, flag
+            assert report["threshold_failures"] == [
+                f"cannot assess {flag}: 0 files from 0 registered codebase entries"
+            ], flag
```
Oleg67, line 276: Good catch, and the reasoning holds up under test. Added in `d1eb7fe3`.

The claim that the per-flag cases couldn't catch this is verifiable, so I checked it: truncating the failure list to its first entry (`for flag in requested_thresholds[:1]`) fails **only** the new combined test — all four per-flag cases stay green. So the gap was real, not hypothetical.

```python
        assert report["threshold_failures"] == [
            f"cannot assess {flag}: 0 files from 0 registered codebase entries"
            for flag in flags[::2]
        ]
```

Suite is 4,443 passed / 4 skipped / 15 xfailed on Python 3.11.

skills/studio/scripts/studio/commands/spec_coverage.py
Oleg67: fix in a new commit

tests/test_enforcement_empty_scan.py
Oleg67: fix in a new commit

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -452,7 +531,12 @@ def _human_spec_coverage(data: dict) -> None:
         ui.blank()
         _show_spec_coverage_files(files)
 
+    applicable = data.get("applicable", True)
```
ainetx: ### Human-readable output prints misleading zero-coverage metrics before the empty-scope disclaimer

**Severity:** Minor

**Problem**
`_human_spec_coverage` unconditionally renders the Files/Coverage/Granularity summary lines (522-524) before checking `applicable` (533), which gates the disclaiming warning (535-536). The disclaimer this PR added always appears after, never instead of, the raw zero metrics.

**How to reproduce**
1. Run `cfs spec-coverage` (no `--json`) against a scope resolving to zero files.
2. Observe "Files: 0/0 covered", "Coverage: 0.0%", "Granularity: 0.0000" printed first.
3. Observe the disclaimer only appears afterward, after a blank line.

**Expected behavior**
When `applicable is False`, the disclaimer should print before (or instead of) the generic zero-metric summary.

**Actual behavior**
The generic numeric summary always prints first; the disclaimer is a secondary afterthought.

```text
_human_spec_coverage()
  +-- Files/Coverage/Granularity lines   <- always printed first
  +-- disclaimer (if applicable=False)   <- printed second
```

**Impact**
A user skimming output may see "Coverage: 0.0%" and conclude a real failure occurred before reaching the clarifying note — undermining this PR's stated goal.

**Suggested correction**
Check `applicable` before rendering the summary lines; skip/replace them with the disclaimer when `applicable is False`, or print the disclaimer first.

**How to verify**
Extend human-output tests to assert the disclaimer appears before any Files/Coverage/Granularity line in the non-applicable case.

Oleg67: Confirmed and fixed in `31857fea`. You're right, and I'd seen this output myself without questioning it — the disclaimer was an afterthought in the literal sense of being printed after.

Before:

```
  Spec Coverage
    Files: 0/0 covered
    Coverage: 0.0%
    Granularity: 0.0000

  ⚠ No codebase entries are registered, so no code files were scanned
```

After:

```
  Spec Coverage
  ⚠ No codebase entries are registered, so no code files were scanned

    Files: 0/0 covered
    Coverage: 0.0%
    Granularity: 0.0000
```

**I took "print the disclaimer first" rather than "skip the summary lines".** Both were on the table in your suggestion, and the deciding argument is that a verdict should still state what it measured — `0 of 0 files` is a materially different statement from `63 of 261 lines`, and suppressing the metrics would drop the denominator instead of explaining it. Printed after the note, the zeroes read as the denominator they are.

Verified on all three empty-scope paths, including the threshold case, where the ordering is now: what wasn't assessed → the zeroes → which guarantee couldn't be given → the verdict.

Covered by the ordering assertion you asked for. One wrinkle worth recording, because it would bite anyone extending these tests: the assertion has to anchor on the metric *values*, not their labels. `out.index("Coverage")` matches the `Spec Coverage` header, which legitimately precedes the disclaimer, so a label-based assertion fails against correct output. It now compares against `0/0 covered`, `0.0%` and `0.0000`.

Gates on the new head, Python 3.11: full suite **4,444 passed, 4 skipped, 15 xfailed**; `validate` 0 errors / 0 warnings, 216/216; `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` all thresholds met; `pylint` and `vulture` clean. Mutation-checked: moving the disclaimer back below the metrics fails the new test and nothing else.

Separately — the follow-up for `validate` is now up as #95, which applies the same separation to the second command: an empty scan there reached no checks at all, so a checked `to_code="true"` ID with no code validated clean. Independent of this PR, different file.

PR #94, opened 2026-08-19
feat(eval): add a deterministic structural compliance scorer

SanjeevSolanki:
## What
Adds the real deterministic structural scorer to the workflow eval-harness, plugged into the `Scorer` seam from #92. `cfs eval --check` now gates on genuine structural compliance instead of the placeholder reference scorer.

## How it works
`StructuralScorer` (kind = DETERMINISTIC) reads a completed run's `plan.toml` manifest and each phase file's `[phase]` frontmatter, then runs a **registry** of independent structural checks: phase numbering (unique, contiguous from 1), manifest agreement (declared totals/entries/files vs. the phase files), phase-total consistency, dependency order (resolve, not forward), declared outputs, and required body sections (per-workflow-configurable).

It returns a compliance %, per-check findings, and the harness's UNKNOWN discipline — a run that can't be loaded or carries no parseable frontmatter scores `UNKNOWN`, never a false 0% ("unscoreable is not zero").

## Notes
- Pure function of the in-memory run artifacts — **no filesystem access, no model call**.
- Checks are tagged callables, skippable by name or tag; the reference scorer stays as the seam's minimal example / gate-contract test fixture.
- CPT-traced; new files at 100% coverage; pylint / vulture / validate / spec-coverage green.

Follows #92; part of the eval-harness track under #72.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic structural evaluation for workflow runs.
  * Validates phase metadata, numbering, dependencies, declared inputs and outputs, required sections, and manifest consistency.
  * Reports PASS, FAIL, or UNKNOWN outcomes with compliance percentages, findings, and coverage details.
  * Supports workflow-specific requirements and selectively skipping checks by name or category.
  * Identifies malformed or incomplete run data that cannot be scored.
  * Updated evaluations to use structural scoring.

* **Documentation**
  * Updated evaluation harness documentation to explain structural scoring, result interpretation, and acceptance criteria.

### DIFF
--- a/skills/studio/scripts/studio/utils/eval_structural.py
+++ b/skills/studio/scripts/studio/utils/eval_structural.py
@@ -1,0 +1,506 @@
+"""Deterministic structural scorer for the workflow eval-harness.
+
+The real ``DETERMINISTIC`` scorer that plugs into the ``Scorer`` seam the scaffold
+(:mod:`studio.utils.eval_harness`) defines. It reads a completed run's ``plan.toml``
+manifest and the ``[phase]`` frontmatter carried by each ``phase-*.md`` file, runs a
+**registry** of independent structural checks (numbering, manifest agreement, dependency
+order, declared outputs, required sections), and returns a ``ScorerResult`` with a
+compliance %, per-check findings, and the scaffold's UNKNOWN discipline.
+
+Design commitments, inherited from the scaffold's gate contract: **unscoreable is not
+zero** (a run that will not load, or whose phases carry no parseable ``[phase]``
+frontmatter, scores ``UNKNOWN`` with a ``None`` score, never a false 0%); and **the
+registry is the unit of extension** — each check is a small tagged callable, skippable by
+name or tag and testable in isolation, rather than one hardcoded function. The scorer is a
+pure function of the in-memory run artifacts: no filesystem, no model call, no way to add
+either to a score.
+
+@cpt-algo:cpt-studio-algo-eval-structural:p1
+"""
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-imports
+from __future__ import annotations
+
+import logging
+import re
+import tomllib
+from dataclasses import dataclass
+from typing import Callable, Dict, List, Optional, Set, Tuple
+
+from .eval_harness import (RunArtifacts, Scenario, ScorerKind, ScorerResult,
+                           VERDICT_FAIL, VERDICT_PASS, VERDICT_UNKNOWN)
+
+logger = logging.getLogger(__name__)
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-imports
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-config
+#: The fenced TOML block at the head of a phase file, carrying its ``[phase]`` table.
+#: Only *horizontal* whitespace is allowed after the opening fence (``[ \t]`` not ``\s``):
+#: letting ``\s`` swallow newlines makes an unterminated fence backtrack quadratically.
+_FRONTMATTER_RE = re.compile(r"^```toml[ \t]*\r?\n(?P<body>.*?)\r?\n```", re.DOTALL)
+
+#: Body sections a self-contained phase file is expected to carry, by default. Measured
+#: against the real plans, not invented: the common shape is ``## Preamble`` / ``## What`` /
+#: ``## Rules``. A workflow whose phases legitimately differ overrides this below.
+DEFAULT_SECTIONS: Tuple[str, ...] = ("Preamble", "What", "Rules")
+
+#: Per-workflow required-section overrides. A verification-style workflow legitimately
+#: carries no ``## Rules`` block, so scoring it against the universal set would be a
+#: false-fail; a workflow absent from this map uses ``DEFAULT_SECTIONS``.
+WORKFLOW_SECTIONS: Dict[str, Tuple[str, ...]] = {
+    "verify": ("Preamble", "What"),
+    "verification": ("Preamble", "What"),
+}
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-config
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-datamodel
+@dataclass
+class StructuralInput:
+    """Everything the checks read: the manifest plus the parsed phase frontmatter."""
+
+    plan_meta: Dict[str, object]                    # plan.toml [plan] table
+    manifest_phases: List[Dict[str, object]]        # plan.toml [[phases]] entries
+    phases: Dict[int, Dict[str, object]]            # phase number -> parsed [phase] frontmatter
+    bodies: Dict[int, str]                          # phase number -> raw phase-file body
+    by_file: Dict[str, int]                         # phase filename -> the number it declares
+    duplicates: List[str]                           # files that re-declare an already-seen number
+    invalid: List[str]                              # files WITH a [phase] block that is broken
+    no_frontmatter: List[str]                       # files with no [phase] block at all
+    required_sections: Tuple[str, ...]              # sections this workflow's phases must carry
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-datamodel
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-parse
+def _as_phase_int(value: object) -> Optional[int]:
+    """A phase number/total as a plain ``int``, or ``None`` when it is not one.
+
+    TOML already types its scalars, so this is a *type check*, not a parse — there is no
+    exception path to swallow. ``bool`` is excluded on purpose (it is an ``int`` subclass,
+    but a flag is never a phase number).
+    """
+    if isinstance(value, bool) or not isinstance(value, int):
+        return None
+    return value
+
+
+#: Leading bytes tolerated before the fence: a UTF-8 BOM and any blank lines / indentation.
+#: Editors and templating commonly prepend these, and rejecting them would look like a
+#: "missing frontmatter" false-negative rather than the near-miss it is.
+_LEADING_NOISE = "﻿ \t\r\n"
+
+
+def _match_frontmatter(text: str):
+    """Match the ``[phase]`` fence, tolerating a leading BOM / blank lines. Returns the match
+    and the normalised text it was matched against (so callers can slice by ``.end()``)."""
+    normalised = text.lstrip(_LEADING_NOISE)
+    return _FRONTMATTER_RE.match(normalised), normalised
+
+
+def _parse_phase_table(body: str) -> Optional[Dict[str, object]]:
+    """Parse an already-extracted fenced ``[phase]`` TOML *body* into its table, or ``None``
+    when the TOML is malformed (surfaced as a warning, not swallowed) or carries no ``[phase]``.
+    """
+    try:
+        data = tomllib.loads(body)
+    except tomllib.TOMLDecodeError as exc:
+        logger.warning("eval: phase file has a malformed [phase] frontmatter block: %s", exc)
+        return None
+    phase = data.get("phase")
+    return phase if isinstance(phase, dict) else None
+
+
+def _collect_phases(phase_texts: Dict[str, str]) -> Tuple[Dict[int, Dict[str, object]],
+                                                          Dict[int, str], Dict[str, int],
+                                                          List[str], List[str], List[str]]:
+    """Parse phase bodies into ``(by_number, bodies, by_file, duplicates, invalid, no_block)``.
+
+    Keying phases by their declared ``number`` (not by filename) is deliberate: it lets the
+    checks reason about ordering and dependencies. ``by_file`` retains the filename→number
+    pairing so a check can verify the manifest maps each number to the right file. A file that
+    cannot contribute a phase is *retained as a finding*, never silently dropped, and the two
+    failure modes are kept apart: a file that **has** a ``[phase]`` block but is broken
+    (unparseable TOML, or a bad number) goes to ``invalid`` — that is a *failing* run; a file
+    with **no** fenced block at all goes to ``no_block`` — that may just be a different workflow
+    shape. A second file re-declaring a number is recorded in ``duplicates``.
+    """
+    phases: Dict[int, Dict[str, object]] = {}
+    bodies: Dict[int, str] = {}
+    by_file: Dict[str, int] = {}
+    duplicates: List[str] = []
+    invalid: List[str] = []
+    no_block: List[str] = []
+    for name in sorted(phase_texts):
+        text = phase_texts[name]
+        match, _ = _match_frontmatter(text)
+        if match is None:
+            no_block.append(f"{name}: no [phase] frontmatter block")
+            continue
+        front = _parse_phase_table(match.group("body"))
+        if front is None:
+            invalid.append(f"{name}: [phase] frontmatter present but unparseable")
+            continue
+        number = _as_phase_int(front.get("number"))
+        if not number or number <= 0:
+            invalid.append(f"{name}: missing or non-positive phase number")
+            continue
+        by_file[name] = number
+        if number in phases:
+            duplicates.append(f"{name} re-declares phase {number}")
+            continue
+        phases[number] = front
+        bodies[number] = text
+    return phases, bodies, by_file, duplicates, invalid, no_block
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-parse
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-check-helpers
+def _capped(items: List[str], cap: int = 3) -> str:
+    """Join the first ``cap`` items, appending ``(+N more)`` so a truncated detail never hides
+    the true count from a reader."""
+    shown = "; ".join(items[:cap])
+    extra = len(items) - cap
+    return f"{shown} (+{extra} more)" if extra > 0 else shown
+
+
+def _manifest_numbers(inp: StructuralInput) -> List[int]:
+    """Every positive phase number the manifest lists, **duplicates preserved** (order kept)."""
+    numbers: List[int] = []
+    for entry in inp.manifest_phases:
+        number = _as_phase_int(entry.get("number"))
+        if number and number > 0:
+            numbers.append(number)
+    return numbers
+
+
+def _manifest_declared(inp: StructuralInput) -> Set[int]:
+    """The set of positive phase numbers the ``plan.toml`` manifest declares."""
+    return set(_manifest_numbers(inp))
+
+
+def _phase_totals(inp: StructuralInput) -> Set[int]:
+    """The distinct ``total`` values the phase frontmatter declares.
+
+    A missing or non-integer ``total`` counts as ``0`` (not dropped), so an absent total
+    lands in the set and trips ``phase-total-consistent`` rather than vanishing.
+    """
+    totals: Set[int] = set()
+    for front in inp.phases.values():
+        total = _as_phase_int(front.get("total"))
+        totals.add(total if total is not None else 0)
+    return totals
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-check-helpers
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-deps
+#: Keys that normalise to a plausible misspelling of ``depends_on`` — flagged so a typo
+#: (e.g. ``dependsOn``) is reported, not silently read as "no dependencies".
+_DEP_KEY_ALIASES = frozenset({"dependson", "dependencies", "depends", "dependency",
+                              "dependsupon", "dependon", "dependenton", "dependings"})
+
+
+def _misspelled_dep_key(front: Dict[str, object]) -> Optional[str]:
+    """A frontmatter key that looks like a misspelled ``depends_on`` (when the real key is
+    absent), or ``None``. Normalises case/underscores/hyphens before comparing."""
+    if "depends_on" in front:
+        return None
+    for key in front:
+        if isinstance(key, str) and key.lower().replace("_", "").replace("-", "") in _DEP_KEY_ALIASES:
+            return key
+    return None
+
+
+def _dependency_problems(inp: StructuralInput) -> Tuple[List[str], List[str]]:
+    """Return ``(unresolved, forward)`` dependency problems across all phases."""
+    unresolved: List[str] = []
+    forward: List[str] = []
+    for number, front in sorted(inp.phases.items()):
+        misspelled = _misspelled_dep_key(front)
+        if misspelled is not None:
+            unresolved.append(f"phase {number}: '{misspelled}' looks like a misspelled depends_on")
+        deps = front.get("depends_on")
+        if deps is None:
+            deps = []
+        elif not isinstance(deps, list):
+            # A present-but-non-list depends_on is a malformed declaration, not "no deps" —
+            # flag it rather than silently treating it as empty (which would vacuously pass).
+            unresolved.append(f"phase {number}: depends_on must be an array")
+            continue
+        for dep in deps:
+            dep_no = _as_phase_int(dep)
+            if dep_no is None:
+                unresolved.append(f"phase {number} -> {dep!r}")
+            elif dep_no not in inp.phases:
+                unresolved.append(f"phase {number} -> {dep_no} (missing)")
+            elif dep_no >= number:
+                forward.append(f"phase {number} -> {dep_no}")
+    return unresolved, forward
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-deps
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-checks
+def _check_phase_frontmatter_valid(inp: StructuralInput) -> Tuple[bool, str]:
+    # Once a run is scoreable (some phase parsed), every other declared phase file must too —
+    # whether it has a broken block (``invalid``) or none at all (``no_frontmatter``), it is a
+    # broken phase and fails here, so it can never be silently dropped into a vacuous pass.
+    problems = inp.invalid + inp.no_frontmatter
+    return not problems, _capped(problems)
+
+
+def _check_phase_numbers_unique(inp: StructuralInput) -> Tuple[bool, str]:
+    return not inp.duplicates, _capped(inp.duplicates)
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-checks
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-checks-manifest
+def _check_manifest_numbers_unique(inp: StructuralInput) -> Tuple[bool, str]:
+    numbers = _manifest_numbers(inp)
+    dupes = sorted({n for n in numbers if numbers.count(n) > 1})
+    return not dupes, f"manifest re-declares phase(s): {dupes}" if dupes else ""
+
+
+def _check_manifest_entries_valid(inp: StructuralInput) -> Tuple[bool, str]:
+    # A [[phases]] entry with a missing or non-positive number is malformed. Other manifest
+    # checks quietly drop such entries (they reduce to the set of *valid* numbers), so without
+    # this a broken manifest entry would never surface — flag each one explicitly.
+    bad = []
+    for index, entry in enumerate(inp.manifest_phases):
+        number = _as_phase_int(entry.get("number"))
+        if number is None or number <= 0:
+            bad.append(f"entry {index}: number={entry.get('number')!r}")
+    return not bad, _capped(bad)
+
+
+def _check_manifest_present(inp: StructuralInput) -> Tuple[bool, str]:
+    # The scaffold already turns an unreadable/`[plan]`-less plan.toml into UNKNOWN before a
+    # scorer runs, so this checks the *other* half — that the [[phases]] manifest actually
+    # lists the phases the files carry. It fails when a plan.toml has a [plan] table but no
+    # numbered [[phases]] entries, a real and distinct defect (not a vacuous always-pass).
+    ok = bool(_manifest_declared(inp))
+    return ok, "" if ok else "plan.toml [[phases]] declares no numbered phase"
+
+
+def _check_manifest_total_matches_entries(inp: StructuralInput) -> Tuple[bool, str]:
+    declared = _manifest_declared(inp)
+    total_declared = _as_phase_int(inp.plan_meta.get("total_phases")) or 0
+    ok = total_declared == len(declared)
+    return ok, f"total_phases={total_declared}, entries={len(declared)}"
+
+
+def _check_manifest_matches_files(inp: StructuralInput) -> Tuple[bool, str]:
+    # Two things must agree, not just the number *set*: (1) the declared and found number sets
+    # match, and (2) each manifest entry's number->file pairing matches the file's own declared
+    # number — a manifest that maps phase 1 to phase-2.md must not pass on set equality alone.
+    problems: List[str] = []
+    declared = _manifest_declared(inp)
+    found = set(inp.phases)
+    if declared != found:
+        problems.append(f"declared={sorted(declared)}, found={sorted(found)}")
+    for entry in inp.manifest_phases:
+        number = _as_phase_int(entry.get("number"))
+        file = entry.get("file")
+        if not isinstance(file, str) or number is None:
+            continue
+        actual = inp.by_file.get(file)
+        if actual is None:
+            problems.append(f"{file} (manifest phase {number}) has no valid phase")
+        elif actual != number:
+            problems.append(f"manifest pairs {file}->{number} but the file declares {actual}")
+    return not problems, _capped(problems, 3)
+
+
+def _check_numbering_contiguous(inp: StructuralInput) -> Tuple[bool, str]:
+    numbers = sorted(inp.phases)
+    ok = numbers == list(range(1, len(numbers) + 1))
+    return ok, f"numbers={numbers}"
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-checks-manifest
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-checks-phase
+def _check_phase_total_consistent(inp: StructuralInput) -> Tuple[bool, str]:
+    totals = _phase_totals(inp)
+    ok = len(totals) == 1 and totals != {0}
+    return ok, f"declared totals={sorted(totals)}"
+
+
+def _check_phase_total_matches_count(inp: StructuralInput) -> Tuple[bool, str]:
+    totals = _phase_totals(inp)
+    count = len(inp.phases)
+    ok = totals == {count}
+    return ok, f"total={sorted(totals)}, count={count}"
+
+
+def _check_dependencies_resolve(inp: StructuralInput) -> Tuple[bool, str]:
+    unresolved, _ = _dependency_problems(inp)
+    return not unresolved, _capped(unresolved)
+
+
+def _check_dependencies_not_forward(inp: StructuralInput) -> Tuple[bool, str]:
+    _, forward = _dependency_problems(inp)
+    return not forward, _capped(forward)
+
+
+def _check_every_phase_declares_output(inp: StructuralInput) -> Tuple[bool, str]:
+    missing = [str(n) for n, front in sorted(inp.phases.items())
+               if not (front.get("outputs") or front.get("output_files"))]
+    return not missing, (f"phases without outputs: {_capped(missing, 5)}" if missing else "")
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-checks-phase
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-prose
+def _prose_headings(raw_body: str) -> str:
+    """The phase body reduced to its real Markdown prose — the leading ``[phase]`` frontmatter
+    and every fenced code block removed — so a ``## Section``-looking line inside TOML or a
+    code sample is not mistaken for a real heading (which would inflate compliance).
+
+    A simple line-based fence toggle, not a Markdown parser: no regex over the whole body, so
+    no backtracking, and it covers the fenced ` ``` ` blocks phase files actually use.
+    """
+    front, normalised = _match_frontmatter(raw_body)
+    body = normalised[front.end():] if front else normalised
+    visible: List[str] = []
+    in_fence = False
+    for line in body.splitlines():
+        if line.lstrip().startswith("```"):
+            in_fence = not in_fence
+            continue
+        if not in_fence:
+            visible.append(line)
+    return "\n".join(visible)
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-prose
+
+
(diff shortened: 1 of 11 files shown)


skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,346 @@
...
+from .eval_harness import (RunArtifacts, Scenario, ScorerKind, ScorerResult,
+                           VERDICT_FAIL, VERDICT_PASS, VERDICT_UNKNOWN)
+
+logger = logging.getLogger(__name__)
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-imports
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-config
+#: The fenced TOML block at the head of a phase file, carrying its ``[phase]`` table.
+_FRONTMATTER_RE = re.compile(r"^```toml\s*\n(?P<body>.*?)\n```", re.DOTALL)
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `5c820f9`. Switched `_FRONTMATTER_RE` to `` ^```toml[ \t]*\r?\n(?P<body>.*?)\r?\n``` `` — only horizontal whitespace after the opening fence (plus CRLF tolerance), so an unterminated fence can no longer backtrack quadratically. This also clears the matching SonarCloud super-linear-regex warning on the same line. Thanks for the catch.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,346 @@
...
+        if front is None:
+            continue
+        number = _as_phase_int(front.get("number"))
+        if not number or number <= 0:
+            continue
+        if number in phases:
+            duplicates.append(f"{name} re-declares phase {number}")
+            continue
+        phases[number] = front
+        bodies[number] = phase_texts[name]
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `5c820f9`. `_collect_phases` now **retains** every declared phase file rather than silently dropping it: a present `phase-*.md` that fails to parse or omits a valid number is recorded and fails a new `phase-frontmatter-valid` check, and duplicate manifest declarations fail a new `manifest-numbers-unique` check (both added to the registry). A structurally broken run can no longer earn a vacuous pass — added a test for the precise case you described (a valid phase alongside an invalid, un-manifested phase file → FAIL). Good find.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,375 @@
...
+        deps = deps if isinstance(deps, list) else []
+        for dep in deps:
+            dep_no = _as_phase_int(dep)
+            if dep_no is None:
+                unresolved.append(f"phase {number} -> {dep!r}")
+            elif dep_no not in inp.phases:
+                unresolved.append(f"phase {number} -> {dep_no} (missing)")
+            elif dep_no >= number:
+                forward.append(f"phase {number} -> {dep_no}")
+    return unresolved, forward
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `a4f2dd0`. `_dependency_problems` now separates the three cases: an absent `depends_on` is treated as no dependencies (fine), while a present **non-list** value is flagged as an unresolved declaration (`phase N: depends_on must be an array`) and fails `dependencies-resolve`, instead of being silently coerced to empty. Added a scalar-`depends_on` test plus one confirming an absent `depends_on` still passes. Same silent-coercion pattern as the previous two — good to have it consistent. Thanks.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,381 @@
...
+               if not (front.get("outputs") or front.get("output_files"))]
+    return not missing, f"phases without outputs: {', '.join(missing[:5])}"
+
+
+def _check_required_sections(inp: StructuralInput) -> Tuple[bool, str]:
+    missing: List[str] = []
+    for number in sorted(inp.phases):
+        body = inp.bodies.get(number, "")
+        for section in inp.required_sections:
+            if not re.search(rf"^##\s+{re.escape(section)}\b", body, re.MULTILINE):
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `39f3658`. `_check_required_sections` now strips the `[phase]` frontmatter (via `_FRONTMATTER_RE`) before scanning for `## Section` headings, so a heading-like line inside a TOML multiline string can no longer satisfy the check and inflate compliance. Added a regression test — a phase carrying `## Rules` only inside its frontmatter (not the Markdown body) now correctly fails `required-sections-present`. Thanks.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,385 @@
...
+    for number in sorted(inp.phases):
+        raw_body = inp.bodies.get(number, "")
+        # Scan only the Markdown body, not the [phase] frontmatter: a heading-like line inside
+        # a TOML multiline string must not count as a real `## Section` and inflate compliance.
+        front = _FRONTMATTER_RE.match(raw_body)
+        body = raw_body[front.end():] if front else raw_body
+        for section in inp.required_sections:
+            if not re.search(rf"^##\s+{re.escape(section)}\b", body, re.MULTILINE):
+                missing.append(f"phase {number}: {section}")
+    return not missing, "; ".join(missing[:3])
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `e2238fd`. Added a `_prose_headings` step that reduces the phase body to real Markdown prose — the `[phase]` frontmatter and any fenced ` ``` ` code blocks removed — before the `## Section` scan, so a heading-like line inside a code sample no longer counts toward compliance. It's a small line-based fence toggle (deliberately not a Markdown parser, and no regex over the whole body → no backtracking), which covers the fences phase files actually use. Added a regression test: `## Rules` inside a code fence with no real heading now correctly FAILs `required-sections-present`. (Re the ast-grep ReDoS note in this thread: `section` is a hardcoded constant from `DEFAULT_SECTIONS`/`WORKFLOW_SECTIONS`, already `re.escape`'d — not user input, so no injection surface.) Thanks.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+
+#: Body sections a self-contained phase file is expected to carry, by default. Measured
+#: against the real plans, not invented: the common shape is ``## Preamble`` / ``## What`` /
+#: ``## Rules``. A workflow whose phases legitimately differ overrides this below.
+DEFAULT_SECTIONS: Tuple[str, ...] = ("Preamble", "What", "Rules")
+
+#: Per-workflow required-section overrides. A verification-style workflow legitimately
+#: carries no ``## Rules`` block, so scoring it against the universal set would be a
+#: false-fail; a workflow absent from this map uses ``DEFAULT_SECTIONS``.
+WORKFLOW_SECTIONS: Dict[str, Tuple[str, ...]] = {
```
ainetx, line 50: ### Unrecognized workflow names silently fall back to default section requirements with no diagnostic

**Severity:** Minor

**Problem**
The structural scorer looks up per-workflow required sections with `WORKFLOW_SECTIONS.get(scenario.workflow, DEFAULT_SECTIONS)`. Any workflow name that isn't an exact key in that table — including a typo of a name that legitimately needs a reduced section set — silently receives the stricter default policy, and nothing in the tool's output distinguishes "this workflow has no explicit policy" from "this phase file is genuinely missing a section."

**How to reproduce**
1. Create a scenario whose workflow is `"verification"` (the second alias declared in the table) or a typo such as `"verifcation"`.
2. Give it phase files that only contain `## Preamble` and `## What` (intentionally omitting `## Rules`, matching the reduced policy meant for verification-style workflows).
3. Run the structural scorer against this scenario.

**Expected behavior**
For a recognized alias like `"verification"`, the reduced section set should be applied and covered by a regression test proving it. For any workflow name that has no explicit policy, the tool should surface a diagnostic (e.g. a log line) so an author can tell the difference between "policy gap" and "actual missing content."

**Actual behavior**
```text
scenario.workflow = "verification"  --> covered by NO test today
scenario.workflow = "verifcation"   --> silently uses DEFAULT_SECTIONS (requires "Rules")
                                         --> false FAIL, no hint that the workflow name wasn't recognized
```

**Impact**
A misspelled or new workflow name produces a false "required section missing" failure that looks identical to a real content defect, costing debugging time. Separately, the `"verification"` alias currently has zero test coverage, so a future refactor could silently break it without any test failing.

**Suggested correction**
Add a log line (e.g. `logger.debug`) when `scenario.workflow` is not a key in `WORKFLOW_SECTIONS`, so the fallback is visible in output. Add a unit test scoring a scenario with `workflow="verification"`, mirroring the existing `"verify"` test.

**How to verify**
Add the missing test and confirm it passes; run the scorer against an unrecognized workflow name and confirm the new log line appears, distinguishing a policy gap from a real structural defect.

ainetx, line 50: ### Per-workflow required-section policy is hardcoded in source with no project-level override

**Severity:** Minor

**Problem**
The set of Markdown sections a phase file must carry, per workflow, is compiled directly into the scorer's Python source as a small dict with two entries. There is no way for a project to add or adjust an entry for a custom workflow without editing and re-shipping the tool's code, even though the project already has an established pattern for exactly this kind of policy — a name-keyed configuration table resolved from a project-editable file with a documented override order.

**How to reproduce**
1. Introduce a new workflow whose phase files legitimately omit one of the default required sections (analogous to how "verify" phases omit "Rules").
2. Run the structural scorer against a scenario for that workflow.
3. Observe that there is no configuration file to declare the new workflow's requirements — the only way to make it pass is to edit the scorer's source code.

**Expected behavior**
Teams should be able to declare custom per-workflow section requirements through project configuration (a TOML file resolved the same way other project policy files are), without modifying the scorer's source.

**Actual behavior**
The mapping is compiled into `eval_structural.py`; extending it for any workflow beyond the two built-in entries requires a source change and redeploy.

**Impact**
Any project or workflow with a phase layout differing from the two hardcoded defaults must either accept false failures on `required-sections-present` or fork the scorer. This raises the cost of onboarding new workflow types and is inconsistent with the project's own precedent for externalizing this shape of policy.

**Suggested correction**
Externalize the per-workflow required-sections table into a project-editable configuration file, using the project's existing override-resolution pattern, and default to the current two entries when no override file is present.

**How to verify**
Confirm a new configuration file can add or override a workflow's required sections without touching the scorer's source, and that behavior is unchanged when no override file exists.

tests/test_eval_cli.py
```diff
@@ -26,8 +26,14 @@ def _write_compliant(root: Path, sid: str = "ok") -> None:
     run.mkdir(parents=True)
     (root / sid / "scenario.toml").write_text(
         f'[scenario]\nid = "{sid}"\nworkflow = "w"\nrun_dir = "run"\nexpect = "compliant"\n')
-    (run / "plan.toml").write_text('[plan]\ntask = "t"\n[[phases]]\nnumber = 1\nfile = "p.md"\n')
-    (run / "p.md").write_text("# p\n")
+    (run / "plan.toml").write_text(
```
ainetx, line 29: ### No end-to-end CLI test covers a plan with zero parseable phase frontmatter

**Severity:** Minor

**Problem**
`cmd_eval` is the real-world entry point users invoke, but the specific failure mode where `plan.toml` loads successfully while every phase file lacks a parseable ` ```toml [phase]` block is only ever exercised through `StructuralScorer` called directly in unit tests — never driven through the actual CLI dispatch path.

**How to reproduce**
1. Search `tests/test_eval_cli.py` for the string "frontmatter" — no matches.
2. Review each `test_cmd_eval_*` test — each builds well-formed or deliberately-malformed baseline files, missing directories, or bad `--min` values, but none constructs a scenarios directory containing phase files with no parseable TOML frontmatter block.
3. Compare with `tests/test_eval_structural.py`, where this exact condition is tested only against `StructuralScorer(...).score(...)` directly, bypassing argument parsing, exit-code mapping, and CLI report formatting.

**Expected behavior**
The CLI exit-code/gating contract for "plan loads, but no phase has usable frontmatter" should be verified at the same layer users interact with, so a regression in argument handling, report formatting, or exit-code mapping would be caught by CI.

**Actual behavior**
This condition is untested at the CLI boundary; a regression here would not be caught by the existing CLI test suite.

**Impact**
Low likelihood, low blast radius — a coverage gap rather than a known defect. If a future change altered how `cmd_eval` maps scorer output to exit codes for this edge case, it could ship silently.

**Suggested correction**
Add a test in `tests/test_eval_cli.py` that builds a scenario with a `plan.toml` and phase file(s) lacking any ` ```toml` frontmatter block, drives it through `cmd_eval([...])`, and asserts on the resulting exit code and printed report content.

**How to verify**
After adding the test, confirm it fails against a deliberately broken `cmd_eval` (e.g. temporarily removing its frontmatter-error handling) and passes against current behavior.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+    unresolved, _ = _dependency_problems(inp)
+    return not unresolved, "; ".join(unresolved[:3])
+
+
+def _check_dependencies_not_forward(inp: StructuralInput) -> Tuple[bool, str]:
+    _, forward = _dependency_problems(inp)
+    return not forward, "; ".join(forward[:3])
+
+
+def _check_every_phase_declares_output(inp: StructuralInput) -> Tuple[bool, str]:
```
ainetx, line 343: ### "Every phase declares an output" check is satisfied by scratch-only `outputs`, undermining its intent

**Severity:** Minor

**Problem**
The structural compliance check named `every-phase-declares-an-output` accepts either the `outputs` field or the `output_files` field in a phase's frontmatter as satisfying the requirement. However, the project's own plan-authoring documentation defines these two fields with different semantics: `output_files` denotes created/modified project deliverables, while `outputs` denotes intermediate scratch files written to a working `out/` directory. A phase that declares only a scratch `outputs` entry and zero real `output_files` passes this check, even though it produced no actual project deliverable.

**How to reproduce**
1. Read the rule in the plan-authoring documentation: `output_files` = created/modified project files, `outputs` = intermediate files in `out/`.
2. Inspect `_check_every_phase_declares_output`: `if not (front.get("outputs") or front.get("output_files"))`.
3. Construct a phase file whose frontmatter sets `outputs = ["out/scratch.md"]` and omits `output_files` entirely (or sets it to `[]`).
4. Run the structural scorer against this phase; observe that `every-phase-declares-an-output` passes.

**Expected behavior**
A check named "every phase declares an output" should verify that a phase declares a real, project-facing deliverable (`output_files`), consistent with how the plan-template documentation distinguishes the two fields.

**Actual behavior**
The check treats `outputs` (documented as disposable, intermediate) as an equally valid substitute for `output_files` (documented as the real deliverable), so a phase with zero real declared output can still pass.

**Impact**
Low-to-moderate — a malformed or lazily-authored phase that declares no real deliverable could pass structural compliance scoring undetected, weakening the guarantee the check is meant to provide. No evidence this has occurred in existing fixtures; it is a latent gap.

**Suggested correction**
Either restrict `_check_every_phase_declares_output` to only accept `output_files` (matching the documented semantics), or, if `outputs`-only phases are intentionally valid for certain phase types, document that exception explicitly and add a test case covering it.

**How to verify**
Add a test using a phase body with only `outputs` set and no `output_files`, and assert the desired pass/fail outcome once the fix (or documented exception) is in place.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+    """One structural check: a name, tags (for grouped skipping), and a pure predicate."""
+
+    name: str
+    tags: Tuple[str, ...]
+    run: Callable[[StructuralInput], Tuple[bool, str]]   # -> (passed, detail)
+
+
+#: The registry. Each check is independent — order affects only the findings list, not any
+#: verdict. Extending the scorer means appending a ``Check`` here, not editing a function.
+CHECKS: List[Check] = [
```
ainetx, line 400: ### Undocumented invariant that structural-check names must not contain a colon

**Severity:** Minor

**Problem**
The structural compliance checks are registered as a list of thirteen named entries, and each check's finding string is formatted as `"<check-name>: <message>"`. The test suite parses these findings by splitting on the first colon to recover the check name for assertions. This creates an implicit requirement that no check name may itself contain a colon — but nothing in the codebase enforces, documents, or tests this constraint.

**How to reproduce**
1. List the registry entries — confirm none of the 13 names currently contain `:`.
2. Inspect the test helper that parses findings by splitting on the first `:`.
3. Add a new check registered with a name containing a colon (e.g. `"phase: has-note"`) that produces a finding on failure.
4. Run the test suite; observe the parsing helper now returns a truncated/wrong key instead of the full check name.

**Expected behavior**
Either check names should be structurally prevented from containing `:` (validated at registry construction time), or the finding format/parsing should use an unambiguous separator that tolerates colons in names.

**Actual behavior**
No validation exists; the constraint is only implicitly satisfied by convention across the current 13 check names.

**Impact**
Low — production code does not rely on colon-splitting, so runtime behavior is unaffected. Only the test helper's parsing would silently misbehave, potentially masking a test regression rather than causing a production bug.

**Suggested correction**
Add a lightweight assertion or registry-construction check that fails loudly if any check name contains `:`, documenting the invariant explicitly.

**How to verify**
After adding the guard, confirm that attempting to register a check name containing `:` raises an error or fails a test, rather than silently corrupting the parsing.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+        passed = 0
+        for check in active:
+            ok, detail = check.run(inp)
+            if ok:
+                passed += 1
+            else:
+                findings.append(f"{check.name}: {detail}" if detail else check.name)
+        return passed, findings
+
+    def _unknown(self, finding: str, coverage: str) -> ScorerResult:
```
ainetx, line 460: ### UNKNOWN scorer results have no structured cause code, only free-text coverage strings

**Severity:** Minor

**Problem**
The structural scorer returns an UNKNOWN verdict from three different code paths (unreadable run artifacts, no phase carries parseable frontmatter, all checks skipped), but the result carries no field that identifies which cause occurred. The only signal is a human-readable "coverage" string, which happens to differ between the three paths but is not a documented or tested contract.

**How to reproduce**
1. Score a run with no loadable artifacts — produces UNKNOWN with coverage "unscoreable: no readable plan.toml".
2. Score a run whose phases have no parseable frontmatter — produces UNKNOWN with coverage "unscoreable: a different workflow shape, not a failing one".
3. Score with every check skipped — produces UNKNOWN with coverage "unscoreable: no checks ran".
4. Observe that in all three cases the verdict is UNKNOWN and score is null; the only way to tell them apart programmatically is to string-match the coverage text.

**Expected behavior**
A consumer (CI gate, dashboard, alerting) that needs to react differently to "run didn't load" vs. "workflow shape doesn't apply" vs. "misconfigured skip list wiped out every check" should be able to do so via a stable, structured field.

**Actual behavior**
The only distinguishing signal is free-text prose, which is not guaranteed stable and isn't covered by any test asserting its exact wording per cause.

**Impact**
Low today — no current consumer parses the coverage field to branch on cause. But it's a latent trap: if a future consumer starts pattern-matching on that text, refactoring the message strings becomes a breaking change without any test to catch it.

**Suggested correction**
Add a lightweight reason enum/string-constant field to the result (e.g. "no_run", "no_parseable_phases", "all_checks_skipped"), or at minimum a stable constants module for the three coverage strings, so future consumers have a contractual value to key off instead of prose.

**How to verify**
After the fix, unit tests can assert the new reason field's exact value for each of the three UNKNOWN code paths, independent of the human-readable message text.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+            elif dep_no not in inp.phases:
+                unresolved.append(f"phase {number} -> {dep_no} (missing)")
+            elif dep_no >= number:
+                forward.append(f"phase {number} -> {dep_no}")
+    return unresolved, forward
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-check-helpers
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-checks
+def _check_phase_frontmatter_valid(inp: StructuralInput) -> Tuple[bool, str]:
```
ainetx, line 242: ### Failing-check detail strings silently truncate to 3–5 items with no indication of the true count

**Severity:** Minor

**Problem**
Six of the thirteen structural checks build their failure detail message by slicing the underlying list of offending items to the first 3 or 5 entries before joining into a string, with no "and N more" suffix or total count anywhere in the finding.

**How to reproduce**
1. Construct a run with, say, 8 phases each missing the "Rules" section.
2. Score it with the structural scorer.
3. Inspect the findings — the relevant entry reads like "required-sections-present: phase 1: Rules; phase 2: Rules; phase 3: Rules".
4. Note there is no text anywhere indicating 5 additional phases also failed this check.

**Expected behavior**
A finding string should communicate (or at least not obscure) the true scale of a defect — e.g. "...; and 5 more" — so a human or downstream tool reading the finding isn't misled into thinking the problem is smaller than it is.

**Actual behavior**
The list is silently capped; the true count is discarded and unrecoverable from the finding string alone.

**Impact**
Low-to-moderate — the pass/fail verdict itself is unaffected, but anyone triaging failures from the report text alone could undercount the scope of a structural problem.

**Suggested correction**
When the source list has more items than the slice, append a suffix such as "; and N more" (or an explicit "X of Y shown") at each of the six truncation sites.

**How to verify**
Add a test case with more than 5 offending items per check and assert the finding string reflects the true count, not just the capped list.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+    Deterministic: its verdict may reach the exit code (the runner enforces that; this
+    class only declares ``kind``). ``skip`` disables checks by name *or* tag, so a suite can
+    turn off a family of checks (e.g. all ``manifest`` checks) without editing the registry.
+    The scorer touches no filesystem — it is a pure function of the in-memory run artifacts.
+    """
+
+    name = "structural"
+    kind = ScorerKind.DETERMINISTIC
+
+    def __init__(self, skip: Tuple[str, ...] = ()) -> None:
```
ainetx, line 432: ### Unrecognized values in `StructuralScorer(skip=...)` silently no-op instead of erroring or warning

**Severity:** Minor

**Problem**
The `skip` parameter accepts any strings and is used purely as a set-difference filter against check names and tags. If a caller passes a name or tag that doesn't exist in the registry (typo, renamed check, stale config), no error or warning is raised — the scorer simply runs every check as if `skip` were empty.

**How to reproduce**
1. Construct the scorer with a typo'd skip value, e.g. `skip=("depedencies-resolve",)` (the real check is `dependencies-resolve`).
2. Score a run where `dependencies-resolve` would fail.
3. Observe the check runs anyway and its failure appears in the findings — the caller's intent to skip it was silently ignored.
4. Compare with the correctly-spelled skip value — that check is properly excluded, with a different verdict/score.

**Expected behavior**
A skip value that matches no check name and no check tag in the registry should surface feedback (a warning log, a raised error, or an entry in the result) so the caller knows their configuration had no effect.

**Actual behavior**
Invalid skip values are absorbed silently; the resulting output looks identical to a correctly-configured run with no skips, giving no signal that the skip configuration itself is broken.

**Impact**
A misconfigured eval suite (e.g. a config file with a typo'd check name meant to suppress a known-flaky check) would silently run the "skipped" check anyway, potentially causing unexpected FAIL verdicts or masking the fact that the intended suppression never took effect.

**Suggested correction**
Compute the set of all valid names/tags from the registry and validate `skip` against it — raise an error for unknown values, or at minimum log a warning for each unrecognized skip token.

**How to verify**
Add a test that constructs the scorer with a nonexistent skip value and asserts either a raised exception or a logged warning; the current codebase has no such test and no such behavior.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+def _check_manifest_total_matches_entries(inp: StructuralInput) -> Tuple[bool, str]:
+    declared = _manifest_declared(inp)
+    total_declared = _as_phase_int(inp.plan_meta.get("total_phases")) or 0
+    ok = total_declared == len(declared)
+    return ok, f"total_phases={total_declared}, entries={len(declared)}"
+
+
+def _check_manifest_matches_files(inp: StructuralInput) -> Tuple[bool, str]:
```
ainetx, line 290: ### Manifest-to-file pairing check only compares number sets, not per-entry mapping

**Severity:** Major

**Problem**
The "manifest matches files" structural check verifies only that the *set* of phase numbers declared in `plan.toml`'s manifest equals the *set* of phase numbers found in the actual phase files. It never verifies that a given manifest entry's `file` field actually points to the file that declares that entry's number.

**How to reproduce**
1. Take a compliant two-phase run with `phase-1.md` declaring number 1 and `phase-2.md` declaring number 2.
2. Edit `plan.toml` so the manifest entries are `{number = 1, file = "phase-2.md"}` and `{number = 2, file = "phase-1.md"}` — swap only the `file` values, keep the `number` values as-is.
3. Run the structural scorer over this run.

**Expected behavior**
A check that exists to confirm manifest/file agreement should flag that the manifest's declared file-to-number pairing does not match reality.

**Actual behavior**
The check (and every other check) passes; the scorer reports 100% compliance because it only ever compares the two number sets.

**Impact**
A `plan.toml` with a genuinely misconfigured phase-to-file mapping scores as fully compliant, silently masking a real authoring/tooling error that other consumers of `plan.toml` (anything that trusts `file` to locate "phase N") would hit.

**Suggested correction**
Add a check (or extend the existing one) that, for each manifest entry, resolves the entry's `file` to the phase actually parsed from that file and asserts its number equals the manifest entry's number.

**How to verify**
Add a test with a swapped file/number manifest, as in the reproduction steps, and assert the scorer reports a failing check, not 100%.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+
+logger = logging.getLogger(__name__)
+# @cpt-end:cpt-studio-algo-eval-structural:p1:inst-structural-imports
+
+
+# @cpt-begin:cpt-studio-algo-eval-structural:p1:inst-structural-config
+#: The fenced TOML block at the head of a phase file, carrying its ``[phase]`` table.
+#: Only *horizontal* whitespace is allowed after the opening fence (``[ \t]`` not ``\s``):
+#: letting ``\s`` swallow newlines makes an unterminated fence backtrack quadratically.
+_FRONTMATTER_RE = re.compile(r"^```toml[ \t]*\r?\n(?P<body>.*?)\r?\n```", re.DOTALL)
```
ainetx, line 40: ### Frontmatter fence regex silently rejects a leading BOM or blank line, with no near-miss diagnostic

**Severity:** Minor

**Problem**
The frontmatter fence pattern requires the ` ```toml ` fence at literal offset 0 of the phase file text. A UTF-8 byte-order-mark or a blank line before the fence causes the match to fail, and the file is treated exactly as it would be if it legitimately had no frontmatter block at all — there is no warning, unlike the sibling malformed-TOML case which does log one.

**How to reproduce**
1. Take a valid phase file with correct frontmatter.
2. Prepend a UTF-8 byte-order-mark or a single blank line before the fence.
3. Run the structural scorer on the run containing this file.

**Expected behavior**
Either the frontmatter still parses (BOM/leading-whitespace tolerant), or the near-miss is surfaced as a warning the same way malformed TOML is.

**Actual behavior**
The phase is treated as if it carried no frontmatter at all, folded into the generic "different shape, unscoreable" bucket with zero diagnostic signal — potentially causing an entire run to score UNKNOWN with no indication of the actual, trivial, fixable cause.

**Impact**
A run degrades from a scoreable PASS/FAIL to an unscoreable UNKNOWN due to an invisible encoding artifact, with no log or finding pointing at the real cause.

**Suggested correction**
Strip a leading BOM and leading blank lines before applying the frontmatter regex, or detect the near-miss and emit the same kind of warning used for malformed TOML.

**How to verify**
Add a test with a BOM-prefixed or blank-line-prefixed valid frontmatter block and assert it still parses (or that a warning-backed finding is produced) rather than silently becoming UNKNOWN.

skills/studio/scripts/studio/utils/eval_structural.py
ainetx, line 400: ### No guard against future duplicate check names in the structural-check registry

**Severity:** Minor

**Problem**
The structural-check registry is a plain list of check objects keyed by name; nothing enforces that names are unique. The skip-by-name mechanism and the pass/total aggregation both implicitly assume uniqueness.

**How to reproduce**
1. Add a second registry entry that accidentally reuses an existing check's name (e.g. copy-paste a new check and forget to rename it).
2. Run the scorer.

**Expected behavior**
Either a startup-time assertion/test failure catching the duplicate name, or well-defined skip/count semantics documented for the duplicate case.

**Actual behavior**
The duplicate check runs and is counted twice in the total (and possibly twice in passed or findings), silently skewing the compliance percentage, and skipping by that name would remove both entries at once, contrary to the apparent intent of skipping one check.

**Impact**
Low today since no duplicate exists, but a future edit could silently corrupt the compliance percentage with no test catching it.

**Suggested correction**
Add a one-line invariant (assertion or a small unit test) asserting the registry's check count equals its distinct-name count.

**How to verify**
Add the assertion/test and confirm it fails if a duplicate name is intentionally introduced in a test-only registry copy.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,406 @@
...
+        totals.add(total if total is not None else 0)
+    return totals
+
+
+def _dependency_problems(inp: StructuralInput) -> Tuple[List[str], List[str]]:
+    """Return ``(unresolved, forward)`` dependency problems across all phases."""
+    unresolved: List[str] = []
+    forward: List[str] = []
+    for number, front in sorted(inp.phases.items()):
+        deps = front.get("depends_on")
```
ainetx, line 221: ### Misspelled `depends_on` key (e.g. `dependsOn`) is silently treated as "no dependencies," with no diagnostic

**Severity:** Major

**Problem**
Phase frontmatter dependency declarations are read via a direct key lookup for `depends_on`. Any key name that doesn't exactly match — including a plausible typo like `dependsOn` or `depend_on` — returns nothing, which the code interprets identically to a phase that has no dependencies at all. Nothing validates the phase frontmatter table's key names, so this typo produces no warning, no failing check, and no other observable signal.

**How to reproduce**
1. Author a two-phase run where phase 2 is meant to depend on phase 1, but write the frontmatter key as `dependsOn = [1]` instead of `depends_on = [1]`.
2. Run the structural scorer over this run.

**Expected behavior**
The scorer should either recognize the intended key despite the typo, or flag an unrecognized/unexpected key in the phase frontmatter so the author notices their ordering constraint was dropped.

**Actual behavior**
The dependency declaration vanishes entirely; both dependency-related checks trivially pass (there's nothing to check), and the run can score 100% compliant despite having no enforced ordering between phases the author intended to be ordered.

**Impact**
An author's ordering intent is silently and completely lost, and the run can score a false 100% "structurally compliant" result even though the dependency the workflow actually relies on for correctness is unenforced — this is precisely the class of defect (dependency order) the scorer exists to validate.

**Suggested correction**
Validate phase frontmatter keys against an explicit allow-list (number, total, depends_on, outputs/output_files, etc.) and add a check/warning for unrecognized keys, so a typo'd key surfaces as a finding rather than disappearing.

**How to verify**
Add a test with a phase using `dependsOn` instead of `depends_on` (with a genuine forward or missing dependency intended) and assert the scorer reports a finding rather than a vacuous pass.

ainetx (commented):
Deep Review: 12 findings (2 Major, 10 Minor) from a check-based structural review of this PR. See individual inline comments below.

_Minor findings not are blockers_

ainetx (changes requested):

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. `manifest-matches-files` now checks two things, not just the number set: (1) declared vs found number sets, and (2) each manifest entry's number->file pairing against the file's own declared number — so a manifest that maps phase 1 to `phase-2.md` now fails. Added a swapped-pairing test. Good catch.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. A phase with no `depends_on` but a key normalizing to a known alias (`dependsOn`, `dependencies`, `depends`, …) is now flagged in `dependencies-resolve` as a likely-misspelled key, instead of silently reading as no-deps. Added a `dependsOn` test. (Covers the common typos, not every possible one.)

tests/test_eval_cli.py
SanjeevSolanki: Fixed in `f3f3c22`. Added an end-to-end `cmd_eval` test: `plan.toml` loads but every phase file lacks a parseable `[phase]` block → the run scores UNKNOWN and `--check` still exits 0 (unscoreable never gates).

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. Truncated detail strings now append `(+N more)` via a shared `_capped` helper, so the true count is never hidden. Applied to all six sliced checks; added a unit test.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. The frontmatter match now tolerates a leading UTF-8 BOM and blank lines (via a small normalizer used by both the parser and the prose extractor), so those parse instead of silently reading as 'no frontmatter'. Added a BOM test.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. `StructuralScorer(skip=...)` now warns when a skip token matches no check name or tag, so a typo'd skip no longer silently no-ops. Added a caplog test.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Fixed in `f3f3c22`. Added a test enforcing check-name uniqueness across the registry (CI fails on a duplicate) and documented the invariant at the registry — the pass/total aggregation and skip-by-name both rely on it.

SanjeevSolanki: Fixed in `f3f3c22`. Documented the 'findings serialise as `name: detail`, so names carry no colon' invariant at the registry, and added a test asserting no `:` in any check name.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Leaving this as-is, deliberately. `WORKFLOW_SECTIONS` is an *override* map, not an allowlist — the vast majority of workflows correctly fall back to the default `Preamble/What/Rules`. A diagnostic on every unmapped workflow would therefore fire on the common, correct case (noise), so silent fallback is the right default here.

SanjeevSolanki: Agreed this is worth having, but as a *feature* (project-level required-section config), not a v1 defect — the in-source map is the minimal viable policy. Tracking it as a follow-up rather than expanding this PR's scope.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Leaving as-is. The phase frontmatter doesn't model output *durability* (scratch vs durable), so the checkable invariant is that a phase *declares* an output at all. Distinguishing scratch outputs would need a frontmatter schema change — noted as a possible follow-up, out of scope here.

skills/studio/scripts/studio/utils/eval_structural.py
SanjeevSolanki: Deferring. The UNKNOWN `coverage` string already states the cause in text (no readable plan / no parseable frontmatter / all checks skipped). A machine-readable cause code would change the shared `ScorerResult` shape (the scaffold's data model), so I'd rather add it when a consumer actually needs to branch on it — tracked as a follow-up.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,480 @@
...
+
+def _misspelled_dep_key(front: Dict[str, object]) -> Optional[str]:
+    """A frontmatter key that looks like a misspelled ``depends_on`` (when the real key is
+    absent), or ``None``. Normalises case/underscores/hyphens before comparing."""
+    if "depends_on" in front:
+        return None
+    for key in front:
+        if isinstance(key, str) and key.lower().replace("_", "").replace("-", "") in _DEP_KEY_ALIASES:
+            return key
+    return None
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `2169f4f`. Added `dependon` / `dependenton` / `dependings` to the alias set, so `depend_on` (and similar) are now flagged, not silently read as no-deps. Parametrized the typo test across `dependsOn` / `depend_on` / `dependencies`. Good catch — a typo-detector with a typo gap.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,480 @@
...
+        number = _as_phase_int(entry.get("number"))
+        file = entry.get("file")
+        if not isinstance(file, str) or number is None:
+            continue
+        actual = inp.by_file.get(file)
+        if actual is None:
+            problems.append(f"{file} (manifest phase {number}) has no valid phase")
+        elif actual != number:
+            problems.append(f"manifest pairs {file}->{number} but the file declares {actual}")
+    return not problems, _capped(problems, 3)
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `2169f4f`. Added a dedicated `manifest-entries-valid` check that fails when any `[[phases]]` entry has a missing or non-positive number — so a malformed manifest entry is surfaced instead of being quietly dropped by the number-set reductions the other manifest checks rely on. Added a test.

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -0,0 +1,480 @@
...
+        """PASS if every active check passes, FAIL if any fails, UNKNOWN if unscoreable."""
+        if run is None:
+            return self._unknown("run artifacts could not be loaded",
+                                 "unscoreable: no readable plan.toml")
+
+        phases, bodies, by_file, duplicates, invalid = _collect_phases(run.phase_texts)
+        if not phases:
+            return self._unknown(
+                "no phase file carries a parseable [phase] frontmatter block",
+                "unscoreable: a different workflow shape, not a failing one")
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed in `2169f4f`. `_collect_phases` now separates two cases: a file with a `[phase]` block that is broken (malformed TOML or a bad number) vs. a file with no fenced block at all. A run whose files carry present-but-broken frontmatter now scores **FAIL** (not UNKNOWN); UNKNOWN is reserved for 'no `[phase]` block anywhere' — a genuinely different workflow shape. Tests updated to assert the FAIL path.

PR #95, opened 2026-08-19
fix(validate): reach the code checks when a scan resolves to no files

Oleg67:
**What.** Makes `cfs validate`'s existing code-traceability checks reachable when a scan resolves to no files, and reports a registered codebase entry that matches nothing. Touches `commands/validate.py`, one new error code, one CDSL instruction, and a new test file.

**Why.** A FEATURE with a checked `to_code = "true"` ID, no `@cpt` markers and no code at all validates clean today:

```json
{"status":"PASS","error_count":0,"warning_count":0,
 "code_files_scanned":0,"to_code_ids_total":1,"code_ids_found":0}
```

`to_code_ids_total: 1` and `code_ids_found: 0` sit in the same object and produce neither an error nor a warning. Adding a single *unmarked* `.py` file to the identical repository produces the correct errors, which is the evidence that nothing is wrong with the checks — cross-validation returned early whenever no code file had been parsed:

```python
if not strict_code_validation or not results.parsed_code_files_full:
    return
```

A registered codebase entry that matches nothing parses no files, so the whole check ran and reported success without having assessed anything.

**How.** The guard now keys on whether a claim exists rather than on how much was scanned. A checked `to_code = "true"` ID is a claim; its marker has to resolve regardless of scan size. `results.to_code_ids` is already populated before this point, so it is exactly the "is anything being claimed?" signal:

```python
if not results.parsed_code_files_full and not results.to_code_ids:
    return
```

**Deleting the file-count check outright would be shorter and wrong.** Cross-validation would then run for repositories that claim nothing, turning every declared CDSL instruction into `code-inst-missing` and failing a spec-first project that has written its design and not yet built it. That case is covered by a test built specifically to expose it (`test_declared_instructions_without_a_code_claim_are_not_errors`) — without it, the over-broad version passes the whole suite.

A registered FULL entry resolving to no files is reported separately as a **warning**, `codebase-entry-empty`, because the subject is the registration rather than any claim — usually a stale or misspelled path. Nothing is owed unless something was claimed, so it does not fail the run. Reporting the two independently is what stops a green exit from meaning two different things:

```
codebase entry `src/api` is registered with FULL traceability but resolved to 0 files
— nothing from it was checked
```

**No new gates and no new vocabulary.** `code-no-marker` and `code-inst-missing` already say the right thing and are unchanged; this makes them reachable. The one new code is in a previously unused `codebase-*` namespace.

**Correction to an earlier version of this description.** It claimed a measured false-positive floor of zero on the strength of this repository's own artifacts being unchanged. That measurement was real but not sufficient: the registry here is flat (both systems declare `children = []`), so it could not exercise nested layouts, and review found a genuine false positive there. It is fixed and covered — see the nested case below. On this repository the numbers are indeed unchanged:

| | Artifacts | Code files | Coverage | Errors | Warnings |
|---|---|---|---|---|---|
| before | 49 | 91 | 219/219 | 0 | 192 |
| after | 49 | 91 | 219/219 | 0 | 192 |

**Nested systems inherit FULL traceability.** A parent that owns the FULL artifact makes the claim, but the code backing it often lives in a child system with no artifact of its own. Traceability was derived per node, so that child was scanned as DOCS-ONLY and never entered the FULL set — and reaching the code checks at zero FULL files then reported a correctly marked implementation as unmarked. Measured on a nested fixture whose child code carries the right marker:

| Child code has the correct marker | `main` | this branch, before the fix | after |
|---|---|---|---|
| verdict | exit 0 | **exit 2 `code-no-marker`** | exit 0 |

An unmarked child still fails, so the inheritance does not become a way to escape the check.

**A passing run names what it did not check.** A non-verbose report omits warning bodies and prints only a count, which left a green run silent about the entry it skipped. The entries are now named in the report itself — `unscanned_codebase_entries` in JSON, and a line on the human surface — since that is the point of reporting them. Workspace entries are qualified by source, so two entries sharing a path stay distinguishable and the reader knows whether to fix the source or the path.

**Tests.** New file `tests/test_enforcement_empty_codebase.py` — 11 tests driving the real CLI through `run_cli_in_project`, covering known-bad going red, known-good staying green, the scan-size invariant (0 files and 1 unmarked file are the same claim, so the same verdict), determinism across repeated runs, and the over-breadth case above.

Each half of the change was mutation-tested, so the corpus is not vacuous:

| Production change reverted | Result |
|---|---|
| Restore the original 0-file bypass | **5 tests fail** |
| Remove the claim check (over-broad) | **over-breadth test fails** |
| Suppress the empty-entry warning | **2 tests fail** |
| Stop inheriting FULL into children | **2 nested tests fail** |
| Drop the report field | **3 non-verbose tests fail** |
| Drop the source qualification | **workspace test fails** |

**Gates**, all on Python 3.11 against the rebased state: full suite **4,573 passed, 4 skipped, 15 xfailed, 58 subtests**; `validate` 0 errors, 219/219; `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` all thresholds met; `pylint` clean; `vulture` clean; `validate-toc` correct. The new engine function is `@cpt`-traced with a declared instruction (`inst-warn-empty-codebase-entry`), and the validate flow's Error Scenarios now state the three empty-scope outcomes. No new dependencies.

**Relationship to other work.** Companion to #89. Independent of #93 — different file, no overlap, either can land first — but the two answer the same question on the two commands, so #93 is useful context: it separates "did anything fail?" from "was there anything to assess?" in `spec-coverage`, and this applies the same separation to `validate`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now detects registered FULL-traceability codebase entries that resolve to no files.
  * Checked code claims are validated even when no code files are found.
  * Added clear warnings and reporting for empty codebase registrations and missing code markers.
  * FULL traceability now carries through to descendant systems.

* **Documentation**
  * Added validation scenarios for empty entries, missing markers, and runs without checked code claims.

* **Tests**
  * Expanded coverage across JSON, human-readable, and workspace validation results.

### DIFF
--- a/skills/studio/scripts/studio/commands/validate.py
+++ b/skills/studio/scripts/studio/commands/validate.py
@@ -757,7 +757,7 @@ def _scan_codebase_entry(
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
     scan_targets = list(_resolve_code_scan_targets(session, entry))
     if traceability == "FULL" and not scan_targets:
-        results.empty_full_codebase_entries.append(str(getattr(entry, "path", "")))
+        results.empty_full_codebase_entries.append(_codebase_entry_path(entry))
     for file_path in scan_targets:
         try:
             rel_path = file_path.resolve().relative_to(session.project_root).as_posix()
@@ -794,6 +794,15 @@ def _scan_codebase_entry(
     # @cpt-end:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
 
 
+def _codebase_entry_path(entry: object) -> str:
+    """Configured path of a codebase entry, which may be a record or a raw dict."""
+    # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
+    if isinstance(entry, dict):
+        return str(entry.get("path", ""))
+    return str(getattr(entry, "path", ""))
+    # @cpt-end:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
+
+
 def _resolve_code_scan_targets(session: _ValidateSession, entry: object) -> List[Path]:
     """Resolve concrete files for one codebase entry."""
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
@@ -801,8 +810,7 @@ def _resolve_code_scan_targets(session: _ValidateSession, entry: object) -> List
     if src_name and session.ws_ctx is not None:
         code_path = session.ws_ctx.resolve_artifact_path(entry, session.project_root)
     else:
-        entry_path = getattr(entry, "path", "") if not isinstance(entry, dict) else entry.get("path", "")
-        code_path = (session.project_root / entry_path).resolve()
+        code_path = (session.project_root / _codebase_entry_path(entry)).resolve()
     if code_path is None or not code_path.exists():
         return []
     if code_path.is_file():
--- a/skills/studio/scripts/studio/commands/validate.py
+++ b/skills/studio/scripts/studio/commands/validate.py
@@ -57,7 +57,7 @@ class _ValidateResults:
     parsed_code_files_full: List[CodeFile] = field(default_factory=list)
     code_ids_found: Set[str] = field(default_factory=set)
     to_code_ids: Set[str] = field(default_factory=set)
-    empty_full_codebase_entries: List[str] = field(default_factory=list)
+    empty_full_codebase_entries: List[Dict[str, str]] = field(default_factory=list)
 
 # @cpt-begin:cpt-studio-algo-workspace-determine-target:p1:inst-validate-source-flag
 def _resolve_source_context(source_name: str, ws_ctx: Optional["WorkspaceContext"]) -> Optional["StudioContext"]:
@@ -757,7 +757,12 @@ def _scan_codebase_entry(
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
     scan_targets = list(_resolve_code_scan_targets(session, entry))
     if traceability == "FULL" and not scan_targets:
-        results.empty_full_codebase_entries.append(_codebase_entry_path(entry))
+        results.empty_full_codebase_entries.append({
+            "path": _codebase_entry_path(entry),
+            "source": str(
+                entry.get("source") or "" if isinstance(entry, dict) else getattr(entry, "source", "") or ""
+            ),
+        })
     for file_path in scan_targets:
         try:
             rel_path = file_path.resolve().relative_to(session.project_root).as_posix()
@@ -830,10 +835,23 @@ def _scan_system_codebase(
     session: _ValidateSession,
     results: _ValidateResults,
     strict_code_validation: bool,
+    inherited_traceability: str = "DOCS-ONLY",
 ) -> None:
-    """Recursively scan a system node's configured codebase entries."""
+    """Recursively scan a system node's configured codebase entries.
+
+    FULL traceability is inherited by descendants. A parent that owns the FULL
+    artifact makes the claim; the code backing it often lives in a child system
+    that has no artifact of its own. Deriving traceability from each node in
+    isolation left such code out of the FULL set, so its markers were never
+    cross-validated and a correctly marked implementation looked unmarked.
+    """
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
-    traceability = "FULL" if any(art.traceability == "FULL" for art in system_node.artifacts) else "DOCS-ONLY"
+    traceability = (
+        "FULL"
+        if inherited_traceability == "FULL"
+        or any(art.traceability == "FULL" for art in system_node.artifacts)
+        else "DOCS-ONLY"
+    )
     for codebase_entry in system_node.codebase:
         _scan_codebase_entry(
             entry=codebase_entry,
@@ -848,29 +866,44 @@ def _scan_system_codebase(
             session=session,
             results=results,
             strict_code_validation=strict_code_validation,
+            inherited_traceability=traceability,
         )
     # @cpt-end:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
 
 
-def _build_empty_codebase_entry_warnings(entry_paths: List[str]) -> List[Dict[str, object]]:
+def _build_empty_codebase_entry_warnings(entries: List[Dict[str, str]]) -> List[Dict[str, object]]:
     """Warn for each FULL codebase entry that is registered but resolves to no files.
 
     A registered entry matching nothing is a configuration mistake -- usually a
-    stale or misspelled path -- and it silently shrinks what the run was able to
-    check. It is reported separately from any unmet ID, because the subject is
-    the registration rather than the claim.
+    stale or misspelled path, or a workspace source that is not reachable -- and
+    it silently shrinks what the run was able to check. It is reported separately
+    from any unmet ID, because the subject is the registration rather than the
+    claim. The workspace source is named when there is one, so two entries that
+    share a path stay distinguishable and the reader knows whether to fix the
+    source or the path.
     """
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-warn-empty-codebase-entry
-    return [
-        code_issue(
-            "structure",
-            f"codebase entry `{path or '<unset>'}` is registered with FULL traceability "
-            "but resolved to 0 files — nothing from it was checked",
-            code=EC.CODEBASE_ENTRY_EMPTY,
-            path=Path(path or "."),
+    warnings: List[Dict[str, object]] = []
+    for entry in entries:
+        path = entry.get("path") or "<unset>"
+        source = entry.get("source") or ""
+        subject = f"`{source}:{path}`" if source else f"`{path}`"
+        cause = (
+            f"workspace source `{source}` did not resolve, or it resolved to an empty tree"
+            if source
+            else "the path does not exist, or it contains no files with the configured extensions"
         )
-        for path in entry_paths
-    ]
+        warnings.append(
+            code_issue(
+                "structure",
+                f"codebase entry {subject} is registered with FULL traceability but "
+                f"resolved to 0 files — nothing from it was checked ({cause})",
+                code=EC.CODEBASE_ENTRY_EMPTY,
+                path=Path(path if path != "<unset>" else "."),
+                source=source or None,
+            )
+        )
+    return warnings
     # @cpt-end:cpt-studio-flow-traceability-validation-validate:p1:inst-warn-empty-codebase-entry
 
 
@@ -1131,6 +1164,13 @@ def _emit_final_validate_report(session: _ValidateSession, results: _ValidateRes
         report["code_ids_found"] = len(results.code_ids_found)
         if results.to_code_ids:
             report["coverage"] = f"{len(results.code_ids_found & results.to_code_ids)}/{len(results.to_code_ids)}"
+        # Named unconditionally: a passing run has to be able to say what it did
+        # not check, and warning bodies are omitted from a non-verbose report.
+        if results.empty_full_codebase_entries:
+            report["unscanned_codebase_entries"] = [
+                f"{entry['source']}:{entry['path']}" if entry.get("source") else entry["path"]
+                for entry in results.empty_full_codebase_entries
+            ]
     if overall_status == "PASS":
         report["next_step"] = (
             "Deterministic validation passed. Now perform semantic validation: "
@@ -1429,6 +1469,22 @@ def _run_content_language_check(
 # ---------------------------------------------------------------------------
 
 # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-format
+def _show_unscanned_codebase_entries(unscanned: List[str]) -> None:
+    """Name the registered entries that resolved to no files, if any."""
+    if not unscanned:
+        return
+    ui.blank()
+    ui.warn(
+        f"{len(unscanned)} registered codebase "
+        f"{'entry' if len(unscanned) == 1 else 'entries'} resolved to 0 files "
+        "— nothing from them was checked"
+    )
+    for entry in unscanned[:10]:
+        ui.substep(f"  {entry}")
+    if len(unscanned) > 10:
+        ui.substep(f"  ... and {len(unscanned) - 10} more")
+
+
 def _human_validate(data: dict) -> None:
     status = data.get("status", "")
     n_art = data.get("artifacts_validated", data.get("artifact_count", 0))
@@ -1444,6 +1500,8 @@ def _human_validate(data: dict) -> None:
         ui.detail("Code files", str(data["code_files_scanned"]))
     if data.get("coverage"):
         ui.detail("Code coverage", str(data["coverage"]))
+
+    _show_unscanned_codebase_entries(data.get("unscanned_codebase_entries", []))
 
     errors = data.get("errors", [])
     if errors:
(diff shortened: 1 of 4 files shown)


skills/studio/scripts/studio/commands/validate.py
```diff
@@ -754,7 +755,10 @@ def _scan_codebase_entry(
 ) -> None:
     """Scan one configured codebase entry and collect code traceability state."""
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
-    for file_path in _resolve_code_scan_targets(session, entry):
+    scan_targets = list(_resolve_code_scan_targets(session, entry))
+    if traceability == "FULL" and not scan_targets:
+        results.empty_full_codebase_entries.append(str(getattr(entry, "path", "")))
+    for file_path in scan_targets:
```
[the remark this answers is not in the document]
Oleg67: Confirmed and fixed in `3bad7844`. `_resolve_code_scan_targets` reads both shapes for `path` *and* `extensions`, so a dict entry would have been resolved for scanning but reported as `<unset>` in the warning — losing the single detail a user needs to fix a stale or misspelled path.

I took the fix a step further than the suggested diff: rather than repeat the `isinstance` branch at the second site, both now share a helper, so the two can't drift apart again — which is how this bug arose in the first place.

```python
def _codebase_entry_path(entry: object) -> str:
    """Configured path of a codebase entry, which may be a record or a raw dict."""
    if isinstance(entry, dict):
        return str(entry.get("path", ""))
    return str(getattr(entry, "path", ""))
```

Covered by a test asserting both the helper and the rendered warning. Mutation-checked: reverting to bare attribute access fails that test and nothing else.

tests/test_enforcement_empty_codebase.py
```diff
@@ -0,0 +1,361 @@
...
+            code, report = run_cli_in_project(root, ["--json", "validate", "--verbose"])
+
+        assert code == 2
+        assert report["status"] == "FAIL"
+        assert "code-no-marker" in _codes(report, "errors")
+
+    def test_the_unmet_claim_is_named_not_just_counted(self):
+        """A verdict has to say which claim is unbacked, or it cannot be acted on."""
+        with TemporaryDirectory() as tmpdir:
+            root = Path(tmpdir)
```
Oleg67, line 246: Agreed, and I'd call this more than a nitpick — it was a real hole in the corpus.

The two checks say different things, and only together do they describe the state honestly: `code-no-marker` says the ID has no marker anywhere, while one `code-inst-missing` per declared step says *which pieces* of the claim have no code block. A change that reached the first but not the second would have looked correct while reporting an unbacked claim without saying how much of it was unbacked. My tests asserted only the first, so that partial outcome would have passed.

Added in `3bad7844`. The fixture now attaches checked CDSL steps to the claimed `to_code` ID — steps bind to the nearest ID above them — and the assertion names the instructions rather than counting them, so it fails on a wrong set as well as a wrong number:

```python
        assert codes.count("code-no-marker") == 1
        assert codes.count("code-inst-missing") == len(INSTRUCTIONS)
        assert missing == set(INSTRUCTIONS)
```

Mutation-checked: passing `artifact_instances=None` to `cross_validate_code` fails this test alone, so it is genuinely pinning the instruction-level path and not riding on the ID check.

Gates on the new head, Python 3.11: full suite **4,575 passed, 4 skipped, 15 xfailed**; `validate` 0 errors / 192 warnings / 219-219 — unchanged from before the PR, so the false-positive floor on this repository is still zero; `spec-coverage` at 90/60/0.46 all thresholds met; `pylint` and `vulture` clean.

skills/studio/scripts/studio/commands/validate.py
```diff
@@ -840,6 +852,28 @@ def _scan_system_codebase(
...
+def _build_empty_codebase_entry_warnings(entry_paths: List[str]) -> List[Dict[str, object]]:
+    """Warn for each FULL codebase entry that is registered but resolves to no files.
+
+    A registered entry matching nothing is a configuration mistake -- usually a
+    stale or misspelled path -- and it silently shrinks what the run was able to
+    check. It is reported separately from any unmet ID, because the subject is
+    the registration rather than the claim.
+    """
+    # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-warn-empty-codebase-entry
+    return [
```
ainetx: ### Empty workspace codebase entries lose their source identity

**Severity:** Minor

**Problem**
When a workspace-scoped codebase entry cannot resolve, the new warning keeps only its bare configured path. It drops both the workspace source and whether the failure came from an unavailable source, a missing path, or an empty directory.

**How to reproduce**

1. Configure a codebase entry with `source = "ghost"` and `path = "src"`.
2. Make the `ghost` workspace source unavailable.
3. Run validation so the entry resolves to no files.

**Expected behavior**
The diagnostic should identify the workspace source and distinguish source-resolution failure from an empty local directory.

**Actual behavior**
The emitted issue is attributed only to `src`, making separate workspace entries with the same path indistinguishable and implying a local empty path.

```text
source=ghost, path=src
    -> workspace resolution fails
    -> [] scan targets
    -> warning records only "src"
```

**Impact**
Users cannot tell which remote source is misconfigured or whether they should fix source availability, the configured path, or the directory contents.

**Suggested correction**
Build the warning from the original entry and preserve source-qualified identity plus a resolution cause in the diagnostic.

**How to verify**
Add a workspace fixture with an unavailable source and assert that its warning identifies both the source and configured path.

skills/studio/scripts/studio/commands/validate.py
```diff
@@ -871,7 +905,12 @@ def _run_code_validation(
             results=results,
             strict_code_validation=strict_code_validation,
         )
-    if not strict_code_validation or not results.parsed_code_files_full:
+    if not strict_code_validation:
+        return
+    results.all_warnings.extend(
```
ainetx, line 951: ### Empty-codebase warnings are hidden on successful default validation

**Severity:** Major

**Problem**
The new `codebase-entry-empty` issue is collected as a warning, but successful non-verbose validation reports only its count. Neither the JSON nor human output identifies the empty codebase entry that the feature is intended to make visible.

**How to reproduce**

1. Configure a FULL codebase entry such as `does/not/exist` with no checked `to_code` claim.
2. Run `cfs validate` without `--verbose`.

**Expected behavior**
The successful run should name the registered codebase entry that resolved to no files, so the user can correct it.

**Actual behavior**
The result says `Warnings: 1` and `All checks passed`; JSON likewise exposes only `warning_count`, without the warning body or configured path.

```text
empty FULL entry
    -> codebase-entry-empty warning collected
    -> PASS and non-verbose report
    -> count emitted, warning body omitted
```

**Impact**
A green run still does not say what it failed to check, defeating the user-facing purpose of the new warning unless the user already knows to request verbose output.

**Suggested correction**
Include warning bodies (or an always-on concise warning summary) whenever a successful final report has warnings, in both JSON and human output.

**How to verify**
Add non-verbose JSON and human-output regressions for an empty FULL entry and assert that the configured path and warning text are visible.

skills/studio/scripts/studio/commands/validate.py
```diff
@@ -754,7 +755,10 @@ def _scan_codebase_entry(
 ) -> None:
     """Scan one configured codebase entry and collect code traceability state."""
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
-    for file_path in _resolve_code_scan_targets(session, entry):
+    scan_targets = list(_resolve_code_scan_targets(session, entry))
```
ainetx, line 768: ### Child codebase entries lose FULL validation unless the child owns a FULL artifact

**Severity:** Major

**Problem**
Nested codebase entries derive traceability only from their own system node. A child that contains implementation but has no local FULL artifact is scanned as DOCS-ONLY, even when its parent owns the FULL claim.

**How to reproduce**

1. Define a parent system with a FULL artifact and a nested child system that owns the codebase entry.
2. Put the implementation marker only in the child codebase.
3. Run `cfs validate`.

**Expected behavior**
The child codebase should participate in FULL validation inherited from the parent’s claim.

**Actual behavior**
Child files are excluded from `parsed_code_files_full`; valid child markers can be ignored, causing false missing-marker/instruction errors. Empty child registrations also miss the new warning.

```text
parent FULL claim
    -> child codebase scan
    -> child has no local FULL artifact
    -> treated as DOCS-ONLY
    -> child code excluded from FULL validation
```

**Impact**
Supported nested layouts can fail validation despite valid implementation, or silently omit the new empty-entry warning for descendant codebases.

**Suggested correction**
Carry effective FULL traceability from ancestor systems into descendant codebase scans, and add nested parent/child coverage.

**How to verify**
Add a nested-system fixture with the FULL artifact on the parent and markers in child code; validation should recognize the markers and warn for an empty child entry.

Oleg67: Confirmed, and this was the important one — thank you. It is a false positive **this branch introduces**, not a pre-existing issue, and I verified it rather than reasoning about it.

Nested fixture: FULL artifact and the checked `to_code` claim on the parent, codebase entry on a child, correct `@cpt-flow` marker in the child's code.

| Child code has the correct marker | `main` | this branch before the fix | after |
|---|---|---|---|
| `code_files_scanned` | 1 | 1 | 1 |
| verdict | exit 0 PASS | **exit 2 `code-no-marker`** | exit 0 PASS |

The file was scanned and the marker was there; it just never entered `parsed_code_files_full`, so reaching the checks at zero FULL files reported it missing. Fixed by carrying FULL from ancestors, as you suggested:

```python
    traceability = (
        "FULL"
        if inherited_traceability == "FULL"
        or any(art.traceability == "FULL" for art in system_node.artifacts)
        else "DOCS-ONLY"
    )
```

An unmarked child still fails, so the inheritance does not become an escape hatch — both directions are pinned, and mutation-checked: removing the propagation fails the nested tests and nothing else.

**Worth stating plainly: my "false-positive floor: zero, measured" claim was wrong**, and your finding shows why. The measurement itself was real — this repository's validate output is byte-identical before and after — but the registry here is flat (both systems declare `children = []`), so it cannot exercise a nested layout at all. I measured the one shape that could not exhibit the bug and then generalised from it. I have corrected the PR description rather than quietly editing the number.

One process note, since it nearly produced a wrong answer for you: my first attempt to establish the `main` baseline used `git stash`, which was a no-op because the work was already committed — so that run was still this branch's code and appeared to show the behaviour as pre-existing. The table above comes from a detached worktree at `upstream/main`.

skills/studio/scripts/studio/commands/validate.py
Oleg67: Confirmed and fixed. You are right that this defeated the point: the warning existed but a default run printed only `Warnings: 1`, so a green run still would not say what it had skipped.

The cause of my missing it is worth recording — every e2e test in the corpus passed `--verbose`, which is the only mode where warning bodies appear. Same shape of blind spot as #93, where the tests ran in JSON mode and the human surface went uncovered.

I did **not** take the broader option of emitting all warning bodies on success. This repository carries 192 pre-existing warnings on every run, so that would turn a passing `make validate` into 192 lines of output — a cross-cutting change to every consumer, well beyond this PR. Instead the specific signal is named in the report itself, unconditionally:

```json
{"status":"PASS","warning_count":1,"unscanned_codebase_entries":["does/not/exist"]}
```

and on the human surface:

```
  ⚠ 1 registered codebase entry resolved to 0 files — nothing from them was checked
      does/not/exist
```

Both are covered by non-verbose regressions — JSON and human — and mutation-checked: dropping the report field fails three tests. If you would rather have the general "show warning bodies on success" behaviour, I am happy to do it, but I would send it as its own change so its blast radius is reviewable on its own.

skills/studio/scripts/studio/commands/validate.py
Oleg67: Fixed. The warning now carries the source and says which of the two things to check:

```
codebase entry `ghost:src` is registered with FULL traceability but resolved to 0 files
— nothing from it was checked (workspace source `ghost` did not resolve, or it
resolved to an empty tree)
```

versus a local entry:

```
codebase entry `src` is registered with FULL traceability but resolved to 0 files
— nothing from it was checked (the path does not exist, or it contains no files
with the configured extensions)
```

The source is also kept as a structured `source` field on the issue, and `unscanned_codebase_entries` in the report uses `source:path` so two workspace entries sharing a path stay distinguishable.

On distinguishing an unresolvable source from an empty one: I have not split those into separate causes. `_resolve_code_scan_targets` collapses both into an empty list — `resolve_artifact_path` returning `None` and a resolved-but-empty tree are indistinguishable by the time my code sees the result — so naming both possibilities is as precise as I can be without changing that resolution path, which felt out of scope here. Say the word if you would rather have it split; it would mean returning a reason from the resolver.

Both message shapes are covered by tests, and mutation-checked: dropping the source qualification fails the workspace test.

skills/studio/scripts/studio/commands/validate.py
```diff
@@ -790,15 +799,23 @@ def _scan_codebase_entry(
...
 def _resolve_code_scan_targets(session: _ValidateSession, entry: object) -> List[Path]:
     """Resolve concrete files for one codebase entry."""
     # @cpt-begin:cpt-studio-flow-traceability-validation-validate:p1:inst-validate-code-scan
     src_name = getattr(entry, "source", None)
     if src_name and session.ws_ctx is not None:
         code_path = session.ws_ctx.resolve_artifact_path(entry, session.project_root)
     else:
-        entry_path = getattr(entry, "path", "") if not isinstance(entry, dict) else entry.get("path", "")
-        code_path = (session.project_root / entry_path).resolve()
+        code_path = (session.project_root / _codebase_entry_path(entry)).resolve()
```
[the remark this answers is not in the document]
Oleg67: Valid, and fixed in `05091e3a` — with one correction to the severity and one to the suggested test.

**The asymmetry is real.** `getattr(entry, "source", None)` returns `None` for a mapping, so a dict entry took the local-path branch. And it would not have been enough to read the source dict-aware in place: `resolve_artifact_path` reads `source` by attribute too, so passing the mapping onward would still ignore it. Normalising is the right fix, so that is what I did — at the single point mappings enter, rather than adding a third dict-aware accessor:

```python
    if isinstance(entry, dict):
        from ..utils.artifacts_meta import CodebaseEntry

        entry = CodebaseEntry.from_dict(entry)
```

Now path, extensions and source are all read the same way, and `resolve_artifact_path` receives the type it is annotated for (`Union[Artifact, CodebaseEntry, Kit]`).

**This also removes a contradiction introduced earlier in this branch**, which is the part I care most about: the empty-entry warning reads `source` from mappings, so before this fix it could have named a workspace source that was never consulted while the resolver scanned a local path. A report that describes something other than what happened is worse than no report.

**On severity — I would call this Minor, not Major.** No mapping reaches this code through configuration: `SystemNode.from_dict` routes every codebase entry through `_parse_autodetect_codebase`, which constructs `CodebaseEntry.from_dict(c)`. So the "validate local code instead of the configured workspace source" outcome is not reachable from a registry today; the dict branches here are defensive. I have fixed it anyway, because defensive code that silently drops a security-relevant selector is worth making correct, but it is a consistency fix rather than a live defect.

**The suggested integration test cannot be written as specified**, for the same reason — there is no way to get a raw mapping into `system_node.codebase` through a registry file. It is covered as a unit test instead: an unavailable source plus a populated local `src/` directory, asserting that nothing is scanned, that the warning names `ghost`, and that the workspace resolver was actually consulted:

```python
        assert results.parsed_code_files_full == []
        assert results.empty_full_codebase_entries == [{"path": "src", "source": "ghost"}]
        session.ws_ctx.resolve_artifact_path.assert_called_once()
```

The decoy file in the local directory is the point: without normalisation that file is scanned and no warning is emitted, which is exactly the failure described.

Mutation-checked: removing the normalisation fails that test and nothing else.

One related observation, not changed here: the `isinstance(entry, dict)` branches inside `_resolve_code_scan_targets` are now unreachable from its only production caller. I have left them alone rather than widen this PR, but they are dead as of this commit and could come out in their own change if you would like.

Gates on the rebased state, Python 3.11: full suite **4,606 passed, 4 skipped, 15 xfailed**; `validate` 0 errors / 192 warnings / 219-219, unchanged from `main`; `spec-coverage` at 90/60/0.46 all thresholds met; `pylint` and `vulture` clean.

PR #103, opened 2026-08-21
spec-coverage: --min-file-granularity cannot be enabled, and whole-file scope claims are invisible

Oleg67:
Related to #76.

**What.** Makes `--min-file-granularity` usable as a gate, and names the files whose coverage rests on a whole-file scope marker. No threshold changes, no metric changes.

**Why.** A file carrying only a scope marker scores granularity **0.0 by definition rather than by measurement** — the metric deliberately refuses to credit a whole-file claim (`utils/coverage.py`: *"Files with only scope markers get granularity 0"*). `_check_min_file_granularity` read that sentinel as a low score, so any positive floor rejected every re-export module and entry point in the tree. The flag is registered, documented in `specs/cli.md`, declared in `features/spec-coverage.md` and covered by tests — and could not be turned on.

Measured here: **10 of 86 files** score 0.0 that way. `--min-file-granularity 0.05` failed all ten before this change; it passes after. A block-traced file below the floor still fails either way — `utils/artifacts_meta.py` at 0.0843 is rejected at 0.10.

**Why the exemption is paired with reporting.** Skipping those files silently would hide the part worth seeing. They are counted as covered and carry no instruction block, so they raise the coverage percentage without being traced to anything, and the summary line cannot tell an entry point from an untraced implementation:

```
  ▸ Whole-file scope claims (10 files, 1612 lines, no instruction tracing)
      skills/studio/scripts/studio/ralphex_export.py  (1154 lines)
      src/studio_proxy/telemetry.py  (165 lines)
      skills/studio/scripts/studio/utils/__init__.py  (127 lines)
      src/studio_proxy/mirrors.py  (113 lines)
      ...
```

The largest is **1,154 lines reported as fully covered with no instruction tracing**, which is the shape this metric exists to discourage. Whether that is worth acting on is the reader's call; the point is that it was not visible.

**Why a per-file floor is worth having at all.** It is judged per file, so adding a well-traced module cannot lower another file's score. A system-wide floor behaves as a shared budget: at zero margin, whichever change rebases last fails regardless of its own tracing quality. See #76 for measurements of that effect.

**Deliberately not attempted.** Classifying which whole-file claims are structurally unmarkable and which are untraced implementations. **Size does not separate them** — the 127-line `utils/__init__.py` has zero functions and zero classes (a pure re-export, nothing to instrument), while the 113-line `mirrors.py` has fourteen. Definition-counting would separate them but needs a parser per language, and the scanner is language-agnostic by design. Naming the files lets a reader decide with the evidence in front of them; a rule that misclassifies on this repository's own data seemed worse than no rule.

**Non-goals.** No change to any threshold default, to the granularity formula, or to what counts as covered. Not excluding structural files from the coverage denominator — that is a separate question.

**Tests.** Two in `TestMinFileGranularity`: a scope-only file is exempt from the per-file floor, and a block-traced file below the floor still fails, so the exemption cannot become a way past the gate. Both mutation-checked — removing the exemption fails the first with `granularity 0.0000 < 0.5000` and leaves the second passing.

**Gates.** `make test` 4652 passed / 4 skipped / 15 xfailed / 58 subtests · `pylint` clean · `vulture-ci` clean · `cfs validate` **0 errors 0 warnings, 220/220** · `validate-toc` correct · per-file line coverage ≥90%.

**This repository's own verdict does not move.** Under the gate `make spec-coverage` runs — `--system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` — the result is coverage 90.4%, granularity 0.4600, all thresholds met, identical to `main`. A change here that altered the existing verdict would be a behaviour change in disguise, so that is asserted rather than assumed.

## Summary by CodeRabbit

* **New Features**
  * Spec coverage reports now list files covered solely by whole-file scope markers, including claimed line counts.
  * These files are ordered by descending claimed line count and clearly identified as lacking instruction tracing.

* **Bug Fixes**
  * Whole-file scope files no longer incorrectly fail minimum per-file granularity checks.

* **Tests**
  * Added coverage for exempt whole-file scope files and low-granularity traced files.

### DIFF
--- a/skills/studio/scripts/studio/commands/spec_coverage.py
+++ b/skills/studio/scripts/studio/commands/spec_coverage.py
@@ -335,6 +335,14 @@ def _check_min_file_granularity(
     for file_coverage in report.per_file:
         if not file_coverage.effective_lines or not file_coverage.covered_lines:
             continue
+        # A scope-only file scores 0.0 by definition rather than by measurement:
+        # the metric deliberately refuses to credit a whole-file claim. Reading
+        # that sentinel as a low score makes any positive floor reject every
+        # re-export module and entry point in the tree, which is why this
+        # threshold is currently unusable as a gate. Those files are reported
+        # under their own heading instead, so exempting them here hides nothing.
+        if file_coverage.has_scope_only:
+            continue
         if file_coverage.granularity >= args.min_file_granularity:
             continue
         failed = True
@@ -489,7 +497,33 @@ def _show_spec_coverage_files(files: dict) -> None:
         ui.step(f"Uncovered files ({len(uncovered)})")
         for path, entry in uncovered.items():
             ui.substep(f"  {path}  ({entry.get('total_lines', 0)} lines)")
+    _show_whole_file_claims(files)
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-human-report-helpers
+
+
+def _show_whole_file_claims(files: dict) -> None:
+    """Name the files whose coverage rests on a whole-file scope marker.
+
+    These are counted as covered but carry no instruction block, so they raise
+    the coverage percentage without being traced to anything. Some are
+    structurally unmarkable -- an entry point, a re-export module -- and some are
+    implementations that were never traced, and the two are indistinguishable
+    from the summary line alone. Listing them by size puts the largest claims in
+    front of the reader instead of leaving them inside an average.
+    """
+    # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-human-report-claims
+    claims = {
+        path: entry for path, entry in files.items()
+        if entry.get("scope_only") and entry.get("covered_lines", 0)
+    }
+    if not claims:
+        return
+    lines_claimed = sum(entry.get("total_lines", 0) for entry in claims.values())
+    ui.blank()
+    ui.step(f"Whole-file scope claims ({len(claims)} files, {lines_claimed} lines, no instruction tracing)")
+    for path, entry in sorted(claims.items(), key=lambda kv: -kv[1].get("total_lines", 0)):
+        ui.substep(f"  {path}  ({entry.get('total_lines', 0)} lines)")
+    # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-human-report-claims
 
 
 def _show_spec_coverage_status(status: str, failures: list, assessed: bool = True) -> None:
--- a/architecture/features/spec-coverage.md
+++ b/architecture/features/spec-coverage.md
@@ -65,7 +65,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - User runs `cfs spec-coverage` → all registered codebase files scanned, coverage report generated with per-file and summary statistics
 - User runs `cfs spec-coverage --min-coverage 80` → same as above, exit code 2 if coverage below threshold
 - User runs `cfs spec-coverage --min-granularity 0.7` → same as above, exit code 2 if granularity below threshold
-- User runs `cfs spec-coverage --min-file-granularity 0.3` → same as above, exit code 2 if any covered file granularity below threshold
+- User runs `cfs spec-coverage --min-file-granularity 0.3` → same as above, exit code 2 if any **block-traced** file's granularity is below threshold. A file whose coverage rests on a whole-file scope marker scores 0.0 by definition rather than by measurement, so it is reported under its own heading instead of judged against this floor — judging it would make any positive floor reject every re-export module and entry point, which is what made this threshold unusable as a gate
 
 **Error Scenarios**:
 - No codebase entries registered → report with `applicable: false` and a hint to configure artifacts.toml; exit code 2 only if a positive threshold was demanded
@@ -96,6 +96,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - [x] - `p1` - Resolve paths relative to project root for human-readable output - `inst-rel-path`
 - [x] - `p1` - Route JSON report to file or terminal UI - `inst-output-report`
 - [x] - `p1` - Format uncovered ranges and render human-friendly per-file/status sections - `inst-human-report-helpers`
+- [x] - `p1` - Name the files whose coverage rests on a whole-file scope marker, largest first, with their count and total claimed lines - `inst-human-report-claims`
 
 ## 3. Processes / Business Logic (CDSL)
 
@@ -248,7 +249,8 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
 - [x] Granularity metric correctly penalizes files with few instructions relative to their size
 - [x] `--min-coverage N` flag causes exit code 2 when coverage is below threshold
 - [x] `--min-granularity N` flag causes exit code 2 when granularity is below threshold
-- [x] `--min-file-granularity N` flag causes exit code 2 when any covered file's granularity is below threshold
+- [x] `--min-file-granularity N` flag causes exit code 2 when any block-traced file's granularity is below threshold; files whose coverage rests on a whole-file scope marker are outside this floor and are listed under "Whole-file scope claims" instead
+- [x] Whole-file scope claims are reported with their count, total claimed lines, and paths ordered largest first, so an untraced claim is visible rather than averaged away
 - [x] `--min-file-coverage N` flag causes exit code 2 when any file's coverage is below threshold
 - [x] `--verbose` flag includes per-file marker details in report
 - [x] Report format mirrors `coverage.py` JSON structure (summary + per-file)
--- a/skills/studio/scripts/studio/utils/coverage.py
+++ b/skills/studio/scripts/studio/utils/coverage.py
@@ -245,6 +245,18 @@ def _scan_scope_markers(lines: List[str]) -> List[int]:
 
 
 # @cpt-begin:cpt-studio-algo-spec-coverage-scan:p1:inst-scan-block-markers
+def _has_block_marker(lines: List[str]) -> bool:
+    """Whether any block marker appears, matched or not.
+
+    Separates "no block tracing here" from "block tracing that did not close",
+    which closed-pair counting cannot tell apart.
+    """
+    return any(
+        _BLOCK_BEGIN_RE.search(line) or _BLOCK_END_RE.search(line)
+        for line in lines
+    )
+
+
 def _scan_block_ranges(lines: List[str]) -> List[Tuple[int, int]]:
     """Scan closed block-marker ranges in a file."""
     return _collect_block_ranges(lines)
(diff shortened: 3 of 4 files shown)


skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -335,6 +335,14 @@ def _check_min_file_granularity(
     for file_coverage in report.per_file:
         if not file_coverage.effective_lines or not file_coverage.covered_lines:
             continue
+        # A scope-only file scores 0.0 by definition rather than by measurement:
+        # the metric deliberately refuses to credit a whole-file claim. Reading
+        # that sentinel as a low score makes any positive floor reject every
+        # re-export module and entry point in the tree, which is why this
+        # threshold is currently unusable as a gate. Those files are reported
+        # under their own heading instead, so exempting them here hides nothing.
+        if file_coverage.has_scope_only:
```
ainetx, line 344: ### Scope-only claims now bypass the documented per-file granularity gate

**Severity:** Major

**Problem**
The new `has_scope_only` branch skips a covered file before the per-file granularity comparison. The feature contract still defines a scope-only file as having granularity `0.0` and says `--min-file-granularity` fails when any covered file is below the requested floor. The new regression test instead expects success for that exact case.

**How to reproduce**
1. Register a codebase containing a file with a whole-file `@cpt-scope` marker and no block markers.
2. Run `cfs spec-coverage --min-file-granularity 0.5`.
3. Observe exit `0` even though the file is covered and has granularity `0.0`.

**Expected behavior**
The command either enforces the documented floor for every covered file, or the feature contract explicitly narrows the floor to measured/block-traced files and tests that new contract.

**Actual behavior**
The scope-only file is skipped from the floor, so the run can pass without evaluating its zero granularity.

```text
scope-only file -> granularity 0.0 -> has_scope_only -> skip floor -> PASS
```

**Impact**
The enabled gate can report success for untraced whole-file claims despite the current artifact contract promising failure, weakening CI's per-file guarantee.

**Suggested correction**
Either remove the exemption, or update the feature contract and acceptance tests to state that scope-only files are intentionally outside the per-file granularity gate; add combined-threshold coverage for the chosen rule.

**How to verify**
Assert both JSON status and exit code for a scope-only file under a positive per-file granularity floor, including a run with per-file coverage enabled.

ainetx, line 344: ### Malformed block markers can inherit the scope-only granularity exemption

**Severity:** Minor

**Problem**
`has_scope_only` is derived from the count of closed block pairs. An unmatched block marker leaves that count at zero, so a file that has both a file-level scope marker and malformed block tracing is treated as scope-only and receives the new granularity exemption.

**How to reproduce**
1. Add a file-level `@cpt-scope` marker and an unmatched `@cpt-begin` or `@cpt-end` marker.
2. Run spec coverage with a positive `--min-file-granularity` floor.
3. The malformed tracing is classified like a deliberate scope-only claim and bypasses the floor.

**Expected behavior**
Only a file with no block-marker attempt should be scope-only; malformed block tracing should remain measurable or produce an explicit invalid-tracing result.

**Actual behavior**
Closed-pair counting erases the distinction between absent and malformed block markers.

```text
scope + unmatched block -> no closed pairs -> has_scope_only -> skip floor
```

**Impact**
Broken instruction tracing can be counted as covered and escape the per-file granularity gate, masking a malformed trace.

**Suggested correction**
Track raw block-marker presence or parse errors separately from closed pairs, and allow the scope-only classification only when no block marker of any kind is present.

**How to verify**
Add a fixture with a scope marker plus an unmatched block marker and assert that it neither becomes scope-only nor bypasses the requested floor.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -489,9 +497,35 @@ def _show_spec_coverage_files(files: dict) -> None:
         ui.step(f"Uncovered files ({len(uncovered)})")
         for path, entry in uncovered.items():
             ui.substep(f"  {path}  ({entry.get('total_lines', 0)} lines)")
+    _show_whole_file_claims(files)
```
ainetx, line 500: ### The new whole-file-claims report path is not defended by tests

**Severity:** Minor

**Problem**
The PR adds a documented human-report requirement and a renderer for whole-file scope claims, but the changed tests only verify the per-file granularity exemption. They do not assert that scope-only files, claim counts, claimed lines, or ordering appear in the human report.

**How to reproduce**
1. Remove the `_show_whole_file_claims(files)` call from `_show_spec_coverage_files`, or make the helper return immediately.
2. Run the current spec-coverage tests.
3. The added tests still pass because they do not exercise the claims-rendering branch.

**Expected behavior**
The documented claims section should have a regression test that fails when the section disappears or misreports its entries.

**Actual behavior**
The human-output test covers only uncovered ranges, while the new claims branch has no `scope_only` fixture or output assertion.

```text
add claims section -> no claims-output assertion -> regression survives CI
```

**Impact**
The visibility feature advertised by this PR can silently regress without a failing test.

**Suggested correction**
Add a human-output test with covered `scope_only` files and assert the claims heading, aggregate claimed lines, listed paths, and largest-first order.

**How to verify**
Delete the claims-helper call or change its heading/filter; the new regression test should fail.

skills/studio/scripts/studio/commands/spec_coverage.py
Oleg67: Accepted, and fixed in `1c24616` by narrowing the contract rather than removing the exemption — with the reasoning, since you offered both.

Removing the exemption puts the flag back where it started: unusable. 10 of 86 files here score 0.0 by definition, so any positive floor rejects every re-export module and entry point, which is the state this PR exists to fix.

So the contract is now narrowed explicitly, in both places you cite:

- the flow success scenario: exit 2 if any **block-traced** file is below the floor, with the reason stated;
- the acceptance criteria: same narrowing, plus a new criterion for the claims report so the exempted files are covered by the contract rather than only by code.

**You are right that I should have done this in the first pass.** The code changed behaviour and the artifact went on describing the old one — the same divergence I corrected on #97, where `inst-return-init-ok` documented a payload no code path emitted. Leaving it would have been that defect, introduced by the person who had just complained about it.

Tests now assert exit code alongside JSON status, and cover the exemption with `--min-file-coverage 50` also requested, so the narrowed floor is not an artefact of running one flag in isolation.

Oleg67: Confirmed and fixed in `1c24616`. `has_scope_only` was `scope_count > 0 and not block_count`, and `block_count` is closed pairs, so a scope marker plus an unmatched `@cpt-begin` was indistinguishable from a deliberate whole-file claim.

Now: `scope_count > 0 and not block_count and not _has_block_marker(lines)` — "scope-only" means no block tracing was *attempted*, not that none of it closed. `_has_block_marker` matches either marker with the existing `_BLOCK_BEGIN_RE`/`_BLOCK_END_RE`, so absent and malformed are separated.

Worth noting the blast radius, since this touches the metric module: **granularity is unchanged in every case.** `_calculate_granularity` already returns 0.0 when no pair closed, whichever way the flag lands, so the score for a malformed file was 0.0 before and is 0.0 now. Only the classification moves — and with it, whether the file is exempt from the floor and whether it appears under "Whole-file scope claims". A malformed file now correctly does neither. This repository is unaffected: coverage 90.4%, granularity 0.4602, unchanged.

Regression test: a file with a scope marker and an unmatched `@cpt-begin` asserts `scope_only` is false and that a `--min-file-granularity 0.5` run exits 2 naming that file. Mutation-checked — reverting to closed-pair counting fails it.

skills/studio/scripts/studio/commands/spec_coverage.py
Oleg67: Fair, and fixed in `1c24616`. I applied mutation checks to the exemption and not to the thing the exemption is justified by, which is the wrong way round: the reporting is what makes exempting those files defensible at all.

New `TestWholeFileClaimsAreReported` asserts the heading, the file count, the aggregate claimed lines, membership, exclusion of block-traced files, and largest-first order. Your mutation is the check I ran: deleting the `_show_whole_file_claims(files)` call now fails it.

**Your mutation also caught a weak assertion I had just written.** My first version compared `output.index("huge.py") < output.rindex("small.py")` across the whole render — and reversing the sort to smallest-first passed anyway, because both files also appear under "Covered files" above, so the indices were found in that section. The assertion is now scoped to the claims section, and reversing the order fails as it should. Recording it here because it is the same failure mode as the finding itself: a test that looks like it defends a behaviour and does not.

Oleg67:
Pushed `fb43aa0`. All three review findings are addressed (replies in their threads), plus the two CI failures from the previous run.

**SonarCloud — real, and mine.** Quality Gate failed on *8.4% duplication on new code* (limit 3%). The four threshold tests I added each repeated the same ~12-line scaffold: temp dir, `src/`, write file, build context, patch twice, parse JSON. Factored out into a `_run_over_src(sources, args)` helper on the base class, so a threshold case is now one fixture and one assertion. The tests are unchanged in behaviour — all four mutations still fail exactly one test each after the rewrite, re-verified.

**Test (Python 3.12) — not from this branch.** The failure was `tests/test_migrate_from_cypilot.py::test_unsupported_legacy_version_update_to_supported_newer_version_migrates`, asserting `status == "PASS"` and getting `"WARN"`. This diff touches `spec_coverage.py`, `utils/coverage.py`, the feature doc and one test file — nothing in the migrate path.

Checked rather than assumed:

- passes locally on 3.12 in isolation, and with `-n 6` across the whole `test_migrate_from_cypilot.py` module;
- full suite on 3.12 locally: 4716 passed;
- `main` already carries `d0f773f3 ci: re-trigger CI (unrelated flaky migrate test on 3.12)`, so this one has been seen before.

So it reads as the known 3.12 flake rather than a regression here. I have not re-triggered CI — say the word if you would like me to, or if you would rather I dig into the flake itself as a separate change. It looks worth doing properly: a test that fails intermittently on one interpreter and gets re-triggered is a gate that does not mean what it says, which is the same family as what this PR is about.

**Gates on `fb43aa0`:** `make test` 4716 passed on 3.11 *and* 3.12 · `pylint` clean · `vulture-ci` clean · `cfs validate` 0 errors 0 warnings 221/221 · `validate-toc` correct · per-file line coverage ≥90% · `spec-coverage --system studio` 90.4% coverage, granularity 0.4602, thresholds met — unchanged from `main`.

ainetx:
## Deep Review Summary

Reviewed at head `466e0bcd` (2026-08-24); author addressed all findings in `fb43aa04` (2026-08-27).

**Findings posted and resolved**

| # | Severity | Finding | Resolution |
|---|---|---|---|
| F-001 | **Major** | Scope-only bypass contradicted the documented per-file granularity gate — any positive `--min-file-granularity` floor silently passed zero-granularity scope-only files | Contract narrowed in `spec-coverage.md` to explicitly scope the floor to block-traced files; rationale documented inline |
| F-002 | Minor | Malformed block markers (unmatched `@cpt-begin`) were classified as whole-file scope claims, inheriting the granularity exemption | `_has_block_marker()` helper added; `has_scope_only` now requires absence of any block marker, not just zero closed pairs; regression test added |
| F-003 | Minor | Whole-file claims reporting path had no regression test — deleting the claims section would not fail CI | `TestWholeFileClaimsAreReported` added, asserting heading, file count, aggregate claimed lines, and largest-first ordering |

All 30 selected checks executed across 4 phases; 27 passed, 3 yielded findings, all verified and resolved by the author.

PR #106, opened 2026-08-27
codebase: three commands walk a registry entry three ways — resolve them all through one exclusion policy

### LINKED ISSUE #100: validate: the code scan bypasses the default ignored directories, and keeps a second copy of init's skip policy
**The gap.** `_resolve_code_scan_targets` in `commands/validate.py` enumerates a codebase entry's files with `rglob` directly:

```python
return [candidate for ext in extensions for candidate in code_path.rglob(f"*{ext}")]
```

Nothing consults `_DEFAULT_IGNORED_DIR_NAMES` from `utils/codebase.py`, so a nested `vendor/`, `dist/`, `build/` or `node_modules/` under a registered entry is scanned as product code. Vendored files then either dilute coverage or produce marker findings against code the project does not own.

**The larger problem.** There are now two skip policies for the same concept, in two files, with no reference between them:

| Consumer | Policy |
|---|---|
| `init` source-root detection | `_CODEBASE_SKIP_ANYWHERE`, `_CODEBASE_SKIP_TOP_LEVEL` (`commands/init.py`) |
| `validate` code scanning | `_DEFAULT_IGNORED_DIR_NAMES` (`utils/codebase.py`), currently unused on this path |

They already disagree, and teaching `rglob` to consult the second list would leave both in place to keep drifting. A user reasonably expects "this directory is not my code" to mean the same thing to detection and to validation.

**The change.** One shared exclusion policy that both detection and scanning consume, applied at any depth, with the scan reporting how many files it excluded so the number is visible rather than inferred.

**Design principles.** A verdict should state its denominator — "scanned N of M, excluded K" is checkable where a bare pass is not. Registration and scanning should not be able to disagree about what counts as code. Exclusions belong in one declared place, not in whichever module needed them first.

**Non-goals.** Not changing the ignore semantics themselves, only where they are defined and that both paths honour them. Not touching per-entry `extensions` resolution. Not adding configuration surface — a project-level override can follow separately if wanted.

**Care needed.** This changes what the gate walks, so it can move coverage numbers repository-wide. It should land with a before/after count of scanned files on this repository and on a fixture with nested vendored code, so the effect is measured rather than assumed. That is also why it was kept out of #97, where it was first noticed: that PR is about `init` registration and does not touch this command.

Oleg67:
Closes #100.

**What.** One resolver for "which files does this codebase entry cover", used by `validate`, `spec-coverage` and `list-ids`/`where-used`. Reports how many candidates it excluded. No threshold changes, no metric changes.

**Why.** Three consumers read the same registry entry and each walked it differently. Only `list-ids`/`where-used` applied `_DEFAULT_IGNORED_DIR_NAMES`; `validate` and `spec-coverage` globbed the entry directly:

```python
return [candidate for ext in extensions for candidate in code_path.rglob(f"*{ext}")]
```

So `vendor/`, `dist/`, `node_modules/` and `__pycache__` beneath a registered root were third-party code to one command and project source to the other two. Registration refuses those directory names and then registers their shallowest parent — scanning that parent recursively puts them straight back.

**Where it bites: a fresh project, not this one.** `cfs init` writes **no `ignore` patterns at all**. So a newly initialised repository has codebase entries and zero exclusions, and vendored code under a registered root was scanned by `validate` and `spec-coverage`, which then demanded `@cpt` markers in third-party files. This repository is unaffected only because its registry hand-lists `skills/studio/scripts/studio/vendor/**` — the two files that pattern covers are now excluded by policy *and counted*, where before they were dropped silently.

**How.** `resolve_entry_code_files()` applies three rules:

| Rule | Why it is separate |
|---|---|
| Conventional non-source directory names, at any depth | The registered parent otherwise re-admits the trees registration refuses |
| Symlinked candidates refused | A link re-opens an excluded tree, and scanning the same file under two names double-counts its lines |
| Resolved containment | A symlinked *directory* leaves the project by a different route than a symlinked file |

It returns `(files, excluded_count)`, and callers report the count: `validate` now prints `Code files: 93 (2 excluded)` and emits `code_files_excluded` in JSON. A scanned total on its own does not say whether the scope was the one the reader assumed.

**Behaviour change worth calling out.** Policy exclusions are counted as *skipped* in `scan_registered_codebase_references` instead of being filtered invisibly. That total is already documented as "ignored, oversized, or unparsable", and an excluded file is the first of those. This also settles an inconsistency the existing tests had encoded — the fail-closed containment path counted its file while the ignored-directory path did not, so `scanned=0, skipped=0` could mean either "nothing here" or "one file we declined to scan". Consequence: `where-used` and `list-ids` will report a non-zero skipped count on projects with vendored code under a registered root.

`_code_paths_for_entry` is removed — it became a wrapper with no callers, and its determinism test now exercises the resolver directly.

**Tests.** New `tests/test_shared_scan_policy.py`: a registered parent not re-admitting `vendor/`/`node_modules/`/`dist/`/`__pycache__` with the excluded count asserted, symlinked files refused, a symlink pointing back *inside* the project still refused because following it double-counts, a single-file entry, a missing entry, overlapping extensions not double-counting, and determinism across repeated runs. Both rules are mutation-checked: removing the ignored-directory rule fails the vendored-code test, removing the symlink/containment rule fails the two symlink tests.

One case is pinned rather than claimed: `rglob`'s `**` does not descend into symlinked directories on the supported versions, so an external tree reached that way is already absent and the containment check never sees it — asserted as `excluded == 0` with the reason stated. The check stays as defence in depth so the guarantee does not rest on that traversal behaviour.

**Gates.** `make test` 4719 passed / 4 skipped / 15 xfailed / 58 subtests · `pylint` clean · `vulture-ci` clean · `cfs validate` **0 errors 0 warnings, 221/221** · `validate-toc` correct · per-file line coverage ≥90%.

**This repository's verdict does not move.** `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` reports 90.4% coverage and granularity 0.4606, all thresholds met, unchanged from `main` apart from the new excluded count. A policy change that altered the existing result would be a behaviour change in disguise, so it is asserted rather than assumed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Code scanning and spec coverage now consistently exclude vendored, generated, cached, and symlinked files.
  * Symlinks resolving outside the project or into excluded directories are no longer scanned.
  * Validation and coverage reports now accurately count and display excluded files, including single-file entries.

* **Tests**
  * Added coverage for exclusions, symlinks, single-file entries, overlapping extensions, repeatability, and consistent scanning results.

### DIFF
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -188,6 +188,8 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 
 **Supporting**:
 - [x] - `p1` - Imports and module setup for query commands (list-ids, where-defined, where-used) - `inst-query-imports`
+- [x] - `p1` - Resolve the code files one registered codebase entry covers, applying the shared exclusion policy and returning how many candidates were excluded - `inst-query-resolve-entry-files`
+- [x] - `p1` - Refuse a candidate that is a symlink or resolves outside the project root, so a link cannot re-open an excluded tree - `inst-query-escapes-project`
 - [x] - `p1` - Argument parsing, context resolution, and artifact collection for query commands - `inst-query-resolve`
 - [x] - `p1` - Deduplicate scan hits per ID, preferring a definition record and surfacing conflicting duplicate definitions - `inst-scan-dedupe`
 - [x] - `p1` - Human-friendly formatters for list-ids, where-defined, and where-used output - `inst-query-format`
--- a/skills/studio/scripts/studio/commands/spec_coverage.py
+++ b/skills/studio/scripts/studio/commands/spec_coverage.py
@@ -11,9 +11,10 @@ import argparse
 import json
 import logging
 from pathlib import Path
-from typing import List
+from typing import List, Tuple
 
 from ..utils import decision_log
+from ..utils.codebase import resolve_entry_code_files
 from ..utils.coverage import (
     FileCoverage,
     calculate_metrics,
@@ -104,7 +105,16 @@ def _collect_codebase_files(
     system_node: object,
     project_root: Path,
     code_files_to_scan: List[Path],
-) -> None:
+) -> int:
+    """Collect this node's code files, returning how many candidates were excluded.
+
+    The count is returned rather than dropped because switching to the shared
+    policy changes the population the coverage percentage is computed over: a
+    registered root holding a vendored subtree contributes fewer files than it
+    used to. A metric whose denominator moves has to say so, or a percentage
+    change looks like the code changed.
+    """
+    excluded = 0
     # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-collect-codebase-files
     for cb_entry in getattr(system_node, "codebase", []):
         path_str = (
@@ -118,17 +128,20 @@ def _collect_codebase_files(
             else cb_entry.get("extensions", None)
         ) or [".py"]
         code_path = _resolve_code_path(project_root, path_str)
-        if not code_path.exists():
-            continue
-        if code_path.is_file():
-            code_files_to_scan.append(code_path)
-            continue
-        for ext in extensions:
-            code_files_to_scan.extend(code_path.rglob(f"*{ext}"))
+        # Resolved through the shared policy rather than a bare rglob, so this
+        # command and `validate` cannot disagree about which files one entry
+        # covers -- and so a registered parent root does not re-admit the
+        # vendored trees that registration itself refuses.
+        entry_files, entry_excluded = resolve_entry_code_files(
+            code_path, extensions, project_root=project_root
+        )
+        excluded += entry_excluded
+        code_files_to_scan.extend(entry_files)
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-collect-codebase-files
     # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-collect-codebase-files
     for child in getattr(system_node, "children", []):
-        _collect_codebase_files(child, project_root, code_files_to_scan)
+        excluded += _collect_codebase_files(child, project_root, code_files_to_scan)
+    return excluded
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-collect-codebase-files
 
 
@@ -148,24 +161,29 @@ def _validate_selected_systems(args, meta) -> tuple[set[str] | None, dict | None
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-validate-systems
 
 
-def _collect_selected_system_files(meta, project_root: Path, system_slugs: set[str] | None) -> List[Path]:
+def _collect_selected_system_files(
+    meta, project_root: Path, system_slugs: set[str] | None
+) -> Tuple[List[Path], int]:
+    """Files the selected systems register, and how many candidates were excluded."""
     # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-resolve-code-files
     code_files_to_scan: List[Path] = []
+    excluded = 0
 
     def visit(node: object) -> None:
+        nonlocal excluded
         if system_slugs is None:
-            _collect_codebase_files(node, project_root, code_files_to_scan)
+            excluded += _collect_codebase_files(node, project_root, code_files_to_scan)
             return
         slug = getattr(node, "slug", "")
         if slug in system_slugs:
-            _collect_codebase_files(node, project_root, code_files_to_scan)
+            excluded += _collect_codebase_files(node, project_root, code_files_to_scan)
             return
         for child in getattr(node, "children", []):
             visit(child)
 
     for system_node in meta.systems:
         visit(system_node)
-    return code_files_to_scan
+    return code_files_to_scan, excluded
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-resolve-code-files
 
 
@@ -392,11 +410,10 @@ def _generate_spec_coverage_report(args, meta, project_root: Path) -> tuple[dict
     if system_slugs == set():
         return {}, 2
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-validate-systems
-    filtered_files = _filter_ignored_files(
-        _collect_selected_system_files(meta, project_root, system_slugs),
-        project_root,
-        meta,
+    collected_files, files_excluded = _collect_selected_system_files(
+        meta, project_root, system_slugs
     )
+    filtered_files = _filter_ignored_files(collected_files, project_root, meta)
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
     if not filtered_files:
         requested = _requested_thresholds(args)
@@ -408,6 +425,9 @@ def _generate_spec_coverage_report(args, meta, project_root: Path) -> tuple[dict
     file_coverages = _scan_file_coverages(filtered_files)
     report = calculate_metrics(file_coverages)
     json_report = generate_report(report, verbose=args.verbose, project_root=project_root)
+    # The population this percentage is computed over, so a shift in it is
+    # attributable rather than looking like a change in the code.
+    json_report["summary"]["files_excluded"] = files_excluded
     status = _apply_thresholds(report, args, project_root, json_report)
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-covered
     if status == "PASS" and report.covered_lines > 0:
@@ -562,7 +582,9 @@ def _human_spec_coverage(data: dict) -> None:
         ui.blank()
 
     summary = data.get("summary", {})
-    ui.detail("Files", f"{summary.get('covered_files', 0)}/{summary.get('total_files', 0)} covered")
+    files_line = f"{summary.get('covered_files', 0)}/{summary.get('total_files', 0)} covered"
+    excluded = summary.get("files_excluded") or 0
+    ui.detail("Files", f"{files_line} ({excluded} excluded)" if excluded else files_line)
     ui.detail("Coverage", f"{summary.get('coverage_pct', 0):.1f}%")
     ui.detail("Granularity", f"{summary.get('granularity_score', 0):.4f}")
 
--- a/skills/studio/scripts/studio/utils/codebase.py
+++ b/skills/studio/scripts/studio/utils/codebase.py
@@ -660,22 +660,100 @@ def _is_in_default_ignored_dir(file_path: Path, root: Path) -> bool:
     return any(part in _DEFAULT_IGNORED_DIR_NAMES for part in rel_parts[:-1])
 
 
-def _code_paths_for_entry(code_path: Path, extensions: List[str]) -> List[Path]:
-    """Return code files covered by one registry codebase entry, sorted for determinism."""
+# @cpt-end:cpt-studio-flow-traceability-validation-query:p1:inst-query-load-context
+
+
+# @cpt-begin:cpt-studio-flow-traceability-validation-query:p1:inst-query-resolve-entry-files
+def resolve_entry_code_files(
+    code_path: Path,
+    extensions: List[str],
+    *,
+    project_root: Path,
+) -> Tuple[List[Path], int]:
+    """Code files one registry codebase entry covers, and how many were excluded.
+
+    Every consumer of a codebase entry resolves it through here, so that a
+    directory the policy excludes cannot be skipped by one command and scanned
+    by another. Three rules apply:
+
+    * conventional non-source directory names, at any depth -- a registered
+      parent root otherwise re-admits the `vendor/` and `dist/` trees that
+      registration itself refuses;
+    * symlinked candidates, so one link cannot re-open an excluded tree or reach
+      outside the project under a name that looks local;
+    * resolved containment, because a symlinked *directory* escapes the project
+      by a different route than a symlinked file.
+
+    The excluded count is returned rather than logged so the caller can report
+    its own denominator: "scanned N, excluded M" is checkable where a bare file
+    count is not.
+    """
+    root = project_root.resolve()
     if not code_path.exists():
-        return []
+        return [], 0
+    # The entry itself is judged before anything is read, so a `../..` entry is
+    # refused rather than walked and then discarded file by file. Checking only
+    # the candidates meant an external tree was traversed first, and every file
+    # in it counted separately toward a total describing this project.
+    #
+    # This also covers the single-file case, which previously returned the path
+    # unconditionally: an entry resolving to a file outside the project was
+    # refused by `list-ids` and read by `validate` and `spec-coverage`.
+    if _escapes_project(code_path, root):
+        return [], 1
     if code_path.is_file():
-        return [code_path]
+        return [code_path], 0
 
-    files: List[Path] = []
+    # Candidates are collected before they are judged, so each one is decided
+    # once. Judging inside the extension loop counted a single excluded file
+    # once per matching extension, which made the excluded total disagree with
+    # the file list it is meant to be the denominator of.
+    candidates: Set[Path] = set()
     for ext in extensions:
-        for candidate in code_path.rglob(f"*{ext}"):
-            if _is_in_default_ignored_dir(candidate, code_path):
-                continue
-            files.append(candidate)
-    return sorted(files, key=str)
+        candidates.update(code_path.rglob(f"*{ext}"))
+
+    files: Set[Path] = set()
+    excluded = 0
+    for candidate in candidates:
+        if _is_in_default_ignored_dir(candidate, code_path):
+            excluded += 1
+            continue
+        if _escapes_project(candidate, root):
+            excluded += 1
+            continue
+        files.add(candidate)
+    return sorted(files, key=str), excluded
 
 
+# @cpt-end:cpt-studio-flow-traceability-validation-query:p1:inst-query-resolve-entry-files
+
+
+# @cpt-begin:cpt-studio-flow-traceability-validation-query:p1:inst-query-escapes-project
+def _escapes_project(candidate: Path, root: Path) -> bool:
+    """Whether *candidate* is a symlink or resolves outside *root*.
+
+    Checked on the resolved path: refusing symlinked files alone still lets a
+    symlinked directory inside a registered root carry the scan outside the
+    project.
+    """
+    try:
+        if candidate.is_symlink():
+            return True
+        resolved = candidate.resolve()
+    except OSError as exc:
+        _warn_codebase(f"failed to resolve {candidate}: {exc}")
+        return True
+    if resolved == root:
+        return False
+    return root not in resolved.parents
+
+
+
+
+# @cpt-end:cpt-studio-flow-traceability-validation-query:p1:inst-query-escapes-project
+
+
+# @cpt-begin:cpt-studio-flow-traceability-validation-query:p1:inst-query-load-context
 def _is_ignored_code_file(file_path: Path, ctx) -> bool:
     """Return whether *file_path* is registry-ignored and should be skipped.
 
@@ -752,7 +830,13 @@ def _scan_codebase_entries(scan_ctx) -> Tuple[List[Dict[str, object]], int, int]
         except ValueError:
             _warn_codebase(f"codebase entry {cb_entry.path!r} resolves outside {root}; skipping")
             continue
-        for file_path in _code_paths_for_entry(code_path, cb_entry.extensions or [".py"]):
+        entry_files, entry_excluded = resolve_entry_code_files(
+            code_path, cb_entry.extensions or [".py"], project_root=root
+        )
+        # Counted as skipped: that total already means "ignored, oversized, or
+        # unparsable", and a policy exclusion is the first of those.
+        skipped += entry_excluded
+        for file_path in entry_files:
             file_hits = _scan_code_file_references(file_path, scan_ctx)
             if file_hits is None:
                 skipped += 1
--- a/tests/test_shared_scan_policy.py
+++ b/tests/test_shared_scan_policy.py
@@ -1,0 +1,323 @@
+"""One policy decides which files a codebase entry covers.
+
+`init` writes `codebase` entries and no `ignore` patterns, so in a fresh project
+the only thing standing between a registered root and its vendored or generated
+subdirectories is the default exclusion policy. That policy lived in
+`utils/codebase.py` and was applied by one of the three consumers of a registry
+entry: `list-ids`/`where-used` honoured it, while `validate` and `spec-coverage`
+globbed the entry directly. The same entry therefore covered different files
+depending on which command read it.
+
+These tests pin the shared resolver so the three cannot drift apart again.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "skills" / "studio" / "scripts"))
+
+from studio.utils.codebase import resolve_entry_code_files
+
+
+def _tree(root: Path, files: list[str]) -> None:
+    for rel in files:
+        target = root / rel
+        target.parent.mkdir(parents=True, exist_ok=True)
+        target.write_text("x = 1\n", encoding="utf-8")
+
+
+def _names(files: list[Path], root: Path) -> list[str]:
+    return sorted(f.relative_to(root).as_posix() for f in files)
+
+
+class TestExcludedDirectoriesStayExcluded:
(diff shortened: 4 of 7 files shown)


skills/studio/scripts/studio/utils/codebase.py
```diff
@@ -660,22 +660,83 @@ def _is_in_default_ignored_dir(file_path: Path, root: Path) -> bool:
...
+    The excluded count is returned rather than logged so the caller can report
+    its own denominator: "scanned N, excluded M" is checkable where a bare file
+    count is not.
+    """
     if not code_path.exists():
-        return []
+        return [], 0
     if code_path.is_file():
-        return [code_path]
+        return [code_path], 0
```
ainetx, line 705: ### `resolve_entry_code_files()` skips every exclusion rule when a codebase entry resolves directly to a file outside `project_root`

**Severity:** Major

**Problem**
This function is meant to be the single shared policy enforcing three rules for every codebase-entry consumer: exclude conventional non-source directories, refuse symlinked candidates, and refuse candidates whose resolved path escapes `project_root`. But the single-file branch here (`if code_path.is_file(): return [code_path], 0`) returns the path immediately without running any of the three checks. The only thing currently preventing an out-of-root single-file entry from being scanned is an unrelated pre-check in `_scan_codebase_entries` (`code_path.relative_to(root)`) that sits *before* this function is ever called, on the `list-ids`/`where-used` path only. `validate.py`'s `_resolve_code_scan_targets` and `spec_coverage.py`'s `_collect_codebase_files` call this function directly with no equivalent pre-check.

**How to reproduce**
1. Register a codebase entry whose path resolves to a single file outside the project (an absolute path, a `../` traversal, or a cross-repo workspace-source resolution landing outside the current root).
2. Run `validate` or `spec-coverage`: they resolve `code_path`, find it `is_file()`, and this function returns `[code_path]` unconditionally.
3. Run `list-ids --include-code` / `where-used --include-code` on the same entry: `_scan_codebase_entries` does `code_path.relative_to(root)` first and skips the entry with a warning before ever reaching this function.
4. `tests/test_shared_scan_policy.py`'s only single-file case (`test_an_entry_naming_a_single_file_resolves_to_it`) uses a file *inside* root, so this path isn't covered by any test.

**Expected behavior**
A single-file codebase entry that resolves outside `project_root` should be refused identically by `validate`, `spec-coverage`, and `list-ids`/`where-used`, since this function is meant to be the one policy all three consumers share.

**Actual behavior**
`validate` and `spec-coverage` read, parse, and include an out-of-project file's path (and its content, via error/coverage messages) in their reports, while `list-ids`/`where-used` silently skip the same entry — the "resolved containment" guarantee in this function's own docstring doesn't hold for the single-file branch.

```text
codebase entry (path resolves to a file outside project_root)
        │
        ├─► validate / spec-coverage
        │        │
        │        └─► resolve_entry_code_files()
        │                 │
        │                 └─► is_file() ──► return [code_path], 0   (no containment/symlink check!)
        │                                         │
        │                                         └─► file scanned & reported
        │
        └─► list-ids / where-used
                 │
                 └─► _scan_codebase_entries()
                          │
                          └─► code_path.relative_to(root) raises ValueError ──► entry skipped
```

**Impact**
A misconfigured or edited codebase-registry entry (including one produced by workspace/cross-repo source resolution) can cause `validate`/`spec-coverage` to read and surface content from arbitrary files outside the project root, and produces silently inconsistent behavior between the three commands this PR sets out to unify.

**Suggested correction**
In `resolve_entry_code_files()`, apply the same containment/symlink checks to the single-file case before returning it — e.g. call `_escapes_project(code_path, root)` in the `code_path.is_file()` branch instead of returning `[code_path]` unconditionally. The redundant pre-check in `_scan_codebase_entries` can then stay as defense-in-depth without being the only enforcement point.

**How to verify**
Add a test mirroring `test_an_entry_naming_a_single_file_resolves_to_it` but with the file entry resolving outside `project_root` (e.g. via a symlink or an absolute path pointing elsewhere), and assert it is refused the same way the directory case already is.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -118,13 +119,14 @@ def _collect_codebase_files(
...
-        if code_path.is_file():
-            code_files_to_scan.append(code_path)
-            continue
-        for ext in extensions:
-            code_files_to_scan.extend(code_path.rglob(f"*{ext}"))
+        # Resolved through the shared policy rather than a bare rglob, so this
+        # command and `validate` cannot disagree about which files one entry
+        # covers -- and so a registered parent root does not re-admit the
+        # vendored trees that registration itself refuses.
+        entry_files, _excluded = resolve_entry_code_files(
```
ainetx: ### `spec-coverage`'s excluded-file count is discarded, hiding a denominator shift

**Severity:** Minor

**Problem**
Before this PR, `_collect_codebase_files` fed `calculate_metrics` with every file matched by a bare `code_path.rglob(f"*{ext}")` — no exclusion of vendor/build/generated directories, symlinks, or path escapes. Switching to `resolve_entry_code_files` (the same policy `validate.py` now uses) means any registered codebase root containing a conventional non-source subdirectory (`vendor/`, `node_modules/`, `dist/`, `.venv/`, `__pycache__/`, etc.) with matching-extension files will have those files silently drop out of the coverage population. `validate.py` was given a `code_files_excluded` field on its report so this is visible there; `spec_coverage.py` binds the same return value to `_excluded` here and never surfaces it anywhere.

**How to reproduce**
1. Register a codebase entry whose root contains a `vendor/` or `node_modules/` subdirectory with files matching the scanned extensions.
2. Run `spec-coverage` before and after this change on that project.
3. Before: vendored files are included in the population feeding `calculate_metrics`. After: they're silently excluded, changing the total file count and therefore the coverage percentage (and pass/fail outcome against `--min-coverage`/`--min-file-coverage` thresholds) — with no field in the JSON or human report indicating anything was excluded, or how much.

**Expected behavior**
Excluding vendor/generated trees from the coverage denominator is a reasonable, arguably overdue correction — but a metric whose file population changes should say so. `spec-coverage`'s report should expose the same kind of excluded-count field `validate`'s report now gets, so a coverage percentage change (or a threshold flipping pass/fail) is traceable to this policy change rather than silently absorbed.

**Actual behavior**
`_excluded` is bound and discarded at this call site; `spec-coverage`'s report carries no scanned/excluded file-count field at all.

```text
   before this PR                          after this PR
   rglob(*.py) ──► all files          resolve_entry_code_files() ──► fewer files
        │                                       │         │
        ▼                                       ▼         ▼
  calculate_metrics(all)              calculate_metrics(filtered)   _excluded (discarded)
        │                                       │
        ▼                                       ▼
   coverage % = X                    coverage % = Y  (X ≠ Y, no signal why)
```

**Impact**
Coverage percentages and `--min-coverage`/`--min-file-coverage`/`--min-granularity` threshold outcomes can shift for existing projects with vendored/generated code inside a registered root, with no reporting signal explaining why — risking a confusing CI pass/fail flip that looks like a regression (or a false green) rather than a deliberate scope correction.

**Suggested correction**
Accumulate the excluded counts returned by `resolve_entry_code_files` across entries in `_collect_codebase_files`, thread the total into the report builder, and add a `code_files_excluded` (or similarly named) field to `spec-coverage`'s JSON and human-readable output, mirroring `validate.py`.

**How to verify**
After adding the field, re-run `spec-coverage` on a project with a vendored subdirectory and confirm the report states how many files were excluded, alongside the (now smaller) scanned count.

architecture/features/traceability-validation.md
```diff
@@ -178,6 +178,8 @@ Catches structural and traceability issues that AI agents miss or hallucinate 
 **Steps**:
 1. [x] - `p1` - User invokes one of: `list-ids [--kind K] [--pattern P]`, `where-defined --id <id>`, `where-used --id <id>`, `get-content --id <id>` - `inst-user-query`
 2. [x] - `p1` - Load project context and resolve all registered artifacts - `inst-query-load-context`
+2. [x] - `p1` - Resolve the code files one registered codebase entry covers, applying the shared exclusion policy and returning how many candidates were excluded - `inst-query-resolve-entry-files`
+2. [x] - `p1` - Refuse a candidate that is a symlink or resolves outside the project root, so a link cannot re-open an excluded tree - `inst-query-escapes-project`
```
ainetx: ### Duplicate step numbers ("2.") in the Query Traceability steps list

**Severity:** Minor

**Problem**
The "Steps" list for the Query Traceability flow numbers three consecutive entries as "2." before jumping to "3.", breaking the sequential step-numbering convention used throughout the rest of this document.

**How to reproduce**
View this section: step `inst-query-load-context` is "2.", the newly added step `inst-query-resolve-entry-files` is also "2.", the newly added step `inst-query-escapes-project` is also "2.", and only the following step `inst-scan-all` resumes at "3.".

**Expected behavior**
Sequential numbering: 1, 2, 3, 4, 5, ... — the two new steps describing the shared exclusion-policy resolution should be numbered 3 and 4, with every subsequent step renumbered accordingly.

**Actual behavior**
Steps are numbered 1, 2, 2, 2, 3, 4, 5, 6, 7, 8, 9 — three steps share the number "2.".

```text
1. user-query
2. load-context
2. resolve-entry-files   ← duplicate
2. escapes-project       ← duplicate
3. scan-all
...
```

**Impact**
Purely cosmetic/documentation — doesn't affect code behavior or CI, but makes the step sequence ambiguous for readers and any tooling that might parse step order by number, and is inconsistent with the rest of the document.

**Suggested correction**
Renumber the steps sequentially: keep 1 as-is, number the two new steps 2 and 3, and increment every subsequent step number (former 3→4, 4→5, 5→6, 6→7, 7→8, 8→9, 9→10) so the list reads 1 through 10 without repeats.

**How to verify**
Re-read the Steps section for Query Traceability and confirm the numeric prefixes run 1, 2, 3, 4, ... with no repeated or skipped numbers.

tests/test_shared_scan_policy.py
```diff
@@ -0,0 +1,175 @@
...
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "skills" / "studio" / "scripts"))
+
+from studio.utils.codebase import resolve_entry_code_files
```
ainetx, line 23: ### No end-to-end test confirms `validate` and `spec-coverage` actually agree on file counts

**Severity:** Minor

**Problem**
This PR's stated motivation is that `validate` and `spec-coverage` used to scan a registry entry differently, so "the same entry therefore covered different files depending on which command read it." The fix routes both through `resolve_entry_code_files`, and this new test module pins that resolver's behavior in isolation — but no test drives `validate`'s and `spec-coverage`'s actual report-generation paths against one shared fixture tree and asserts their file counts/lists agree.

**How to reproduce**
Inspect this file and `tests/test_codebase.py`: every test case calls `resolve_entry_code_files` (or `scan_registered_codebase_references`) directly and asserts on its return value. Neither file constructs a fixture project and invokes both commands' report-building logic against it to compare outputs.

**Expected behavior**
A regression test that builds one fixture tree, runs both commands' report/collection logic against it, and asserts the resulting code-file counts (or file sets) are identical — directly guarding the "commands must agree" property this PR exists to fix.

**Actual behavior**
Coverage stops at the shared resolver; nothing exercises the two call sites end-to-end together, so a future change that has one command pass different resolver arguments (extensions, `project_root`) than the other, or post-process the resolver's output differently, wouldn't be caught by any existing test.

```text
   validate.py ──┐
                 ├──► resolve_entry_code_files()  ← pinned by unit tests
   spec_coverage.py ──┘
                 │
                 X   no test asserts validate's and spec-coverage's
                     own report outputs agree on the same fixture
```

**Impact**
Low risk today, since both commands call the exact same function with directly comparable arguments — the resolver's own unit tests already cover most regression scenarios. The gap only matters for a narrower future failure mode: divergence introduced in how each command *calls* or *post-processes* the resolver's output, which current unit tests can't detect. Not a functional defect in this PR.

**Suggested correction**
Add an integration-style test (e.g. in this file) that builds one temp project fixture, invokes `validate`'s scan/report path and `spec-coverage`'s collection path against it, and asserts the two report equal file counts (and ideally identical file sets) for the same registry entry.

**How to verify**
After adding the test, deliberately pass mismatched arguments (e.g. different `extensions`) from one command's call site and confirm the new test fails, proving it actually detects cross-command divergence.

skills/studio/scripts/studio/utils/codebase.py
Oleg67: Confirmed and fixed in `0a3a91d`. This is a hole in my own work, and your trace of it is exactly right: the single-file branch returned before any of the three rules ran, so the containment guarantee in that function's own docstring did not hold for the case the docstring most obviously covered.

Worse than the omission is what hid it. I wrote `test_an_entry_naming_a_single_file_resolves_to_it` using a file *inside* the root, so the single-file branch looked tested. A test that exercises only the passing side of a policy is the failure mode this whole PR is about, and I put one in the PR that fixes it.

Fixed as you suggested — `_escapes_project(code_path, root)` now guards the single-file branch, returning `([], 1)` so the refusal is counted rather than silent. `root` is resolved before the early returns so both branches share it. The pre-check in `_scan_codebase_entries` stays as defence in depth rather than being the only enforcement.

New test `test_a_single_file_entry_outside_the_project_is_refused` builds the file outside the project and asserts refusal plus the count. Mutation-checked — restoring the unconditional `return [code_path], 0` fails it and nothing else.

skills/studio/scripts/studio/commands/spec_coverage.py
Oleg67: Correct on both counts, and the second one matters more than the missing field: **my PR body claimed "no metric changes", and that is wrong in general.** It is true for this repository only because its registry hand-lists `vendor/**`, so the excluded files were already outside the population. On any project with a vendored subtree under a registered root and no matching ignore pattern — which is every freshly `init`-ed project, since `init` writes no ignore patterns — the coverage denominator moves. Thank you for catching a claim I should not have made.

Fixed in `0a3a91d`. `_collect_codebase_files` now returns its excluded count, `_collect_selected_system_files` sums it, and it reaches the report:

```
  Files: 87/87 covered (2 excluded)
```

plus `summary.files_excluded` in JSON, matching the shape `validate` got. So a coverage percentage that shifts is attributable to the policy rather than looking like the code changed.

I put it in `summary` beside the other population figures rather than at the top level, since it describes the same denominator as `total_files`. Say the word if you would rather it sat elsewhere in the schema.

architecture/features/traceability-validation.md
Oleg67: Fixed in `0a3a91d`, though not by renumbering — worth explaining in case you prefer your version.

The two entries I added are not stages of the query flow; they are helpers the flow uses. This document already has a **Supporting** section for exactly that, holding `inst-query-imports`, `inst-query-resolve`, `inst-scan-dedupe` and `inst-query-format`. So I moved both there instead. The Steps list is back to 1, 2, 3 … 9 with no repeats and no renumbering of the seven existing steps, and the two instructions are now classified the way the document classifies things of their kind.

That also avoids a diff touching seven unrelated step numbers, which would have made the change harder to review than the change deserves. `validate-toc` passes and `cfs validate` still resolves 221/221 markers, so both instruction declarations are still bound to their code.

If you would rather they were numbered flow steps after all, I will renumber instead — it is a one-line preference either way.

tests/test_shared_scan_policy.py
Oleg67: Fair, and added in `0a3a91d` — with your verification, not just your suggestion.

`TestTheCommandsAgreeOnOneEntry::test_validate_and_spec_coverage_select_the_same_files` builds one fixture (real source, nested `vendor/`, `node_modules/`, `dist/`, plus a symlink), drives `validate`'s `_resolve_code_scan_targets` and `spec-coverage`'s `_collect_codebase_files`, and asserts the two resolved sets are identical *and* that the policy is actually in force in both (`{main.py, mod.py}`).

I drove the two call sites rather than the two full commands. The failure mode you describe is divergence in how each command *calls or post-processes* the resolver, and the call sites are where that lives; going through both full report builders would add context and registry scaffolding without covering anything more of that surface. If you would rather see it at full-command level, I will take it further.

Ran your verification exactly: changing `validate`'s call site to pass `[".ts"]` instead of the entry's extensions fails the new test, so it does detect cross-command divergence rather than just re-testing the resolver.

**Gates on `0a3a91d`** (rebased onto current `main`): 4767 passed on 3.11 and 3.12 · `pylint` clean · `vulture-ci` clean · `cfs validate` 0 errors 0 warnings, 221/221, `Code files: 93 (2 excluded)` · `validate-toc` correct · per-file coverage ≥90% · `spec-coverage --system studio` 90.4%, granularity 0.4613, thresholds met.

Oleg67:
Pushed `0a3a91d`, rebased onto current `main`. All four findings addressed; replies in their threads.

**One correction to this PR's own description, which the review earned.** The body said "no metric changes". That is true for *this* repository and not in general: its registry hand-lists `skills/studio/scripts/studio/vendor/**`, so the newly-excluded files were already outside the coverage population. On a project with a vendored subtree under a registered root and no matching ignore pattern the denominator does move — and since `cfs init` writes **no** ignore patterns, that is every freshly initialised project. Both commands now report the count so the shift is attributable:

```
cfs validate         Code files: 93 (2 excluded)
cfs spec-coverage    Files: 87/87 covered (2 excluded)
```

Summary of the four:

| Finding | Resolution |
|---|---|
| Single-file entry bypassed all three rules (Major) | `_escapes_project` now guards that branch, returning `([], 1)` so the refusal is counted |
| `spec-coverage` discarded the excluded count | Threaded through to `summary.files_excluded` and the human line |
| Duplicate step numbers | Both entries moved to **Supporting**, where this document keeps helpers — no renumbering of existing steps |
| No cross-command agreement test | Added, driving both call sites over one fixture |

Both new tests are mutation-checked, including the verification suggested for the agreement test: passing `[".ts"]` from `validate`'s call site makes it fail, so it detects divergence rather than re-testing the resolver.

**Gates:** 4767 passed on 3.11 *and* 3.12 · `pylint` clean · `vulture-ci` clean · `cfs validate` 0 errors 0 warnings 221/221 · `validate-toc` correct · per-file line coverage ≥90% · `spec-coverage --system studio` 90.4% coverage, granularity 0.4613, all thresholds met.

skills/studio/scripts/studio/utils/codebase.py
```diff
@@ -660,22 +660,90 @@ def _is_in_default_ignored_dir(file_path: Path, root: Path) -> bool:
...
             if _is_in_default_ignored_dir(candidate, code_path):
+                excluded += 1
                 continue
-            files.append(candidate)
-    return sorted(files, key=str)
+            if _escapes_project(candidate, root):
+                excluded += 1
+                continue
+            files.add(candidate)
+    return sorted(files, key=str), excluded
```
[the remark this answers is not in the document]
Oleg67: Two claims here — one is a real bug I have fixed, the other I am declining with reasoning. Both checked empirically first.

**Double-counted exclusions: correct, fixed in `28ab682e`.** Reproduced before fixing:

```
one ext : files=1 excluded=1
dup ext : files=1 excluded=2      <- files deduplicated, count not
```

The file set was a `set` and the counter was incremented inside the extension loop, so a single excluded file was counted once per matching extension. The denominator disagreed with the list it is meant to describe. Candidates are now collected into a set and judged once each, as you suggest.

Notably my own `test_overlapping_extensions_do_not_double_count` asserted the file list and not the count, so it passed while the bug was live — a test named for the property it failed to check. It now asserts that repeating an extension changes neither the list nor the count. Mutation-checked: moving the judgement back inside the extension loop fails it.

**Ignored-directory check relative to `project_root`: declining, and here is the case that decides it.**

```
entry = "src/vendor/mylib"  ->  files=[core.py]  excluded=0
```

That is a project deliberately registering code that lives under a conventionally-ignored name. Under `root`, `vendor` becomes an ignored component of the path and the entry resolves to **zero files** — an explicit registration silently emptied by a default name list. Vendoring your own library under `vendor/`, or keeping generated-then-committed sources under `dist/`, stops working with no diagnostic beyond an empty scan.

The current semantics read as deliberate to me: the exclusion policy applies *within* what you registered, not to your choice of what to register. An entry is an explicit statement that this is product code, and a default list of conventional names should not override it.

Two further reasons to keep it out of this PR: the `_is_in_default_ignored_dir(candidate, code_path)` call predates this change — I preserved it while converging the three consumers onto one resolver — so switching its base is a behaviour change beyond the convergence, and it would newly break explicit registrations that work today.

Happy to be wrong about the intent. If maintainers prefer `root`, the honest version is a separate change that keeps explicit registration working — e.g. exempting the entry root's own path components from the check, so `src/vendor/mylib` scans while `src` still refuses its nested `vendor/`. I would rather ship that deliberately than fold it in here.

**Gates on `28ab682e`:** 4767 passed · `pylint` clean · `vulture` clean · `cfs validate` 0 errors 0 warnings, 221/221, `Code files: 93 (2 excluded)` · `spec-coverage --system studio` 90.4%, granularity 0.4613, thresholds met.

skills/studio/scripts/studio/utils/codebase.py
```diff
@@ -660,22 +660,97 @@ def _is_in_default_ignored_dir(file_path: Path, root: Path) -> bool:
...
+    # once per matching extension, which made the excluded total disagree with
+    # the file list it is meant to be the denominator of.
+    candidates: Set[Path] = set()
     for ext in extensions:
-        for candidate in code_path.rglob(f"*{ext}"):
-            if _is_in_default_ignored_dir(candidate, code_path):
-                continue
-            files.append(candidate)
-    return sorted(files, key=str)
+        candidates.update(code_path.rglob(f"*{ext}"))
```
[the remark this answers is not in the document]
Oleg67: Confirmed and fixed in `bb49f40`. `_escapes_project(code_path, root)` now runs before the `is_file()` branch and before any traversal, returning `([], 1)`.

Two consequences beyond the traversal itself, which is why this was worth more than moving one line:

- an escaping *directory* entry was walked and then discarded file by file, so `excluded` was incremented once per file in an external tree — a count that is supposed to describe this project was reporting the size of someone else's;
- my own earlier fix guarded containment for the single-file case only, so the asymmetry was mine: files judged, directories walked first.

Regression test `test_an_escaping_directory_entry_is_refused_without_being_walked` patches `Path.rglob` to raise, so it proves traversal never starts rather than merely that the result came back empty. Mutation-checked — moving the check back below `is_file()` fails it.

tests/test_shared_scan_policy.py
```diff
@@ -0,0 +1,252 @@
...
+            session.ws_ctx = None
+            session.code_files_excluded = 0
+            validate_files = {p.resolve() for p in _resolve_code_scan_targets(session, entry)}
+
+            node = MagicMock()
+            node.codebase = [entry]
+            node.children = []
+            collected: list = []
+            _collect_codebase_files(node, root, collected)
+            coverage_files = {p.resolve() for p in collected}
```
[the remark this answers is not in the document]
Oleg67: This is the sharpest finding on the PR, and you are right. Fixed in `bb49f40`.

`validate._resolve_code_scan_targets` had `if code_path.is_file(): return [code_path]` **before** it ever called the resolver. So the single-file containment fix I made earlier in response to the maintainer's Major finding did not apply to `validate` at all — the exact command that finding was about. `spec-coverage` called the resolver, `validate` did not, and the two still disagreed on the one shape I had just claimed to have unified.

My agreement test could not catch it because I wrote the fixture with a directory entry. A test asserting "the commands agree" that only exercises the shape where they already agreed.

Removed the early return, as you suggest, so file entries go through the resolver in both commands. Added `test_validate_and_spec_coverage_agree_on_a_single_file_entry`, parametrised over an inside-project file and an external one, asserting the two commands select identical sets in both cases and that the external entry is refused. Mutation-checked — restoring the early return fails it.

**One existing test had to change, and I want to flag it rather than let it pass as noise.** `test_validate_code_scan_branches_and_failure_includes_warnings` registered its codebase entry as a file in a *second temp directory*, outside the project root, and asserted `validate` scanned it and exited 2. That fixture encoded the behaviour this thread asks me to refuse. Its purpose is the code-scan failure branch, which does not need an out-of-project path, so I moved the file inside the project root and left a comment saying why. If you would rather that test keep asserting external scanning, this change needs re-thinking — but I read it as incidental setup rather than intended contract.

**Gates on `bb49f40`:** 4769 passed on 3.11 (×3 runs) and 3.12 · `pylint` clean · `vulture` clean · `cfs validate` 0 errors 0 warnings, 221/221, `Code files: 93 (2 excluded)` · per-file coverage ≥90% · `spec-coverage --system studio` 90.4%, granularity 0.4613.

Unrelated: `test_init_update_legacy_studio_yes_updates_baseline_then_migrates` failed once in a parallel run and passes in isolation and in every run since — the same intermittent shape as the migrate test `main` already re-triggers CI over.

Issue #104, opened 2026-08-24
Large-document context overload in Constructor Studio: diagnosis and a benchmarked read-once-per-file token-reduction approach


tkcoding:
## Summary
This issue covers two connected investigations: (1) diagnosing why Constructor Studio cannot read large/oversized documents without exploding LLM context, and (2) building and benchmarking a read-once-per-file approach to reduce context window usage once a document can be read at all.

---

## Part 1 — The context-overload problem (diagnosis)

Neither `cf-explain` nor `cfs chunk-input` can handle a large document without either blocking outright or crashing — tested against a real 166-page / 3.6 MB PDF.

- **`cf-explain`**: E0 size guard blocks on *raw binary file size* before any content is read (`block_too_large` at >256 KB / >3000 lines). Even after clean text extraction (600 KB / 8439 lines), the plain-text version still exceeds the guard — the threshold itself is too low for a document this size, not just a binary-format problem.
- **`cfs chunk-input`**: pure UTF-8 text tool, hard-crashes on binary input (`'utf-8' codec can't decode byte...`). No PDF/binary path at all. Works fine on plain text (verified: 4050-line file → 14 chunks cleanly).
- **Dependency gap**: `pdftotext`/`pdfinfo` come from Homebrew's `poppler` — not a Studio dependency, just happened to be installed on the test machine. Not guaranteed on a fresh clone, CI, or a locked-down environment.

---

## Part 2 — Reducing context windows: read once per file, not per query

Once a document's *text* is available (Markdown, or any already-plain-text artifact), the second problem is avoiding a full-context load on every query. Two things were built, then benchmarked against a real 166-page document.

### Built

- **`cfs validate-toc` (extended)**: 4 new structural JIT-readiness checks — duplicate headings, depth jumps, oversized sections, missing description — warnings only, zero regressions, 132 tests passing.
- **`cfs doc-index` (new)**: a cached, read-once-per-file heading/section index, keyed by content etag. Builds once, reused on every subsequent call until the file actually changes (proven: 17.3ms cold build vs 7.9ms cache hit, correct invalidation on content edits).

### Benchmarked (converted 166-page document, 5 representative queries)

| | Baseline (whole doc) | Heading-nav | OKF bundle |
|---|---:|---:|---:|
| Total (5 queries) | 2,842,545 bytes | 161,725 bytes | 35,983 bytes |
| Reduction vs baseline | — | 94.3% | **98.7%** |
[...]

PR #108, opened 2026-08-28
feat(toc,doc-index): add JIT-retrieval readiness checks and cached doc index

tkcoding:
## Summary
- `toc.py` gains four warning-only JIT-retrieval readiness signals: duplicate heading titles, heading depth jumps, oversized sections (configurable `--max-section-lines`, default 300), and a missing top-of-file description — structural properties that make heading-based JIT retrieval harder without invalidating otherwise-valid documents.
- New `doc_index.py` (util + `cfs doc-index` command): a cached, etag-invalidated structural index (headings + section line ranges) so navigation reads a file's structure once per file, not once per query, with an `annotate_section_summary` hook for a future LLM caller to attach per-section summaries.

See constructorfabric/studio#104.

## Test plan
- [x] `pytest tests/test_toc.py tests/test_doc_index.py` — 132 passed
- [x] Verified `doc-index`/`eval` both resolve correctly via the CLI dispatch table after rebasing onto current `main`

## Summary by CodeRabbit

* **New Features**
  * Added a `doc-index` command to build, reuse, inspect, and rebuild structural indexes for Markdown documents.
  * Cached indexes refresh automatically when source content changes.

* **Validation Improvements**
  * TOC validation now warns about duplicate headings, abrupt heading-depth changes, oversized sections, and missing document descriptions.
  * Added a configurable maximum section length, defaulting to 300 lines.
  * Warning-only findings are clearly reported without failing validation.

### DIFF
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -451,7 +451,8 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 
 A cached, read-once-per-file structural index for Markdown JIT retrieval (see
 constructorfabric/studio#104): parsing a file's headings/section boundaries
-happens once, not once per query, until the file's content actually changes.
+happens once, not once per query, until its stat fingerprint (`mtime_ns` +
+size) changes.
 The cache-validity fingerprint is deliberately metadata-only (`mtime` + file
 size via `Path.stat()`), never a content hash â€” the point of the cache is to
 avoid reading the file at all on a hit, and a content hash would defeat that
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -467,6 +467,8 @@ by requiring the read it's meant to save.
 **Supporting**:
 - [x] - `p1` - Stat-based cache-validity fingerprint (`mtime_ns` + size); resolved from the file's own path, never a content hash - `inst-doc-index-etag`
 - [x] - `p1` - Resolve the cache file location within the Studio directory owning the indexed file, resolved from the file's own path (not the process's working directory) - `inst-doc-index-cache-path`
+- [x] - `p1` - `cfs doc-index` CLI wrapper: parse arguments, build the JSON output payload, reporting a clean error for a missing or unreadable file - `inst-doc-index-cmd`
+- [x] - `p1` - Human-friendly formatter for `cfs doc-index` output - `inst-doc-index-cmd-format`
 
 ### Markdown Parsing Utilities
 
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -399,11 +399,11 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
    2. [x] - `p1` - Generate expected TOC from headings - `inst-toc-generate-expected`
    3. [x] - `p1` - Compare existing vs expected: check anchor validity, heading coverage, staleness - `inst-toc-compare`
    4. [x] - `p1` - **IF** mismatch, record error with diff details - `inst-toc-if-mismatch`
-4. [x] - `p1` - **RETURN** JSON: `{status, files_checked, errors}` - `inst-toc-return`
+4. [x] - `p1` - **RETURN** JSON: `{status, files_validated, error_count, warning_count, results}`, each `results[]` entry `{file, status, error_count, warning_count}` plus `errors`/`warnings` arrays when `--verbose` or non-empty - `inst-toc-return`
 
 **Supporting**:
 - [x] - `p1` - Imports and module setup for validate-toc command - `inst-toc-imports`
-- [x] - `p1` - Human-friendly formatter for validate-toc output - `inst-toc-format`
+- [x] - `p1` - Human-friendly formatter for validate-toc output: a WARN-only file prints its warnings the same way a FAIL file prints its errors, not just the bare status - `inst-toc-format`
 
 ### TOC Utilities
 
@@ -417,9 +417,9 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 4. [x] - `p1` - Insert/update TOC using heading-based insertion (`## Table of Contents`) for kit file generator - `inst-toc-util-insert-heading`
 5. [x] - `p1` - Process file: strip manual TOC, insert marker-based TOC, write if changed - `inst-toc-util-process-file`
 6. [x] - `p1` - Validate TOC: check existence, anchor validity, completeness, staleness - `inst-toc-util-validate`
-7. [x] - `p1` - Parse headings with line numbers, fence-aware (shared by doc-index and the JIT-retrieval readiness checks, which need section boundaries the plain heading list doesn't carry) - `inst-toc-util-parse-headings-lines`
+7. [x] - `p1` - Parse headings with line numbers, fence-aware and front-matter-aware (skips a leading YAML block so a `#`-prefixed front-matter line is never mistaken for a heading); shared by doc-index and the JIT-retrieval readiness checks, which need section boundaries the plain heading list doesn't carry -- `parse_headings` itself now delegates here, stripping the line number, so both share one fence/heading-match implementation - `inst-toc-util-parse-headings-lines`
 8. [x] - `p1` - Collect JIT-retrieval readiness warnings for a document: gathers all four signals below over *every* heading level, independent of whatever level cap the caller configured for TOC-completeness checking - `inst-toc-jit-readiness-collect`
-9. [x] - `p1` - Compute the four JIT-retrieval readiness signals -- duplicate heading titles, heading depth jumps, oversized sections (`--max-section-lines`), and a missing top-of-file description/frontmatter block -- all warning-only, never errors (see constructorfabric/studio#104) - `inst-toc-jit-readiness`
+9. [x] - `p1` - Compute the four JIT-retrieval readiness signals -- duplicate heading titles (compared case-insensitively, with internal whitespace collapsed and Unicode-normalized, though the original text is still shown in the warning), heading depth jumps, oversized sections (`--max-section-lines`, default 300, validated against non-finite/non-positive input independent of the CLI's own argparse guard), and a missing top-of-file description/frontmatter block -- all warning-only, never errors (see constructorfabric/studio#104) - `inst-toc-jit-readiness`
 
 **Supporting**:
 - [x] - `p1` - Imports, constants, fence tracking, GitHub anchor slug generation - `inst-toc-util-datamodel`
@@ -447,7 +447,7 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 
 **Input**: Markdown file path
 
-**Output**: A structural index (headings, section line ranges, per-section summary slots) cached per file, keyed by a stat-based fingerprint
+**Output**: `cfs doc-index`'s JSON is `{file, cache_hit, total_lines, section_count, sections}`, each `sections[]` entry `{level, heading, line_start, line_end, summary}`. The underlying index dict additionally carries `schema_version`, `path`, and `etag`.
 
 A cached, read-once-per-file structural index for Markdown JIT retrieval (see
 constructorfabric/studio#104): parsing a file's headings/section boundaries
@@ -457,9 +457,9 @@ size via `Path.stat()`), never a content hash â€” the point of the cache is to
 avoid reading the file at all on a hit, and a content hash would defeat that
 by requiring the read it's meant to save.
 
-1. [x] - `p1` - Build a fresh structural index: parse headings + line ranges from current content, compute the stat-based fingerprint - `inst-doc-index-build`
-2. [x] - `p1` - Load a cached index for a file, validated against current stat metadata (no content read on a hit); returns `None` if missing, stale, or corrupt - `inst-doc-index-load`
-3. [x] - `p1` - Persist an index to its cache location; no-ops silently outside a Studio-adapted project - `inst-doc-index-save`
+1. [x] - `p1` - Build a fresh structural index: parse headings + line ranges from current content, compute the stat-based fingerprint, stamp the current schema version - `inst-doc-index-build`
+2. [x] - `p1` - Load a cached index for a file, validated against current stat metadata (no content read on a hit) and against the required-field shape at the current schema version; returns `None` if missing, stale, corrupt, or an incomplete/outdated shape - `inst-doc-index-load`
+3. [x] - `p1` - Persist an index to its cache location atomically (temp file + `os.replace`, so a concurrent reader never observes a torn write); no-ops silently outside a Studio-adapted project - `inst-doc-index-save`
 4. [x] - `p1` - Return the cached index or build-and-cache a fresh one; reports cache hit/miss for benchmarking - `inst-doc-index-get-or-build`
 5. [x] - `p1` - Attach a one-line, LLM-authored summary to a cached section by its `line_start`, for a future per-section-summary caller - `inst-doc-index-annotate`
 
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -21,6 +21,7 @@
   - [List ID Kinds](#list-id-kinds)
   - [Validate TOC](#validate-toc)
   - [TOC Utilities](#toc-utilities)
+  - [Document Index](#document-index)
   - [Markdown Parsing Utilities](#markdown-parsing-utilities)
   - [Fixing Prompt Enrichment](#fixing-prompt-enrichment)
   - [Headings Contract Validation](#headings-contract-validation)
@@ -416,6 +417,9 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 4. [x] - `p1` - Insert/update TOC using heading-based insertion (`## Table of Contents`) for kit file generator - `inst-toc-util-insert-heading`
 5. [x] - `p1` - Process file: strip manual TOC, insert marker-based TOC, write if changed - `inst-toc-util-process-file`
 6. [x] - `p1` - Validate TOC: check existence, anchor validity, completeness, staleness - `inst-toc-util-validate`
+7. [x] - `p1` - Parse headings with line numbers, fence-aware (shared by doc-index and the JIT-retrieval readiness checks, which need section boundaries the plain heading list doesn't carry) - `inst-toc-util-parse-headings-lines`
+8. [x] - `p1` - Collect JIT-retrieval readiness warnings for a document: gathers all four signals below over *every* heading level, independent of whatever level cap the caller configured for TOC-completeness checking - `inst-toc-jit-readiness-collect`
+9. [x] - `p1` - Compute the four JIT-retrieval readiness signals -- duplicate heading titles, heading depth jumps, oversized sections (`--max-section-lines`), and a missing top-of-file description/frontmatter block -- all warning-only, never errors (see constructorfabric/studio#104) - `inst-toc-jit-readiness`
 
 **Supporting**:
 - [x] - `p1` - Imports, constants, fence tracking, GitHub anchor slug generation - `inst-toc-util-datamodel`
@@ -436,6 +440,32 @@ Catches structural and traceability issues that AI agents miss or hallucinate â€
 - [x] - `p1` - Heading-based TOC replace branch: overwrite content under existing `## Table of Contents` heading - `inst-toc-util-insert-heading-replace`
 - [x] - `p1` - Heading-based TOC new-insert branch: inject `## Table of Contents` before first heading when absent - `inst-toc-util-insert-heading-new`
 - [x] - `p1` - TOC validate init: build heading list and expected TOC string before comparison checks - `inst-toc-util-validate-init`
+
+### Document Index
+
+- [x] `p1` - **ID**: `cpt-studio-algo-traceability-validation-doc-index`
+
+**Input**: Markdown file path
+
+**Output**: A structural index (headings, section line ranges, per-section summary slots) cached per file, keyed by a stat-based fingerprint
+
+A cached, read-once-per-file structural index for Markdown JIT retrieval (see
+constructorfabric/studio#104): parsing a file's headings/section boundaries
+happens once, not once per query, until the file's content actually changes.
+The cache-validity fingerprint is deliberately metadata-only (`mtime` + file
+size via `Path.stat()`), never a content hash â€” the point of the cache is to
+avoid reading the file at all on a hit, and a content hash would defeat that
+by requiring the read it's meant to save.
+
+1. [x] - `p1` - Build a fresh structural index: parse headings + line ranges from current content, compute the stat-based fingerprint - `inst-doc-index-build`
+2. [x] - `p1` - Load a cached index for a file, validated against current stat metadata (no content read on a hit); returns `None` if missing, stale, or corrupt - `inst-doc-index-load`
+3. [x] - `p1` - Persist an index to its cache location; no-ops silently outside a Studio-adapted project - `inst-doc-index-save`
+4. [x] - `p1` - Return the cached index or build-and-cache a fresh one; reports cache hit/miss for benchmarking - `inst-doc-index-get-or-build`
+5. [x] - `p1` - Attach a one-line, LLM-authored summary to a cached section by its `line_start`, for a future per-section-summary caller - `inst-doc-index-annotate`
+
+**Supporting**:
+- [x] - `p1` - Stat-based cache-validity fingerprint (`mtime_ns` + size); resolved from the file's own path, never a content hash - `inst-doc-index-etag`
+- [x] - `p1` - Resolve the cache file location within the Studio directory owning the indexed file, resolved from the file's own path (not the process's working directory) - `inst-doc-index-cache-path`
 
 ### Markdown Parsing Utilities
 
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -16,6 +16,7 @@ from ..utils.ui import ui
 logger = logging.getLogger(__name__)
 
 
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cmd
 def cmd_doc_index(argv: List[str]) -> int:
     """Build (or reuse the cached) structural index for a Markdown file."""
     p = argparse.ArgumentParser(
@@ -60,8 +61,10 @@ def cmd_doc_index(argv: List[str]) -> int:
     }
     ui.result(output, human_fn=_human_doc_index)
     return 0
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cmd
 
 
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cmd-format
 def _human_doc_index(data: dict) -> None:
     ui.header("Doc Index")
     ui.substep(data["file"])
@@ -72,3 +75,4 @@ def _human_doc_index(data: dict) -> None:
         summary = f" — {s['summary']}" if s.get("summary") else ""
         ui.substep(f"  H{s['level']} [{s['line_start']}-{s['line_end']}] {s['heading']}{summary}")
     ui.blank()
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cmd-format
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -1,0 +1,64 @@
+"""Studio doc-index command — build/reuse a cached structural index for a
+Markdown file, so heading-based JIT retrieval reads a file's structure once,
+not once per query.
+
+Thin CLI wrapper around ``studio.utils.doc_index``.
+
+@cpt-flow:cpt-studio-flow-traceability-validation-validate:p1
+"""
+
+import argparse
+from pathlib import Path
+from typing import List
+
+from ..utils.doc_index import get_or_build_doc_index
+from ..utils.ui import ui
+
+
+def cmd_doc_index(argv: List[str]) -> int:
+    """Build (or reuse the cached) structural index for a Markdown file."""
+    p = argparse.ArgumentParser(
+        prog="cfs doc-index",
+        description=(
+            "Build or reuse a cached heading/section index for a Markdown file, "
+            "so navigation reads the file's structure once, not once per query."
+        ),
+    )
+    p.add_argument("file", help="Markdown file path")
+    p.add_argument(
+        "--rebuild",
+        action="store_true",
+        help="Force a fresh build even if a valid cached index exists",
+    )
+    args = p.parse_args(argv)
+
+    filepath = Path(args.file).resolve()
+    if not filepath.is_file():
+        ui.result(
+            {"file": str(filepath), "status": "ERROR", "message": "File not found"},
+            human_fn=lambda d: ui.error(f"{d['file']}: {d['message']}"),
+        )
+        return 2
+
+    index = get_or_build_doc_index(filepath, force_rebuild=args.rebuild)
+
+    output = {
+        "file": str(filepath),
+        "cache_hit": index["cache_hit"],
+        "total_lines": index["total_lines"],
+        "section_count": len(index["sections"]),
+        "sections": index["sections"],
+    }
+    ui.result(output, human_fn=_human_doc_index)
+    return 0
+
+
+def _human_doc_index(data: dict) -> None:
+    ui.header("Doc Index")
+    hit = "cache hit — reused existing index" if data["cache_hit"] else "cache miss — built fresh index"
+    ui.substep(hit)
+    ui.substep(f"{data['section_count']} section(s), {data['total_lines']} total lines")
+    for s in data["sections"]:
+        summary = f" — {s['summary']}" if s.get("summary") else ""
+        ui.substep(f"  H{s['level']} [{s['line_start']}-{s['line_end']}] {s['heading']}{summary}")
+    ui.blank()
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -3,8 +3,6 @@ Markdown file, so heading-based JIT retrieval reads a file's structure once,
 not once per query.
 
 Thin CLI wrapper around ``studio.utils.doc_index``.
-
-@cpt-flow:cpt-studio-flow-traceability-validation-validate:p1
 """
 
 import argparse
@@ -55,6 +53,7 @@ def cmd_doc_index(argv: List[str]) -> int:
 
 def _human_doc_index(data: dict) -> None:
     ui.header("Doc Index")
+    ui.substep(data["file"])
     hit = "cache hit — reused existing index" if data["cache_hit"] else "cache miss — built fresh index"
     ui.substep(hit)
     ui.substep(f"{data['section_count']} section(s), {data['total_lines']} total lines")
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -6,11 +6,14 @@ Thin CLI wrapper around ``studio.utils.doc_index``.
 """
 
 import argparse
+import logging
 from pathlib import Path
 from typing import List
 
 from ..utils.doc_index import get_or_build_doc_index
 from ..utils.ui import ui
+
+logger = logging.getLogger(__name__)
 
 
 def cmd_doc_index(argv: List[str]) -> int:
@@ -38,7 +41,15 @@ def cmd_doc_index(argv: List[str]) -> int:
         )
         return 2
 
-    index = get_or_build_doc_index(filepath, force_rebuild=args.rebuild)
+    try:
+        index = get_or_build_doc_index(filepath, force_rebuild=args.rebuild)
+    except (OSError, UnicodeDecodeError) as exc:
+        logger.warning("doc-index: cannot read %s: %s", filepath, exc)
+        ui.result(
+            {"file": str(filepath), "status": "ERROR", "message": f"Cannot read file: {exc}"},
+            human_fn=lambda d: ui.error(f"{d['file']}: {d['message']}"),
+        )
+        return 2
 
     output = {
         "file": str(filepath),
--- a/skills/studio/scripts/studio/commands/validate_toc.py
+++ b/skills/studio/scripts/studio/commands/validate_toc.py
@@ -130,6 +130,10 @@ def _human_validate_toc(data: dict) -> None:
                 ui.substep(f"  ✗ {e}")
             for w in r.get("warnings", []):
                 ui.substep(f"  ⚠ {w}")
+        elif status == "WARN":
+            ui.warn(f"{path}: {warns} warning(s)")
+            for w in r.get("warnings", []):
+                ui.substep(f"  ⚠ {w}")
         else:
             ui.substep(f"{path}: {status}")
     overall = data.get("status", "")
--- a/skills/studio/scripts/studio/commands/validate_toc.py
+++ b/skills/studio/scripts/studio/commands/validate_toc.py
@@ -14,7 +14,7 @@ import argparse
 from pathlib import Path
 from typing import List
 
-from ..utils.toc import add_toc_max_level_argument, validate_toc
+from ..utils.toc import DEFAULT_MAX_SECTION_LINES, add_toc_max_level_argument, validate_toc
 from ..utils.ui import ui
 # @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-imports
 
@@ -31,6 +31,12 @@ def cmd_validate_toc(argv: List[str]) -> int:
         help="Markdown file path(s) to validate",
     )
     add_toc_max_level_argument(p)
+    p.add_argument(
+        "--max-section-lines",
+        type=int,
+        default=DEFAULT_MAX_SECTION_LINES,
+        help=f"Warn when a section exceeds this many lines (default: {DEFAULT_MAX_SECTION_LINES})",
+    )
     p.add_argument(
         "--verbose",
         action="store_true",
@@ -63,6 +69,7 @@ def cmd_validate_toc(argv: List[str]) -> int:
             content,
             artifact_path=filepath,
             max_heading_level=args.max_level,
+            max_section_lines=args.max_section_lines,
         )
 
         errors = report.get("errors", [])
--- a/skills/studio/scripts/studio/utils/doc_index.py
+++ b/skills/studio/scripts/studio/utils/doc_index.py
@@ -1,0 +1,211 @@
+"""Cached, read-once-per-file document index for Markdown JIT retrieval.
+
+Builds a structural index (headings + section line ranges) for a Markdown
+file exactly once, persists it keyed by an etag of the file's own state, and
+reuses that cached index on every subsequent call against the same file --
+until the file actually changes. This is the "read once per file, not once
+per query" mechanism: parsing/etag work never repeats across queries, and
+optional per-section summaries (written by an LLM caller, not by this
+module) accumulate in the same cached artifact instead of being
+re-derived each time.
+
+Scope: Markdown only. PDF/DOCX conversion is a separate concern (Layer 1);
+this module operates purely on already-plain-text content (Layer 2).
+
+See constructorfabric/studio#104.
+
+@cpt-algo:cpt-studio-algo-traceability-validation-doc-index:p1
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import time
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+from .toc import parse_headings_with_lines
+
+logger = logging.getLogger(__name__)
+
+_CACHE_SUBDIR = ".cache"
+_INDEX_CACHE_DIR = "doc-index"
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-etag
+def _compute_etag(path: Path) -> str:
+    """Compute a cheap cache-validity fingerprint from filesystem metadata.
+
+    Deliberately *not* a content hash: ``Path.stat()`` is metadata-only (no
+    file read), which is what lets a cache *hit* stay free of a full read --
+    the whole point of a read-once-per-file index. mtime + size changes on
+    any real edit (including a same-size, same-line-count text swap, since
+    a write always advances mtime), so it still catches content changes; a
+    byte-count/line-count-only fingerprint would not (two edits that happen
+    to preserve both would silently look unchanged, and computing either
+    requires reading the entire file that this check exists to avoid
+    reading).
+    """
+    st = path.stat()
+    return f"{st.st_mtime_ns}:{st.st_size}"
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-etag
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cache-path
+def _index_cache_path(path: Path) -> Optional[Path]:
+    """Resolve ``<studio-dir>/.cache/doc-index/<slug>.json`` for a file.
+
+    Resolved from ``path`` itself (not the process's current working
+    directory), so indexing a file outside the caller's cwd still resolves
+    -- and always resolves -- the Studio directory that actually owns it.
+
+    Returns ``None`` when no Studio directory can be found (e.g. outside a
+    Studio-adapted project) -- callers should fall back to an uncached build.
+    """
+    from .files import find_studio_directory
+
+    try:
+        studio_dir = find_studio_directory(path.resolve().parent)
+    except OSError as exc:
+        # A file whose parent can't be stat'd (permissions, a race) is not a
+        # reason to fail the caller -- just an uncached build, like "no
+        # Studio directory found".
+        logger.debug("doc-index cache path lookup skipped for %s: %s", path, exc)
+        studio_dir = None
+    if studio_dir is None:
+        return None
+
+    slug = hashlib.sha256(str(path.resolve()).encode("utf-8")).hexdigest()[:16]
+    return studio_dir / _CACHE_SUBDIR / _INDEX_CACHE_DIR / f"{slug}.json"
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cache-path
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-build
+def build_doc_index(path: Path) -> Dict[str, Any]:
+    """Build a fresh structural index for a Markdown file.
+
+    Purely deterministic -- headings, section line ranges, and an etag.
+    Contains no LLM-generated content; per-section ``summary`` fields start
+    as ``None`` and are filled in later via :func:`annotate_section_summary`.
+    """
+    canonical_path = path.resolve()
+    content = canonical_path.read_text(encoding="utf-8")
+    lines = content.split("\n")
+    line_count = len(lines)
+
+    headings = parse_headings_with_lines(lines)
+    sections: List[Dict[str, Any]] = []
+    for i, (level, text, line_start) in enumerate(headings):
+        line_end = headings[i + 1][2] - 1 if i + 1 < len(headings) else line_count
+        sections.append({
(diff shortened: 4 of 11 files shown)


skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,178 @@
...
+
+def _compute_etag(canonical_path: Path, byte_size: int, line_count: int) -> str:
+    """Compute a content-state fingerprint for a file.
+
+    Same formula as the storytelling content-pack's own etag
+    (``sha256(canonical_path + ":" + byte_size + ":" + line_count)``), reused
+    here for consistency across Studio's caching mechanisms.
+    """
+    payload = f"{canonical_path}:{byte_size}:{line_count}"
+    return hashlib.sha256(payload.encode("utf-8")).hexdigest()
```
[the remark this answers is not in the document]
tkcoding: Fixed in d5e2be72 — replaced the etag with `Path.stat()` (`mtime_ns` + size), which also fixes a second issue from the same review: `load_doc_index()` was reading the entire file on every cache hit just to compute the old etag, defeating the point of the cache. Added `test_same_size_same_line_count_edit_is_still_detected_as_stale`, which replaces a heading with same-size, same-line-count text and confirms it's now caught as stale.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,178 @@
...
+def _index_cache_path(path: Path) -> Optional[Path]:
+    """Resolve ``<studio-dir>/.cache/doc-index/<slug>.json`` for a file.
+
+    Returns ``None`` when no Studio directory can be found (e.g. outside a
+    Studio-adapted project) -- callers should fall back to an uncached build.
+    """
+    from .files import find_studio_directory
+
+    try:
+        studio_dir = find_studio_directory(Path.cwd())
```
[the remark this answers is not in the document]
tkcoding: Fixed in d5e2be72 — `find_studio_directory` is now called with `path.resolve().parent` instead of `Path.cwd()`. Added `test_studio_directory_resolved_from_file_path_not_cwd`, which spies on the call and asserts the resolved file path is what's passed, not cwd.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,148 @@ def _record_missing_toc_error(
...
+    ``MIN_LINES_FOR_DESCRIPTION_CHECK`` lines — a trivial file doesn't need
+    a description, and flagging every small file drowns the signal.
+    """
+    from . import error_codes as EC
+    from .constraints import error
+
+    if len(lines) < MIN_LINES_FOR_DESCRIPTION_CHECK:
+        return []
+    if _find_frontmatter_end(lines) > 0:
+        return []
```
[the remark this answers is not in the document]
tkcoding: Fixed in d5e2be72 — added `_frontmatter_has_description`, which checks for a non-empty `description:` field within the frontmatter block rather than accepting any frontmatter's mere presence. Added `test_frontmatter_without_description_field_still_warns` (frontmatter with only a `title` field still warns).

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -796,8 +981,16 @@ def validate_toc(
         max_heading_level,
     )
 
+    # JIT-retrieval readiness signals (warning-only, run regardless of
+    # TOC presence below — independent of the TOC-filtered `headings`).
+    warnings.extend(_check_missing_description(lines, path))
+    all_headings = parse_headings_with_lines(lines, max_level=max_heading_level)
+    warnings.extend(_check_duplicate_heading_titles(all_headings, path))
+    warnings.extend(_check_heading_depth_jumps(all_headings, path))
+    warnings.extend(_check_section_lengths(all_headings, len(lines), path, max_section_lines))
```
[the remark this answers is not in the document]
tkcoding: Fixed in d5e2be72 — the readiness checks now always parse every heading level via `parse_headings_with_lines(lines)` with no level cap, independent of `max_heading_level`. Worth noting: this is the same bug I'd independently found earlier in the session testing against a real PDF-converted document, where the CLI's own `--max-level 3` default was hiding real duplicate/depth-jump/oversized-section issues in H4+ headings — good catch, confirmed from a second angle. Added `test_readiness_signals_see_headings_deeper_than_max_heading_level`.

skills/studio/scripts/studio/commands/validate_toc.py
```diff
@@ -63,6 +69,7 @@ def cmd_validate_toc(argv: List[str]) -> int:
             content,
             artifact_path=filepath,
             max_heading_level=args.max_level,
+            max_section_lines=args.max_section_lines,
```
ainetx, line 72: ### JIT-readiness warnings invisible in human CLI output

**Severity:** Major

**Problem**
The human-readable output branch for a `WARN`-only file (`_human_validate_toc`'s final `else:`, line 134) prints only `f"{path}: {status}"` and never iterates `r.get("warnings", [])`, unlike the `FAIL` branch just above it which does. Since this PR's four new JIT-retrieval readiness checks (duplicate headings, heading-depth jumps, oversized sections, missing description) are warning-only by design, any file that trips *only* these new checks lands in this silent branch. (Anchored here at the new `max_section_lines` wiring since line 134 itself isn't touched by this diff, but its behavior is what this PR's new warnings now depend on.) Separately, `_human_doc_index` in `commands/doc_index.py` never prints the target file name either.

**How to reproduce**
1. Create a Markdown file with one duplicate heading title and no other TOC issues.
2. Run `cfs validate-toc file.md` (human mode, no `--json`).

**Expected behavior**
The new warning (e.g. `toc-heading-duplicate: ...`) is printed, the same way an error would be under `FAIL`.

**Actual behavior**
Output is just `file.md: WARN` with no indication of what's wrong — the JSON output (`--json`) does contain the warning, so the information exists but never reaches the default human-facing path.

```text
cmd_validate_toc()
   -> validate_toc() returns {status: "WARN", warnings: [...]}
   -> _human_validate_toc()
        status == "WARN"  -->  else: branch
                                   prints "path: WARN" only
                                   (warnings list never read)
```

**Impact**
The PR's headline feature (JIT-readiness warnings) is effectively invisible to anyone using the default CLI output instead of `--json`.

**Suggested correction**
Add a branch for `status == "WARN"` (or extend the `else`) that iterates `r.get("warnings", [])` the same way the `FAIL` branch iterates `r.get("errors", [])`; print `data["file"]` in `_human_doc_index`.

**How to verify**
Add/extend a `TestCmdValidateToc` case: a file with exactly one warning and no errors, run through `cmd_validate_toc`'s human-output path, and assert the warning text appears in the captured output.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,211 @@
...
+    canonical_path = path.resolve()
+    try:
+        current_etag = _compute_etag(canonical_path)
+    except OSError as exc:
+        logger.debug("doc-index staleness check failed for %s: %s", path, exc)
+        return None
+
+    if cached.get("etag") != current_etag:
+        return None
+    return cached
```
ainetx, line 188: ### Cache read has no shape/key validation, causing unhandled KeyError

**Severity:** Major

**Problem**
`load_doc_index` validates only the `etag` key before returning the cached dict as-is. It never checks that the other required keys (`total_lines`, `sections`, etc.) are present. Downstream code (`commands/doc_index.py`, `annotate_section_summary`) indexes into the result with bare `[...]`, not `.get(...)`.

**How to reproduce**
1. Run `cfs doc-index file.md` once to create the cache file under `.cache/doc-index/<slug>.json`.
2. Hand-edit that JSON to remove the `"sections"` key, leaving `"etag"` untouched (source file unmodified, so the etag still matches).
3. Run `cfs doc-index file.md` again without `--rebuild`.

**Expected behavior**
A malformed/incomplete cache should be treated like a stale or corrupt one — logged and rebuilt (the module's own docstring says `load_doc_index` returns `None` "if missing/stale/absent").

**Actual behavior**
`load_doc_index` returns the truncated dict as a valid hit, and `commands/doc_index.py`'s `len(index["sections"])` raises an unhandled `KeyError: 'sections'`.

```text
load_doc_index(path)
   etag matches -> return cached   (no key-shape check)
                        |
                        v
cmd_doc_index(): index["sections"]  -->  KeyError: 'sections'
```

**Impact**
A partially-written or hand-edited cache file crashes the CLI with a raw traceback instead of a documented error or automatic rebuild.

**Suggested correction**
After the etag check in `load_doc_index`, validate the presence of required keys (`{"etag", "total_lines", "sections"}` at minimum); treat a mismatch the same as stale/corrupt (log and return `None`).

**How to verify**
Add a test that writes a cache file with a matching etag but a missing required key, then asserts `load_doc_index` returns `None` (and/or that `cmd_doc_index` transparently rebuilds rather than raising).

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,211 @@
...
+def _compute_etag(path: Path) -> str:
+    """Compute a cheap cache-validity fingerprint from filesystem metadata.
+
+    Deliberately *not* a content hash: ``Path.stat()`` is metadata-only (no
+    file read), which is what lets a cache *hit* stay free of a full read --
+    the whole point of a read-once-per-file index. mtime + size changes on
+    any real edit (including a same-size, same-line-count text swap, since
+    a write always advances mtime), so it still catches content changes; a
+    byte-count/line-count-only fingerprint would not (two edits that happen
+    to preserve both would silently look unchanged, and computing either
```
ainetx: ### mtime+size cache etag can silently miss same-size edits

**Severity:** Major

**Problem**
The cache-validity fingerprint is `f"{st.st_mtime_ns}:{st.st_size}"` — filesystem metadata only, no content hash. The docstring above it asserts this is safe unconditionally ("a write always advances mtime"), but that isn't universally true: filesystems with coarse mtime resolution (FAT32, classic HFS+, many NFS mounts) can give two same-size edits within one mtime tick an identical etag.

**How to reproduce**
On a filesystem with ≥1s mtime granularity:
1. Write a file with heading "Section A"; call `get_or_build_doc_index` (caches etag `T:N`).
2. Within the same mtime tick, overwrite with same-length content renaming the heading to "Section Z".
3. Call `get_or_build_doc_index` again.

**Expected behavior**
The cache should detect the content change and rebuild.

**Actual behavior**
`_compute_etag` returns the same `T:N` string, so the stale cache (still showing "Section A") is returned as a valid hit. Notably, this PR's own regression test (`test_same_size_same_line_count_edit_is_still_detected_as_stale`) has to manually call `os.utime(f, ns=(..., st.st_mtime_ns + 1))` to force the mtime forward — direct evidence that the "always advances mtime" assumption doesn't hold reliably even in this PR's own test environment.

```text
write("...Section A...")  -> etag = T:N  -> cache saved
write("...Section Z...")  (same size, same mtime tick)
   -> etag = T:N  (unchanged!)
   -> load_doc_index(): cached.etag == current_etag -> HIT (stale data returned)
```

**Impact**
Silent false cache hit — stale headings/sections/summaries served with no error surfaced anywhere.

**Suggested correction**
Either document this as an explicit, accepted limitation (call out filesystem mtime granularity in the docstring), or add a stronger invalidation signal (e.g. a monotonic write-generation counter, or falling back to a content hash when size matches but the edit is suspiciously fast).

**How to verify**
On a filesystem/test harness that can simulate coarse mtime resolution (or by monkeypatching `Path.stat()` to return identical `st_mtime_ns`/`st_size` across two different contents), confirm the cache is invalidated regardless.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,167 @@ def _record_missing_toc_error(
...
+# are otherwise fine. See constructorfabric/studio#104.
+
+DEFAULT_MAX_SECTION_LINES = 300
+# Below this size, a missing description is not worth flagging — the whole
+# point of a description is to let a caller pick the right *file* before
+# reading it, among many; a trivial file doesn't need that.
+MIN_LINES_FOR_DESCRIPTION_CHECK = 100
+
+
+def _check_duplicate_heading_titles(
```
ainetx, line 761: ### Duplicate-heading check misses case/whitespace variants and doesn't skip front-matter

**Severity:** Major

**Problem**
`_check_duplicate_heading_titles` compares raw heading text via a plain dict with no case-folding or whitespace normalization, and the heading list it operates on (`parse_headings_with_lines`) has no YAML front-matter skip (unlike `_check_missing_description`, which does skip front-matter). No `unicodedata.normalize` call exists anywhere in the file either, so NFC/NFD-distinct but visually-identical headings are also missed by the same root cause.

**How to reproduce**
Parse a document containing:
```markdown
---
title: Foo
# this is a note
---
## Section
## section
## Setup  Guide
## Setup Guide
```

**Expected behavior**
`"Section"`/`"section"` and `"Setup  Guide"`/`"Setup Guide"` (double space) should be flagged as duplicates; the `# this is a note` line inside front-matter should not be parsed as a real heading.

**Actual behavior**
None of the four headings above are flagged as duplicates, and the front-matter comment line is parsed into the heading list, potentially throwing off depth-jump and section-length checks that consume the same list.

```text
parse_headings_with_lines(lines)
  no front-matter awareness --> "# this is a note" counted as H1
_check_duplicate_heading_titles(headings)
  seen[text] keyed on raw text --> "Section" != "section" (case)
                                --> "Setup  Guide" != "Setup Guide" (whitespace)
```

**Impact**
Silently defeats the duplicate-heading check (and pollutes the shared heading list used by the other three JIT-readiness checks) for a common class of real-world documents.

**Suggested correction**
Normalize heading text (casefold + collapse internal whitespace + NFC-normalize) before using it as the `seen` dict key, while still reporting the original text in the warning payload; skip front-matter lines in `parse_headings_with_lines` (reuse the existing `_find_frontmatter_end` helper).

**How to verify**
Add test cases for case-only duplicates, whitespace-only duplicates, NFC/NFD-only duplicates, and a front-matter block containing a `#`-prefixed line, asserting the expected warnings/non-warnings.

skills/studio/scripts/studio/commands/doc_index.py
```diff
@@ -0,0 +1,64 @@
+"""Studio doc-index command — build/reuse a cached structural index for a
+Markdown file, so heading-based JIT retrieval reads a file's structure once,
+not once per query.
+
+Thin CLI wrapper around ``studio.utils.doc_index``.
+
+@cpt-flow:cpt-studio-flow-traceability-validation-validate:p1
```
ainetx: ### CPT `@cpt-flow` tag references the wrong, unrelated flow

**Severity:** Major

**Problem**
This module docstring adds `@cpt-flow:cpt-studio-flow-traceability-validation-validate:p1`. That ID is the pre-existing "Validate Artifacts" flow (`cfs validate`), already correctly tagged as a bare `#`-comment directly above `cmd_validate` in `commands/validate.py:1230`. None of that flow's steps (load context, resolve artifacts, cross-validate, etc.) relate to doc-index in any way.

**How to reproduce**
Run `cfs where-used --id cpt-studio-flow-traceability-validation-validate` after this PR merges.

**Expected behavior**
Only `commands/validate.py`'s `cmd_validate` should be attributed to this flow ID; `cfs doc-index` should either have no `@cpt-flow` tag, or a real one describing its own flow.

**Actual behavior**
`doc_index.py` is now also attributed to the Validate-Artifacts flow, polluting `cfs where-used`/`cfs validate` traceability/coverage data for that flow, while the real `cfs doc-index` feature ends up with zero flow-level traceability of its own. The tag is also placed inside the module docstring rather than as a bare comment directly above the entry-point function, unlike every other `@cpt-flow` usage in this codebase.

```text
cpt-studio-flow-traceability-validation-validate
   correctly tagged:   commands/validate.py:1230  (above cmd_validate)
   ALSO tagged:        commands/doc_index.py:7     (unrelated command, docstring)
   -> traceability tooling now double-counts this flow ID
```

**Impact**
Corrupts the traceability data this feature exists to protect — the exact opposite of what a traceability-validation PR should do.

**Suggested correction**
Remove the stray `@cpt-flow` tag from `doc_index.py`'s docstring. If a flow-level definition is actually wanted for `cfs doc-index`, add a real `cpt-studio-flow-...` entry to `architecture/features/traceability-validation.md` and reference that ID as a bare `#`-comment directly above `def cmd_doc_index`.

**How to verify**
Run `cfs where-used --id cpt-studio-flow-traceability-validation-validate` (or the equivalent traceability query) and confirm it reports only `commands/validate.py`.

ainetx (commented):
Automated Deep Review — 5 Major findings out of 66 checks run across 7 thematic phases (each independently adversarially verified). Full plan and 12 additional Minor findings available on request.

architecture/features/traceability-validation.md
```diff
@@ -416,6 +417,9 @@ Catches structural and traceability issues that AI agents miss or hallucinate 
 4. [x] - `p1` - Insert/update TOC using heading-based insertion (`## Table of Contents`) for kit file generator - `inst-toc-util-insert-heading`
 5. [x] - `p1` - Process file: strip manual TOC, insert marker-based TOC, write if changed - `inst-toc-util-process-file`
 6. [x] - `p1` - Validate TOC: check existence, anchor validity, completeness, staleness - `inst-toc-util-validate`
+7. [x] - `p1` - Parse headings with line numbers, fence-aware (shared by doc-index and the JIT-retrieval readiness checks, which need section boundaries the plain heading list doesn't carry) - `inst-toc-util-parse-headings-lines`
+8. [x] - `p1` - Collect JIT-retrieval readiness warnings for a document: gathers all four signals below over *every* heading level, independent of whatever level cap the caller configured for TOC-completeness checking - `inst-toc-jit-readiness-collect`
+9. [x] - `p1` - Compute the four JIT-retrieval readiness signals -- duplicate heading titles, heading depth jumps, oversized sections (`--max-section-lines`), and a missing top-of-file description/frontmatter block -- all warning-only, never errors (see constructorfabric/studio#104) - `inst-toc-jit-readiness`
```
ainetx: ### Spec doc field-shape mismatch and missing defaults/output fields

**Severity:** Minor

**Problem**
This spec doc describes `--max-section-lines` by name here but never states its default (300, `DEFAULT_MAX_SECTION_LINES` in `utils/toc.py`). The pre-existing "Validate TOC" return-shape line a few lines up (`**RETURN** JSON: {status, files_checked, errors}`, line 402) doesn't match the actual output shape (`{status, files_validated, error_count, warning_count, results}`) even before this PR — and this PR's new warning codes (`toc-heading-duplicate`, `toc-heading-depth-jump`, `toc-section-too-long`, `toc-missing-description`) aren't named anywhere either. Separately, the "Document Index" section below never enumerates `cfs doc-index`'s exact JSON output fields.

**How to reproduce**
Read the "Validate TOC" section expecting to find the four new warning-code strings or the `--max-section-lines` default; read the "Document Index" section expecting the `cfs doc-index` output field list.

**Expected behavior**
The doc should state the default value and the exact field/error-code names as the authoritative contract.

**Actual behavior**
Neither is present — the default and the literal codes/field names must be discovered by reading source/tests instead of the spec.

```text
traceability-validation.md line 402: RETURN JSON: {status, files_checked, errors}   <- stale, wrong field names
actual:                              {status, files_validated, error_count, warning_count, results}
line 422: mentions --max-section-lines by name, no default value stated
Document Index section: no output-field enumeration for `cfs doc-index`
```

**Impact**
Spec doc is misleading/incomplete as a contract reference for both commands touched by this PR.

**Suggested correction**
Update the stale return-shape line at 402 to the actual field names, add the `--max-section-lines` default here, and add an output-fields line to the Document Index section.

**How to verify**
Cross-read the updated doc against `commands/validate_toc.py`'s and `commands/doc_index.py`'s actual output dicts and confirm every field/code name matches.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,167 @@ def _record_missing_toc_error(
...
+                line=line,
+                heading_text=text,
+                from_level=prev_level,
+                to_level=level,
+            ))
+        prev_level = level
+    return warnings
+
+
+def _check_section_lengths(
```
ainetx, line 824: ### `--max-section-lines` unguarded against nan/negative at the gate-function layer

**Severity:** Minor

**Problem**
`_check_section_lengths`/`validate_toc`'s `max_section_lines` parameter has no independent validation inside this function — only the CLI's `argparse(type=int)` layer blocks non-numeric junk.

**How to reproduce**
Call `validate_toc(content, max_section_lines=float('nan'))` or `validate_toc(content, max_section_lines=-1)` directly as a library call, bypassing the CLI.

**Expected behavior**
A non-numeric/out-of-range value should be rejected or explicitly classified by the function itself, independent of any CLI wrapper.

**Actual behavior**
`float('nan')` silently disables the oversized-section check entirely (`x > nan` is always `False`), and a negative value flags virtually every section — no explicit error in either case.

```text
validate_toc(content, max_section_lines=float('nan'))  # oversized-section warnings never fire
validate_toc(content, max_section_lines=-1)             # every section flagged
```

**Impact**
A caller using the Python API directly (not the CLI) gets silently wrong behavior instead of a clear error.

**Suggested correction**
Clamp/validate `max_section_lines` inside `validate_toc`/`_check_section_lengths` itself, independent of the CLI's argparse type guard.

**How to verify**
Add unit tests calling `validate_toc`/`_check_section_lengths` directly with `nan`, `inf`, and a negative value, asserting either a raised error or documented classification.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,211 @@
...
+        line_end = headings[i + 1][2] - 1 if i + 1 < len(headings) else line_count
+        sections.append({
+            "level": level,
+            "heading": text,
+            "line_start": line_start,
+            "line_end": line_end,
+            "summary": None,
+        })
+
+    return {
```
ainetx, line 129: ### No schema-version marker on the doc-index cache

**Severity:** Minor

**Problem**
The index dict returned here has no `schema_version`/`version` field, and `load_doc_index` only compares `etag`, never a version marker.

**How to reproduce**
Ship a future Studio version that changes the `sections` shape; run against a project with an existing cache built by the old version and an unchanged source file (etag still matches).

**Expected behavior**
A future schema change should be detectable and trigger an explicit rebuild/error rather than silently returning old-shape data.

**Actual behavior**
This is a brand-new, local-only, gitignored cache with no existing incompatible format to worry about yet, so today there is no active bug — but nothing would catch a future mismatch either.

```text
build_doc_index() -> {"path","etag","built_at","total_lines","sections"}   # no schema_version
load_doc_index()  -> only checks cached.get("etag") != current_etag
```

**Impact**
Forward-looking hardening gap, not an active bug today.

**Suggested correction**
Add a `SCHEMA_VERSION` constant, embed it here, and check it alongside `etag` in `load_doc_index`.

**How to verify**
Bump the constant in a follow-up schema change and confirm old caches are rejected/rebuilt rather than returned as-is.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,211 @@
...
+
+    fresh = build_doc_index(path)
+    save_doc_index(path, fresh)
+    fresh["cache_hit"] = False
+    return fresh
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-get-or-build
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-annotate
+def annotate_section_summary(path: Path, line_start: int, summary: str) -> bool:
```
ainetx, line 234: ### `annotate_section_summary` is unreachable from any production code path

**Severity:** Minor

**Problem**
`annotate_section_summary` is not called from any production code path in this PR — only from tests, and imported into `vulture_whitelist.py` purely to silence the unused-symbol linter.

**How to reproduce**
Search the repo for callers of `annotate_section_summary` outside `tests/test_doc_index.py` and `vulture_whitelist.py`.

**Expected behavior**
A shipped, non-test-only function should have at least one reachable production caller, or be clearly marked as forward-looking API.

**Actual behavior**
Zero production callers exist; it's genuinely speculative code for a documented "future LLM caller" per its own docstring — transparent and tested, not hidden, but unreachable outside tests today.

```text
annotate_section_summary()
   callers: tests/test_doc_index.py (6 sites), vulture_whitelist.py (import only)
   callers in commands/doc_index.py or cli.py: none
```

**Impact**
No runtime risk; purely a maintainability/clarity note.

**Suggested correction**
No change required for merge; consider tracking a follow-up issue for when the LLM-caller integration actually lands, since right now there's no way to reach this function from `cfs doc-index`.

**How to verify**
N/A — informational; nothing to verify until a real caller is wired in.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -170,6 +170,43 @@ def parse_headings(
     return headings
 # @cpt-end:cpt-studio-algo-traceability-validation-toc-utils:p1:inst-toc-util-parse-headings
 
+# @cpt-begin:cpt-studio-algo-traceability-validation-toc-utils:p1:inst-toc-util-parse-headings-lines
+def parse_headings_with_lines(
```
ainetx, line 157: ### Heading-parsing logic duplicated between `parse_headings` and `parse_headings_with_lines`

**Severity:** Minor

**Problem**
This function duplicates the fence-tracking/heading-matching loop of the pre-existing `parse_headings` almost line-for-line, rather than having one delegate to the other (the docstring explains this was a deliberate choice to avoid touching `parse_headings`'s existing 2-tuple call sites).

**How to reproduce**
Compare `parse_headings` (existing) and `parse_headings_with_lines` (new) side by side — both independently call `_fence_update` and `_HEADING_RE` in the same sequence.

**Expected behavior**
A future fix to fence-detection or the heading regex should apply to both functions automatically.

**Actual behavior**
A future fix applied to only one of the two would silently diverge them, since there is no shared implementation.

```text
parse_headings()              parse_headings_with_lines()
  loop: _fence_update()          loop: _fence_update()   <- duplicated
        _HEADING_RE.match()           _HEADING_RE.match() <- duplicated
```

**Impact**
Maintenance/drift risk, not an active bug.

**Suggested correction**
Have `parse_headings` delegate to `parse_headings_with_lines` internally and strip the line numbers, or extract the shared fence/heading-match loop into one private generator both functions consume.

**How to verify**
After refactoring, confirm both functions' existing test suites still pass unchanged (behavior-preserving refactor).

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -764,12 +962,38 @@ def _append_stale_toc_warning(
...
+    warnings.extend(_check_missing_description(lines, path))
+    all_headings = parse_headings_with_lines(lines)
+    warnings.extend(_check_duplicate_heading_titles(all_headings, path))
+    warnings.extend(_check_heading_depth_jumps(all_headings, path))
+    warnings.extend(_check_section_lengths(all_headings, len(lines), path, max_section_lines))
+    return warnings
+# @cpt-end:cpt-studio-algo-traceability-validation-toc-utils:p1:inst-toc-jit-readiness-collect
+
 # @cpt-begin:cpt-studio-algo-traceability-validation-toc-utils:p1:inst-toc-util-validate
 def validate_toc(
```
ainetx, line 1037: ### No opt-out for the four new warning-only checks

**Severity:** Minor

**Problem**
The four new JIT-readiness checks run unconditionally inside `validate_toc()` — there's no flag to disable them as a group (only `--max-section-lines` tunes one threshold).

**How to reproduce**
Run `cfs validate-toc` on a large legacy doc corpus with many long-standing oversized sections or duplicate headings.

**Expected behavior**
A team should be able to quiet noisy warnings for files/repos where they aren't actionable, without weakening the check globally.

**Actual behavior**
Every run emits `WARN` for these files with no way to suppress all four short of raising `--max-section-lines` globally (which weakens that one check for everyone, and does nothing for the other three checks).

```text
validate_toc(content)
   _collect_jit_readiness_warnings() always runs, no disable flag
   only knob: --max-section-lines (threshold, not on/off)
```

**Impact**
Low severity since the checks are warning-only and never affect exit code — a real but non-blocking UX gap.

**Suggested correction**
Optional — consider a `--skip-readiness-checks` flag if teams later report warning fatigue; not blocking for this PR.

**How to verify**
N/A unless implemented — then verify the flag suppresses all four warning codes.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,211 @@
...
+        return None
+
+    if cached.get("etag") != current_etag:
+        return None
+    return cached
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-load
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-save
+def save_doc_index(path: Path, index: Dict[str, Any]) -> None:
```
ainetx, line 193: ### Cache write is not atomic

**Severity:** Minor

**Problem**
`save_doc_index` writes the cache file directly (`cache_path.write_text(...)`) rather than atomically (temp file + `os.replace`). Two concurrent processes building/saving the same file's cache could interleave writes.

**How to reproduce**
Run two processes concurrently calling `get_or_build_doc_index` for the same file with `force_rebuild=True`, racing on the same cache path.

**Expected behavior**
A reader should never observe a torn/partial write, ideally via an atomic rename.

**Actual behavior**
A reader mid-write could see truncated/invalid JSON. This matches the rest of the codebase's existing convention (not unique to this PR), and `load_doc_index`'s existing `except (json.JSONDecodeError, OSError): return None` — backed by a dedicated passing test — already turns a torn/corrupt read into a safe, self-healing rebuild rather than propagated corruption, so the practical risk is low.

```text
Process A: write_text(partial...) ─┐
Process B: read_text() ------------┘--> JSONDecodeError -> load_doc_index returns None -> safe rebuild
```

**Impact**
Low — bounded by the existing corrupt-cache fallback; worst case is a wasted rebuild, not silent corruption reaching a caller.

**Suggested correction**
Optional hardening — write to a temp file and `os.replace()` into place.

**How to verify**
A concurrency stress test writing/reading the same cache path from multiple processes/threads, asserting no unhandled exception ever surfaces to the caller.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,167 @@ def _record_missing_toc_error(
...
+    (one past the closing ``---``); the body being scanned is
+    ``lines[1:frontmatter_end - 1]``, excluding both delimiter lines.
+    """
+    return any(
+        _DESCRIPTION_FIELD_RE.match(line.strip())
+        for line in lines[1:frontmatter_end - 1]
+    )
+
+
+def _check_missing_description(
```
ainetx, line 903: ### Missing-description check is front-matter-only, flags badge/blurb READMEs

**Severity:** Minor

**Problem**
`_check_missing_description` only recognizes a description declared inside a YAML front-matter `description:` field.

**How to reproduce**
Create a >100-line document with a title, badge images, and a genuine one-line prose blurb directly under the title, but no YAML front-matter block, followed by a normal `## Table of Contents`.

**Expected behavior**
A document with a clear, human-readable description near the top should not be flagged as missing one, regardless of whether it uses YAML front-matter.

**Actual behavior**
The document is still flagged `toc-missing-description` — a plausible false positive for common README styles. (A nearby test's docstring references "CodeRabbit PR #108," suggesting this front-matter-only scope was already discussed and may be intentional design rather than an oversight.)

```text
Doc: # Title\n[badge][badge]\nThis tool does X for people who need Y.\n## Table of Contents
_check_missing_description(): no YAML front-matter found -> toc-missing-description (even though a real blurb exists)
```

**Impact**
Possible noisy false positives on README-style docs that don't use YAML front-matter.

**Suggested correction**
Optional — rename the check/message to explicitly say "no YAML front-matter description field" so it reads as scoped rather than "no description at all"; or extend detection to accept a short prose blurb before the first `##` heading.

**How to verify**
Add a test with a badge+blurb, no-front-matter document and assert the intended (documented) behavior, whichever is chosen.

tests/test_toc.py
```diff
@@ -707,6 +707,255 @@ def test_toml_comments_in_fence_between_toc_and_heading(self):
...
+            "## Table of Contents\n\n"
+            "1. [A](#a)\n\n"
+            "---\n\n"
+            f"## A\n\n{filler}\n"
+        )
+        result = validate_toc(content, max_heading_level=2)
+        codes = [w["code"] for w in result["warnings"]]
+        assert "toc-missing-description" in codes
+
+    def test_jit_readiness_warnings_are_never_errors(self):
```
ainetx, line 1016: ### Test coverage gap: warning-only guarantee not proven via full CLI path

**Severity:** Minor

**Problem**
This is the only test exercising all four new warning codes together, but it also deliberately trips an unrelated pre-existing TOC-completeness error, and it calls `validate_toc()` directly rather than `cmd_validate_toc`.

**How to reproduce**
Read `test_jit_readiness_warnings_are_never_errors` and check whether it asserts an exit code via the CLI entry point.

**Expected behavior**
A dedicated test should prove "all four warnings, zero errors -> `cmd_validate_toc` returns 0" end-to-end.

**Actual behavior**
No such test exists — the behavior looks correct by code inspection (`cmd_validate_toc` returns `2` only `if total_errors`), but it's unverified end-to-end for the all-four-warnings, zero-errors case.

```text
test_jit_readiness_warnings_are_never_errors()
   calls validate_toc() directly, also trips an unrelated error
   -> does not prove: cmd_validate_toc([...]) == 0 when ONLY warnings fire
```

**Impact**
A regression in `cmd_validate_toc`'s exit-code logic for the warn-only case would not be caught by the current suite.

**Suggested correction**
Add a `TestCmdValidateToc` case with a document tripping all four JIT codes and a clean/complete TOC (no completeness/staleness errors), asserting `rc == 0` and `status == "WARN"` with `warning_count >= 4`.

**How to verify**
Run the new test and confirm it fails if `cmd_validate_toc`'s exit-code branch is deliberately broken (mutation check).

tests/test_doc_index.py
```diff
@@ -0,0 +1,331 @@
...
+        assert rc == 2
+        out = json.loads(capsys.readouterr().out)
+        assert out["status"] == "ERROR"
+
+    def test_basic(self, tmp_path: Path, capsys, monkeypatch):
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path)
+        rc = cmd_doc_index([str(f)])
+        assert rc == 0
+        out = json.loads(capsys.readouterr().out)
```
ainetx, line 377: ### Test coverage gap: corrupted-cache auto-rebuild not proven at CLI/exit-code level

**Severity:** Minor

**Problem**
This test only asserts `load_doc_index(f) is None` for a corrupted cache file — it never calls `cmd_doc_index([...])` afterward to confirm the command-level exit code stays `0` when transparently rebuilding after corruption.

**How to reproduce**
Read `test_load_returns_none_on_corrupt_cache_file` and check whether it calls `cmd_doc_index` after corrupting the cache.

**Expected behavior**
A dedicated test should prove the full CLI path recovers cleanly (`rc == 0`, fresh rebuild) from a corrupted on-disk cache.

**Actual behavior**
No such end-to-end test exists — the underlying behavior looks safe by inspection (the `except (json.JSONDecodeError, OSError)` path returns `None`, triggering a rebuild), but it's unverified at the CLI/exit-code layer.

```text
test_load_returns_none_on_corrupt_cache_file()
   asserts load_doc_index(f) is None only
   -> does not call cmd_doc_index([str(f)]) to confirm rc == 0
```

**Impact**
A regression in the CLI's handling of the rebuild-after-corruption path would not be caught by the current suite.

**Suggested correction**
Add a test that corrupts the cache file the same way, then calls `cmd_doc_index([str(f)])` and asserts `rc == 0` and `cache_hit is False`.

**How to verify**
Run the new test and confirm it fails if the corrupt-cache fallback path is deliberately broken (mutation check).

ainetx (commented):
Automated Deep Review — 12 Minor findings (10 inline below). Two more without a diff-anchorable line:

- **No Unicode (NFC/NFD) normalization in duplicate-heading comparison** (same root cause and same fix as the Major finding on `_check_duplicate_heading_titles` in the prior review — add `unicodedata.normalize("NFC", text)` alongside the casefold/whitespace fix).
- **`skills/studio/studio.clispec` not updated** for the new `doc-index` command or the `validate-toc --max-section-lines` flag. This file isn't touched by this PR's diff at all, so it can't be anchored inline. It's the repo's maintained per-command CLI reference (COMMAND/SYNOPSIS/OPTIONS/EXIT CODES/OUTPUT blocks for all 38 dispatchable commands) — no `COMMAND doc-index` entry exists, and `validate-toc`'s OPTIONS block doesn't list the new flag. Doc-only gap, no evidence of CI enforcement, so kept Minor.

ainetx:
### No Unicode (NFC/NFD) normalization in duplicate-heading comparison

**Severity:** Minor

**Problem**
`_check_duplicate_heading_titles` (skills/studio/scripts/studio/utils/toc.py) uses plain Python string equality with no `unicodedata.normalize` call anywhere in the file. This is the same root cause as the Major finding on this function (missing case/whitespace normalization) — Python does not consider NFC and NFD forms of the same visual text equal.

**How to reproduce**
Create two headings that are visually identical but encoded differently: `## café` (NFC, `b'caf\xc3\xa9'`) and `## café` (NFD, `b'cafe\xcc\x81'`).

**Expected behavior**
Both should be flagged as duplicate headings, since they render identically to a reader.

**Actual behavior**
Neither is flagged — `text in seen` compares the raw byte-distinct strings, which are unequal.

```text
seen = {}
"café" (NFC) -> seen["café(NFC)"] = line
"café" (NFD) -> "café(NFD)" not in seen -> not flagged as duplicate (false negative)
```

**Impact**
Genuinely duplicate/near-duplicate Unicode headings silently pass unflagged; a narrower instance of the same missing-normalization problem as the case/whitespace finding.

**Suggested correction**
Normalize with `unicodedata.normalize("NFC", text)` in addition to the casefold/whitespace-collapse fix already suggested for the case/whitespace finding on this same function.

**How to verify**
Add a test with one heading in NFC form and a visually-identical one in NFD form, asserting they're flagged as duplicates.

---

### `studio.clispec` CLI reference not updated for the new command/flag

**Severity:** Minor

**Problem**
`skills/studio/studio.clispec` is the repo's maintained per-command CLI reference (COMMAND/SYNOPSIS/DESCRIPTION/ARGUMENTS/OPTIONS/EXIT CODES/OUTPUT/EXAMPLE/RELATED blocks for all 38 dispatchable commands, including `toc` and `validate-toc`). This file is not touched anywhere in this PR's diff.

**How to reproduce**
Search `studio.clispec` for `doc-index` (no `COMMAND doc-index` entry exists) and check the existing `validate-toc` entry's OPTIONS block for `--max-section-lines` (absent).

**Expected behavior**
Every dispatchable command and its flags should have a corresponding entry in this reference.

**Actual behavior**
The new `cfs doc-index` command has no entry at all, and the existing `validate-toc` entry's OPTIONS block doesn't list the new `--max-section-lines` flag.

```text
studio.clispec:
  COMMAND toc          (exists)
  COMMAND validate-toc (exists, OPTIONS missing --max-section-lines)
  COMMAND doc-index    (missing entirely)
```

**Impact**
Doc-only gap — no evidence this file is enforced by CI or read at runtime, so it doesn't break anything functionally, but it misleads anyone (human or agent) consulting it for the current CLI surface.

**Suggested correction**
Add a `COMMAND doc-index` block (synopsis, description, exit codes, JSON output shape, example) mirroring neighboring entries, and add `--max-section-lines <number> [default: 300]` to the existing `validate-toc` OPTIONS block.

**How to verify**
After updating, confirm the new/updated blocks follow the same structure as sibling command entries in the file.

skills/studio/scripts/studio/commands/validate_toc.py
tkcoding: Fixed in 1854c123: _human_validate_toc now has an explicit `elif status == "WARN":` branch that iterates `r.get("warnings", [])` the same way the FAIL branch iterates errors, and _human_doc_index now prints `data['file']`. Regression tests: `test_warn_only_file_prints_warnings_in_human_output`, `test_all_four_jit_warnings_zero_errors_exits_clean_via_cli`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 1854c123: added `_has_schema_current_index()` (schema_version + required-field shape check), called by `load_doc_index()` before returning a cache hit -- an incomplete/hand-edited/pre-schema cache is now rebuilt instead of returned malformed. Regression tests: `test_cache_missing_a_required_field_is_rebuilt_not_returned`, `test_cache_from_an_older_schema_version_is_rebuilt_not_returned`, `test_cmd_doc_index_rebuilds_cleanly_after_corrupt_cache`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Addressed via documentation, per your first suggested option: the module's own docstring overclaimed "a write always advances mtime" unconditionally. Fixed in 1854c123 to state the real, accepted limitation explicitly (coarse-mtime-resolution filesystems can share an etag across two same-size edits within one tick), and explains why a full fix (content hash) would defeat this module's entire reason to exist (avoiding a read on cache hit). Not adding a stronger invalidation signal, since that would reintroduce the read this cache exists to avoid.

skills/studio/scripts/studio/utils/toc.py
tkcoding: Fixed in 1854c123: added `_normalize_heading_key()` (casefold + whitespace-collapse + NFC-normalize) used as the `seen` dict key in `_check_duplicate_heading_titles`, while the warning still displays the original text; `parse_headings_with_lines` now skips a leading YAML front-matter block via `_find_frontmatter_end`. Regression tests: `test_duplicate_detection_is_case_insensitive`, `test_duplicate_detection_collapses_internal_whitespace`, `test_duplicate_warning_still_shows_the_original_heading_text`, `test_frontmatter_hash_line_is_not_parsed_as_a_heading`.

skills/studio/scripts/studio/commands/doc_index.py
tkcoding: Fixed in 1854c123: removed the stray `@cpt-flow` tag from `commands/doc_index.py`'s docstring. Confirmed via `cfs validate` (0 errors) that this doesn't orphan anything -- `cpt-studio-flow-traceability-validation-validate` remains correctly attributed only to `commands/validate.py::cmd_validate`. Side effect worth flagging: since that tag was this file's only traceability marker, removing it moves `commands/doc_index.py` from a (wrongly) "covered" whole-file claim to fully uncovered in `cfs spec-coverage`'s per-file report -- the overall gate still passes ("All thresholds met"), and I didn't add a new, real flow definition since none was requested and inventing one felt like scope beyond this fix.

architecture/features/traceability-validation.md
tkcoding: Fixed in 1854c123: updated the stale RETURN-shape line to the real fields (`{status, files_validated, error_count, warning_count, results}`, with each `results[]` entry's actual shape), added the `--max-section-lines` default (300) next to its mention, and added a real output-field enumeration to the Document Index section (`{file, cache_hit, total_lines, section_count, sections}`, plus the underlying index's `schema_version`/`path`/`etag`).

skills/studio/scripts/studio/utils/toc.py
tkcoding: Fixed in 1854c123: `_check_section_lengths` now validates `max_section_lines` itself (`math.isfinite(...) and > 0`, falling back to `DEFAULT_MAX_SECTION_LINES` otherwise), independent of the CLI's argparse guard. Regression tests: `test_nan_max_section_lines_falls_back_to_the_default_instead_of_disabling_the_check`, `test_negative_max_section_lines_falls_back_to_the_default_instead_of_flagging_everything`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 1854c123: added `_SCHEMA_VERSION`/`schema_version`, checked alongside the etag and required fields in `_has_schema_current_index()` -- a future incompatible schema change will now invalidate an old cache instead of silently returning old-shape data. Regression test: `test_cache_from_an_older_schema_version_is_rebuilt_not_returned`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Acknowledged, no change made -- you already flagged this as informational with no suggested correction required for merge. `annotate_section_summary` is a documented external-caller hook (the future per-section-summary integration), same category as `write_concept_file`/`diff_stale_sections` added in later PRs in this stack; tracked as forward-looking API, not dead code (it's tested and whitelisted in `vulture_whitelist.py` with that rationale).

skills/studio/scripts/studio/utils/toc.py
tkcoding: Fixed in 1854c123: `parse_headings_with_lines` now accepts `skip_first`/`skip_toc_heading` (matching `parse_headings`'s own options), and `parse_headings` is now a thin wrapper delegating to it and stripping the line number -- one fence/heading-match implementation instead of two that could silently diverge. All 112 pre-existing `test_toc.py` tests continue to pass unchanged, confirming this is behavior-preserving.

skills/studio/scripts/studio/utils/toc.py
tkcoding: Acknowledged, not implemented -- you flagged this as optional/non-blocking with no evidence yet of warning fatigue on a real corpus. Will pick this up as a real follow-up if a team actually reports noise; adding a suppression flag speculatively, before there's a concrete need, felt like the wrong direction to add surface area in.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 1854c123: `save_doc_index` now writes to a per-process temp file (`<cache>.<pid>.tmp`) and `os.replace()`s it into place, so a concurrent reader observes either the complete old file or the complete new one, never a torn write.

skills/studio/scripts/studio/utils/toc.py
tkcoding: Investigated, not changed -- checked the actual warning text and it already says "Document has no frontmatter/description block at the top" (not "no description at all"), and the function's own docstring is explicit that this only recognizes a YAML frontmatter description field. Expanding detection to a non-frontmatter prose blurb is a real, separate feature with its own design questions (what counts as a clear blurb, how to handle badge images before it, etc.) rather than a message-wording fix, so I'm leaving this as a documented scope boundary rather than inventing prose-blurb detection as a quick addition. Open to a follow-up if this turns out to cause real false positives on READMEs in practice.

tests/test_toc.py
tkcoding: Fixed in 1854c123: added `test_all_four_jit_warnings_zero_errors_exits_clean_via_cli`, which drives a warn-only, zero-error document (all headings present in an otherwise-complete TOC) through `cmd_validate_toc` end to end and asserts `rc == 0`, `status == "WARN"`, `error_count == 0`.

tests/test_doc_index.py
tkcoding: Fixed in 1854c123: added `test_cmd_doc_index_rebuilds_cleanly_after_corrupt_cache`, which corrupts the on-disk cache the same way the existing unit test does, then calls `cmd_doc_index([str(f)])` and asserts `rc == 0` and `cache_hit is False` -- proving the rebuild-after-corruption path at the CLI/exit-code level, not just `load_doc_index()` in isolation.

architecture/features/traceability-validation.md
```diff
@@ -437,6 +441,32 @@ Catches structural and traceability issues that AI agents miss or hallucinate 
...
+
+**Output**: `cfs doc-index`'s JSON is `{file, cache_hit, total_lines, section_count, sections}`, each `sections[]` entry `{level, heading, line_start, line_end, summary}`. The underlying index dict additionally carries `schema_version`, `path`, and `etag`.
+
+A cached, read-once-per-file structural index for Markdown JIT retrieval (see
+constructorfabric/studio#104): parsing a file's headings/section boundaries
+happens once, not once per query, until the file's content actually changes.
+The cache-validity fingerprint is deliberately metadata-only (`mtime` + file
+size via `Path.stat()`), never a content hash — the point of the cache is to
+avoid reading the file at all on a hit, and a content hash would defeat that
+by requiring the read it's meant to save.
```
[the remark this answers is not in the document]
tkcoding: Fixed in f8774901: reworded the cache guarantee in `traceability-validation.md` to state what the fingerprint actually detects (`mtime_ns` + size), not "until content actually changes." The code docstring in `doc_index.py` (`_compute_etag`) already made this same honest limitation explicit in an earlier round; this brings the spec doc's wording in line with it.

skills/studio/scripts/studio/commands/doc_index.py
```diff
@@ -0,0 +1,63 @@
...
+
+    filepath = Path(args.file).resolve()
+    if not filepath.is_file():
+        ui.result(
+            {"file": str(filepath), "status": "ERROR", "message": "File not found"},
+            human_fn=lambda d: ui.error(f"{d['file']}: {d['message']}"),
+        )
+        return 2
+
+    index = get_or_build_doc_index(filepath, force_rebuild=args.rebuild)
```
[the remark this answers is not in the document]
tkcoding: Fixed in f8774901: `cmd_doc_index` now wraps `get_or_build_doc_index` in `except (OSError, UnicodeDecodeError)`, logs a warning, and returns the same clean exit-code-2 JSON error contract the missing-file case already uses. Added regression tests for both a non-UTF-8 file and an OSError raised from the underlying call.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,251 @@
...
+            "heading": text,
+            "line_start": line_start,
+            "line_end": line_end,
+            "summary": None,
+        })
+
+    return {
+        "schema_version": _SCHEMA_VERSION,
+        "path": str(canonical_path),
+        "etag": _compute_etag(canonical_path),
```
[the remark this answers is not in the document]
tkcoding: Fixed in f8774901: `build_doc_index` now computes the etag before reading the file's content (not after), so a concurrent write can only ever make the stored etag look *older* than what got parsed -- which the next `load_doc_index` call's stat comparison always catches -- rather than newer, which no later check could detect. Added a test that asserts the etag-then-read call order directly (not just the end-to-end behavior), to guard the ordering itself against regressing.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +730,189 @@ def _record_missing_toc_error(
...
+                path=path,
+                line=line,
+                heading_text=text,
+                section_length=section_length,
+                max_section_lines=max_section_lines,
+            ))
+    return warnings
+
+
+_DESCRIPTION_FIELD_RE = re.compile(r"^description\s*:\s*\S")
```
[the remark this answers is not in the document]
tkcoding: Fixed in f8774901: replaced the bare `\S`-after-colon check with `_is_real_description_value()`, which explicitly rejects an empty quoted scalar (`""`/`''`) and a comment-only value (`# TODO`) while still accepting real text (including a quoted value followed by a trailing comment). Added parametrized regression tests covering all three rejected forms.

skills/studio/scripts/studio/commands/doc_index.py
ainetx: Thanks for looking at this — but I don't think this is fixed, just moved to a different wrong flow.

`cpt-studio-flow-developer-experience-self-check` is defined in `architecture/features/developer-experience.md` as the **`cfs self-check`** flow (validate kit examples against templates/constraints — `inst-user-self-check`, `inst-load-registry`, `inst-for-each-kit`, etc.). None of those steps relate to `cfs doc-index` either, same class of mismatch as the original `...traceability-validation-validate` reference.

`cfs doc-index` doesn't appear to have its own flow definition anywhere yet. Suggest either removing the `@cpt-flow` tag entirely (an algo-level tag alone, which this file already has correctly via the `utils/doc_index.py` side, may be sufficient), or adding a real `cpt-studio-flow-...` definition for `cfs doc-index` and pointing here at that.

(Verified by an automated Deep Review re-check after this PR's new commits — see the note this triggers on the review side.)

ainetx: Confirmed fixed in the latest commit (`1e8384b`) — the stray `@cpt-flow` tag has been removed entirely from the module docstring, and `cmd_doc_index`/`_human_doc_index` now carry correct algo-level `@cpt-begin`/`@cpt-end` markers instead. No more incorrect flow attribution. Thanks for the fix.

(Automated Deep Review re-verification.)

PR #109, opened 2026-08-28
feat(doc-index): section-granularity inference, hashing, and caching fixes

tkcoding:
## Stacked on #108 (constructorfabric/studio#108)

This branches from `jit-retrieval-doc-index` (#108) — `doc_index.py`/`toc.py` don't exist on `main` yet, so until #108 merges this diff includes #108's changes too. Once #108 lands, I'll rebase this branch onto the new `main`; the diff here will shrink to just what this PR adds.

## Summary

Infers real per-document section boundaries for JIT retrieval, hashes them for per-section staleness detection, and fixes a caching race condition, a legacy-cache crash, and two description-check gaps found during review.

### Core feature

- `infer_section_level()` picks which heading level represents one real retrievable section, using the level's frequency as the signal — PDF-to-Markdown conversion assigns heading levels by font-size heuristics, not semantic depth, so a document's real recurring structure (its chapters) shows up as the level used most often, while an occasional heading at an anomalous level is rare precisely because it's noise. Verified against the real PDF-converted document that originally exposed this bug: 12 correctly-sized sections instead of one fake 6,601-line mega-section.
- `retrieval_sections` groups headings at exactly the inferred level, each with a SHA-256 hash of its own text.
- `diff_stale_sections()` compares a file's current content against its last cached build at this granularity and reports which sections actually changed, matched by position — heading text alone can't disambiguate a duplicate title.

### Fixes found during review (not part of the original scope)

- `load_doc_index()` now schema-validates a cached index, not just its etag — a cache written before `section_level`/`retrieval_sections` existed could otherwise pass a matching-etag check and then crash the caller with a `KeyError`.
- `_read_with_stable_etag()` closes a real race: a write landing between reading a file's content and computing its etag could previously save headings from the old content stamped with the new etag, serving stale data until a later edit changed the etag again.
- `diff_stale_sections()`'s changed/unchanged entries now carry `line_start` alongside heading text, since two sections sharing a duplicate title otherwise can't be told apart.
- `annotate_section_summary()` now updates the matching entry in both `sections` and `retrieval_sections`, not just the former.
- `cmd_doc_index()` now exposes `retrieval_sections`/`section_level` in both JSON and human output — they existed in the index but weren't reachable through the CLI.
- `toc.py`'s description check now rejects a YAML comment (`description: # TODO`), an empty quoted string (`description: ""`), and an empty block-scalar (`description: |` with no indented content) — previously all three satisfied the check despite carrying no real value.

## Test plan

- [x] `pytest tests/test_doc_index.py tests/test_toc.py` — 172 passed, 100% coverage on touched files
- [x] Full suite — 4831 passed; the same pre-existing macOS-local/flaky failures seen on #108, none in files touched here
- [x] `pylint` / `vulture` clean; `cfs validate` 0 errors; `spec-coverage` thresholds met
- [x] `infer_section_level`/`retrieval_sections` re-verified against the real PDF-converted document after every fix round — still 12 correct sections
- [x] All 7 CodeRabbit findings across two review rounds fixed and confirmed on-thread

### DIFF
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -23,7 +23,10 @@ def cmd_doc_index(argv: List[str]) -> int:
         prog="cfs doc-index",
         description=(
             "Build or reuse a cached heading/section index for a Markdown file, "
-            "so navigation reads the file's structure once, not once per query."
+            "so navigation reads the file's structure once, not once per query. "
+            "section_level is inferred from the most frequently repeated heading "
+            "level (ties prefer the shallower level); a level used only once is "
+            "never chosen."
         ),
     )
     p.add_argument("file", help="Markdown file path")
--- a/skills/studio/scripts/studio/commands/doc_index.py
+++ b/skills/studio/scripts/studio/commands/doc_index.py
@@ -58,6 +58,9 @@ def cmd_doc_index(argv: List[str]) -> int:
         "total_lines": index["total_lines"],
         "section_count": len(index["sections"]),
         "sections": index["sections"],
+        "section_level": index["section_level"],
+        "retrieval_section_count": len(index["retrieval_sections"]),
+        "retrieval_sections": index["retrieval_sections"],
     }
     ui.result(output, human_fn=_human_doc_index)
     return 0
@@ -74,5 +77,13 @@ def _human_doc_index(data: dict) -> None:
     for s in data["sections"]:
         summary = f" — {s['summary']}" if s.get("summary") else ""
         ui.substep(f"  H{s['level']} [{s['line_start']}-{s['line_end']}] {s['heading']}{summary}")
+    ui.blank()
+
+    level = data["section_level"]
+    ui.step(f"Retrieval sections (level {level}, {data['retrieval_section_count']} section(s))" if level is not None
+             else "Retrieval sections (no headings — none inferred)")
+    for s in data["retrieval_sections"]:
+        summary = f" — {s['summary']}" if s.get("summary") else ""
+        ui.substep(f"  [{s['line_start']}-{s['line_end']}] {s['heading']} ({s['hash'][:12]}){summary}")
     ui.blank()
 # @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-cmd-format
--- a/skills/studio/scripts/studio/commands/validate_toc.py
+++ b/skills/studio/scripts/studio/commands/validate_toc.py
@@ -14,9 +14,57 @@ import argparse
 from pathlib import Path
 from typing import List
 
+from ..utils import error_codes as EC
 from ..utils.toc import DEFAULT_MAX_SECTION_LINES, add_toc_max_level_argument, validate_toc
 from ..utils.ui import ui
 # @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-imports
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-validate-one
+def _validate_one_file(filepath: Path, args: argparse.Namespace) -> dict:
+    """Validate a single file, returning its result dict. Never raises --
+    a missing file or a read failure (permission denied, binary/non-UTF-8
+    content, a TOCTOU race) is reported as an ERROR result instead, so one
+    bad file in a batch can't abort validation of the rest.
+    """
+    if not filepath.is_file():
+        return {
+            "file": str(filepath),
+            "status": "ERROR",
+            "message": "File not found",
+            "code": EC.FILE_LOAD_ERROR,
+        }
+
+    try:
+        content = filepath.read_text(encoding="utf-8")
+    except (OSError, UnicodeDecodeError) as exc:
+        return {
+            "file": str(filepath),
+            "status": "ERROR",
+            "message": f"Could not read file: {exc}",
+            "code": EC.FILE_READ_ERROR,
+        }
+
+    report = validate_toc(
+        content,
+        artifact_path=filepath,
+        max_heading_level=args.max_level,
+        max_section_lines=args.max_section_lines,
+    )
+    errors = report.get("errors", [])
+    warnings = report.get("warnings", [])
+    file_result: dict = {
+        "file": str(filepath),
+        "status": "FAIL" if errors else ("WARN" if warnings else "PASS"),
+        "error_count": len(errors),
+        "warning_count": len(warnings),
+    }
+    if args.verbose or errors:
+        file_result["errors"] = errors
+    if args.verbose or warnings:
+        file_result["warnings"] = warnings
+    return file_result
+# @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-validate-one
+
 
 def cmd_validate_toc(argv: List[str]) -> int:
     """Validate Table of Contents in markdown files."""
@@ -46,50 +94,19 @@ def cmd_validate_toc(argv: List[str]) -> int:
     # @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-parse-args
 
     # @cpt-begin:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-resolve-files
-    results = []
-    total_errors = 0
-    total_warnings = 0
     files_to_validate = [Path(f).resolve() for f in args.files]
     # @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-resolve-files
 
     # @cpt-begin:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-foreach-file
-    for filepath in files_to_validate:
-
-        if not filepath.is_file():
-            results.append({
-                "file": str(filepath),
-                "status": "ERROR",
-                "message": "File not found",
-            })
+    results = [_validate_one_file(filepath, args) for filepath in files_to_validate]
+    total_errors = 0
+    total_warnings = 0
+    for file_result in results:
+        if file_result["status"] == "ERROR":
             total_errors += 1
-            continue
-
-        content = filepath.read_text(encoding="utf-8")
-        report = validate_toc(
-            content,
-            artifact_path=filepath,
-            max_heading_level=args.max_level,
-            max_section_lines=args.max_section_lines,
-        )
-
-        errors = report.get("errors", [])
-        warnings = report.get("warnings", [])
-        total_errors += len(errors)
-        total_warnings += len(warnings)
-
-        file_result: dict = {
-            "file": str(filepath),
-            "status": "FAIL" if errors else ("WARN" if warnings else "PASS"),
-            "error_count": len(errors),
-            "warning_count": len(warnings),
-        }
-
-        if args.verbose or errors:
-            file_result["errors"] = errors
-        if args.verbose or warnings:
-            file_result["warnings"] = warnings
-
-        results.append(file_result)
+        else:
+            total_errors += file_result["error_count"]
+            total_warnings += file_result["warning_count"]
     # @cpt-end:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-foreach-file
 
     # @cpt-begin:cpt-studio-algo-traceability-validation-validate-toc:p1:inst-toc-return
--- a/skills/studio/scripts/studio/utils/doc_index.py
+++ b/skills/studio/scripts/studio/utils/doc_index.py
@@ -269,7 +269,7 @@ def _read_cache_file(cache_path: Path) -> Optional[Dict[str, Any]]:
 
 
 # @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-load
-_REQUIRED_INDEX_FIELDS = ("total_lines", "sections")
+_REQUIRED_INDEX_FIELDS = ("total_lines", "sections", "section_level", "retrieval_sections")
 
 
 def _has_schema_current_index(cached: Dict[str, Any]) -> bool:
@@ -293,6 +293,13 @@ def load_doc_index(path: Path) -> Optional[Dict[str, Any]]:
     read (the property the whole cache exists to provide). Only a stale or
     absent cache falls through to :func:`build_doc_index`, which does the
     one real read.
+
+    A matching etag alone isn't enough: a cache written by an older version
+    of this module (before ``section_level``/``retrieval_sections``
+    existed) can have a matching etag if the file hasn't changed since, but
+    a caller reading those fields on it would hit a ``KeyError`` rather
+    than a clean rebuild. Treated the same as a stale cache -- rebuilt,
+    not crashed on.
     """
     cache_path = _index_cache_path(path)
     if cache_path is None or not cache_path.is_file():
--- a/skills/studio/scripts/studio/utils/doc_index.py
+++ b/skills/studio/scripts/studio/utils/doc_index.py
@@ -12,6 +12,18 @@ re-derived each time.
 Scope: Markdown only. PDF/DOCX conversion is a separate concern (Layer 1);
 this module operates purely on already-plain-text content (Layer 2).
 
+Schema contract: adding a new top-level key to the index dict is always
+additive and does not require bumping ``_SCHEMA_VERSION`` -- an older
+build of this module simply never wrote that key, which
+``_has_schema_current_index``'s required-field check already treats as
+"predates the current schema" and rebuilds. ``_SCHEMA_VERSION`` exists for
+the other kind of change: an existing key's *meaning* or *shape* changing
+incompatibly (e.g. what ``retrieval_sections`` entries contain), which a
+field-presence check alone can't detect since the field is still there,
+just holding something a new reader would misinterpret. Bump
+``_SCHEMA_VERSION`` for that kind of change; a plain new field needs only
+listing in ``_REQUIRED_INDEX_FIELDS`` if a consumer reads it unconditionally.
+
 See constructorfabric/studio#104.
 
 @cpt-algo:cpt-studio-algo-traceability-validation-doc-index:p1
@@ -85,8 +97,11 @@ def _index_cache_path(path: Path) -> Optional[Path]:
     except OSError as exc:
         # A file whose parent can't be stat'd (permissions, a race) is not a
         # reason to fail the caller -- just an uncached build, like "no
-        # Studio directory found".
-        logger.debug("doc-index cache path lookup skipped for %s: %s", path, exc)
+        # Studio directory found". Warning, not debug: this is a genuine
+        # anomaly (unlike the ordinary, unlogged "no Studio directory"
+        # case below), and should be visible at the CLI's default log
+        # level rather than indistinguishable from a routine cache miss.
+        logger.warning("doc-index cache path lookup failed for %s: %s", path, exc)
         studio_dir = None
     if studio_dir is None:
         return None
@@ -149,10 +164,15 @@ def _build_retrieval_sections(
     whichever ``section_level`` section it falls under, rather than
     splitting a real section apart.
 
-    Each section's ``hash`` is a SHA-256 of its own text slice -- the
-    per-section granularity :func:`diff_stale_sections` needs to tell "this
-    one section changed" from "the whole file changed", which a whole-file
-    fingerprint structurally cannot do.
+    Each section's ``hash`` is a SHA-256 of its own text slice, with each
+    line's trailing whitespace stripped before hashing -- a harmless
+    "trim trailing whitespace on save" edit (a common editor/IDE default)
+    changes no meaningful content and must not look like a real edit to
+    :func:`diff_stale_sections`, which is the whole point of hashing at
+    section granularity in the first place. The per-section granularity
+    :func:`diff_stale_sections` needs to tell "this one section changed"
+    from "the whole file changed", which a whole-file fingerprint
+    structurally cannot do.
     """
     if section_level is None:
         return []
@@ -161,12 +181,12 @@ def _build_retrieval_sections(
     sections: List[Dict[str, Any]] = []
     for i, (text, line_start) in enumerate(marks):
         line_end = marks[i + 1][1] - 1 if i + 1 < len(marks) else line_count
-        section_text = "\n".join(lines[line_start - 1:line_end])
+        hash_text = "\n".join(line.rstrip() for line in lines[line_start - 1:line_end])
         sections.append({
             "heading": text,
             "line_start": line_start,
             "line_end": line_end,
-            "hash": hashlib.sha256(section_text.encode("utf-8")).hexdigest(),
+            "hash": hashlib.sha256(hash_text.encode("utf-8")).hexdigest(),
             "summary": None,
         })
     return sections
@@ -264,7 +284,12 @@ def _read_cache_file(cache_path: Path) -> Optional[Dict[str, Any]]:
     try:
         return json.loads(cache_path.read_text(encoding="utf-8"))
     except (json.JSONDecodeError, OSError) as exc:
-        logger.debug("doc-index cache unreadable at %s: %s", cache_path, exc)
+        # Reached only once the caller has already confirmed the cache file
+        # exists, so a failure here is real corruption or a permissions
+        # problem, not a routine cache miss -- warning, not debug, so it's
+        # visible at the CLI's default log level instead of masquerading
+        # as an ordinary first-time build.
+        logger.warning("doc-index cache unreadable at %s: %s", cache_path, exc)
         return None
 
 
@@ -313,7 +338,10 @@ def load_doc_index(path: Path) -> Optional[Dict[str, Any]]:
     try:
         current_etag = _compute_etag(canonical_path)
     except OSError as exc:
-        logger.debug("doc-index staleness check failed for %s: %s", path, exc)
+        # The cache file was just confirmed to exist, so a stat() failure
+        # on the *source* file here means it vanished or became unreadable
+        # since -- a real anomaly, not a routine miss.
+        logger.warning("doc-index staleness check failed for %s: %s", path, exc)
         return None
 
     if cached.get("etag") != current_etag:
@@ -374,7 +402,13 @@ def _compute_fresh_retrieval_sections(path: Path) -> Optional[List[Dict[str, Any
     canonical_path = path.resolve()
     try:
         content = canonical_path.read_text(encoding="utf-8")
-    except OSError as exc:
+    except (OSError, UnicodeDecodeError) as exc:
+        # Deliberately still debug, unlike this module's other fallback
+        # logs: this one path has a genuinely expected trigger ("the file
+        # was deleted after it was cached", per this function's own
+        # contract) alongside the anomalous ones, so promoting it would
+        # make a normal outcome noisy rather than making a real anomaly
+        # visible.
         logger.debug("doc-index section diff failed for %s: %s", path, exc)
         return None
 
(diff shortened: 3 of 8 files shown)


skills/studio/scripts/studio/commands/doc_index.py
```diff
@@ -0,0 +1,64 @@
...
+
+    index = get_or_build_doc_index(filepath, force_rebuild=args.rebuild)
+
+    output = {
+        "file": str(filepath),
+        "cache_hit": index["cache_hit"],
+        "total_lines": index["total_lines"],
+        "section_count": len(index["sections"]),
+        "sections": index["sections"],
+    }
```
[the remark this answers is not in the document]
tkcoding: Fixed in 66b9cc27 — `cmd_doc_index()` now includes `section_level`, `retrieval_section_count`, and `retrieval_sections` in the JSON output, and `_human_doc_index` lists them the same way it already lists the finer-grained `sections`. Added `test_json_output_exposes_retrieval_sections` and `test_human_output_lists_retrieval_sections`.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,378 @@
...
+            "line_end": line_end,
+            "summary": None,
+        })
+
+    section_level = infer_section_level(headings)
+
+    return {
+        "path": str(canonical_path),
+        "etag": _compute_etag(canonical_path),
+        "built_at": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
```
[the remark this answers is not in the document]
tkcoding: Fixed in 66b9cc27 — added `_read_with_stable_etag()`, which brackets the content read with an etag snapshot on each side and retries on mismatch, so the saved etag is provably the one that matches what was actually parsed. Added `test_retries_when_the_file_changes_mid_read` and `test_gives_up_after_max_attempts_under_sustained_contention` (the latter confirms it still returns a real read rather than erroring under sustained contention, failing in the safe direction).

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,378 @@
...
+            "structural_change": True,
+            "unchanged": [],
+            "changed": [s["heading"] for s in fresh_sections],
+        }
+
+    unchanged: List[str] = []
+    changed: List[str] = []
+    for old, new in zip(old_sections, fresh_sections):
+        (unchanged if old["hash"] == new["hash"] else changed).append(new["heading"])
+    return {"structural_change": False, "unchanged": unchanged, "changed": changed}
```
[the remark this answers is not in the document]
tkcoding: Fixed in 66b9cc27 — both `changed` and `unchanged` entries are now `{"heading": str, "line_start": int}` instead of bare heading text. Added `test_duplicate_headings_are_disambiguated_by_line_start`, and updated the existing tests' assertions to the new shape. Also added the `strict=True` you flagged via Ruff on the `zip()` call.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,378 @@
...
+    matched = False
+    for section in index["sections"]:
+        if section["line_start"] == line_start:
+            section["summary"] = summary
+            matched = True
+            break
+    if not matched:
+        return False
+
+    save_doc_index(path, index)
```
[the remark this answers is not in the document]
tkcoding: Fixed in 66b9cc27 — `annotate_section_summary()` now also updates the matching `retrieval_sections` entry when one exists at the same `line_start` (an off-level heading that matches only `sections` correctly leaves `retrieval_sections` untouched, since there's no corresponding entry to update). Added `test_updates_matching_retrieval_section_too` and `test_off_level_heading_leaves_retrieval_sections_untouched`.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,167 @@ def _record_missing_toc_error(
...
+    """Check whether a YAML frontmatter block declares a non-empty ``description``.
+
+    ``frontmatter_end`` is the index returned by :func:`_find_frontmatter_end`
+    (one past the closing ``---``); the body being scanned is
+    ``lines[1:frontmatter_end - 1]``, excluding both delimiter lines.
+    """
+    return any(
+        _DESCRIPTION_FIELD_RE.match(line.strip())
+        for line in lines[1:frontmatter_end - 1]
+    )
```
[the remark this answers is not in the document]
tkcoding: Fixed in 66b9cc27 — `_frontmatter_has_description()` now parses the field's actual value: a YAML comment (`# ...`) or an empty/whitespace-only quoted string no longer counts as a description. Added `test_comment_only_description_value_still_warns`, `test_empty_quoted_description_value_still_warns`, and `test_real_description_after_regex_tightening_still_suppresses_warning` (confirming the tightened check doesn't overcorrect against a genuinely populated one).

skills/studio/scripts/studio/commands/doc_index.py
```diff
@@ -0,0 +1,75 @@
...
+
+    output = {
+        "file": str(filepath),
+        "cache_hit": index["cache_hit"],
+        "total_lines": index["total_lines"],
+        "section_count": len(index["sections"]),
+        "sections": index["sections"],
+        "section_level": index["section_level"],
+        "retrieval_section_count": len(index["retrieval_sections"]),
+        "retrieval_sections": index["retrieval_sections"],
```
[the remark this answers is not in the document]
tkcoding: Fixed in 9cf51f4c — `load_doc_index()` now also checks that the cached dict has `section_level` and `retrieval_sections` before returning it; a legacy cache missing either is treated the same as a stale one (rebuilt, not returned). Added `test_legacy_cache_with_matching_etag_but_old_schema_is_rebuilt_not_returned` and `test_cmd_doc_index_does_not_crash_on_a_legacy_cache`.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,185 @@ def _record_missing_toc_error(
...
+        value = match.group(1).strip()
+        if not value or value.startswith("#"):
+            continue
+        if value[0] in "\"'":
+            quote = value[0]
+            closing = value.find(quote, 1)
+            inner = value[1:closing] if closing != -1 else value[1:]
+            if not inner.strip():
+                continue
+        return True
```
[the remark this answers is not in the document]
tkcoding: Fixed in 9cf51f4c — added block-scalar detection (`|`, `>`, and modifiers like `|-`/`>+`): the marker line itself is never treated as content now, only real indented lines beneath it are (blank lines in between are skipped correctly, not mistaken for the block ending). Added tests for an empty block scalar (still warns), a populated one (suppresses the warning), a leading blank line before real content, and the folded (`>-`) variant.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+    section_level = infer_section_level(headings)
+
+    return {
+        "path": str(canonical_path),
+        "etag": etag,
+        "built_at": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
+        "total_lines": line_count,
+        "sections": sections,
+        "section_level": section_level,
+        "retrieval_sections": _build_retrieval_sections(headings, lines, section_level),
```
ainetx, line 270: ### No documented additive-only / schema-version contract for the new retrieval_sections/section_level cache fields

**Severity:** Minor

**Problem**
The PR adds `retrieval_sections`, `section_level`, and per-section `hash` fields to the persisted doc-index cache JSON (`build_doc_index` in doc_index.py). Nowhere in the codebase — not in doc_index.py's docstrings, not in `architecture/features/traceability-validation.md` — is there a documented contract stating that this addition is additive-only and safe for strict-schema readers, nor is there any schema-version marker a strict reader could check against.

**How to reproduce**
Grep doc_index.py and architecture/features/traceability-validation.md for "additive", "schema_version", "backward-compat" — none found relating to the doc-index cache format.

**Expected behavior**
A schema-evolution contract (e.g. a documented statement that new top-level keys are always additive, or an explicit `schema_version`/`format_version` field) so that any strict-schema consumer has a documented, checkable way to know new fields won't appear unannounced.

**Actual behavior**
No such contract or version marker exists anywhere in the codebase or docs. (Note: as of commit 9cf51f4c, `load_doc_index` does now check for required-field presence to detect an old-format cache, which mitigates the practical staleness risk — but there is still no explicit version field or documented additive-only contract for future schema evolution.)

**Impact**
Future consumers with strict schema validation have no documented guarantee and no version field to gate on; a future breaking change to this cache format also has no version field to bump.

**Suggested correction**
Add a `format_version` (or `schema_version`) field to the doc-index cache dict, and document in the module docstring that new top-level keys are additive-only within a given format_version.

**How to verify**
After the fix, grep for the new version field in build_doc_index's returned dict and confirm docstring language stating the additive-only guarantee.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+    describes the content just read. If they don't, retry. After
+    ``_MAX_READ_ATTEMPTS`` under sustained contention, return the last read
+    anyway, stamped with its own trailing etag -- the safe direction to
+    fail in, since a file still being rewritten that fast will simply look
+    stale again on the very next check, never silently wrong.
+    """
+    etag_after = _compute_etag(path)
+    for _ in range(_MAX_READ_ATTEMPTS):
+        etag_before = etag_after
+        content = path.read_text(encoding="utf-8")
```
ainetx, line 220: ### Non-UTF-8/binary Markdown input crashes build_doc_index with an uncaught UnicodeDecodeError

**Severity:** Minor

**Problem**
`_read_with_stable_etag()` in `skills/studio/scripts/studio/utils/doc_index.py` reads file content with `path.read_text(encoding="utf-8")` and no exception handling around the decode. This is the sole content-read path feeding `build_doc_index()`, which is in turn called by `get_or_build_doc_index()` and the `cfs doc-index` CLI command. None of these call sites check the file's content type or catch `UnicodeDecodeError`.

**How to reproduce**
```python
from pathlib import Path
from studio.utils.doc_index import build_doc_index

p = Path("/tmp/binary.md")
p.write_bytes(b"\xff\xfe\x00\x01garbage")
build_doc_index(p)  # raises UnicodeDecodeError, uncaught
```
Equivalently, `cfs doc-index /tmp/binary.md` crashes with a raw Python traceback instead of a clean CLI error.

**Expected behavior**
A non-Markdown or binary file passed to `doc-index` (or fed into `build_doc_index`/`get_or_build_doc_index`) should produce a clear, actionable error (e.g. a `ui.error(...)`-style message via the same error-reporting convention already used for "File not found" in `cmd_doc_index`), not an unhandled stack trace.

**Actual behavior**
`path.read_text(encoding="utf-8")` raises `UnicodeDecodeError` uncaught, propagating out of `build_doc_index` → `get_or_build_doc_index` → `cmd_doc_index`, producing a raw traceback and non-zero-but-uninformative failure for the CLI user.

**Impact**
Minor but real robustness gap: any caller pointing `doc-index`/`build_doc_index` at a binary file, a non-UTF-8-encoded text file, or an accidentally-wrong path (e.g. a PDF instead of its converted Markdown, which this feature explicitly exists to support) gets an unhandled crash instead of a diagnosable error message.

**Suggested correction**
Wrap the `path.read_text(...)` call in `_read_with_stable_etag` (or at the `build_doc_index`/`cmd_doc_index` boundary) in a `try/except UnicodeDecodeError`, and surface a clear error consistent with the existing "File not found" pattern in `cmd_doc_index`.

**How to verify**
Add a regression test that writes a file with invalid UTF-8 bytes and asserts `build_doc_index`/`cmd_doc_index` raises or reports a clear, typed error rather than an unhandled `UnicodeDecodeError`.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+    marks = [(text, line_start) for level, text, line_start in headings_with_lines if level == section_level]
+    sections: List[Dict[str, Any]] = []
+    for i, (text, line_start) in enumerate(marks):
+        line_end = marks[i + 1][1] - 1 if i + 1 < len(marks) else line_count
+        section_text = "\n".join(lines[line_start - 1:line_end])
+        sections.append({
+            "heading": text,
+            "line_start": line_start,
+            "line_end": line_end,
+            "hash": hashlib.sha256(section_text.encode("utf-8")).hexdigest(),
```
ainetx: ### Trailing-whitespace-only edits are not staleness-inert

**Severity:** Minor

**Problem**
`_build_retrieval_sections` hashes `section_text = "\n".join(lines[line_start-1:line_end])` verbatim via `hashlib.sha256(section_text.encode("utf-8"))`, with no trailing-whitespace normalization. (Line-ending differences, e.g. LF vs CRLF, are already safely handled by Python's universal-newline translation in `read_text()` — this finding is scoped only to trailing spaces/tabs, not newline style.)

**How to reproduce**
1. Build a doc index for a markdown file with at least two heading sections at the inferred `section_level`.
2. Without changing any word/heading/newline, add a single trailing space to the end of one existing line inside a section, and save.
3. Call `diff_stale_sections(path)`.

**Expected behavior**
A whitespace-only edit that doesn't change meaningful content should not cause `diff_stale_sections` to report the section as `changed`.

**Actual behavior**
The modified section's hash differs from the cached hash (trailing whitespace is part of the hashed bytes), so `diff_stale_sections` places it in `changed`, not `unchanged`.

**Impact**
Defeats the partial-rebuild/no-false-staleness optimization for a common, harmless class of edits (editor/IDE "trim trailing whitespace on save"), potentially triggering unnecessary re-summarization on every such save.

**Suggested correction**
Either strip trailing whitespace per line before hashing in `_build_retrieval_sections` (and document the normalization), or explicitly document that hashing is byte-sensitive to trailing whitespace and scope the "staleness-inert" claim to newline-sequence differences only.

**How to verify**
Add a regression test that builds an index, rewrites the file changing only trailing whitespace on one line within one retrieval section, and asserts the desired `diff_stale_sections` behavior for that case.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+from .toc import parse_headings_with_lines
+
+logger = logging.getLogger(__name__)
+
+_CACHE_SUBDIR = ".cache"
+_INDEX_CACHE_DIR = "doc-index"
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-etag
+def _compute_etag(path: Path) -> str:
```
ainetx, line 58: ### _compute_etag can collide across genuinely different content on coarse-mtime-resolution filesystems

**Severity:** Minor

**Problem**
`_compute_etag()` derives the cache-validity fingerprint solely from `st_mtime_ns` and `st_size`. The docstring justifies this by claiming "a write always advances mtime," but that doesn't hold universally: on filesystems/mounts with coarse mtime resolution (some network mounts, overlay/FUSE filesystems, older FAT-family filesystems), two distinct writes to the same file within one mtime tick, each producing content of the same byte size, report an identical `(mtime_ns, size)` pair even though the content differs.

**How to reproduce**
On a filesystem with coarse mtime granularity (or by mocking `Path.stat()` to return the same `st_mtime_ns`/`st_size` for two different byte contents): cache an index for content A via `get_or_build_doc_index(path)`, overwrite the file with same-length content B within the same mtime tick, then call `load_doc_index(path)` again — it returns the cached index built from content A.

**Expected behavior**
A cache-validity check should not return a hit when the file's actual content has changed, regardless of filesystem timestamp resolution.

**Actual behavior**
`load_doc_index()` compares only `cached.get("etag") != current_etag`; a same-size write within one mtime tick is silently treated as "unchanged" and stale cached `sections`/`retrieval_sections` are returned.

**Impact**
Minor but real: on affected filesystems/mounts, a caller can silently retrieve or diff against stale section content and stale summaries after a genuine edit, with no error or staleness signal. Likelihood is low on typical local disks with nanosecond mtime resolution.

**Suggested correction**
Either document this as an accepted, bounded limitation specific to coarse-mtime filesystems, or add an inexpensive secondary signal less prone to same-tick collisions (e.g. combining `st_mtime_ns` with `st_ctime_ns`/inode info where available).

**How to verify**
Add a regression test that monkeypatches `Path.stat()` (or the etag computation) to return identical `(mtime_ns, size)` for two different content strings and confirms the current stale-return behavior, then re-run after any fix.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+        "built_at": time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
+        "total_lines": line_count,
+        "sections": sections,
+        "section_level": section_level,
+        "retrieval_sections": _build_retrieval_sections(headings, lines, section_level),
+    }
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-build
+
+
+def _read_cache_file(cache_path: Path) -> Optional[Dict[str, Any]]:
```
ainetx, line 275: ### All new disk-cache fallback paths in utils/doc_index.py are silent (logger.debug only), indistinguishable from a clean "no cache" state

**Severity:** Minor

**Problem**
Four new `except` blocks in `skills/studio/scripts/studio/utils/doc_index.py` handle real failure conditions (unreadable/corrupt cache file, `OSError` during path resolution or staleness check) but only call `logger.debug(...)` before falling back — no warning-level log, no field on the returned index, no signal to any caller. Sites: `_index_cache_path` (OSError on Studio-directory lookup), `_read_cache_file` (JSONDecodeError/OSError on the cache file), `load_doc_index` (OSError computing etag), `_compute_fresh_retrieval_sections` (OSError reading the file for a diff).

**How to reproduce**
Corrupt an existing `.cache/doc-index/<slug>.json` file (truncate to invalid JSON) or revoke read permission, then call `get_or_build_doc_index(path)` at default (non-debug) logging.

**Expected behavior**
A broken/corrupt/permission-denied cache should be distinguishable from a normal first-time cache miss — e.g. via a warning-level log by default, or a field on the returned index similar to `cache_hit`.

**Actual behavior**
All four paths degrade identically and silently, logged only at `debug` level (off by default), so a systemic problem (e.g. a permissions misconfiguration) looks exactly like normal cache misses.

**Impact**
Repeated, unexplained full rebuilds (defeating the "read once per file" caching this feature exists to provide) would go unnoticed without manually enabling debug logging.

**Suggested correction**
Promote these four to `logger.warning`, and/or add a boolean field such as `index["cache_error"]` alongside the existing `cache_hit` field.

**How to verify**
Add a regression test that corrupts a cache file and asserts a warning-level log record is emitted, or that the returned index carries an error-fallback marker.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,458 @@
...
+        return None
+    if any(field not in cached for field in _REQUIRED_INDEX_FIELDS):
+        logger.debug("doc-index cache for %s predates the current schema; rebuilding", path)
+        return None
+    return cached
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-load
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-save
+def save_doc_index(path: Path, index: Dict[str, Any]) -> None:
```
ainetx, line 357: ### save_doc_index() writes the cache file non-atomically with no cross-process lock, risking lost updates under concurrent writers

**Severity:** Minor-to-moderate

**Problem**
`save_doc_index()` persists the index with a plain truncating write (`cache_path.write_text(json.dumps(index, indent=2), encoding="utf-8")`) — no write-to-temp-then-`os.replace()` atomic-rename pattern, and no file lock coordinating concurrent writers. `annotate_section_summary()` goes through this same function after a read-modify-write cycle, which is a classic lost-update window: two processes annotating different sections of the same file's index around the same time can each load the same base index, mutate their own section, and whichever writes last silently overwrites the other's update.

**How to reproduce**
Two processes call `annotate_section_summary(path, line_start=10, summary="A")` and `annotate_section_summary(path, line_start=40, summary="B")` concurrently — both load the same cached index (no summaries yet), mutate their own section, and write back; whichever `write_text()` call lands second overwrites the other's `line_start` entry entirely.

**Expected behavior**
Either the write should be atomic (temp file + `os.replace()`), and/or writes should be serialized (e.g. via a lock file) so two concurrent read-modify-write cycles don't clobber each other.

**Actual behavior**
`write_text()` truncates and writes in place with no atomicity guarantee beyond a single `write()` syscall, and no locking — concurrent `annotate_section_summary` calls (or a save racing a rebuild) can race and lose updates. Note: torn reads are already safe (a partial write is caught by `_read_cache_file`'s JSON-parse-error handling, triggering a clean rebuild) — the residual risk is specifically full read-modify-write races losing an update, not corruption.

**Impact**
Concurrent annotation of different sections of the same large document's index can silently drop one side's summary with no error or warning.

**Suggested correction**
Write to a sibling temp file and `os.replace()` it onto the cache path for atomicity; consider a lock file around the annotate read-modify-write cycle specifically.

**How to verify**
Add a regression test that runs two `annotate_section_summary` calls for different `line_start` values against the same index (simulating concurrency), and assert both summaries are present in the final saved index.

skills/studio/scripts/studio/commands/validate_toc.py
```diff
@@ -63,6 +69,7 @@ def cmd_validate_toc(argv: List[str]) -> int:
             content,
             artifact_path=filepath,
             max_heading_level=args.max_level,
+            max_section_lines=args.max_section_lines,
```
ainetx: ### cmd_validate_toc aborts the entire multi-file validation batch on any unhandled per-file read error, discarding already-collected results

**Severity:** Minor-to-moderate

**Problem**
In `skills/studio/scripts/studio/commands/validate_toc.py::cmd_validate_toc`, the per-file loop only guards the "file not found" case (`if not filepath.is_file()`). The actual read, `content = filepath.read_text(encoding="utf-8")`, has no exception handling. Any other read failure on any file in the batch — permission denied, non-UTF-8/binary content, or a TOCTOU race — raises uncaught out of the loop, aborting the command before it ever reaches `ui.result(...)`. This discards results already accumulated for every file validated earlier in the same invocation. Separately, `FILE_READ_ERROR`/`FILE_LOAD_ERROR` codes already exist in `utils/error_codes.py` but are unused here (`error_codes` isn't even imported in `validate_toc.py`).

**How to reproduce**
```python
# files = [good.md, bad.md, good2.md]; bad.md has invalid UTF-8 bytes or revoked permissions
cmd_validate_toc([str(good), str(bad), str(good2)])
# raises uncaught after good.md was already validated and appended to `results` -- that
# result is lost, and good2.md is never reached
```

**Expected behavior**
A read failure on one file in a `cfs validate-toc a.md b.md c.md` batch should be recorded as a per-file `ERROR` result (using `FILE_READ_ERROR`/`FILE_LOAD_ERROR`), consistent with the existing "File not found" handling, and the batch should continue validating the remaining files.

**Actual behavior**
An unhandled exception propagates out of the loop, the command crashes with a raw traceback, and results already computed for earlier files are never emitted.

**Impact**
In CI or batch-validation contexts, a single bad file (encoding issue, permission glitch, or a file removed mid-run) silently destroys the whole run's output rather than surfacing that one file's problem alongside the rest.

**Suggested correction**
Wrap the `filepath.read_text(...)` call in a `try/except (OSError, UnicodeDecodeError)` inside the loop, append an `ERROR` entry to `results` (using `error_codes.FILE_READ_ERROR`/`FILE_LOAD_ERROR`) analogous to the existing "File not found" branch, increment `total_errors`, and `continue` to the next file.

**How to verify**
Add a regression test with a batch of 2+ files where one has invalid UTF-8; assert the command still emits results for the other files, marks the bad file as an `ERROR` with a stable error code, and does not raise.

skills/studio/scripts/studio/commands/doc_index.py
```diff
@@ -0,0 +1,75 @@
...
+from ..utils.doc_index import get_or_build_doc_index
+from ..utils.ui import ui
+
+
+def cmd_doc_index(argv: List[str]) -> int:
+    """Build (or reuse the cached) structural index for a Markdown file."""
+    p = argparse.ArgumentParser(
+        prog="cfs doc-index",
+        description=(
+            "Build or reuse a cached heading/section index for a Markdown file, "
```
ainetx, line 25: ### cfs doc-index --help never explains how section_level is inferred

**Severity:** Minor

**Problem**
`cfs doc-index` picks `section_level` via a frequency-based heuristic in `infer_section_level()`: the heading level used most often wins, levels appearing exactly once are excluded as candidates, and ties (or an all-singletons document) fall back to the shallowest level. This choice directly determines which lines get grouped into `retrieval_sections` in both JSON and human-readable output. None of this is mentioned anywhere a CLI user would see it.

**How to reproduce**
Run `cfs doc-index --help` (or read the `argparse.ArgumentParser` description in `commands/doc_index.py`). The description reads only: "Build or reuse a cached heading/section index for a Markdown file, so navigation reads the file's structure once, not once per query." No mention of `section_level`, how it's chosen, or that it can differ from the "obvious" chapter level in irregularly-leveled documents (the exact PDF-conversion scenario `infer_section_level()`'s own docstring cites as its motivating case).

**Expected behavior**
A user seeing `"section_level": 5` (say) instead of the level they expected should have a way, via `--help` or the command's own output, to understand that the level was inferred from heading-frequency, not simply "H1/H2 is always the section level."

**Actual behavior**
`--help` gives no indication that section-level selection is heuristic at all.

**Impact**
Minor UX/discoverability gap — doesn't cause incorrect behavior, but makes an already-nontrivial, silently-applied heuristic harder to audit or trust from the CLI alone.

**Suggested correction**
Extend the `argparse.ArgumentParser(description=...)` in `cmd_doc_index()` (or add a short note to `_human_doc_index()`'s output) with one sentence, e.g.: "`section_level` is inferred from the most frequently repeated heading level (ties prefer the shallower level); a level used only once is never chosen."

**How to verify**
Run `cfs doc-index --help` after the fix and confirm the heuristic is described.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 8b80cd14: added an explicit schema-evolution contract to the module docstring -- a new top-level key is always additive (an older build simply never wrote it, which the required-field check already treats as 'predates the current schema' and rebuilds); `_SCHEMA_VERSION` (already added on PR #108, inherited here via rebase) is reserved for the other kind of change: an existing key's meaning/shape changing incompatibly, which a field-presence check alone can't catch.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 8b80cd14: `cmd_doc_index` now catches `UnicodeDecodeError` around `get_or_build_doc_index(...)` and reports a clean ERROR result (with a diagnostic log line, matching this codebase's user-facing-error-with-logging convention), instead of crashing with a raw traceback. `_compute_fresh_retrieval_sections`'s existing OSError handler now also catches `UnicodeDecodeError` (the same latent gap existed there too). Regression test: `test_non_utf8_file_reports_a_clean_error_not_a_raw_traceback`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 8b80cd14: each line is now `.rstrip()`'d before being joined for hashing in `_build_retrieval_sections`, so a "trim trailing whitespace on save" edit no longer changes the hash. Regression test: `test_trailing_whitespace_only_edit_does_not_change_the_hash`.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Addressed via documentation (same fix already applied on PR #108 for the identical underlying concern, inherited here via rebase): the etag docstring now explicitly states this as a known, accepted limitation on coarse-mtime filesystems, and explains why closing it fully (a content hash) would defeat this module's whole reason to exist (avoiding a read on cache hit).

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 8b80cd14, with one deliberate exception: promoted 3 of the 4 sites to `logger.warning` (cache-directory lookup failure, corrupt cache file, source file vanishing mid-etag-check -- all genuine anomalies with no legitimate 'expected' trigger). Left `_compute_fresh_retrieval_sections`'s debug log as-is: its own docstring documents 'the file was deleted after it was cached' as a legitimate, expected trigger (diffing against an arbitrarily-old cached build during batch reconciliation, unlike load_doc_index's freshly-validated-then-immediately-reused pattern), so promoting it would make a normal outcome noisy rather than making a real anomaly visible -- explained inline in a comment at that site. Regression tests assert WARNING-level records for the 3 promoted sites.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 8b80cd14: added `_with_cache_lock()`, an exclusive-`fcntl`-lock wrapper mirroring `decision_log.py`'s existing `_append_locked` pattern exactly (same POSIX-lock-else-unlocked-fallback shape), and `annotate_section_summary`'s read-modify-write cycle now runs under it -- serializing two concurrent calls annotating different sections of the same document so neither silently drops the other's update. (Atomicity of each individual write was already fixed on PR #108 via temp-file + os.replace, inherited here; this addresses the separate lost-update race you correctly identified as the residual risk.) Regression test: a real threading-based test (`test_concurrent_annotations_of_different_sections_do_not_lose_either_update`) that injects a delay inside the locked section to force a genuine overlap window -- verified it actually fails without the lock before confirming it passes with it.

skills/studio/scripts/studio/commands/validate_toc.py
tkcoding: Fixed in 8b80cd14: extracted `_validate_one_file()`, which now wraps the file read in `try/except (OSError, UnicodeDecodeError)` and returns an ERROR result (using `FILE_READ_ERROR`/`FILE_LOAD_ERROR` from `error_codes.py`, now actually imported and used) instead of raising -- one bad file in a batch no longer discards results already collected for files validated earlier, and the batch continues to the remaining files. (The extraction also fixed a pylint too-many-locals this change would otherwise have tripped.) Regression test: `test_a_read_failure_on_one_file_does_not_abort_the_batch` (good.md, then a binary file, then good2.md -- asserts all three get their own correct per-file result).

skills/studio/scripts/studio/commands/doc_index.py
tkcoding: Fixed in 8b80cd14: extended `cmd_doc_index`'s `argparse` description with a one-sentence explanation of the heuristic (most-frequent heading level wins; a level used only once is never chosen). Regression test: `test_help_explains_how_section_level_is_inferred`.

ainetx (approved):
Approving after a full Deep Review pass (56 checks across 7 categories: CLI conventions, I/O/schema compatibility, core algorithm correctness, hashing/staleness/performance, error handling/concurrency, traceability, and test quality).

All 9 findings raised during review (docs staleness, binary-input handling, hash normalization, etag collision risk, silent error logging, batch-abort behavior, cache write atomicity, missing `--help` docs, and the human-output warning gap) have been independently verified as fixed in the current code at this head — not just from commit messages. CI is fully green and the branch is clean to merge.

Nice work resolving everything across the review rounds.

PR #110, opened 2026-08-31
feat(tfidf,okf): add TF-IDF scoring and a local OKF bundle

tkcoding:
## Depends on #108 and #109 — diff will shrink once those merge

This branches from `jit-retrieval-cascade` (on top of #109, which is on top of #108) — `doc_index.py`'s `retrieval_sections` don't exist on `main` yet, so until #108/#109 merge this diff includes their changes too. Once they land, I'll rebase this branch onto the new `main`; the diff here will shrink to just what this PR adds.

## Summary

Two independently-testable JIT-retrieval mechanisms, both built on top of `doc_index.py`'s `retrieval_sections` (#109) rather than re-deriving section boundaries themselves.

### `tfidf.py` — TF-IDF scoring

Purely mechanical, no LLM call. Scores each retrieval section as sum(term-frequency x inverse-document-frequency) over a query's terms, and returns a margin/unambiguous confidence signal alongside the ranking, not just the ranking alone — a routing layer built on top of this needs to know when the ranking itself isn't trustworthy, not just what it is.

Verified against the real PDF-converted document referenced throughout this feature's design: the "KAPING" query is unambiguous (0.0016 vs. 0.0000 everywhere else); the "zero-shot" query reproduces the documented real failure exactly (margin 1.06x, wrong section on top, since term frequency is normalized by section length and the real answer lives in a longer section than the one that wins).

### `okf.py` — a local OKF bundle

Deterministic cache/storage infrastructure, matching `doc_index.py`'s own contract of containing no LLM-generated content — writing an actual summary is an external caller's job (an agent, dispatched outside this codebase), the same role as `doc_index.annotate_section_summary` one layer up.

Tracks which concept files should exist against a document's *current* retrieval sections, detects staleness via the section hash recorded when a concept file was written (not a separate cache mechanism — reuses the per-section hashing from #109), and regenerates `index.md` deterministically from the manifest.

**The whole bundle lives under `.cache/okf/` and is gitignored.** Unlike the content of a summary (expensive, real LLM tokens), the bundle not surviving a fresh clone just means it rebuilds the same way `doc_index.py`'s own cache does — nothing here assumes the bundle survives across clones, only across calls on the same machine.

### New CLI commands

- `cfs tfidf-score <file> <query>` — rank a file's retrieval sections against a query
- `cfs okf-status <file>` — report missing/stale/current per section

## Test plan

[...]

PR #111, opened 2026-08-31
feat(cascade): heading-nav, two-tier retrieval routing, and large-read gate

tkcoding:
## Summary

Stacked on #108, #109, and #110 (still open) — combines the mechanical ingredients those PRs built (`doc_index.py`'s structural index, `tfidf.py`'s scoring, `okf.py`'s bundle status) into a callable two-tier JIT-retrieval routing decision, and closes the two remaining gaps from the design document (see constructorfabric/studio#104): the missing heading-nav mechanism Tier 1 depends on, and the token-tracking/large-read confirmation gate, which is now actually wired into the cascade's baseline fallback path instead of sitting as an unconnected prototype.

- `utils/heading_nav.py` + `commands/heading_nav.py` (`cfs heading-nav`) — a case-insensitive literal-substring search of a query against each retrieval section's own text, mirroring `grep -i` — the "grep the query's words, then read the enclosing section" mechanism the design's Tier 1 table needs but nothing in the codebase implemented yet.
- `utils/cascade.py` + `commands/cascade.py` (`cfs retrieve`) — `route_query()` runs Tier 1 (heading-nav + TF-IDF agreement/margin), and only escalates to Tier 2 (OKF vs. baseline) when Tier 1 can't resolve confidently. Verified end-to-end against the real 166-page source PDF this design was built against, reproducing the three documented real scenarios exactly: KAPING resolves at Tier 1 (unambiguous), "making up" escalates (heading-nav zero hits), "zero-shot" escalates (diffuse margin, not unambiguous).
- `utils/read_gate.py` + `commands/read_gate.py` (`cfs read-gate`) — the large-read confirmation threshold check (default 5,000 lines, the real number measured designing this), now wired into `route_query()`'s baseline fallback so a Tier 2 baseline recommendation always reports whether it crosses the confirmation threshold.
- `decision_log.py` gets a new `"read"` event, `record_read()`, and `summarize_reads()` — activating its existing (previously uncalled) `summarize()`/`read_events()` read API — plus `commands/usage_report.py` (`cfs usage-report`) for the per-method token table the design's token-tracking prototype produced.
- `doc_index.py` gains a small shared `section_text()` helper (promoted out of `tfidf.py`'s private copy) so `heading_nav.py` doesn't duplicate the section-slicing logic.

## Two design decisions this PR resolves

**Tier 1's margin threshold**: the design's two real data points (an infinite/unambiguous margin on a correct pick, and a 1.06x-1.58x margin on two independently wrong picks) don't support a specific numeric cutoff. `route_tier1()` requires `unambiguous=True` for a large-margin Tier 1 resolution by default; a `margin_threshold` parameter exists to opt into a numeric cutoff later, once there's real evidence for one.

**Staleness policy (block vs. serve-stale-and-rebuild-async)**: this codebase has no background job runner and no LLM-calling capability anywhere (by design — see `okf.py`'s own docstring). `route_tier2()` never recommends an OKF concept file it knows is stale/missing for Tier 1's candidate section; it falls back to baseline with `okf_needs_rebuild: true` instead, since serving a known-stale pointer with no mechanism to ever correct it would be a silent wrong answer.

A real bug was caught and fixed during manual verification before writing tests: `get_okf_status()` returns one entry per retrieval section even when nothing has ever been summarized (all `status: "missing"`), so an early `bundle_exists` check based on "bundle_dir resolves and entries is non-empty" was always true inside a Studio project — fixed to require at least one entry with `status != "missing"`.

## Test plan

- [x] `python3 -m pytest tests/ -q` — full suite passes (only the same 12 pre-existing, unrelated failures present on `main` before this branch, confirmed via `git stash` diff)
- [x] 100% coverage on every new/changed file (`cascade.py`, `heading_nav.py`, `read_gate.py`, their command wrappers, `decision_log.py`'s additions, `doc_index.py`, `tfidf.py`)
- [x] `pylint` clean on all new/changed files
- [x] `make vulture-ci` clean (one expected whitelist addition: `record_read`, an external-caller hook with no in-repo caller yet, same pattern as `write_concept_file`)
- [x] `cfs validate` — 0 errors
- [x] `cfs spec-coverage` — granularity 0.4638, above the 0.46 floor
- [x] Live end-to-end verification against the real 166-page PDF (freshly reconverted via `pymupdf4llm`) for all three documented scenarios (KAPING, "making up", "zero-shot")

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added seven Studio utility commands for document indexing, TF-IDF scoring, heading navigation, retrieval, read confirmation, OKF status, and usage reporting.
  * Added cached document indexing, section search, retrieval scoring, cascade routing, and regenerable concept bundles.
  * Added large-read confirmation checks and read-usage summaries.
* **Bug Fixes**
  * Improved file validation, parsing errors, read-error reporting, TOC validation, and frontmatter handling.
  * Added safer atomic file updates and locking.
* **Tests**
  * Added comprehensive coverage for the new commands, retrieval workflows, caching, validation, and error handling.


skills/studio/scripts/studio/commands/usage_report.py
```diff
@@ -0,0 +1,58 @@
...
+        "summary": decision_log.summarize(),
+        "reads": decision_log.summarize_reads(),
+    }
+    ui.result(output, human_fn=_human_usage_report)
+    return 0
+# @cpt-end:cpt-studio-algo-core-infra-decision-log:p1:inst-usage-report-cmd
+
+
+# @cpt-begin:cpt-studio-algo-core-infra-decision-log:p1:inst-usage-report-cmd-format
+def _human_usage_report(data: dict) -> None:
```
ainetx, line 36: ### Human-readable usage-report and okf-status output silently drop several fields present in JSON

**Severity:** Major

**Problem**
The human-readable renderers for two new commands render only a subset of the data their own JSON output includes, so a person using the default human mode sees strictly less information than someone parsing `--json`. In `commands/usage_report.py`, `_human_usage_report` prints `summary['path']`, `summary['exists']`, `summary['total_events']`, and `summary['runs']`, but never touches `summary['event_counts']`, `summary['first_ts']`, or `summary['last_ts']` — all of which `utils/decision_log.py`'s `summarize()` computes and returns. Symmetrically, in `commands/okf.py`, `_human_okf_status` prints each entry's `status`, `line_start`, `line_end`, and `heading`, but never `entry['concept_file']`, even though `utils/okf.py`'s `get_okf_status()` includes it in every entry.

**How to reproduce**
1. In a Studio-adapted project with a non-empty decision log, run `cfs usage-report` (human mode, the default) and separately `cfs usage-report --json`.
2. Compare the two outputs' field coverage.
3. Similarly run `cfs okf-status <file>` and `cfs okf-status <file> --json` and compare per-entry fields.

**Expected behavior**
The human-readable view should surface the same substantive information as the JSON view (formatted for readability), or at minimum not silently omit fields a user would need to answer basic questions like "what event types were logged?", "over what time range?", or "which concept file does this section map to?".

**Actual behavior**
`event_counts`, `first_ts`, and `last_ts` never appear in human `usage-report` output; `concept_file` never appears in human `okf-status` output. These fields exist only in the JSON path.

```text
cfs usage-report            cfs usage-report --json
  |                            |
  v                            v
_human_usage_report()      output dict
  prints: path,                { path, exists, total_events, runs,
  exists, total_events,          event_counts, first_ts, last_ts }
  runs                            ^
  (missing: event_counts,         |
   first_ts, last_ts) -----------/  <- silently dropped in human mode
```

**Impact**
Interactive users (the default, non-scripted use case) cannot see event-type breakdowns or log time range from `usage-report`, and cannot see which on-disk concept file corresponds to a given section from `okf-status` — forcing them to re-run with `--json` and parse manually, which defeats the purpose of a human-formatted mode.

**Suggested correction**
Add a line (or a few) to each human renderer: e.g. in `_human_usage_report`, print event-type counts and the `first_ts`/`last_ts` range next to the existing summary line; in `_human_okf_status`, include `entry['concept_file']` alongside `heading`/`status`/line range in the per-entry line.

**How to verify**
Add/extend unit tests asserting the human-mode captured output (via `capsys`) contains the event-count breakdown and timestamp range for `usage-report`, and contains the concept filename for each `okf-status` entry.

tests/test_okf.py
```diff
@@ -0,0 +1,257 @@
+"""Tests for the local, regenerable OKF bundle (okf.py).
```
ainetx, line 1: ### No test exercises the argparse required-argument-omitted path for the seven new commands

**Severity:** Major

**Problem**
Across `tests/test_cascade.py`, `tests/test_doc_index.py`, `tests/test_heading_nav.py`, `tests/test_okf.py`, `tests/test_read_gate.py`, and `tests/test_tfidf.py`, every "missing" test exercises the case where a file path *is* supplied but doesn't exist on disk (asserting `rc == 2` via `ui.require_existing_file`). None of these files contain a `pytest.raises(SystemExit)`, meaning the distinct code path where a required positional CLI argument is omitted entirely — which argparse handles by printing usage and calling `sys.exit(2)` before the command's own logic ever runs — is untested for any of the seven new commands. Other, pre-existing test files in this same repo (e.g. `tests/test_cli_integration.py`) do cover this pattern, showing it's an established convention this PR's new tests don't follow.

**How to reproduce**
1. Run `cmd_okf_status([])` (or the CLI equivalent with no file argument) in a test.
2. Observe argparse's `SystemExit(2)` fires before `ui.require_existing_file` is ever reached — a different failure mode than the "path given, file missing" case already tested.
3. Search the six new-command test files for `SystemExit`/`pytest.raises` — none appear.

**Expected behavior**
Each new command's test suite should include at least one test asserting `pytest.raises(SystemExit)` when the required positional argument is omitted, distinguishing "no argument" from "bad/missing-file argument."

**Actual behavior**
Only the file-not-found path is tested; the missing-argument path has zero coverage for any of the 7 new commands.

```text
cmd_okf_status([])
  |
  v
argparse.parse_args([])  -- required positional missing
  |
  v
SystemExit(2)  <-- never reached by any existing test
  (existing tests instead pass a nonexistent-file argv,
   reaching ui.require_existing_file's own rc==2 path)
```

**Impact**
A future change to any of these commands' `argparse.ArgumentParser` setup (e.g. accidentally making the argument optional, or changing its name) could silently break required-argument enforcement without any test catching it.

**Suggested correction**
Add one small test per command file, e.g.:
```python
def test_missing_required_argument(self):
    with pytest.raises(SystemExit):
        cmd_okf_status([])
```
mirroring the pattern already used in `tests/test_cli_integration.py`.

**How to verify**
Run the new tests and confirm they fail if the argparse argument is temporarily made optional (`nargs="?"`), and pass against current code.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -0,0 +1,198 @@
...
+            rec = _baseline_recommendation(expected_future_queries)
+            rec["okf_needs_rebuild"] = True
+            return rec
+
+    return {"recommendation": "okf", "reason": "okf_bundle_current", "bundle_dir": status["bundle_dir"]}
+# @cpt-end:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-tier2
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-route
+def route_query(
```
ainetx, line 191: ### Tier 2's returned schema has no stability guarantee documented, despite live internal consumers

**Severity:** Minor

**Problem**
`route_query()` in `skills/studio/scripts/studio/utils/cascade.py` builds its return dict ad hoc — `{"query": query, **tier1}`, then conditionally merges in `tier2` and `read_gate` — and its docstring only describes control flow, with no note that the field names or their nesting are a stable contract. Two real consumers already key directly into these fields by name: `commands/cascade.py`'s `_human_retrieve` reads `data['tier']`, `data['reason']`, `data['candidates']`, `data['tier2']['recommendation']`, `data['tier2']['reason']`, `data['tier2'].get('okf_needs_rebuild')`, and `data['read_gate']['needs_confirmation']`; `tests/test_cascade.py` asserts on `result["tier"]`, `result["tier2"]["recommendation"]`, and `result["read_gate"]["needs_confirmation"]`/`["total_lines"]`. Because the schema is undocumented as a contract, a future change to `route_tier1`/`route_tier2`'s returned keys could silently break these consumers without any doc to flag the break.

**How to reproduce**
N/A — this is a documentation gap, not a runtime failure.

**Expected behavior**
The module or function docstring should state explicitly that the returned dict's shape (top-level `tier`/`reason`/`candidates`, plus conditional `tier2` and `read_gate` sub-dicts with their own named fields) is a stable, versioned contract relied on by CLI formatting and tests, so future edits know to preserve or intentionally version-bump it.

**Actual behavior**
The docstring only narrates behavior with no mention of schema stability, versioning, or the existing internal consumers that depend on specific field names.

**Impact**
Low risk of an accidental breaking change to the internal CLI formatter and test suite when the cascade's return shape is modified, since nothing documents that these fields are load-bearing outside `cascade.py` itself. Purely a maintainability/documentation gap — no current functional defect.

**Suggested correction**
Add a short note to `route_query`'s docstring (or a small schema block/comment) listing the guaranteed top-level keys and their consumers.

**How to verify**
Confirm the docstring update lists the exact keys used by `commands/cascade.py::_human_retrieve` and the assertions in `tests/test_cascade.py`, and that no behavior change is required (this is a comment-only fix).

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -0,0 +1,252 @@
...
+    if bundle_dir is None:
+        return False
+    bundle_dir.mkdir(parents=True, exist_ok=True)
+    (bundle_dir / _MANIFEST_NAME).write_text(json.dumps(manifest, indent=2), encoding="utf-8")
+    return True
+# @cpt-end:cpt-studio-algo-traceability-validation-okf:p1:inst-okf-manifest-io
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-okf:p1:inst-okf-status
+def get_okf_status(path: Path) -> Dict[str, Any]:
```
ainetx, line 200: ### OKF status reports "current" for a corrupted concept file it never reads

**Severity:** Major

**Problem**
`get_okf_status()` in `skills/studio/scripts/studio/utils/okf.py` determines each section's status purely from manifest metadata and a physical-presence check (`(bundle_dir / concept_file).is_file()`). It never opens or validates the concept file's actual content. When a manifest entry exists and its `built_from_hash` matches the section's current hash, the status is set to `"current"` without any read of the file the status vouches for.

**How to reproduce**
1. Run the OKF pipeline against a document so a section gets a manifest entry and a written concept file, with `built_from_hash` matching the section's current hash.
2. Truncate, corrupt, or empty the concept file's content on disk (leave the manifest untouched).
3. Call `get_okf_status(path)`.

**Expected behavior**
A concept file whose content is unreadable, truncated, or corrupted should not be reported as `"current"` — it should be flagged as needing regeneration (e.g. `"missing"` or a new `"corrupt"` status).

**Actual behavior**
The entry is reported `"current"`, identical to a genuinely valid file, because only physical presence and manifest hash agreement are checked.

```text
manifest.built_from_hash == section.hash?  -- yes (untouched)
  |
  v
(bundle_dir/concept_file).is_file()?  -- yes (corrupted but present)
  |
  v
status = "current"  <-- content never read/validated
```

**Impact**
A caller trusting `"current"` status (e.g. the retrieval cascade in `cascade.py`) can serve a corrupted or empty concept file to a query as if it were a valid, trustworthy summary, producing a silently wrong answer with no detection mechanism.

**Suggested correction**
Add a minimal content-validity check before returning `"current"` — e.g. verify the file is non-empty and/or parses per its expected frontmatter/body structure, falling back to `"missing"` or a distinct `"corrupt"` status on failure.

**How to verify**
Add a test that writes a valid manifest entry with a matching hash but replaces the concept file's content with empty/garbage bytes, then asserts `get_okf_status` no longer reports `"current"` for that section.

skills/studio/scripts/studio/utils/heading_nav.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Dict
+
+from .doc_index import get_or_build_doc_index, section_text
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-heading-nav:p1:inst-heading-nav-search
+def find_sections(path: Path, query: str) -> Dict[str, Any]:
```
ainetx, line 44: ### Heading-nav substring search counts hits inside fenced code blocks as prose matches

**Severity:** Minor

**Problem**
`find_sections`/`section_text` in `skills/studio/scripts/studio/utils/heading_nav.py` count query occurrences via a plain case-insensitive substring count over each section's raw text. There is no fenced-code-block or inline-code-span awareness, even though `utils/toc.py` already implements fence-tracking (`_fence_update`) elsewhere in the same codebase.

**How to reproduce**
1. Create a Markdown document with a retrieval section whose only occurrence of a query term is inside a fenced code block (e.g. a sample command or variable name), not in prose.
2. Call `find_sections(path, query)` with that query term.

**Expected behavior**
A term appearing only inside a code sample arguably should not count identically to a genuine prose hit, or at minimum this distinction should be a documented, tested design choice.

**Actual behavior**
The code-block occurrence is counted the same as a prose hit, with no fence-awareness and no test coverage for this case in `tests/test_heading_nav.py`.

**Impact**
Low — this can inflate `hit_count` and, in edge cases, cause a section to be selected by Tier 1 routing based solely on an incidental code-sample match rather than genuine prose relevance.

**Suggested correction**
Reuse the existing `_fence_update` fence-tracking from `utils/toc.py` to exclude fenced code blocks (and optionally inline code spans) from the substring count, or explicitly document this as an accepted limitation if intentional.

**How to verify**
Add a test with a section containing the query term only inside a fenced code block and assert the expected hit-count/match behavior once a decision is made.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+    """
+    from .files import find_studio_directory
+
+    try:
+        studio_dir = find_studio_directory(path.resolve().parent)
+    except OSError as exc:
+        # A file whose parent can't be stat'd (permissions, a race) is not a
+        # reason to fail the caller -- just an uncached build, like "no
+        # Studio directory found".
+        logger.debug("doc-index cache path lookup skipped for %s: %s", path, exc)
```
ainetx: ### Cache/manifest read failures are logged only at an unreachable debug level

**Severity:** Minor

**Problem**
Six new except blocks (`utils/doc_index.py` lines 76, 266, 315, 370; `utils/okf.py` lines 63, 100) catch `OSError`/`json.JSONDecodeError`, log via `logger.debug(...)`, and silently fall back (return `None` / disable caching). The CLI's own logging setup pins the `studio` logger at `WARNING`, and there is no `--debug`/`--verbose` flag anywhere in the CLI to lower that threshold.

**How to reproduce**
1. Inspect `cli.py`'s logging configuration, which sets the studio logger to `WARNING` unconditionally.
2. Search the CLI for any flag that adjusts log level — none exists.
3. Trigger one of the six except paths (e.g. make the doc-index cache file unreadable/corrupt) and observe no diagnostic output at any log level a user can enable.

**Expected behavior**
A real, unexpected failure (permission error, disk corruption, race condition) reading or writing the cache should be visible to a user who wants to diagnose why caching silently stopped working.

**Actual behavior**
The failure is logged at `DEBUG`, a level that is structurally unreachable through the CLI's own logging configuration, so it is fully invisible in practice.

**Impact**
Silent cache degradation (e.g., permission issues, unexpected corruption) is undiagnosable without editing source code or attaching a custom log handler.

**Suggested correction**
Either expose a `--debug`/`--verbose` CLI flag that lowers the logger threshold, or log unexpected fallback conditions at `WARNING` (reserving `DEBUG` only for fully-expected states like "no Studio directory found").

**How to verify**
After the fix, trigger one of the six fallback conditions and confirm the resulting message is visible through the CLI's normal (non-flag) output or via a documented verbosity flag.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+    describes the content just read. If they don't, retry. After
+    ``_MAX_READ_ATTEMPTS`` under sustained contention, return the last read
+    anyway, stamped with its own trailing etag -- the safe direction to
+    fail in, since a file still being rewritten that fast will simply look
+    stale again on the very next check, never silently wrong.
+    """
+    etag_after = _compute_etag(path)
+    for _ in range(_MAX_READ_ATTEMPTS):
+        etag_before = etag_after
+        content = path.read_text(encoding="utf-8")
```
ainetx, line 260: ### Unhandled decode error crashes every file-consuming CLI command on binary or non-UTF-8 input

**Severity:** Major

**Problem**
`_read_with_stable_etag()` in `utils/doc_index.py` calls `path.read_text(encoding="utf-8")` with no exception handling. Every command that consumes a file (`doc-index`, `heading-nav`, `tfidf-score`, `retrieve`, `read-gate`, and `okf-status` once inside a Studio project) routes through this function via `get_or_build_doc_index`. The CLI's top-level dispatcher only catches `SystemExit`, so nothing intercepts a `UnicodeDecodeError`.

**How to reproduce**
1. Create a file with invalid UTF-8 bytes, e.g. `python3 -c "open('bad.md','wb').write(b'\xff\xfe\x00Hello world')"`.
2. Run `doc-index` (or `heading-nav`, `tfidf-score`, `retrieve`, `read-gate`) with that file.
3. Observe the process crash with a raw Python traceback and exit code 1.

**Expected behavior**
The CLI should catch the decode failure and print a clean, structured error (consistent with its other file-validation error paths), exiting with a controlled non-zero code.

**Actual behavior**
An unhandled `UnicodeDecodeError` traceback is printed straight to stderr, exposing internal file paths and stack frames, for any binary or non-UTF-8 file passed to any of these commands. Reproduced live against all five directly-testable commands.

**Impact**
Any binary file, PDF, image, or non-UTF-8-encoded document handed to these commands crashes the tool ungracefully instead of failing cleanly.

**Suggested correction**
Wrap the `read_text` call in `_read_with_stable_etag` (or at the CLI dispatch boundary) with a `try/except UnicodeDecodeError` that reports a standard file-validation error, matching the existing pattern for missing files.

**How to verify**
Re-run the reproduction steps above against each of the six commands after the fix; each should print a structured error message and exit non-zero without a raw traceback.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+        return None
+    if not _has_schema_current_sections(cached):
+        logger.debug("doc-index cache for %s predates the current schema; rebuilding", path)
+        return None
+    return cached
+# @cpt-end:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-load
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-doc-index:p1:inst-doc-index-save
+def save_doc_index(path: Path, index: Dict[str, Any]) -> None:
```
ainetx: ### Persistence functions' failure signal is discarded, so callers report false success

**Severity:** Minor

**Problem**
`save_doc_index()` has no return value (declared to return `None`) and no-ops silently when a file is outside a Studio project. It's called from `get_or_build_doc_index` and `annotate_section_summary`, with no signal available to detect the no-op. `annotate_section_summary` then unconditionally returns `True`. Similarly, `save_okf_manifest()` returns `bool` (`False` on no-op), but its only call site, `write_concept_file()`, discards the result and unconditionally returns `True`.

**How to reproduce**
1. Call `annotate_section_summary(path, line_start, summary)` for a file located outside any Studio-adapted project (so the underlying cache/bundle path resolves to `None`).
2. Observe the function returns `True` even though `save_doc_index` never wrote anything to disk.
3. Similarly call `write_concept_file(...)` outside a Studio project and observe it returns `True` even though `save_okf_manifest` returned `False` and no manifest was persisted.

**Expected behavior**
A caller should be able to tell, from the return value, whether the summary/manifest was actually persisted.

**Actual behavior**
Both functions report unconditional success regardless of whether the underlying save actually wrote anything.

**Impact**
An external caller (e.g., an LLM-driven summarization pass) believes a summary or concept file was saved when it silently wasn't, leading to repeated wasted work (re-summarizing the same section every run) with no error surfaced.

**Suggested correction**
Have `save_doc_index` return a `bool` matching `save_okf_manifest`'s convention, and have both `annotate_section_summary` and `write_concept_file` propagate that result instead of returning `True` unconditionally.

**How to verify**
Call both functions against a file outside a Studio-adapted project and confirm they now return `False`/report failure instead of `True`.

skills/studio/scripts/studio/commands/cascade.py
```diff
@@ -0,0 +1,69 @@
...
+# @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-cmd
+def cmd_retrieve(argv: List[str]) -> int:
+    """Route a query against a Markdown file through the JIT-retrieval cascade."""
+    p = argparse.ArgumentParser(
+        prog="cfs retrieve",
+        description="Route a query through the two-tier JIT-retrieval cascade and report the decision.",
+    )
+    p.add_argument("file", help="Markdown file path")
+    p.add_argument("query", help="Query text")
+    p.add_argument(
```
ainetx, line 50: ### `--margin-threshold` accepts values that defeat the cascade's own documented safety margin

**Severity:** Major

**Problem**
`commands/cascade.py`'s `--margin-threshold` argument uses `type=float` with no validator, so any float — negative, zero, or non-finite — reaches `utils/cascade.py`'s `route_tier1()`. The module's own docstring states the design basis explicitly: real measured data showed only an *infinite* margin was safe on a correct pick, while finite margins of 1.06x-1.58x still occurred on two independently wrong picks — "unambiguous is safe, anything finite isn't yet proven safe." A negative or near-zero `--margin-threshold` makes `tfidf_result["margin"] >= margin_threshold` true for virtually any margin, so `agree_large_margin` fires and Tier 1 returns a confident `"resolved"` instead of escalating — exactly the case the module's design notes say is unsafe. (A `nan` threshold does not have this effect: NaN comparisons are always false, so `margin >= nan` never fires and Tier 1 simply falls back to the conservative `unambiguous`-only path — that is a fail-safe no-op, not a defeat of the escalation logic.) The codebase already has an established pattern for exactly this hazard: `commands/eval.py`'s `_compliance_arg` rejects any value where `not math.isfinite(parsed)` before accepting it as an argparse type. `cascade.py` does not reuse or mirror that pattern.

**How to reproduce**
1. Run `cfs cascade <file> <query> --margin-threshold -1`.
2. Construct or find a query where heading-nav and TF-IDF agree on a section but with only a small measured margin (not `unambiguous`).
3. Observe `route_tier1` returns `{"tier": "resolved", ...}` instead of `{"tier": "escalate", "reason": "diffuse_margin", ...}`.

**Expected behavior**
A negative, zero, or otherwise out-of-range `--margin-threshold` should be rejected at the CLI boundary, consistent with the project's own stated design intent that no finite margin has yet been proven safe.

**Actual behavior**
Any float, including negative values, is silently accepted and used directly in the safety-relevant comparison, letting an operator (accidentally or otherwise) force Tier 1 to resolve confidently on cases the module's own design notes classify as unsafe.

**Impact**
Silently weakens the one documented safety guard in the cascade's decision logic, with no error or warning at invocation time. Given this is optional/off-by-default, impact is limited to callers who explicitly pass the flag, but there is nothing stopping an unvalidated or accidentally-negative value from being passed.

**Suggested correction**
Add a small validator function for `--margin-threshold` (mirroring `_compliance_arg` in `commands/eval.py`) that requires `math.isfinite(value) and value > 0`, and use it as the argparse `type=`.

**How to verify**
Add a unit test asserting the CLI rejects `--margin-threshold -1`, `--margin-threshold 0` with a clear argparse error, and that a valid positive finite value still passes through to `route_tier1` unchanged.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -0,0 +1,198 @@
...
+    candidates = tier1_result.get("candidates", [])
+    if candidates:
+        by_line_start = {entry["line_start"]: entry for entry in status["entries"]}
+        relevant = [by_line_start[c["line_start"]] for c in candidates if c["line_start"] in by_line_start]
+        if any(entry["status"] != "current" for entry in relevant):
+            rec = _baseline_recommendation(expected_future_queries)
+            rec["okf_needs_rebuild"] = True
+            return rec
+
+    return {"recommendation": "okf", "reason": "okf_bundle_current", "bundle_dir": status["bundle_dir"]}
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: `route_tier2()` now checks *every* entry in the bundle when Tier 1 named no candidate (row 1), not just entries the tier1 candidates happened to reference -- if any section is stale/missing, it falls back to baseline with `okf_needs_rebuild: true` instead of recommending OKF for the external selector to potentially land on the untrustworthy one. Regression tests: `test_no_candidate_with_a_partially_summarized_bundle_falls_back_to_baseline`, `test_no_candidate_with_a_fully_current_bundle_recommends_okf`.

skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -440,4 +454,27 @@ def summarize(path: Optional[Path] = None) -> Dict[str, Any]:
...
+    """
+    methods: Dict[str, Dict[str, int]] = {}
+    total_tokens = 0
+    for obj in read_events(path, event="read"):
+        payload = obj.get("payload") or {}
+        method = str(payload.get("method", "?"))
+        entry = methods.setdefault(method, {"count": 0, "total_tokens": 0, "total_lines": 0})
+        entry["count"] += 1
+        entry["total_tokens"] += int(payload.get("tokens", 0) or 0)
+        entry["total_lines"] += int(payload.get("lines", 0) or 0)
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: `summarize_reads()` now skips a "read" record whose payload isn't a dict, or whose `tokens`/`lines` aren't numeric, instead of raising -- the same tolerance `read_events()` already gives an unparseable line, extended to a parseable line with a malformed payload shape. Regression tests: `test_summarize_reads_skips_a_non_dict_payload`, `test_summarize_reads_skips_non_numeric_tokens_or_lines`.

skills/studio/scripts/studio/utils/doc_index.py
```diff
@@ -0,0 +1,484 @@
...
+    """``True`` only if every field in ``_REQUIRED_INDEX_FIELDS`` is present
+    *and* every ``retrieval_sections`` entry carries a ``hash`` -- that field
+    was added after ``retrieval_sections`` itself, so a cache from that
+    intermediate schema passes a field-presence check on the container but
+    still raises ``KeyError`` the moment a consumer (e.g.
+    :func:`diff_stale_sections`) reads ``entry["hash"]``.
+    """
+    if any(field not in cached for field in _REQUIRED_INDEX_FIELDS):
+        return False
+    return all("hash" in section for section in cached["retrieval_sections"])
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: `_REQUIRED_INDEX_FIELDS` now also includes `sections`, so a matching-etag cache that omits it is rebuilt rather than returned and later KeyError-ing in `cmd_doc_index()`. Same class of gap as the per-section `hash` fix on PR #110. Regression test: `test_cache_missing_sections_field_is_rebuilt_not_returned`.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -0,0 +1,252 @@
...
+def load_okf_manifest(path: Path) -> Optional[Dict[str, Any]]:
+    """Load the OKF bundle manifest for ``path``, or ``None`` if absent/corrupt/unavailable."""
+    bundle_dir = _okf_bundle_dir(path)
+    if bundle_dir is None:
+        return None
+    manifest_path = bundle_dir / _MANIFEST_NAME
+    if not manifest_path.is_file():
+        return None
+    try:
+        return json.loads(manifest_path.read_text(encoding="utf-8"))
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: added `_has_valid_manifest_shape()`, checked by `load_okf_manifest()` before returning -- a manifest that isn't a dict, whose `entries` isn't a list, or whose entries are missing `position`/`heading`/`concept_file`/`built_from_hash` is now treated as absent (clean rebuild) rather than returned malformed. Regression tests: `test_returns_none_on_malformed_entry_shape`, `test_returns_none_when_entries_is_not_a_list`, `test_returns_none_when_manifest_top_level_is_not_a_dict`, `test_empty_entries_list_is_a_valid_manifest`.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -0,0 +1,252 @@
...
+
+    index = get_or_build_doc_index(path)
+    manifest = load_okf_manifest(path) or {"entries": []}
+    by_line_start = {entry["line_start"]: entry for entry in manifest.get("entries", [])}
+
+    entries = []
+    for position, section in enumerate(index["retrieval_sections"], start=1):
+        manifest_entry = by_line_start.get(section["line_start"])
+        concept_file = _concept_filename(position, section["heading"])
+        if manifest_entry is None or not (bundle_dir / concept_file).is_file():
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: manifest entries are now matched by the section's document *position* (1-based ordinal among retrieval sections) instead of `line_start`, and `get_okf_status()` uses the persisted `concept_file` from a matched entry instead of re-deriving one. Position only shifts on a genuine structural change (a section added/removed) -- exactly when losing identity is correct -- while line_start shifting from unrelated content resizing elsewhere no longer misreports an untouched section as missing. Regression test: `test_content_inserted_above_an_unchanged_section_keeps_it_current`.

skills/studio/scripts/studio/utils/tfidf.py
```diff
@@ -0,0 +1,135 @@
...
+            "score": score,
+        })
+    ranked.sort(key=lambda entry: entry["score"], reverse=True)
+    return ranked
+
+
+def _confidence(ranked: List[Dict[str, Any]]) -> tuple:
+    """See :func:`score_sections` for what ``margin``/``unambiguous`` mean."""
+    if len(ranked) < 2 or ranked[0]["score"] <= 0:
+        return None, False
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: `_confidence()` now returns `unambiguous=True` for a single ranked section with a positive score -- nothing else exists for it to be confused with, same as beating every rival's zero score. Regression tests: `test_single_section_with_positive_score_is_unambiguous` (updated from the previous, now-incorrect expectation), `test_single_section_with_zero_score_is_not_unambiguous` (confirms the fix doesn't overcorrect).

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,207 @@ def _record_missing_toc_error(
...
+                line=line,
+                heading_text=text,
+                section_length=section_length,
+                max_section_lines=max_section_lines,
+            ))
+    return warnings
+
+
+_DESCRIPTION_FIELD_RE = re.compile(r"^description\s*:\s*(.*)$")
+_BLOCK_SCALAR_RE = re.compile(r"^[|>][+\-]?\d*$")
```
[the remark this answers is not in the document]
tkcoding: Fixed in 7af88aca: `_BLOCK_SCALAR_RE` now accepts both chomping/indentation indicator orders and an optional trailing "# comment", matching YAML 1.2.2's actual grammar -- adopted your verified regex directly. Regression tests: `test_empty_block_scalar_with_trailing_comment_still_warns`, `test_empty_block_scalar_with_reversed_indicator_order_still_warns`.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -687,6 +724,207 @@ def _record_missing_toc_error(
...
+    ``_unique_slug``) and are NOT errors, but they make it impossible to
+    unambiguously address a section by its heading text alone.
+    """
+    from . import error_codes as EC
+    from .constraints import error
+
+    seen: Dict[str, int] = {}
+    warnings: List[Dict[str, Any]] = []
+    for _level, text, line in headings_with_lines:
+        if text in seen:
```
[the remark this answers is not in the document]
tkcoding: Investigated, not fixed -- this looks like a mix-up between two unrelated pieces of code. The description ("`find_sections` lowercases both query and section text") matches my new `heading_nav.find_sections()`, a mechanical retrieval search with no duplicate-detection concept at all. But the proposed diff patches `_check_duplicate_heading_titles()` in `toc.py` (line ~759), a pre-existing, unrelated CDSL structural-hygiene check (`toc-heading-duplicate`) that predates this PR entirely and has nothing to do with query matching. Case-folding that check's semantics would be a real behavior change to a different feature (whether "Intro" and "intro" count as duplicate document headings), with its own test surface and design tradeoffs, not a quick win alongside this PR's retrieval work. Happy to pick this up as its own follow-up if the maintainers want case-insensitive duplicate-heading detection, but leaving `toc.py`'s existing behavior untouched here.

vulture_whitelist.py
```diff
@@ -72,3 +87,9 @@
 record_escalation  # noqa: B018
 record_invocation  # noqa: B018
 summarize  # noqa: B018
+
+# record_read: called by a future external caller once a read-and-answer step
+# actually fires (an agent doing the real read), not yet reached from
+# production paths. Exercised by tests. See
+# skills/studio/scripts/studio/utils/decision_log.py.
+record_read  # noqa: B018
```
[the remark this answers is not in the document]
tkcoding: Investigated, not fixed as suggested -- this is a considered design choice, not an oversight, and I don't think it's fixable the way proposed. `cfs retrieve` (cascade.py) deliberately never performs the actual read: its own module docstring states it "never reads more of the document than heading-nav/TF-IDF/OKF status already needed" and that "the actual answering step... stays an external caller's job." A `record_read` event needs a real `tokens` count, which only exists once something actually loads the picked section(s) into an LLM's context -- `cfs retrieve` has no such number to log honestly (estimating one would violate this whole effort's own working principle of never presenting an estimate as a measured value, and would conflict with the existing `record_routing`/`record_escalation` events that already exist for logging a routing decision itself, not a read's cost). `record_read` is the same category as `write_concept_file`/`annotate_section_summary`: an external-caller hook for whoever actually performs the real read (an agent) to call with the real number. `cfs usage-report` staying empty until that caller exists is the correct, honest state, not a bug -- the alternative is fabricating a number this deterministic CLI can't actually know.

skills/studio/scripts/studio/commands/usage_report.py
tkcoding: Fixed in 5ffeb43b: `_human_usage_report` now prints the time range (`first_ts`/`last_ts`) and an event-type breakdown (`event_counts`) alongside the existing summary line. `_human_okf_status` already included `concept_file` (fixed in an earlier round on this same branch), so only the usage-report half needed this fix.

tests/test_okf.py
tkcoding: Fixed in 5ffeb43b: added missing-required-argument coverage for all six new commands. `retrieve`/`heading-nav`/`read-gate` use plain argparse, so each gets a `pytest.raises(SystemExit)` test as suggested; `doc-index` uses `JsonSafeArgumentParser` (converted in an earlier PR #110 round), so omitting its required argument does NOT raise SystemExit -- it already got the correct JSON-error-result test instead (mirroring the ones `okf-status`/`tfidf-score` already had). `usage-report` has no positional argument to omit, so it's out of scope for this finding.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 5ffeb43b: `route_query`'s docstring now states its returned shape (top-level `tier`/`reason`/`candidates`, conditional `tier2`/`read_gate` sub-dicts) is a stable contract, and names the two real consumers (`commands/cascade.py::_human_retrieve`, `tests/test_cascade.py`) that key into it by field name. Comment-only, no behavior change.

skills/studio/scripts/studio/utils/okf.py
tkcoding: Fixed in 5ffeb43b: added `_concept_file_is_valid()`, a minimal content check (the file must open with the YAML frontmatter block `_build_frontmatter` always writes) applied before reporting `current` -- a truncated/corrupted-but-present concept file now reports `missing` instead. Added a regression test that corrupts a written concept file's content in place and asserts the status falls back to missing.

skills/studio/scripts/studio/utils/heading_nav.py
tkcoding: Fixed in 5ffeb43b: added `_strip_fenced_code_lines()`, reusing `toc.py`'s own `_fence_update` fence-tracking, and apply it to a section's lines before counting query hits -- a term appearing only inside a fenced code block no longer counts as a prose hit. Added regression tests for a code-only hit (now zero matches) and a real prose hit alongside an unrelated code block (still counted).

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 5ffeb43b: promoted the remaining genuinely-anomalous debug-level fallback logs to warning in `doc_index.py` (the schema-mismatch-on-matching-etag case) and `okf.py` (the bundle-dir-lookup OSError and manifest-unreadable cases), matching the reasoning already applied to their sibling checks in earlier rounds on this branch. The one remaining debug call in `doc_index.py` (`_compute_fresh_retrieval_sections`) stays debug deliberately -- its own inline comment explains why (a documented, expected trigger: "file deleted after caching", distinct from the anomalous cases). No CLI verbosity flag was added; that's a larger, separate feature this fix doesn't attempt.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 5ffeb43b: `heading-nav`, `retrieve`, `read-gate`, and `okf-status` now catch `OSError`/`UnicodeDecodeError` around their file-reading call and report a clean ERROR result, via a new shared `ui.call_with_read_error_handling()` helper (`doc-index`/`tfidf-score` already had this from earlier rounds; extracting the helper avoided a 4-way pylint duplicate-code finding). Verified against all four commands with a non-UTF-8 fixture.

skills/studio/scripts/studio/utils/doc_index.py
tkcoding: Fixed in 5ffeb43b: `save_doc_index` now returns `bool` (was always `None`), and both `annotate_section_summary` and `write_concept_file` propagate their underlying save's real result instead of returning `True` unconditionally. Note: the specific reproduction in this finding (calling either function outside a Studio project) was already caught by an earlier early-return guard added in a prior round, so that exact case didn't silently report success -- but the general API-consistency gap (save_doc_index carrying no signal at all, unlike save_okf_manifest) was real and is now closed, including the narrow window between the two internal directory-resolution calls. Added tests that mock the underlying save to fail and assert the wrapper now propagates that.

skills/studio/scripts/studio/commands/cascade.py
tkcoding: Fixed in 5ffeb43b: added `_margin_threshold_arg()`, mirroring `commands/eval.py`'s `_compliance_arg`, rejecting any non-finite or non-positive `--margin-threshold` value at the argparse boundary. Added tests asserting `-1`, `0`, `nan`, and `inf` are all rejected.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -0,0 +1,290 @@
...
+    (hand-edited, or from a schema this module predates) triggers a clean
+    rebuild instead of a ``KeyError`` deep in a consumer.
+    """
+    if not isinstance(manifest, dict):
+        return False
+    entries = manifest.get("entries")
+    if not isinstance(entries, list):
+        return False
+    return all(
+        isinstance(entry, dict) and all(field in entry for field in _REQUIRED_MANIFEST_ENTRY_FIELDS)
```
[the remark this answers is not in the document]
tkcoding: Fixed in 5ffeb43b, adapted to this branch's actual schema: the manifest scheme here is keyed by `line_start` (not `position` -- an independent, earlier fix on this same PR used position-keying, but it was superseded during the rebase onto PR #110's hash-primary-matching fix, which is strictly more reorder-tolerant). `_is_valid_manifest_entry()` now checks that `line_start` is a non-bool `int` and `concept_file`/`built_from_hash` are `str`, closing the same class of bug: an unhashable value (e.g. a list) used as `line_start` would otherwise raise `TypeError` when used as a dict key in `get_okf_status`. Added a regression test with `line_start: []`.

ainetx:
## Deep Review — final status

All 9 findings from this review have been independently re-verified as fixed against the current head (commit referenced here as `9a5d33aa`):

- **Major** — OKF status reported "current" for a corrupted concept file it never read → fixed, content-validity check added, regression test corrupts a real file and confirms fallback to "missing".
- **Major** — Unhandled decode error crashed every file-consuming command on binary/non-UTF-8 input → fixed, verified live against all 6 affected commands with an actual malformed file: zero crashes, clean structured errors.
- **Major** — `--margin-threshold` accepted values that defeated the cascade's documented safety margin → fixed, verified live: `-1`, `0`, `nan`, `inf` are all now cleanly rejected at the CLI boundary.
- **Major** — `usage-report`/`okf-status` human output silently dropped fields present in JSON → fixed.
- **Major** — No test covered the argparse required-argument-omitted path → fixed, mechanism-appropriate coverage added for all 6 affected commands.
- **Minor** — `route_query()`'s returned schema had no documented stability contract → fixed (docstring only, no behavior change).
- **Minor** — Heading-nav counted fenced-code-block matches the same as prose → fixed, reuses existing fence-tracking, regression tests pass.
- **Minor** — Cache/manifest read failures logged only at an unreachable debug level → fixed, 5 of 6 promoted to warning with one legitimate exception.
- **Minor** — Persistence functions discarded their failure signal, reporting false success → fixed, real result now propagated.

No open findings remain from this review.

🤖 Deep Review verification — Claude Code

ainetx (approved):
All 9 Deep Review findings independently re-verified as fixed. Approving.

skills/studio/scripts/studio/commands/cascade.py
```diff
@@ -0,0 +1,98 @@
...
+    if not math.isfinite(parsed) or parsed <= 0:
+        raise argparse.ArgumentTypeError(
+            f"--margin-threshold must be a finite number > 0, got {value!r}")
+    return parsed
+
+
+# @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-cmd
+def cmd_retrieve(argv: List[str]) -> int:
+    """Route a query against a Markdown file through the JIT-retrieval cascade."""
+    p = argparse.ArgumentParser(
```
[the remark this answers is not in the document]
tkcoding: Fixed in 6c7cfc7e: `retrieve`, `heading-nav`, `read-gate`, and `usage-report` now all use `JsonSafeArgumentParser` (matching `doc-index`/`tfidf-score`/`okf-status`), so a missing or malformed argument returns the project's own `--json` ERROR contract instead of exiting via argparse's plain-text usage banner. Added regression tests for the missing-required-argument case on all four, plus the malformed-margin-threshold values (`-1`, `0`, `nan`, `inf`) now going through the same JSON-error path instead of a bare SystemExit.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -0,0 +1,220 @@
...
+    # every entry "missing" means no concept file has actually been written
+    # yet, which is "no bundle" for this decision, not "bundle exists."
+    bundle_exists = status["available"] and any(entry["status"] != "missing" for entry in status["entries"])
+    if not bundle_exists:
+        return _baseline_recommendation(expected_future_queries)
+
+    candidates = tier1_result.get("candidates", [])
+    if candidates:
+        by_line_start = {entry["line_start"]: entry for entry in status["entries"]}
+        relevant = [by_line_start[c["line_start"]] for c in candidates if c["line_start"] in by_line_start]
```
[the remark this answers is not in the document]
tkcoding: Fixed in 6c7cfc7e: `route_tier2` now treats a Tier 1 candidate that no longer maps to any current OKF status entry (the document changed structurally between Tier 1 picking it and this re-derived status) as unverifiable rather than silently dropping it from `relevant` -- previously, if every candidate dropped out this way, the empty list made `any(...)` vacuously False and the code fell through to recommending OKF on a candidate that was never actually checked. An unresolved candidate now triggers the same baseline+`okf_needs_rebuild` fallback as a genuinely stale one. Added a regression test with a bogus candidate line_start that matches no current section.

skills/studio/scripts/studio/utils/toc.py
```diff
@@ -327,11 +370,19 @@ def add_toc_max_level_argument(parser: argparse.ArgumentParser) -> None:
...
+    (the explicit end-of-document marker) -- accepting only ``---`` meant
+    frontmatter closed with ``...`` was never recognized as closed at all,
+    so every line after it (including every real heading) was treated as
+    still inside frontmatter and skipped entirely.
+    """
     if not lines or lines[0].strip() != "---":
         return 0
     idx = 1
-    while idx < len(lines) and lines[idx].strip() != "---":
+    while idx < len(lines) and lines[idx].strip() not in ("---", "..."):
```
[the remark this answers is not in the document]
tkcoding: Fixed in 6c7cfc7e (same fix applied on PR #110, since this was independently flagged on both): `_find_frontmatter_end`'s terminator check now uses `rstrip()` (trailing whitespace only), never `strip()`, so an indented `---`/`...` inside a YAML block scalar is no longer mistaken for the real terminator. Added a regression test with an indented `...` followed by a still-inside-frontmatter `#`-prefixed line, asserting only the genuine headings are parsed.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -0,0 +1,231 @@
...
+            "tier": "resolved_multi",
+            "reason": "heading_nav_tfidf_disagree",
+            "candidates": [_as_candidate(nav_pick), _as_candidate(tfidf_pick)],
+        }
+
+    agree_large_margin = tfidf_result["unambiguous"] or (
+        margin_threshold is not None
+        and tfidf_result["margin"] is not None
+        and tfidf_result["margin"] >= margin_threshold
+    )
```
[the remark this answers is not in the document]
tkcoding: Fixed in 33a37620 — added `_validate_margin_threshold`, called at the top of `route_tier1` (and therefore covering `route_query`, which delegates to it), rejecting any non-finite or non-positive `margin_threshold` with a `ValueError`. This closes the gap where a direct Python caller bypassed `commands/cascade.py`'s `_margin_threshold_arg` CLI validation entirely. Added `test_margin_threshold_rejects_invalid_values_at_the_callable_api` (parametrized over 0, -1, nan, inf) and `test_route_query_also_rejects_an_invalid_margin_threshold` — verified both fail without the fix.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -0,0 +1,531 @@
...
+    truncated, emptied, or corrupted after the fact -- this catches that
+    without needing full YAML parsing, which is more than this check needs
+    to answer "is there real content here at all".
+    """
+    try:
+        content = concept_path.read_text(encoding="utf-8")
+    except (OSError, UnicodeDecodeError) as exc:
+        logger.debug("okf concept file unreadable at %s: %s", concept_path, exc)
+        return False
+    return content.startswith("---\n")
```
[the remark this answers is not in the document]
tkcoding: Fixed in 33a37620 — `_concept_file_is_valid` now requires the closing `---` delimiter too (`content.startswith("---\n") and "\n---\n" in content[4:]`), not just the opening one, so a file truncated right after the opening delimiter is correctly reported as missing rather than current. Added `test_truncated_right_after_the_opening_delimiter_reports_missing` — verified it fails without the fix.

PR #105, opened 2026-08-27
feat(spec-coverage): add an advisory semantic-coverage engine

SanjeevSolanki:
## What

Adds an **advisory semantic-coverage engine** (`utils/eval_semantic.py`). Structural `spec-coverage` scores marker *density* — a file can read 100% covered while the code inside its markers does the wrong thing. This engine adds the layer density can't reach: **does a marked block implement the requirement it cites?**

## How

- **Rank first, judge last.** A deterministic, stdlib-only token-overlap pre-filter scores every marked block against its requirement text and surfaces the *weak links* (low overlap). Only those go to a model, so a large repo triggers zero model calls per block. The pre-filter is a *budget heuristic*, not a correctness oracle — a strong-overlap `presumed_covered` block buys itself out of a model call, it is never evidence the block is correct (lexical overlap is gameable; comments/string literals are stripped before scoring so prose echoing the requirement can't inflate it).
- **Seam, not transport.** The model call is a pluggable `SemanticJudgeFn` supplied out-of-tree; with none wired, every weak link is `unjudgeable` (never a false verdict) and nothing gates. This module contains no model client.
- **Advisory, never gates.** A verdict here never touches an exit code.
- **Honest coverage.** Blocks with no retrievable requirement, or too little text to compare, are reported `unjudgeable` — never a silent "covered". Every block is accounted for exactly once: `assessed + presumed_covered + unjudgeable + skipped_excluded`. Each verdict carries an evidence check (a quote absent from the code sets `evidence_ok=false`).
- **Scoped by the coverage-report contract.** Files a human declared `excluded` are skipped; `whole_file_claims` files are always judged regardless of overlap — that's where a green structural number most plausibly hides wrong code.
- **Optional calibration** over gold-labelled pairings reports the judge's accuracy + run-to-run consistency, honestly excluding unscoreable / crashed / no-majority (tie) cases and reporting effective sample size rather than deflating or inflating the metric.

## Tests & gates

- New tests span unit, adversarial (wrong code surfaced; comment- and escaped-quote-literal masking; non-ASCII/Cyrillic), fail-safe (a judge that raises, and a reply object whose attribute access itself raises), honest-accounting, calibration integrity, and **property-based invariants** — seed-deterministic generated inputs asserting overlap stays in `[0,1]∪{None}`, the accounting identity always holds, `tokenize` never emits short/stopword tokens, and `assess`/`calibrate` never raise.
[...]

Issue #107, opened 2026-08-27
spec-coverage: emit the excluded[] / whole_file_claims[] report fields the semantic layer reads


Oleg67:
**The gap.** #105 adds a semantic-coverage layer that scopes itself from the coverage report:

```python
return CoverageScope(excluded=set(_paths_from(report, "excluded")),
                     prioritised=_paths_from(report, "whole_file_claims"))
```

Nothing emits either field. #103 added the human-readable "Whole-file scope claims" section and #106 added scanned/excluded **counts** (`files_excluded`, `code_files_excluded`), but no array reaches the JSON report.

The consumer degrades safely — a report without those keys yields an empty scope rather than an error — which means the failure is silent: the semantic layer **skips nothing and prioritises nothing**, and because it is advisory, nothing goes red. The feature is wired and inert. That is the shape of defect this repository already tracks elsewhere: a check that reports success without having assessed anything.

**The change.** Emit both fields from `spec-coverage`, in the object form the consumer parses (`[{"path": ...}, ...]`, extra keys ignored):

| Field | Contents |
|---|---|
| `whole_file_claims[]` | One entry per file whose coverage rests on a whole-file scope marker — scope markers present, zero closed block pairs. Exactly the population the human section already prints. Fields: `path`, `total_lines`, `effective_lines`, `scope_marker_count`, `block_marker_count`. **Ordered largest first.** |
| `excluded[]` | Only files a human declared. Emitted as `[]` until that configuration surface exists — never inferred. |

**Design principles.**

- *Report on the same numbers.* No threshold changes, no change to how coverage or granularity is computed, no change to what counts as covered. This is reporting.
- *Exclusion is declared, not deduced.* An inferred "non-instructable" list fails unsafe: a wrongly excluded file is silently skipped by every consumer downstream. Attempts to infer it do not survive contact with real trees — on this repository a 127-line re-export module has zero functions while a 113-line module has fourteen, so size misclassifies, and counting definitions needs a parser per language where the marker scanner is deliberately language-agnostic.
[...]

PR #122, opened 2026-08-31
feat(spec-coverage): add an advisory `--semantic` coverage pass

SanjeevSolanki:
## What

Adds an opt-in `--semantic` pass to `cfs spec-coverage` that wires the advisory semantic-coverage engine (`utils/eval_semantic.py`, landed in #105) into the command. It builds the engine's pairings from the real marked code blocks and their resolved requirements, runs the advisory `assess`, and attaches a `semantic` section to the JSON report plus a one-line human summary.

Off by default; nothing changes unless `--semantic` is passed.

## Why

Structural `spec-coverage` scores marker *density* — a file can read fully covered while a block implements the wrong behaviour. This pass adds the advisory signal density can't reach: *does a marked block actually implement the requirement it cites?* (covered / partial / wrong / unjudgeable).

## Advisory — never gates

The section is attached **after** the structural status and exit code are computed, so a semantic verdict can never change them. Enforced structurally and tested:

- A forced `wrong` verdict leaves status/exit byte-for-byte unchanged.
- The pass is exception-isolated: any failure inside it (imports, context, the pass itself) is swallowed, logged, and recorded as `{"advisory": true, "error": …}` — it can never crash the run.
- With no judge wired, weak links are `unjudgeable` (never a silent zero); no model is called.
- Pairing paths are project-relative POSIX, matching the report's own scope arrays.

## Design notes

- Requirement granularity is **per-algo** (a block's id is its algo id; the feature-doc requirement is scoped to that id).
- Scope (`excluded` / `whole_file_claims`) is read from the coverage report and tolerated absent (degrades to an empty scope, never an error).
- Library-level logic in `utils/semantic_coverage.py`, CPT-traced; the command wiring is a thin resolver.

## Testing

- 14 unit tests: pairing construction, the no-judge default, honest serialisation, scope consumption, the advisory-never-gates invariant (including a **raising** pass), relative-path emission, and human/JSON rendering.
- Gates green: `cfs validate` 223/223 (0 errors), `spec-coverage` thresholds met (granularity 0.4616 ≥ 0.46 floor), pylint, vulture, full suite (4865 passed), per-file coverage (`semantic_coverage.py` 100%, `spec_coverage.py` ≥90%).

Follows the DCO + conventional-commit conventions in `CONTRIBUTING.md`; single signed-off commit.

## Summary by CodeRabbit

- **New Features**
  - Added an optional semantic coverage check for the `spec-coverage` command, enabled with `--semantic`.
  - Reports now include semantic assessments, requirement-to-code findings, identified gaps, and a concise summary.
  - Semantic coverage is available in both JSON and human-readable output, including when no files are found.

- **Bug Fixes**
  - Semantic findings and processing errors remain advisory and do not change structural coverage results or command exit status.

- **Documentation**
  - Updated spec-coverage documentation to describe the optional semantic assessment and its non-gating behavior.

### DIFF
--- a/architecture/features/spec-coverage.md
+++ b/architecture/features/spec-coverage.md
@@ -14,6 +14,7 @@
   - [Calculate Coverage Metrics](#calculate-coverage-metrics)
   - [Calculate Granularity Score](#calculate-granularity-score)
   - [Generate Coverage Report](#generate-coverage-report)
+  - [Assess Semantic Coverage (advisory)](#assess-semantic-coverage-advisory)
 - [4. States (CDSL)](#4-states-cdsl)
   - [Coverage Report Lifecycle](#coverage-report-lifecycle)
 - [5. Definitions of Done](#5-definitions-of-done)
@@ -72,7 +73,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - No code files found → report with `applicable: false` naming how many registered entries resolved to no files, and 0% coverage
 
 **Steps**:
-1. [x] - `p1` - User invokes `cfs spec-coverage [--min-coverage N] [--min-file-coverage N] [--min-granularity N] [--min-file-granularity N] [--verbose]` - `inst-user-spec-coverage`
+1. [x] - `p1` - User invokes `cfs spec-coverage [--min-coverage N] [--min-file-coverage N] [--min-granularity N] [--min-file-granularity N] [--verbose] [--semantic]` - `inst-user-spec-coverage`
 2. [x] - `p1` - Load project context: studio config, registry, systems, codebase entries - `inst-load-context`
 3. [x] - `p1` - Resolve all code files from registered codebase entries - `inst-resolve-code-files`
 4. [x] - `p1` - **FOR EACH** code file, scan for `@cpt-*` markers using `cpt-studio-algo-spec-coverage-scan` - `inst-foreach-file`
@@ -97,6 +98,7 @@ Without spec coverage, teams have no visibility into which parts of the codebase
 - [x] - `p1` - Route JSON report to file or terminal UI - `inst-output-report`
 - [x] - `p1` - Format uncovered ranges and render human-friendly per-file/status sections - `inst-human-report-helpers`
 - [x] - `p1` - Name the files whose coverage rests on a whole-file scope marker, largest first, with their count and total claimed lines - `inst-human-report-claims`
+- [x] - `p1` - When `--semantic` is set, run the advisory semantic pass (`cpt-studio-algo-semantic-coverage-pass`) and attach its section to the report **after** status/exit are computed, so it can never gate - `inst-attach-semantic`
 
 ## 3. Processes / Business Logic (CDSL)
 
@@ -167,6 +169,23 @@ A file with good granularity has approximately 1 CDSL instruction (`@cpt-begin`/
 
 **Supporting**:
 - [x] - `p1` - Report function signature and relative path helper - `inst-report-datamodel`
+
+### Assess Semantic Coverage (advisory)
+
+- [x] `p1` - **ID**: `cpt-studio-algo-semantic-coverage-pass`
+
+**Input**: registered artifacts (for id→doc resolution), the scanned code files, the coverage report (for scope)
+
+**Output**: an advisory `semantic` report section — never affects status/exit. Success shape: `assessed` / `presumed_covered` / `unjudgeable[]` / `findings[]` / `skipped_excluded` / `schema_version` / `advisory`. If the pass raises, it degrades to `{advisory: true, error: <message>}` instead.
+
+**Steps**:
+1. [x] - `p1` - Build the id→declaring-artifact map from cpt definition hits across the registered artifacts - `inst-scov-defmap`
+2. [x] - `p1` - Build one pairing per marked block: code = the block's lines, requirement = the algo declaration resolved from the block's id (unjudgeable when unresolved) - `inst-scov-pairings`
+3. [x] - `p1` - Run the advisory engine (`assess`) over the pairings, passing the coverage report for scope, and serialise the result as the `semantic` section (`advisory: true`) - `inst-scov-run`
+4. [x] - `p1` - Render a one-line advisory human summary (counts + weak/wrong tally) - `inst-scov-summary`
+
+**Supporting**:
+- [x] - `p1` - Module imports and setup for the semantic-coverage pass - `inst-scov-imports`
 
 ## 4. States (CDSL)
 
@@ -240,6 +259,7 @@ The system **MUST** output a JSON report with: summary (total files, covered fil
 | Report Generator | `skills/.../utils/coverage.py` | JSON report assembly |
 | Codebase Utils | `skills/.../utils/codebase.py` | Existing code scanning infrastructure (reused) |
 | Language Config | `skills/.../utils/language_config.py` | Language-specific comment patterns (reused) |
+| Semantic Coverage Pass | `skills/.../utils/semantic_coverage.py` | Advisory: build pairings, run the semantic engine, serialise the `semantic` section (never gates) |
 
 ## 7. Acceptance Criteria
 
--- a/skills/studio/scripts/studio/commands/spec_coverage.py
+++ b/skills/studio/scripts/studio/commands/spec_coverage.py
@@ -75,6 +75,10 @@ def _build_spec_coverage_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument("--verbose", action="store_true", help="Include per-file marker details and covered ranges")
     parser.add_argument("--output", default=None, help="Write report to file instead of stdout")
+    parser.add_argument("--semantic", action="store_true",
+                        help="Attach the advisory semantic-coverage pass — assesses covered/partial/"
+                             "wrong/unjudgeable per marked block; never gates status/exit "
+                             "(see architecture/features/spec-coverage.md)")
     return parser
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-build-parser
 
@@ -402,6 +406,27 @@ def _load_spec_coverage_context():
     # @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-load-context
 
 
+# @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-attach-semantic
+def _attach_semantic_section(args, filtered_files, json_report) -> None:
+    """Attach the advisory semantic pass AFTER status/exit are set, so a verdict can never gate."""
+    if not getattr(args, "semantic", False):
+        return
+    try:
+        # Everything advisory lives inside the guard — imports, context resolution, and the pass
+        # itself — so NOTHING (an import failure included) can change the structural status/exit
+        # already computed above. A failure is recorded as an advisory error rather than crashing.
+        from ..utils.context import get_context
+        from ..utils.semantic_coverage import run_semantic_pass
+        ctx = get_context()
+        if ctx is None:
+            return
+        json_report["semantic"] = run_semantic_pass(ctx, filtered_files, json_report)
+    except Exception as exc:  # pylint: disable=broad-except
+        logger.warning("semantic pass failed (advisory, ignored): %s", exc)
+        json_report["semantic"] = {"advisory": True, "error": str(exc)}
+# @cpt-end:cpt-studio-flow-spec-coverage-report:p1:inst-attach-semantic
+
+
 def _generate_spec_coverage_report(args, meta, project_root: Path) -> tuple[dict, int]:
     # @cpt-begin:cpt-studio-flow-spec-coverage-report:p1:inst-validate-systems
     system_slugs, validation_error = _validate_selected_systems(args, meta)
@@ -417,18 +442,20 @@ def _generate_spec_coverage_report(args, meta, project_root: Path) -> tuple[dict
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
     if not filtered_files:
         requested = _requested_thresholds(args)
-        empty = _empty_coverage_result(
-            _count_selected_codebase_entries(meta, system_slugs), requested
-        )
-        return empty, 2 if requested else 0
+        json_report = _empty_coverage_result(
+            _count_selected_codebase_entries(meta, system_slugs), requested)
+        # --semantic still attaches its (empty) advisory section here, so the flag's presence
+        # is consistent whether or not the codebase resolved to any files.
+        _attach_semantic_section(args, filtered_files, json_report)
+        return json_report, 2 if requested else 0
     # @cpt-end:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
-    file_coverages = _scan_file_coverages(filtered_files)
-    report = calculate_metrics(file_coverages)
+    report = calculate_metrics(_scan_file_coverages(filtered_files))
     json_report = generate_report(report, verbose=args.verbose, project_root=project_root)
     # The population this percentage is computed over, so a shift in it is
     # attributable rather than looking like a change in the code.
     json_report["summary"]["files_excluded"] = files_excluded
     status = _apply_thresholds(report, args, project_root, json_report)
+    _attach_semantic_section(args, filtered_files, json_report)
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-covered
     if status == "PASS" and report.covered_lines > 0:
         return json_report, 0
@@ -587,6 +614,17 @@ def _human_spec_coverage(data: dict) -> None:
     ui.detail("Files", f"{files_line} ({excluded} excluded)" if excluded else files_line)
     ui.detail("Coverage", f"{summary.get('coverage_pct', 0):.1f}%")
     ui.detail("Granularity", f"{summary.get('granularity_score', 0):.4f}")
+
+    # Advisory semantic line — never part of the gate; only shown when --semantic ran.
+    semantic = data.get("semantic")
+    if semantic and semantic.get("error"):
+        # The advisory pass recorded a failure (possibly an import failure of semantic_coverage
+        # itself) — render it WITHOUT importing summary_line, which could re-raise here, after the
+        # structural status/exit are already set.
+        ui.info(f"semantic (advisory, never gates): pass errored, skipped — {semantic['error']}")
+    elif semantic:
+        from ..utils.semantic_coverage import summary_line
+        ui.info(summary_line(semantic))
 
     # Per-file details — files is a dict {path: entry_dict}
     files = data.get("files", {})
--- a/skills/studio/scripts/studio/utils/semantic_coverage.py
+++ b/skills/studio/scripts/studio/utils/semantic_coverage.py
@@ -1,0 +1,149 @@
+"""Advisory semantic-coverage pass — wire the semantic engine into ``spec-coverage``.
+
+Structural ``spec-coverage`` scores marker *density*; the semantic engine
+(:mod:`studio.utils.eval_semantic`) asks the layer density cannot reach — *does a marked block
+implement the requirement it cites?* This module is the thin glue between them: it builds the
+engine's ``Pairing`` list from the real marked blocks and their resolved requirements, runs the
+advisory ``assess``, and serialises the result for the coverage report plus a one-line human summary.
+
+**Advisory, never gates.** Nothing here touches the coverage status or exit code — the caller
+attaches the returned section to the report *after* the structural gate is computed.
+
+Requirement granularity is **per-algo**, not per-instruction: a block marker's ``id`` is its algo id,
+and the feature-doc requirement text is scoped to that id (the instruction slug is not independently
+scoped in the doc format). Every block of an algo therefore pairs with the algo's declaration text;
+blocks whose code diverges from that vocabulary surface as weak links.
+
+@cpt-algo:cpt-studio-algo-semantic-coverage-pass:p1
+"""
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-imports
+from __future__ import annotations
+
+import logging
+import os
+from pathlib import Path
+from typing import Dict, List, Optional, Sequence
+
+from .codebase import CodeFile
+from .context import collect_artifacts_to_scan
+from .document import scan_cpt_ids
+from . import eval_semantic
+
+logger = logging.getLogger(__name__)
+# @cpt-end:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-imports
+
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-defmap
+def _definition_map(ctx: object) -> Dict[str, Path]:
+    """Map every cpt **definition** id to its declaring artifact path.
+
+    This is the id→doc lookup the requirement side needs: a block cites an algo id, and the
+    requirement text lives in whichever artifact *defines* that id. Built from the registered
+    artifacts (``collect_artifacts_to_scan``); the first definition wins on the rare duplicate.
+    """
+    out: Dict[str, Path] = {}
+    artifacts, _sources = collect_artifacts_to_scan(ctx)
+    for artifact_path, _kind in artifacts:
+        for hit in scan_cpt_ids(artifact_path):
+            if hit.get("type") == "definition" and isinstance(hit.get("id"), str):
+                existing = out.get(hit["id"])
+                if existing is not None and existing != artifact_path:
+                    logger.warning("semantic: cpt id %s defined in both %s and %s; keeping the first",
+                                   hit["id"], existing, artifact_path)
+                out.setdefault(hit["id"], artifact_path)
+    return out
+# @cpt-end:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-defmap
+
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-pairings
+def _relative_posix(path: Path, project_root: Optional[Path]) -> str:
+    """A **project-relative POSIX** path, never absolute — so no local path leaks into the report or
+    the out-of-tree judge prompt. Degrades to a ``..``-relative path for a file outside the root,
+    then to the bare filename if even that is impossible (Windows cross-drive)."""
+    if project_root is None:
+        return path.as_posix()
+    root, resolved = Path(project_root).resolve(), path.resolve()
+    try:
+        return resolved.relative_to(root).as_posix()
+    except ValueError:
+        logger.debug("semantic: %s is outside project_root; using a relative path", path)
+    try:
+        return Path(os.path.relpath(resolved, root)).as_posix()
+    except ValueError:
+        return Path(path.name).as_posix()
+
+
+def _pairings_for_files(files: Sequence[Path], definitions: Dict[str, Path],
+                        project_root: Optional[Path] = None) -> List[eval_semantic.Pairing]:
+    """One ``Pairing`` per marked block: code = the block's own lines, requirement = the algo
+    declaration resolved from the block's id, or ``None`` (unjudgeable) when the id declares no
+    retrievable text. ``block_id`` is ``<algo>:<inst>`` so each block is individually identifiable
+    even though the requirement (and ``bm.id``) is shared across an algo's blocks.
+
+    ``path`` is emitted **project-relative POSIX** (see ``_relative_posix``) so it matches the
+    coverage report's own scope arrays — the code files arrive absolute, so without this the scope
+    reader would compare an absolute pairing path against a relative report path and never match."""
+    pairings: List[eval_semantic.Pairing] = []
+    for code_path in dict.fromkeys(files):        # dedup duplicate registrations, keep order
+        code_file, errs = CodeFile.from_path(code_path)
+        if code_file is None:
+            logger.warning("semantic: skipping unparseable file %s: %s", code_path, errs)
+            continue
+        path_posix = _relative_posix(code_file.path, project_root)
+        for block in code_file.block_markers:
+            doc = definitions.get(block.id)
+            requirement = eval_semantic.resolve_requirement(doc, block.id) if doc is not None else None
+            pairings.append(eval_semantic.Pairing(
+                block_id=f"{block.id}:{block.inst}",
+                inst=block.inst,
+                path=path_posix,
+                start_line=block.start_line,
+                code="\n".join(block.content),
+                requirement=requirement))
+    return pairings
+# @cpt-end:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-pairings
+
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-run
+def run_semantic_pass(ctx: object, files: Sequence[Path], coverage_report: Dict[str, object],
+                      judge_fn: Optional[eval_semantic.SemanticJudgeFn] = None) -> Dict[str, object]:
+    """Build pairings from the marked blocks, run the advisory engine, return the ``semantic`` section.
+
+    ``coverage_report`` is passed straight to ``assess`` for scoping (``excluded`` / ``whole_file_claims``,
+    tolerated absent → empty scope). With no ``judge_fn`` wired, weak links are ``unjudgeable`` and no
+    model is called. The returned dict is advisory on its face (``"advisory": True``) and is never read
+    by the coverage gate — the caller attaches it after the status/exit are set.
+    """
+    definitions = _definition_map(ctx)
+    pairings = _pairings_for_files(files, definitions, getattr(ctx, "project_root", None))
+    result = eval_semantic.assess(pairings, judge_fn=judge_fn, report=coverage_report)
+    return {
+        "assessed": result.assessed,
+        "presumed_covered": result.presumed_covered,
+        "unjudgeable": [{"block_id": gap.block_id, "path": gap.path,
+                         "start_line": gap.start_line, "reason": gap.reason}
+                        for gap in result.unjudgeable],
+        "findings": [{"block_id": finding.block_id, "path": finding.path,
+                      "start_line": finding.start_line, "verdict": finding.verdict,
+                      "rationale": finding.rationale, "evidence_ok": finding.evidence_ok,
+                      "forced": finding.forced}
+                     for finding in result.findings],
+        "skipped_excluded": result.skipped_excluded,
+        "schema_version": result.schema_version,
+        "advisory": True,
+    }
+# @cpt-end:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-run
+
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-summary
+def summary_line(semantic: Dict[str, object]) -> str:
+    """A one-line advisory human summary: counts + the weak/wrong finding tally, labelled advisory."""
+    if semantic.get("error"):
+        return f"semantic (advisory, never gates): pass errored, skipped — {semantic['error']}"
+    findings = semantic.get("findings") or []
+    weak = sum(1 for finding in findings
+               if finding.get("verdict") in (eval_semantic.SEM_WRONG, eval_semantic.SEM_PARTIAL))
+    return (f"semantic (advisory, never gates): {semantic.get('assessed', 0)} judged, "
+            f"{semantic.get('presumed_covered', 0)} presumed-covered, "
+            f"{len(semantic.get('unjudgeable') or [])} unjudgeable, {weak} weak/wrong")
+# @cpt-end:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-summary
--- a/tests/test_semantic_coverage.py
+++ b/tests/test_semantic_coverage.py
@@ -1,0 +1,337 @@
+"""Tests for the advisory semantic-coverage pass (utils.semantic_coverage) and its wiring into
+``spec-coverage``.
+
+The headline invariant: attaching a semantic section — even one carrying a ``wrong`` verdict — never
+changes the coverage status or exit code. The rest pin pairing construction, the no-judge default,
+honest serialisation, and scope consumption.
+"""
+import json
+import sys
+from io import StringIO
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "skills" / "studio" / "scripts"))
+
+from studio.utils import semantic_coverage as sc
+from studio.utils import eval_semantic as sem
+from studio.commands.spec_coverage import cmd_spec_coverage
+from studio.utils.artifacts_meta import ArtifactsMeta, CodebaseEntry, Kit, SystemNode
+
+_ALGO = "cpt-studio-algo-fixture"
+_REQ = "validate the user email address format and reject a malformed address before saving the record"
+
+
+def _marked(tmp_path: Path, name: str = "mod.py") -> Path:
+    """A file with two blocks of one algo: a strong block whose identifiers echo the requirement, and
+    a weak block (compute_tax) with no overlap."""
+    p = tmp_path / name
+    p.write_text(
+        f"# @cpt-begin:{_ALGO}:p1:inst-strong\n"
+        "def validate_email_address_and_reject_malformed(record):\n"
+        "    return save_record(record)\n"
+        f"# @cpt-end:{_ALGO}:p1:inst-strong\n"
+        f"# @cpt-begin:{_ALGO}:p1:inst-weak\n"
+        "def compute_tax(amount):\n"
+        "    return amount * lookup_rate(amount)\n"
+        f"# @cpt-end:{_ALGO}:p1:inst-weak\n",
+        encoding="utf-8")
+    return p
+
+
+# --- pairing construction --------------------------------------------------
+
+def test_pairings_one_per_block_with_fields(tmp_path: Path) -> None:
+    pairings = sc._pairings_for_files([_marked(tmp_path)], {})
+    assert {p.block_id for p in pairings} == {f"{_ALGO}:strong", f"{_ALGO}:weak"}
+    weak = next(p for p in pairings if p.block_id.endswith(":weak"))
+    assert weak.inst == "weak"
+    assert weak.path.endswith("mod.py")
+    assert weak.start_line > 0
+    assert "compute_tax" in weak.code
+    assert weak.requirement is None            # empty definitions map → unresolved → unjudgeable
+
+
+def test_pairings_resolve_requirement_from_definitions(tmp_path: Path, monkeypatch) -> None:
+    monkeypatch.setattr(sem, "resolve_requirement", lambda doc, bid: _REQ if bid == _ALGO else None)
+    pairings = sc._pairings_for_files([_marked(tmp_path)], {_ALGO: tmp_path / "doc.md"})
+    assert all(p.requirement == _REQ for p in pairings)
+
+
+# --- run_semantic_pass -----------------------------------------------------
+
+def test_no_judge_default_reports_unjudgeable_not_zero(tmp_path: Path, monkeypatch) -> None:
+    monkeypatch.setattr(sc, "_definition_map", lambda ctx: {})   # no requirements resolve
+    section = sc.run_semantic_pass(None, [_marked(tmp_path)], {}, judge_fn=None)
+    assert section["advisory"] is True
+    assert section["assessed"] == 0
+    assert section["findings"] == []
+    assert len(section["unjudgeable"]) == 2                      # both blocks: no requirement
+
+
+def test_serialises_a_wrong_finding_with_the_full_shape(tmp_path: Path, monkeypatch) -> None:
+    monkeypatch.setattr(sc, "_definition_map", lambda ctx: {_ALGO: tmp_path / "doc.md"})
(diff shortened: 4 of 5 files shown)


skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -417,18 +440,17 @@ def _generate_spec_coverage_report(args, meta, project_root: Path) -> tuple[dict
     # @cpt-begin:cpt-studio-state-spec-coverage-report:p1:inst-state-uncovered
     if not filtered_files:
         requested = _requested_thresholds(args)
-        empty = _empty_coverage_result(
+        return _empty_coverage_result(
             _count_selected_codebase_entries(meta, system_slugs), requested
-        )
-        return empty, 2 if requested else 0
+        ), 2 if requested else 0
```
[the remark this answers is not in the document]
SanjeevSolanki: Fixed. The empty-scope early return now attaches the (empty) advisory section via `_attach_semantic_section` before returning, so a `--semantic` run always carries a `semantic` section regardless of whether the codebase resolved to any files. Added `test_semantic_section_attaches_on_empty_scope`.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -588,6 +610,12 @@ def _human_spec_coverage(data: dict) -> None:
     ui.detail("Coverage", f"{summary.get('coverage_pct', 0):.1f}%")
     ui.detail("Granularity", f"{summary.get('granularity_score', 0):.4f}")
 
+    # Advisory semantic line — never part of the gate; only shown when --semantic ran.
+    semantic = data.get("semantic")
+    if semantic:
+        from ..utils.semantic_coverage import summary_line
```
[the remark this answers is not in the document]
SanjeevSolanki: Good catch — fixed. The human renderer now renders a recorded advisory `error` inline without importing `summary_line`, closing the re-raise-after-status hole on the human path (it mirrors the exception isolation already on the JSON path). Added `test_human_error_section_renders_without_summary_line`, which asserts `summary_line` is never called for an error section.

architecture/features/spec-coverage.md
```diff
@@ -168,6 +170,23 @@ A file with good granularity has approximately 1 CDSL instruction (`@cpt-begin`/
 **Supporting**:
 - [x] - `p1` - Report function signature and relative path helper - `inst-report-datamodel`
 
+### Assess Semantic Coverage (advisory)
+
+- [x] `p1` - **ID**: `cpt-studio-algo-semantic-coverage-pass`
+
+**Input**: registered artifacts (for id→doc resolution), the scanned code files, the coverage report (for scope)
+
+**Output**: an advisory `semantic` report section (`assessed` / `presumed_covered` / `unjudgeable[]` / `findings[]`) — never affects status/exit
```
ainetx: ### Feature doc's semantic-section output description is incomplete

**Severity:** Minor

**Problem**
This section documents the `semantic` JSON output as only `assessed` / `presumed_covered` / `unjudgeable[]` / `findings[]`. The actual `run_semantic_pass()` function returns a 7-key dict, adding `skipped_excluded`, `schema_version`, and `advisory`. On an internal failure, `_attach_semantic_section`'s except-branch emits a completely different 2-key shape instead: `{"advisory": true, "error": "<message>"}`. Neither the three extra keys nor the error shape are mentioned anywhere in the doc.

**How to reproduce**
1. Read the `**Output**` line above.
2. Open `skills/studio/scripts/studio/utils/semantic_coverage.py`, `run_semantic_pass`, and compare its literal return dict.
3. Open `skills/studio/scripts/studio/commands/spec_coverage.py`, `_attach_semantic_section`, and note the `except Exception` branch's alternate shape.

**Expected behavior**
The documented output contract enumerates every key a consumer might encounter in `json_report["semantic"]`, including the error-degraded shape.

**Actual behavior**
The doc lists 4 of 7 success-path keys and never mentions the alternate error shape.

```text
 doc claims:                      code actually emits (success path):
 { assessed,                      { assessed,
   presumed_covered,                presumed_covered,
   unjudgeable[],                   unjudgeable[],
   findings[] }                     findings[],
                                     skipped_excluded,   <- undocumented
                                     schema_version,     <- undocumented
                                     advisory }          <- undocumented

                                  code emits (failure path, entirely different):
                                  { advisory: true, error: "<msg>" }   <- undocumented at all
```

**Impact**
Low risk to correctness — behavior is unaffected — but a downstream tool or reader integrating against the JSON report off this doc will be surprised by the extra fields and won't know to defensively check for the error shape before assuming the 4-key success shape.

**Suggested correction**
Update the `**Output**` line to enumerate the full 7-key success shape and add a line documenting the `{"advisory": true, "error": ...}` degraded shape emitted on internal failure.

**How to verify**
Re-read the updated doc line against `run_semantic_pass()`'s return dict and `_attach_semantic_section`'s except-branch; confirm every key literal in both code paths appears in the doc.

tests/test_semantic_coverage.py
```diff
@@ -0,0 +1,262 @@
+"""Tests for the advisory semantic-coverage pass (utils.semantic_coverage) and its wiring into
```
ainetx, line 1: ### `--semantic` is untested against a structurally-failing run (exit code 2)

**Severity:** Major

**Problem**
This PR's headline guarantee is that the semantic pass is purely advisory and "never gates" the structural result. The test suite verifies this only for the exit-0/PASS case (`test_semantic_section_attaches_and_a_wrong_verdict_never_gates`, `test_semantic_pass_that_raises_never_gates` — both run `--semantic` with no threshold flags against a fully-covered fixture). Neither this file nor `tests/test_spec_coverage.py` combines `--semantic` with a threshold flag (e.g. `--min-coverage`) set to force exit code 2.

**How to reproduce**
1. Search both test files for any co-occurrence of a `min_*` threshold flag with `--semantic`.
2. Observe zero matches — the only `--semantic` invocations run with no threshold flags against a fully-covered fixture, so exit is always 0 regardless of what the semantic pass does.

**Expected behavior**
At least one test should assert that `--semantic` attached to a run that structurally fails (exit code 2, e.g. via `--min-coverage 100` against a partially-covered fixture) still exits 2 and carries the correct `status`, exactly matching the equivalent run without `--semantic`.

**Actual behavior**
No such test exists — the invariant is verified only on the exit-0 side.

```text
tested:      --semantic + (no thresholds, full coverage) --> exit 0  == exit 0 without --semantic   [OK, covered]
untested:    --semantic + (--min-coverage above actual)   --> exit 2 == exit 2 without --semantic?   [UNKNOWN]
                                                              ^
                                                   this half of the "never gates" claim
                                                   has no regression coverage
```

**Impact**
A future change to the attach-point ordering could silently break the never-gates guarantee specifically on the failure path — the path where a false PASS or a false FAIL is most damaging to CI — without any test catching it.

**Suggested correction**
Add a test mirroring `test_semantic_section_attaches_and_a_wrong_verdict_never_gates` but with a threshold flag (e.g. `--min-coverage 100`) forcing exit 2, asserting `code_sem == code_plain == 2` and `rep_sem.get("status") == rep_plain.get("status")`.

**How to verify**
Add the suggested test, temporarily reorder the threshold-check and semantic-attach calls to confirm the new test fails when the invariant is violated, then restore the original ordering and confirm it passes.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
+"""Advisory semantic-coverage pass — wire the semantic engine into ``spec-coverage``.
```
ainetx, line 1: ### `excluded`/`whole_file_claims` scoping never activates on a real coverage report

**Severity:** Major

**Problem**
`eval_semantic.py`'s `coverage_scope()` reads `report.get("excluded")` and `report.get("whole_file_claims")`, expecting each to be a list of `{"path": ...}` dicts, to implement the documented behavior "files a human declared excluded are skipped; files flagged whole_file_claims are prioritised." But the real report producer, `generate_report()` in `utils/coverage.py`, never emits either key in that name or shape — it only produces `summary["files_excluded"]` (an integer count) and `flagged_files` (a list of plain path strings). Since the reader tolerates a missing/malformed key by returning an empty list, `coverage_scope()` always falls back to an empty scope in every real invocation — only exercised in tests via a hand-constructed report dict that stubs a shape the real code never produces.

**How to reproduce**
1. Run `spec-coverage --semantic` on any real project (no test doubles).
2. Inspect the `coverage_report` dict passed into `run_semantic_pass`.
3. Confirm `coverage_report.get("excluded")` is `None` and `coverage_report.get("whole_file_claims")` is `None`.
4. Trace into `coverage_scope()` — both lookups return `[]`, yielding an empty scope regardless of what files are actually excluded or scope-only.

**Expected behavior**
A file a human excluded from structural coverage should be skipped by the semantic pass, and a file flagged as scope-only should be prioritized for judging, in every real invocation.

**Actual behavior**
The scoping logic never activates on a real coverage report; the excluded-count is always 0 and prioritization is always empty in production.

```text
generate_report() actually emits:              eval_semantic.coverage_scope() expects:
{ summary: { files_excluded: <int> }, ... }    report["excluded"]        = [{"path": ...}, ...]
{ flagged_files: ["a.py", "b.py"] }            report["whole_file_claims"] = [{"path": ...}, ...]
       (no "excluded" key)                            ^ never matches -> always []
       (no "whole_file_claims" key)                    ^ never matches -> always []
                                                        |
                                                        v
                                        CoverageScope(excluded=set(), prioritised=[])
                                        every real run -- documented skip/prioritize is a no-op
```

**Impact**
Files a maintainer explicitly excluded from spec-coverage still get semantically judged (wasted judge budget and potentially confusing findings on code deliberately opted out of), and files whose green coverage is most likely to be hiding wrong code are not prioritized as intended — undermining the pass's stated purpose.

**Suggested correction**
Either update `generate_report()` to emit `excluded` and `whole_file_claims` in the shape `coverage_scope()` expects, or change `coverage_scope()` to read the shape the report actually produces today. Add an integration test that runs the semantic pass against real `generate_report()` output.

**How to verify**
After the fix, run `spec-coverage --semantic` against a fixture with one excluded and one whole-file-scope-only file; confirm the excluded-skip count matches and the scope-only file's blocks are ranked ahead of others.

skills/studio/scripts/studio/commands/spec_coverage.py
```diff
@@ -75,6 +75,8 @@ def _build_spec_coverage_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument("--verbose", action="store_true", help="Include per-file marker details and covered ranges")
     parser.add_argument("--output", default=None, help="Write report to file instead of stdout")
+    parser.add_argument("--semantic", action="store_true",
```
ainetx, line 78: ### `--semantic` help text doesn't explain what it classifies

**Severity:** Minor

**Problem**
This flag's argparse help string is a single terse clause — "Attach the advisory semantic-coverage pass (never gates)" — with no mention of what the pass reports or how classification works. The heuristic and its caveats are documented only in `eval_semantic.py`'s module docstring, never surfaced in CLI help.

**How to reproduce**
1. Run `spec-coverage --help`.
2. Read the `--semantic` line.
3. Search the repo for where the classification is explained — only in module docstrings, never referenced from CLI help.

**Expected behavior**
`--help` output gives a user enough information to understand what output the flag produces, or at minimum points them at the feature doc.

**Actual behavior**
The help text says only that the pass is "advisory" and "never gates," with no indication of what it measures or classifies.

```text
--help output today:
  --semantic   Attach the advisory semantic-coverage pass (never gates)
                                     ^
                       no mention of: covered / partial / wrong / unjudgeable
                       no pointer to: eval_semantic.py docstring or the feature doc
```

**Impact**
Low — does not affect correctness or gating, but increases the chance a user runs `--semantic`, sees unfamiliar output, and has no in-tool way to learn what it means.

**Suggested correction**
Expand the help string to something like: "Attach the advisory semantic-coverage pass (assesses covered/partial/wrong/unjudgeable per marked block; never gates status/exit — see architecture/features/spec-coverage.md)", or add a short `epilog` summarizing the classification scheme.

**How to verify**
Run `spec-coverage --help` after the change and confirm the description names the verdict categories or references the doc.

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx, line 1: ### The covered/partial/wrong verdict boundary has no documented rubric

**Severity:** Minor

**Problem**
`eval_semantic.py` documents and constant-ifies the unjudgeable/weak-link boundary (`_WEAK_LINK_THRESHOLD`). But the substantive classification a real judge must perform — covered vs. partial vs. wrong — has no rubric documented anywhere. That decision is entirely delegated to an out-of-tree judge function. The one concrete example shipped in-tree, `reference_stub_judge`, explicitly disclaims itself as not representing real classification behavior.

**How to reproduce**
1. Search `eval_semantic.py` and the repo for a definition distinguishing `SEM_PARTIAL` from `SEM_WRONG`, or `SEM_COVERED` from `SEM_PARTIAL`.
2. Find only the three bare constants, the unrelated weak-link threshold, and the self-disclaimed stub judge.
3. Confirm no doc, docstring, or comment specifies criteria for the actual judge to apply.

**Expected behavior**
Since this classification is the core value proposition of the semantic pass, the repo should document — even at a high level — what should make a judge call something wrong vs. partial vs. covered.

**Actual behavior**
The classification boundary is left entirely to whatever judge is plugged in, with the only in-repo example explicitly disclaiming itself as representative.

```text
documented boundary:                 undocumented boundary:
 unjudgeable | judged                  covered | partial | wrong
      ^                                     ^        ^        ^
 _WEAK_LINK_THRESHOLD                   no rubric anywhere in-tree;
 (named constant,                       reference_stub_judge explicitly
  pre-filter logic)                     disclaims representativeness
```

**Impact**
Low direct risk today, but every real judge integration starts from a blank slate with no in-tree acceptance criteria to test conformance against.

**Suggested correction**
Add a short rubric (even a few bullet examples) to this module's docstring or the feature doc describing the intended semantics of covered/partial/wrong for judge implementers.

**How to verify**
After adding the rubric, check that existing gold-label test cases are consistent with the documented definitions.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+    declaration resolved from the block's id, or ``None`` (unjudgeable) when the id declares no
+    retrievable text. ``block_id`` is ``<algo>:<inst>`` so each block is individually identifiable
+    even though the requirement (and ``bm.id``) is shared across an algo's blocks.
+
+    ``path`` is emitted **project-relative POSIX** when ``project_root`` is given, matching the
+    coverage report's own scope arrays (``excluded`` / ``whole_file_claims``) — the code files
+    arrive absolute, so without this the scope reader would compare an absolute pairing path against
+    a relative report path and silently never match."""
+    pairings: List[eval_semantic.Pairing] = []
+    for code_path in files:
```
ainetx: ### Semantic pass double-counts pairings for files with duplicate registrations

**Severity:** Major

**Problem**
The structural coverage pass explicitly dedupes before scanning: `_scan_file_coverages` iterates `sorted(set(filtered_files))`. The semantic pass, however, is fed the same `filtered_files` list *before* any dedup — `_pairings_for_files` then does `for code_path in files:` with zero deduplication, building one `CodeFile` and one set of block pairings per occurrence in the list, not per unique path. `filtered_files` can legitimately contain the same path more than once whenever codebase-entry registrations overlap between a parent system and a selected child system (or between two sibling systems with overlapping globs).

**How to reproduce**
1. Register a codebase entry that is covered by both a parent system and one of its children (or by two systems whose codebase globs overlap).
2. Run `spec-coverage --semantic` scoped to include both systems (or with no system filter, which walks the whole tree).
3. Observe that `filtered_files` contains the duplicated path twice.
4. Observe that the structural coverage percentage is computed off the deduped file set, while the semantic section's `assessed`/`findings` count every block in the duplicated file twice.

**Expected behavior**
The semantic pass should process each unique file's marked blocks exactly once, so per-file pairing/finding counts are consistent with the (deduped) structural pass over the same file set.

**Actual behavior**
A file appearing twice in `filtered_files` produces two identical pairing entries per block, so its findings (including any wrong/partial verdicts) and its contribution to assessed/presumed-covered counts are doubled in the semantic report, even though the structural percentage for the same file is only counted once.

```text
_collect_selected_system_files (system tree walk)
        |
        |  visit(parent-system)  -> _collect_codebase_files(file.py) appended
        |  visit(child-system)   -> _collect_codebase_files(file.py) appended again
        v
   collected_files = [file.py, file.py, ...]
        |
        v
_filter_ignored_files -> filtered_files (duplicates preserved)
        |
        +--------------------------+---------------------------+
        v                                                       v
_scan_file_coverages                                   _attach_semantic_section
  sorted(set(filtered_files))  <- DEDUPED               run_semantic_pass(ctx, filtered_files, ...)
        |                                                       |
        v                                                       v
  structural % (file.py counted once)             _pairings_for_files: for code_path in files
                                                          (file.py processed twice)
                                                                |
                                                                v
                                                   semantic findings/assessed doubled for file.py
```

**Impact**
The advisory semantic section's counts become inflated and inconsistent with the structural percentage whenever the project's system tree has overlapping/nested codebase registrations — a plausible, not-exotic project shape. Since the section is advisory-only it doesn't flip pass/fail, but it silently misrepresents semantic coverage counts and could double-report the same finding to a human reading the report.

**Suggested correction**
Dedupe the file list before building pairings, mirroring the structural pass — e.g. call `run_semantic_pass(ctx, sorted(set(filtered_files)), json_report)`, or dedupe inside `_pairings_for_files` itself (`for code_path in dict.fromkeys(files):` to preserve order while removing repeats).

**How to verify**
Add a test that constructs a system tree where two system nodes resolve to the same source file path (or directly calls `_pairings_for_files`/`run_semantic_pass` with a `files` list containing a duplicate path) and assert pairings/findings for that path are produced once, not once per occurrence in the input list.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+        code_file, _errs = CodeFile.from_path(code_path)
+        if code_file is None:
+            continue
+        block_path = code_file.path
+        if project_root is not None:
+            try:
+                block_path = block_path.resolve().relative_to(Path(project_root).resolve())
+            except ValueError:
```
ainetx: ### Path-relativization fallback reintroduces the absolute-vs-relative scope-matching bug it was meant to fix

**Severity:** Minor (currently latent; becomes a real correctness bug the moment a separately-flagged producer gap is closed)

**Problem**
This function computes a project-relative path for scope matching via `block_path.resolve().relative_to(Path(project_root).resolve())`, but on `ValueError` (raised when the file isn't under `project_root` — different drive, symlinked/mounted subtree, etc.) it falls back to the original absolute `code_file.path`. The docstring above warns: "the code files arrive absolute, so without this the scope reader would compare an absolute pairing path against a relative report path and silently never match" — yet the fallback branch does exactly that for any file outside `project_root`. This is currently masked because the coverage report producer doesn't yet emit `excluded`/`whole_file_claims` keys at all, so scope matching degrades to empty regardless of path format today — but the bug is latent and will silently reactivate once that gap is closed.

**How to reproduce**
1. Register a codebase entry pointing at a file outside `project_root` (e.g. a symlinked or mounted path, or a different drive on Windows).
2. Run `spec-coverage --semantic` with that file excluded via a (future) `excluded` scope entry emitted as a project-relative path.
3. `_pairings_for_files` hits the `ValueError` branch and emits an absolute path for that file's pairings.
4. Scope matching compares the absolute pairing path against the relative scope entry and never matches — the exclusion silently fails to apply.

**Expected behavior**
When a file falls outside `project_root`, the fallback path representation should still be normalizable/comparable against the relative scope arrays the coverage report emits, so scope exclusion/prioritization keeps working regardless of why relativization failed.

**Actual behavior**
The fallback silently reverts to the exact absolute-path shape the surrounding code comment says must never reach the scope reader, for the case "project_root given, but file outside it" — untested by the existing suite, which only covers a same-filesystem unrelated-directory case and asserts merely that no exception is raised, not that the emitted path is correct or that a genuine cross-drive/mount trigger is handled.

```text
project_root given, file OUTSIDE project_root
        |
        v
block_path.resolve().relative_to(project_root)  --raises ValueError-->
        |
        v
except ValueError: block_path = code_file.path   (ABSOLUTE)
        |
        v
pairing.path = absolute path string
        |
        v
coverage_scope(report).excluded = {relative paths from report}
        |
        v
_norm_path(pairing.path) not in scope.excluded   --> ALWAYS TRUE (never matches)
        |
        v
file that should have been excluded/prioritized is silently treated as in-scope/unprioritized
```

**Impact**
Currently inert (no producer emits `excluded`/`whole_file_claims` yet), but it is a latent, silent correctness bug: the moment the coverage-report producer is extended to populate those scope arrays, any codebase file outside `project_root` will bypass exclusion/prioritization scoping with no error, warning, or test failure.

**Suggested correction**
On the `ValueError` fallback, keep emitting a path shape comparable to what the scope arrays use (e.g. via `os.path.relpath`, which doesn't raise for the same class of cases, or by normalizing both sides consistently). At minimum, add a comment/TODO tying this fallback to the still-open scope-emission gap so the two aren't fixed independently and silently re-break each other. Also strengthen the existing test to cover an actual cross-drive/mount/symlink trigger, not just a same-filesystem sibling-directory case, and to assert the emitted path's correctness rather than just absence of a crash.

**How to verify**
Once the coverage-report producer emits `excluded`/`whole_file_claims`, add a test with a file outside `project_root` that is also listed in the report's `excluded` array (using the same relative-path form the producer emits) and assert the corresponding pairing is actually dropped — not just that no exception is raised.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+    This is the id→doc lookup the requirement side needs: a block cites an algo id, and the
+    requirement text lives in whichever artifact *defines* that id. Built from the registered
+    artifacts (``collect_artifacts_to_scan``); the first definition wins on the rare duplicate.
+    """
+    out: Dict[str, Path] = {}
+    artifacts, _sources = collect_artifacts_to_scan(ctx)
+    for artifact_path, _kind in artifacts:
+        for hit in scan_cpt_ids(artifact_path):
+            if hit.get("type") == "definition" and isinstance(hit.get("id"), str):
+                out.setdefault(hit["id"], artifact_path)
```
ainetx, line 53: ### Duplicate cpt definitions for the same algo id are silently discarded with no warning

**Severity:** Minor

**Problem**
This builds the id→doc lookup via `out.setdefault(hit["id"], artifact_path)` while iterating artifacts in a stable order. This makes resolution deterministic (first-registered artifact wins), which is good, but if two artifacts both declare a DEFINITION for the same algo id — a data-authoring mistake that can genuinely occur (a copy-pasted feature doc, or a doc moved without removing the old copy) — the second definition is dropped with no log line, warning, or any other signal that a collision occurred. Only the zero-definition case is tested; there is no test exercising the two-artifacts-same-id case at all.

**How to reproduce**
1. Author two feature docs, both declaring the same algo id as a DEFINITION.
2. Register both artifacts so both are scanned, with the first one registered first.
3. Call `_definition_map`.
4. Observe the mapping resolves only to the first artifact; the second artifact's definition (and its requirement text) is invisible to the semantic pass, with nothing in logs or output indicating the collision happened.

**Expected behavior**
A duplicate DEFINITION for the same id across artifacts is a spec-authoring problem worth surfacing — at minimum a warning noting the id and both paths, so a human can notice and fix the duplicate declaration; ideally a regression test locks in the "first wins, deterministically" behavior explicitly.

**Actual behavior**
The collision is silently absorbed; there is no signal in logs, in the returned mapping, or in any existing test that a second definition existed or was discarded.

```text
collect_artifacts_to_scan(ctx)
        |
        v
artifacts = [(feature-a.md, "feature"), (feature-b.md, "feature")]
        |
        v
for artifact_path, _kind in artifacts:
    for hit in scan_cpt_ids(artifact_path):
        if hit["type"] == "definition":
            out.setdefault(hit["id"], artifact_path)
        |
        |  feature-a.md defines cpt-studio-algo-example --> out[id] = feature-a.md
        |  feature-b.md ALSO defines cpt-studio-algo-example --> setdefault is a no-op, SILENTLY DROPPED
        v
mapping = {"cpt-studio-algo-example": feature-a.md}   (feature-b.md's definition invisible, no warning)
```

**Impact**
Low likelihood (requires an authoring mistake) but invisible when it happens: the semantic pass will quietly resolve requirement text from the wrong/stale doc for any code blocks tagged with that id, and nothing in the tool's output hints that two definitions existed.

**Suggested correction**
Add a warning in `_definition_map` when a duplicate DEFINITION id is encountered (naming the id and both competing paths), and add a test asserting both that the first-registered artifact wins deterministically and that the warning fires.

**How to verify**
Add a test with two artifacts both declaring the same algo id as a DEFINITION, assert `_definition_map` returns the first artifact's path, and assert a warning was logged naming both paths and the colliding id.

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx: ### Swallowed ValueError in path relativization has no logging

**Severity:** Minor

**Problem**
This `except ValueError:` block catches a failed `relative_to()` call and silently falls back to the absolute path, with no log call, no re-raise, and no other visible signal. This is inconsistent with the sibling `except Exception as exc:` handler in `_attach_semantic_section` (`spec_coverage.py`), which does log via `logger.warning(...)`.

**How to reproduce**
1. Call `_pairings_for_files` where one file resolves to a path outside `project_root` (e.g. a symlinked/mounted subtree, or a different drive).
2. `relative_to()` raises `ValueError`.
3. Execution enters the except block — no logger call occurs anywhere in this path.
4. Nothing in stdout, logs, or the returned pairings indicates relativization was attempted and failed for this file.

**Expected behavior**
The except block should log the fallback event so operators/maintainers have a visible trail when relativization silently degrades.

**Actual behavior**
The exception is caught and discarded with zero observability.

```text
project_root given, file OUTSIDE project_root
        |
        v
block_path.resolve().relative_to(project_root)  --raises ValueError-->
        |
        v
except ValueError:                    <-- caught, no logger call, no signal
        |
        v
block_path = code_file.path (absolute, degraded)
        |
        v
pairing constructed and returned as if nothing happened
```

**Impact**
Any future regression or unexpected trigger of this fallback will be completely invisible in logs, making it harder to diagnose reports of the semantic pass silently mis-scoping files.

**Suggested correction**
Add a log call in the except block, e.g. `logger.debug("semantic_coverage: %s is outside project_root %s; using absolute path", code_file.path, project_root)`.

**How to verify**
Add a unit test that constructs a code path outside a given project root, calls `_pairings_for_files`, and asserts (via `caplog`) that a log record is emitted for the fallback branch.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+    retrievable text. ``block_id`` is ``<algo>:<inst>`` so each block is individually identifiable
+    even though the requirement (and ``bm.id``) is shared across an algo's blocks.
+
+    ``path`` is emitted **project-relative POSIX** when ``project_root`` is given, matching the
+    coverage report's own scope arrays (``excluded`` / ``whole_file_claims``) — the code files
+    arrive absolute, so without this the scope reader would compare an absolute pairing path against
+    a relative report path and silently never match."""
+    pairings: List[eval_semantic.Pairing] = []
+    for code_path in files:
+        code_file, _errs = CodeFile.from_path(code_path)
```
ainetx: ### Parse errors from `CodeFile.from_path` are discarded instead of surfaced

**Severity:** Major

**Problem**
When `CodeFile.from_path(code_path)` returns `(None, errs)` for a file that fails to parse (e.g. an unclosed `@cpt-end` marker), the code does `if code_file is None: continue`. The `errs` return value — structured diagnostic information about why the file failed to parse — is bound and then dropped entirely. It is never logged, never attached to the semantic report's `unjudgeable` list, and never threaded through the output in any form.

**How to reproduce**
1. Create a source file with an unbalanced marker, e.g. `@cpt-begin:...` with no matching `@cpt-end`.
2. Include that file in a run of `spec-coverage --semantic`.
3. `CodeFile.from_path` returns `(None, errs)`.
4. The loop hits `continue` and moves on — no pairing, no unjudgeable entry, no log line references this file at all.
5. The existing test `test_pairings_skip_a_file_that_does_not_parse` asserts only that the bad file's blocks are absent — nothing about any diagnostic being recorded, because none is produced.

**Expected behavior**
A file that cannot be assessed should be visibly reported as unjudgeable (or via an equivalent explicit diagnostic channel), not silently vanish from the assessment with no trace — consistent with the module's own "never a silent covered" principle.

**Actual behavior**
A file that fails to parse is dropped with zero diagnostic trail; the returned semantic section makes no mention of it, and a reviewer has no way to learn a file was skipped due to a parse error.

```text
files = [bad.py (unclosed @cpt-end), good.py]
        |
        v
CodeFile.from_path(bad.py) --> (None, errs)     [errs: structured parse-failure info]
        |
        v
code_file is None --> continue                  [errs discarded]
        |
        v
pairings = [ ...only from good.py... ]
        |
        v
semantic section: {assessed, presumed_covered, unjudgeable, findings, ...}
        |
        v
reviewer sees NO mention of bad.py at all --
cannot distinguish "not applicable" from "silently failed to parse"
```

**Impact**
A file with a genuine, unnoticed marker-syntax bug will be silently excluded from scrutiny indefinitely, with the report implying full coverage of that codebase entry when it is not.

**Suggested correction**
Thread the parse errors into the report instead of discarding them — log a warning and either surface a `parse_failures` list or fold entries into `unjudgeable` with a distinct `parse_error` reason, so both the log stream and the JSON report reflect it.

**How to verify**
Extend the existing test to assert the parse failure is surfaced — via a logged warning, a returned diagnostics structure, or an `unjudgeable` entry with a `parse_error`-style reason — rather than asserting only that no exception is raised.

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-pairings
+def _pairings_for_files(files: Sequence[Path], definitions: Dict[str, Path],
+                        project_root: Optional[Path] = None) -> List[eval_semantic.Pairing]:
+    """One ``Pairing`` per marked block: code = the block's own lines, requirement = the algo
+    declaration resolved from the block's id, or ``None`` (unjudgeable) when the id declares no
+    retrievable text. ``block_id`` is ``<algo>:<inst>`` so each block is individually identifiable
+    even though the requirement (and ``bm.id``) is shared across an algo's blocks.
+
+    ``path`` is emitted **project-relative POSIX** when ``project_root`` is given, matching the
```
ainetx: ### Absolute local file path leaks into the external judge-model prompt when a file resolves outside project_root

**Severity:** Major

**Problem**
This function tries to make each block's path project-relative for scope matching. When the file resolves outside `project_root` (a symlinked/mounted/federated source — this project's own multi-repo workspace tooling explicitly supports such layouts), `relative_to()` raises `ValueError` and the fallback reassigns the absolute path (e.g. `/Users/alice/other-repo/mod.py`). This absolute string becomes `pairing.path`, which is not just used for internal scope comparisons — `build_semantic_request` in `eval_semantic.py` embeds it verbatim into the natural-language prompt handed to the model: `f"CODE ({pairing.path}:{pairing.start_line}):\n{_capped(pairing.code)}\n\n"`. When a real (non-stub) judge_fn is wired, this prompt is sent to whatever LLM backend the host supplies — disclosing the local OS username and directory layout to that external service, with no filtering, warning, or opt-out.

This is not a narrow, defensive dead branch: `_filter_ignored_files` in `spec_coverage.py` explicitly tolerates a codebase file whose resolved path falls outside `project_root` — it only logs a warning and keeps the file in scope rather than dropping it, so this fallback is reachable in ordinary operation.

**How to reproduce**
1. Configure a codebase entry (or workspace federation/symlink) that resolves to a file outside `project_root`.
2. Run `spec-coverage --semantic` with a real judge_fn wired.
3. The scope filter warns but keeps the file; the path-relativization fallback sets an absolute path.
4. If that block ranks as a weak link, the prompt builder embeds the absolute path into the text sent to the judge.

**Expected behavior**
No locally-identifying filesystem information should be sent to an external judge model — paths handed to the prompt-building layer should always be relative/redacted.

**Actual behavior**
The raw absolute path is silently embedded in the model prompt whenever a file legitimately resolves outside `project_root` and ranks as a weak link, with no logging or warning.

```text
codebase entry resolves OUTSIDE project_root (symlink / mount / federated source)
        |
        v
_filter_ignored_files (spec_coverage.py): warns, but does NOT drop the file
        |
        v
_pairings_for_files: block_path.resolve().relative_to(project_root)  --raises ValueError-->
        |
        v
except ValueError: block_path = code_file.path      (ABSOLUTE, e.g. /Users/alice/other-repo/mod.py)
        |
        v
pairing.path = absolute path string
        |
        v
weak-link ranking selects this pairing for judging
        |
        v
build_semantic_request(): f"CODE ({pairing.path}:{start_line}):..."
        |
        v
prompt text (containing absolute path) sent verbatim to judge_fn (real/external LLM backend)
        |
        v
local OS username / directory structure disclosed to a third-party service
```

**Impact**
Any team running `--semantic` with a real judge wired against a workspace that legitimately includes files outside `project_root` will transmit local filesystem paths — including OS usernames and directory layout — to whatever external judge model is configured, with no logging or consent step.

**Suggested correction**
Never let an unrelativized absolute path reach judge-facing text: derive a path safe for external consumption in the fallback (e.g. `os.path.relpath`, or a stable synthetic identifier), and/or have the prompt builder strip/redact any non-project-relative path as a defense-in-depth backstop.

**How to verify**
Add a test that drives a pairing for a file outside `project_root`, ranks it as a weak link, builds the judge request, and asserts the resulting prompt contains no absolute filesystem path.

skills/studio/scripts/studio/commands/spec_coverage.py
ainetx, line 78: ### `--semantic` is never exercised end-to-end by CI, only through mocked unit tests

**Severity:** Major

**Problem**
This flag wires in a real import chain (`utils.semantic_coverage.run_semantic_pass` → `utils.eval_semantic.assess`) and scans the real project's marked code blocks and artifacts. Neither of the two places that invoke `spec-coverage` in CI ever passes this flag: the `spec-coverage` Makefile target and the CI job that calls it. The only place `--semantic` is ever invoked is the test suite, and every test there either monkeypatches `_definition_map`/`run_semantic_pass` directly or patches `context.get_context`, so the real definition-scanning, real artifact-collection, and real `eval_semantic` import path are never actually run against this repository's real project tree by any CI step.

**How to reproduce**
1. Read the `spec-coverage` Makefile target: `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` — no `--semantic` anywhere.
2. Read the CI workflow's `spec-coverage` job — it just runs `make spec-coverage`, i.e. exactly the invocation above.
3. Search the test suite for any `--semantic` invocation that does not patch `run_semantic_pass`, `_definition_map`, or `context.get_context` — none exists.

**Expected behavior**
Since `--semantic` performs a real import of `eval_semantic`, a real scan of the project's registered artifacts, and a real walk of the project's marked code blocks, some CI step should exercise that exact path against the real repository tree at least once, so a regression in any of those real code paths would be visible somewhere.

**Actual behavior**
No CI step ever passes `--semantic`. The only exercised path is the unit-test suite, which substitutes mocks for exactly the pieces that would need to be real to catch a wiring regression. Compounding this, the exception handler around the semantic pass catches any exception — including an `ImportError` — and downgrades it to a warning log line plus an advisory error section, so even a manual run would silently report a degraded advisory section rather than fail loudly.

```text
CI: make spec-coverage
        |
        v
Makefile target: spec-coverage --system studio --min-coverage 90 ...   (no --semantic)
        |
        v
CI workflow "spec-coverage" job: run: make spec-coverage   (same, no --semantic)
        |
        v
   real eval_semantic import chain, real artifact scan  --  NEVER EXERCISED BY CI


tests/test_semantic_coverage.py: --semantic invocations
        |
        v
   monkeypatch _definition_map / run_semantic_pass / get_context
        |
        v
   real code path replaced by a mock  -->  regression in the real path is invisible to CI
```

**Impact**
A regression that breaks the real `--semantic` wiring — a broken import in `eval_semantic`, a definition-map that silently resolves nothing against the real docs tree, a scope-matching bug against the real coverage report shape — would ship and persist indefinitely with zero CI signal. Because the broad exception handler degrades any such failure to a warning log line, even a maintainer running `--semantic` manually would likely not notice unless they read stderr closely.

**Suggested correction**
Add a CI step (or extend the existing `spec-coverage` job) that runs `spec-coverage --semantic --system studio` against the real repository tree at least once, asserting the command exits successfully and that the semantic section does not contain an `error` key. This does not need to assert on specific findings — it only needs to prove the real import chain and real artifact/code scan succeed end-to-end.

**How to verify**
Add the suggested CI step, then temporarily break `eval_semantic`'s import or the definition-map's artifact-scanning call, and confirm the new step fails; then restore the code and confirm it passes again.

architecture/features/spec-coverage.md
SanjeevSolanki: Fixed. The **Output** line now lists all seven success keys (`assessed`/`presumed_covered`/`unjudgeable[]`/`findings[]`/`skipped_excluded`/`schema_version`/`advisory`) and notes the degraded `{advisory: true, error: <message>}` shape emitted when the pass raises.

tests/test_semantic_coverage.py
SanjeevSolanki: Fixed. Added `test_semantic_never_gates_a_structurally_failing_run`: it forces a structural FAIL (exit 2, status FAIL) via an uncovered function + `--min-coverage 100`, then asserts `--semantic` leaves exit code and status identical to the plain run while still attaching the advisory section. Locks the never-gates invariant on the exit-2 path (which the ordering already guaranteed).

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: You're right it's inert on real reports today — and that's intentional and safe: `coverage_scope` degrades to an empty scope (judge everything, prioritise nothing), and the `excluded[]`/`whole_file_claims[]` schema is frozen bilaterally with the producer, whose catch-up is tracked in #107. I've softened the module docstring to say the scoping activates once the producer emits those keys (citing #107), so it no longer reads as active-today. Deliberately NOT retro-fitting the consumer to today's `flagged_files` shape, since that shape is slated to change under the frozen schema.

skills/studio/scripts/studio/commands/spec_coverage.py
SanjeevSolanki: Fixed. The `--semantic` help now names the verdicts (covered/partial/wrong/unjudgeable per marked block), states it never gates status/exit, and points at `architecture/features/spec-coverage.md`.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed. Added a one-line rubric for covered/partial/wrong directly into the judge prompt in `build_semantic_request`, so the out-of-tree judge gets the classification definitions rather than only the verdict names.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed. `_pairings_for_files` now iterates `dict.fromkeys(files)` (order-preserving dedup), so a file registered more than once no longer double-counts its blocks. Added `test_pairings_dedup_duplicate_file_registrations`.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Acknowledged — the abs-vs-rel scope mismatch is real but currently inert (the producer doesn't emit `excluded[]`/`whole_file_claims[]` until #107) and the pass is advisory-only. I did close the concrete leak half now: the fallback emits a **relative** path (never absolute) via `os.path.relpath`, with a bare-name fallback for the Windows cross-drive case, so no absolute local path can reach the report or the judge prompt (see the reply on the path-leak thread). The deeper scope-matching lands with #107.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed. `_definition_map` now logs a warning when the same cpt id is defined in two different artifacts (keeping the first), mirroring the pattern in `list_ids`. Added `test_definition_map_warns_on_duplicate_cpt_definition`.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed. The relativization fallback now logs at `debug` when a file is outside `project_root`.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed. When `CodeFile.from_path` returns `None`, the pass now logs a warning with the parse errors before skipping, rather than discarding them silently.

skills/studio/scripts/studio/utils/semantic_coverage.py
SanjeevSolanki: Fixed — good catch, this was a real leak (`pairing.path` is interpolated into the judge prompt at `eval_semantic.py`). The out-of-project fallback no longer emits an absolute path: it uses `os.path.relpath` (with a bare-name fallback for the Windows cross-drive case), so `pairing.path` is always relative. Added a test asserting the path is never absolute, plus a cross-drive-failure test proving one off-drive file can't sink the whole advisory section.

skills/studio/scripts/studio/commands/spec_coverage.py
SanjeevSolanki: Fixed. Added `test_semantic_pass_runs_end_to_end_unmocked`, which runs `cmd_spec_coverage(['--semantic'])` with `run_semantic_pass` NOT mocked — real `_definition_map` + `_pairings_for_files` + `assess` over a real ctx, asserting a real (non-error) section is produced. Closes the mocked-only gap for the pass logic.

SanjeevSolanki:
Thanks for the thorough review, @ainetx — sharp catches, several on the parts I'd just changed. All 12 are addressed in the latest push; per-thread replies have the specifics. Summary:

**Code**
- Dedup files before pairing (`dict.fromkeys`) so a doubly-registered file no longer double-counts its blocks.
- The out-of-project path fallback now emits a **relative** path (never absolute) — `pairing.path` is interpolated into the judge prompt, so this closes a real absolute-local-path leak. A bare-name fallback covers the Windows cross-drive case so one off-drive file can't sink the whole advisory section.

**Docs / prompt**
- Feature-doc output now lists all seven success keys plus the degraded `{advisory, error}` shape.
- `--semantic` help names the verdict vocabulary and points at the feature doc.
- The judge prompt now carries a one-line covered/partial/wrong rubric.
- The engine docstring notes the `excluded[]`/`whole_file_claims[]` scoping activates once the producer emits those keys (#107).

**Logging**
- Warn on duplicate cpt definitions (keeping the first), warn on unparseable files before skipping, and debug-log the outside-root path fallback.

**Tests (+6)**
- `--semantic` never-gates on a **structurally-failing** run (exit 2), an **unmocked** end-to-end pass, dedup, duplicate-def warning, path-never-absolute, and the cross-drive fallback.

**Two by design (replies on those threads):** the scope arrays being empty on today's real reports is intentional and safe (empty-scope = judge everything; the schema is frozen with the producer, catch-up tracked in #107); the deeper scope-matching lands with #107 rather than coupling to today's shape.

All gates green: `cfs validate` 223/223, spec-coverage thresholds met, pylint, vulture, full suite, per-file coverage. (One note: the rubric + docstring touches reached `eval_semantic.py` since they were raised in this PR's context.)

SanjeevSolanki:
Also addressed CodeRabbit's outside-diff note on `architecture/features/spec-coverage.md` (line 76): the primary invocation syntax in step 1 now includes `[--semantic]` alongside the existing flags.

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx: Acknowledged — deferring the scope-activation to #107 with an honest docstring is a reasonable interim call given it's advisory-only and the current behavior (judge everything, prioritize nothing) is the safe default. Leaving this thread open until #107 lands.

ainetx (approved):
Deep Review complete — 56 checks across 6 phases (22 required project-config rules + 34 LLM-suggested), 12 findings posted as inline comments.

11/12 verified fixed on 1e8ed1a and resolved:
- Doc output shape fully documented (incl. error shape)
- `--semantic` now tested against a structurally-failing run (exit 2)
- `--help` names the verdict categories
- Judge prompt now carries a covered/partial/wrong rubric
- Duplicate file registrations deduped
- Path-relativization fallback no longer leaks an absolute path (fixes both the scope-matching bug and the judge-prompt PII leak)
- Duplicate cpt definitions now logged, not silently dropped
- Swallowed exception now logged
- Parse failures now logged instead of silently discarded
- New unmocked end-to-end test closes the "only mocked tests" CI blind spot

1 remains open by design: `excluded`/`whole_file_claims` scoping doesn't activate on real reports yet — deliberately deferred to #107 with a safe default (judge everything, prioritize nothing) and an honest docstring. Agreed this is a reasonable interim call.

All CI green, no Critical or unaddressed Major findings remain. Approving.

PR #123, opened 2026-09-01
test(eval): add example scenarios so the harness ships with something to score

SanjeevSolanki:
The eval harness reads scenarios from `<project>/eval`, and none were included. Running
`eval` out of the box therefore discovered nothing, scored nothing, and reported success —
indistinguishable from a healthy run, and the same shape of false-green the harness exists
to catch.

This adds two scenarios so the command ships with something to score, and so a reader can
see what a scenario looks like without reverse-engineering the loader.

- `compliant-two-phase` — a well-formed two-phase run (`expect = "compliant"`).
- `non-compliant-numbering-gap` — phases numbered 1 and 3, so exactly one check
  (`numbering-contiguous-from-1`) fails (`expect = "non_compliant"`).

The broken one is deliberate. Without a run that is known to be bad, a green result proves
nothing about a checker: one that never fails is indistinguishable from one that cannot.
`eval/README.md` records which check it is meant to trip, so a later reader can tell a
fixture from a defect.

**Verified before pushing**

| | result |
|---|---|
| `compliant-two-phase` | PASS, 100%, 14/14 structural checks, no findings |
| `non-compliant-numbering-gap` | FAIL, 92.86%, one finding: `numbering-contiguous-from-1: numbers=[1, 3]` |
| plain run | exit **0** — the bad run is reported, not gated |
| `--check` | exit **2** — gates only when asked |
| `--calibrate` | reference-stub, 0 gold-backed, accuracy n/a (no gold ships; it says so rather than inventing a number) |

No engine changes — test data and a README only.

## Summary by CodeRabbit

- **New Features**
  - Added evaluation scenarios for compliant and intentionally non-compliant multi-phase workflows.
  - Added a health-check endpoint design and implementation exercise, including response-shape requirements and tests for healthy and degraded states.
  - Added validation coverage for detecting gaps in phase numbering.

- **Documentation**
  - Documented evaluation scenario structure, phase metadata, expected content, command usage, and optional enforcement checks.

### DIFF
--- a/eval/README.md
+++ b/eval/README.md
@@ -1,0 +1,32 @@
+# Example scenarios for `cfs eval`
+
+Two runs the eval harness can score, so the command has something to work on out of the
+box — and so a reader can see what a scenario looks like without reverse-engineering the
+loader.
+
+| Scenario | `expect` | Purpose |
+|---|---|---|
+| `compliant-two-phase` | `compliant` | A well-formed two-phase run. Every structural check should pass. |
+| `non-compliant-numbering-gap` | `non_compliant` | **Deliberately broken.** Its phases are numbered 1 and 3, so `numbering-contiguous` fails. Everything else about it is valid. |
+
+The second one is deliberate, not a mistake to be tidied up: without a run that is known
+to be bad, a green result proves nothing — a checker that never fails is indistinguishable
+from one that cannot fail. If you change it, keep exactly one check failing and say which
+one here.
+
+## Layout
+
+```
+eval/<scenario>/
+  scenario.toml        [scenario] id · workflow · run_dir · expect
+  run/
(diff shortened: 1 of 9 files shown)


SanjeevSolanki:
Closing — no review needed.

These scenarios were built as fixtures for our own QA verification of the eval
harness, and they do that job perfectly well from the branch. They don't need to
land upstream to be useful, and I'd rather not add review load for test data that
isn't blocking anything.

The underlying observation stands and I'll raise it separately if it's worth it:
with no scenarios present, the harness discovers nothing, scores nothing and
reports success — which reads the same as a healthy run. If you'd like example
scenarios shipped in-tree at some point, this branch is ready and green.

Issue #134, opened 2026-09-03
JIT-retrieval cascade: auto-triggered OKF build and multi-signal fusion from #104 not implemented


tkcoding:
## Summary

#104 was closed by pointing at PRs #108–#111 as delivering the Part 2 retrieval work in full, including [@Oleg67's four suggestions](https://github.com/constructorfabric/studio/issues/104#issuecomment-5470230878). Verified against `main` @ `6e0bcad3`: only 2 of the 4 actually landed. This issue tracks the other two, plus two small hardening fixes from #111's own review that fell through a merge race.

### Confirmed done — no action needed here

- **Suggestion #2 (anchors into source).** Each OKF concept file's frontmatter embeds `{source_path}#L{line_start}-L{line_end}` (`okf.py:421`), and the cascade matches Tier-1 candidates back to OKF entries by `line_start` (`cascade.py:162`). Summary locates, source verifies, as proposed.
- **Suggestion #3 (never mix stale summary with source).** `get_okf_status` reports missing/stale/current per section, and `route_tier2` refuses to recommend a stale/missing concept file (`cascade.py:151-184`) — it downgrades to baseline instead. One nuance: it falls back to a full baseline read rather than reusing the heading-nav candidate Tier 1 already computed (row 4 case), which is more conservative than proposed but not incorrect. Minor, not blocking this issue.

---

## Problem 1 — Suggestion #1 (lazy, auto-triggered OKF build) is not implemented

`route_tier2`'s break-even math (`cascade.py:117-127`) only runs off `expected_future_queries` — a number a human types in via `--expected-future-queries` on the CLI (`commands/cascade.py:53-55`). Nothing tracks real per-document query volume, and nothing ever triggers a build automatically — the exact "policy, not threshold" problem the suggestion was trying to remove.

It's also not currently buildable from existing data:
- `decision_log.record_read` redacts the `target` field (`decision_log.py:372`).
- `summarize_reads` aggregates logged reads only by `method`, never by document (`decision_log.py:486-490`).

So today nobody — human or automated — can answer "how many times has this specific document been queried," which is the input suggestion #1 depends on.

**Proposed fix:**
- Add a per-document (etag-keyed, not raw-path) query counter, incremented on every `route_query` call — likely living alongside `doc_index`'s existing etag-keyed cache rather than the redacted decision log.
[...]

PR #135, opened 2026-09-03
fix(cascade,okf): resolve PR #111 round-3 CodeRabbit findings that missed main

tkcoding:
## Summary
- Addresses Problem 3 of #134: two CodeRabbit-flagged fixes from #111's own review that never landed on `main` because the fix commit was pushed ~49 minutes after a maintainer merged #111.
- `route_tier1`/`route_query` now validate `margin_threshold` themselves (`_validate_margin_threshold`), rejecting non-finite or non-positive values — `commands/cascade.py`'s CLI-level validator only guarded that one entry point, so any direct Python caller could pass `0`/negative/`nan`/`inf` and defeat the "no finite margin is yet proven safe" design basis documented in `cascade.py`'s own module docstring.
- `_concept_file_is_valid` now also requires the closing frontmatter delimiter, not just the opening one — a concept file truncated right after `"---\n"` previously passed validation and could be handed to Tier 2 as a usable OKF summary.
- Cherry-picked from this session's pre-merge branch (original commit `33a37620`), applied cleanly against current `main`, no conflicts.

## Test plan
- [x] `pytest tests/test_cascade.py tests/test_okf.py -q` — 73 passed

## Update: review round (ainetx) — 3 Minor, all fixed
- **Non-numeric `margin_threshold` raised `TypeError` instead of the documented `ValueError`.** `_validate_margin_threshold` called `math.isfinite()` before checking the input was even numeric, so a direct Python caller passing a string (or any other non-numeric) hit an unhandled `TypeError`. **Fix:** added an `isinstance` guard (bool excluded, since it subclasses int but isn't a meaningful threshold), plus a test proving a non-numeric/bool input now raises the documented `ValueError`.
- **`route_query`'s invalid-threshold test coverage was a single hard-coded value while `route_tier1`'s was a full parametrized matrix** (`[0, -1, nan, inf]`) — the two share the same validator via the same call path, so a future refactor that decoupled them could slip through unnoticed. **Fix:** mirrored the same matrix (now including the non-numeric/bool cases above) onto the `route_query` test.
- **Both tests asserted only a substring match on the error message** (`match="margin_threshold"`), which would still pass even if the message lost its actual explanation. **Fix:** tightened both to a full-message regex, and added an accept-path test (a large finite threshold must not itself be rejected) for both `route_tier1` and `route_query`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid margin thresholds—including zero, negative, infinite, non-numeric, and boolean values—are now rejected consistently with a clear validation error.
  * Valid finite thresholds, including very small and large values, continue to route requests as expected.
  * Tier 1 routing now evaluates heading navigation and TF-IDF together, resolving clear matches and escalating ambiguous results appropriately.
  * Incomplete concept files—including missing frontmatter fields, closing delimiters, or non-blank content—are now correctly identified as invalid or missing.
  * Duplicate Tier 2 escalations are now avoided, with graceful handling when escalation resources are unavailable.

### DIFF
--- a/skills/studio/scripts/studio/commands/cascade.py
+++ b/skills/studio/scripts/studio/commands/cascade.py
@@ -8,10 +8,9 @@ Thin CLI wrapper around ``studio.utils.cascade``.
 """
 
 import argparse
-import math
 from typing import List
 
-from ..utils.cascade import _TIER2_BREAK_EVEN_ESCALATIONS, route_query
+from ..utils.cascade import _TIER2_BREAK_EVEN_ESCALATIONS, _check_margin_threshold_value, route_query
 from ..utils.doc_index import _MAX_ESCALATION_KEY_LENGTH
 from ..utils.ui import ui
 
@@ -28,14 +27,25 @@ def _margin_threshold_arg(value: str) -> float:
     result, defeating that safety margin entirely; only a genuine positive
     finite number is accepted, mirroring ``commands/eval.py``'s
     ``_compliance_arg`` validator for the same class of hazard.
+
+    Parses the argparse string to ``float`` first (a string input like
+    ``"0.5"`` is a legitimate CLI value -- unlike the direct Python API in
+    ``utils/cascade.py``'s :func:`~studio.utils.cascade._validate_margin_threshold`,
+    which requires an already-numeric value and rejects a string outright),
+    then delegates the finite/positive check to the same shared
+    ``_check_margin_threshold_value`` helper :func:`_validate_margin_threshold`
+    uses, so both entry points enforce one identical policy instead of two
+    independently drifting copies of it.
     """
     try:
         parsed = float(value)
     except ValueError as exc:
         raise argparse.ArgumentTypeError(f"invalid float value: {value!r}") from exc
-    if not math.isfinite(parsed) or parsed <= 0:
+    try:
+        _check_margin_threshold_value(parsed)
+    except ValueError as exc:
         raise argparse.ArgumentTypeError(
-            f"--margin-threshold must be a finite number > 0, got {value!r}")
+            f"--margin-threshold must be a finite number > 0, got {value!r}") from exc
     return parsed
 
 
--- a/skills/studio/scripts/studio/utils/cascade.py
+++ b/skills/studio/scripts/studio/utils/cascade.py
@@ -126,6 +126,22 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
+    """Reject a non-finite or non-positive ``margin_threshold`` before it can
+    reach row 4's comparison.
+
+    ``commands/cascade.py``'s ``_margin_threshold_arg`` enforces this same
+    rule for the CLI, but a direct Python caller of :func:`route_tier1`/
+    :func:`route_query` bypasses argparse entirely -- without this check
+    here too, a non-positive or non-finite threshold would make
+    ``tfidf_result["margin"] >= margin_threshold`` fire on virtually any
+    finite margin, silently defeating the "no finite value is yet proven
+    safe" design basis this module's own docstring documents.
+    """
+    if margin_threshold is not None and not (math.isfinite(margin_threshold) and margin_threshold > 0):
+        raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
+
+
 # @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-tier1
 def _route_tier1_heading_nav_miss(
     tfidf_ranked: List[Dict[str, Any]], has_tfidf_signal: bool, tfidf_result: Dict[str, Any],
@@ -177,6 +193,7 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
     assumption (validated against a single ~166-page, ~9-section document,
     deliberately no cap on section count or file size) this cost relies on.
     """
+    _validate_margin_threshold(margin_threshold)
     nav_first_match = find_sections(path, query)["first_match"]
     # Unconditional, even when heading-nav already found a hit: fusing the
     # two signals (this function's whole purpose, see the module docstring)
--- a/skills/studio/scripts/studio/utils/cascade.py
+++ b/skills/studio/scripts/studio/utils/cascade.py
@@ -138,7 +138,22 @@ def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
     finite margin, silently defeating the "no finite value is yet proven
     safe" design basis this module's own docstring documents.
     """
-    if margin_threshold is not None and not (math.isfinite(margin_threshold) and margin_threshold > 0):
+    if margin_threshold is None:
+        return
+    # isinstance guards math.isfinite() itself: it raises TypeError for any
+    # non-numeric argument, which would propagate out before the ValueError
+    # below ever runs -- defeating this function's own documented contract
+    # for exactly the direct-Python-caller case its docstring exists for
+    # (the CLI's argparse type already coerces to float before this is
+    # ever reached, so only a direct caller can hit this). bool is
+    # deliberately excluded even though it subclasses int: True/False are
+    # not meaningful margin thresholds, and isfinite(True) would otherwise
+    # silently accept one.
+    if (
+        isinstance(margin_threshold, bool)
+        or not isinstance(margin_threshold, (int, float))
+        or not (math.isfinite(margin_threshold) and margin_threshold > 0)
+    ):
         raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
 
 
--- a/skills/studio/scripts/studio/utils/cascade.py
+++ b/skills/studio/scripts/studio/utils/cascade.py
@@ -67,6 +67,7 @@ option that fits what this codebase can actually guarantee.
 
 from __future__ import annotations
 
+import logging
 import math
 from pathlib import Path
 from typing import Any, Dict, List, Optional, cast
@@ -76,6 +77,8 @@ from .heading_nav import find_sections
 from .okf import get_okf_status
 from .read_gate import check_gate
 from .tfidf import score_sections
+
+logger = logging.getLogger(__name__)
 
 #: Real, measured per-query token rates (see the design session's findings).
 _OKF_BUILD_COST_TOKENS = 301_187
@@ -126,35 +129,61 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def _check_margin_threshold_value(value: float) -> None:
+    """Raise ``ValueError`` unless ``value`` is a finite number > 0.
+
+    The single source of truth for the "finite and positive" rule, shared
+    by :func:`_validate_margin_threshold` (the direct Python API, which
+    requires ``value`` to already be numeric -- no string coercion) and
+    ``commands/cascade.py``'s ``_margin_threshold_arg`` (the CLI, which
+    parses the argparse string to ``float`` *first* and then delegates the
+    finite/positive check here, translating a ``ValueError`` into
+    ``argparse.ArgumentTypeError``). Before this was extracted, both call
+    sites duplicated this exact check independently -- accepting different
+    input domains (the CLI coerced ``"0.5"`` to ``0.5`` and accepted it,
+    while this function rejects a bare string outright) was an accident of
+    that duplication, not an intentional difference in policy, and the two
+    copies could silently drift further apart on any future edit to just
+    one of them.
+
+    isinstance guards ``math.isfinite()`` itself: it raises ``TypeError``
+    for any non-numeric argument, which would propagate out before the
+    ``ValueError`` below ever runs. bool is deliberately excluded even
+    though it subclasses int: True/False are not meaningful margin
+    thresholds, and ``isfinite(True)`` would otherwise silently accept one.
+    """
+    if (
+        isinstance(value, bool)
+        or not isinstance(value, (int, float))
+        or not (math.isfinite(value) and value > 0)
+    ):
+        raise ValueError(f"margin_threshold must be a finite number > 0, got {value!r}")
+
+
 def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
     """Reject a non-finite or non-positive ``margin_threshold`` before it can
     reach row 7's comparison.
 
     ``commands/cascade.py``'s ``_margin_threshold_arg`` enforces this same
-    rule for the CLI, but a direct Python caller of :func:`route_tier1`/
-    :func:`route_query` bypasses argparse entirely -- without this check
-    here too, a non-positive or non-finite threshold would make
+    rule for the CLI (via the shared :func:`_check_margin_threshold_value`),
+    but a direct Python caller of :func:`route_tier1`/:func:`route_query`
+    bypasses argparse entirely -- without this check here too, a
+    non-positive or non-finite threshold would make
     ``tfidf_result["margin"] >= margin_threshold`` fire on virtually any
     finite margin, silently defeating the "no finite value is yet proven
     safe" design basis this module's own docstring documents.
+
+    A rejection is logged (not just raised) so a direct Python caller --
+    which has no argparse error surface of its own to fall back on -- still
+    leaves a structured trace of what was rejected and why.
     """
     if margin_threshold is None:
         return
-    # isinstance guards math.isfinite() itself: it raises TypeError for any
-    # non-numeric argument, which would propagate out before the ValueError
-    # below ever runs -- defeating this function's own documented contract
-    # for exactly the direct-Python-caller case its docstring exists for
-    # (the CLI's argparse type already coerces to float before this is
-    # ever reached, so only a direct caller can hit this). bool is
-    # deliberately excluded even though it subclasses int: True/False are
-    # not meaningful margin thresholds, and isfinite(True) would otherwise
-    # silently accept one.
-    if (
-        isinstance(margin_threshold, bool)
-        or not isinstance(margin_threshold, (int, float))
-        or not (math.isfinite(margin_threshold) and margin_threshold > 0)
-    ):
-        raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
+    try:
+        _check_margin_threshold_value(margin_threshold)
+    except ValueError:
+        logger.warning("rejected invalid margin_threshold: %r", margin_threshold)
+        raise
 
 
 # @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-tier1
--- a/skills/studio/scripts/studio/utils/cascade.py
+++ b/skills/studio/scripts/studio/utils/cascade.py
@@ -128,7 +128,7 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
 
 def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
     """Reject a non-finite or non-positive ``margin_threshold`` before it can
-    reach row 4's comparison.
+    reach row 7's comparison.
 
     ``commands/cascade.py``'s ``_margin_threshold_arg`` enforces this same
     rule for the CLI, but a direct Python caller of :func:`route_tier1`/
@@ -207,6 +207,13 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
     oversight -- see ``tfidf.py``'s own module docstring for the scale
     assumption (validated against a single ~166-page, ~9-section document,
     deliberately no cap on section count or file size) this cost relies on.
+
+    Raises ``ValueError`` if ``margin_threshold`` is not ``None`` and not a
+    finite number > 0 (see :func:`_validate_margin_threshold`). This is
+    checked unconditionally as the very first step, before ``nav_first_match``
+    is even computed -- so it raises even for a query that would otherwise
+    hit row 1 (``no_signal_from_either_method``), which never reads
+    ``margin_threshold`` at all.
     """
     _validate_margin_threshold(margin_threshold)
     nav_first_match = find_sections(path, query)["first_match"]
@@ -431,6 +438,10 @@ def route_query(
     query attempt, so a caller retrying this same query after a transient
     failure or timeout can pass the same key again and not double-count
     the Tier-2 escalation (constructorfabric/studio#136).
+
+    Raises ``ValueError`` for an invalid ``margin_threshold`` -- this calls
+    :func:`route_tier1` first, which validates it unconditionally before
+    doing anything else (see that function's own docstring).
     """
     tier1 = route_tier1(path, query, margin_threshold=margin_threshold)
     result: Dict[str, Any] = {"query": query, **tier1}
--- a/skills/studio/scripts/studio/utils/okf.py
+++ b/skills/studio/scripts/studio/utils/okf.py
@@ -292,19 +292,21 @@ def _match_sections_by_hash(
 
 def _concept_file_is_valid(concept_path: Path) -> bool:
     """Minimal content-validity check for a concept file already confirmed
-    to exist on disk: real content always opens with the YAML frontmatter
-    block :func:`_build_frontmatter` writes. A physical-presence check
-    alone (``is_file()``) can't tell a genuine concept file from one
-    truncated, emptied, or corrupted after the fact -- this catches that
-    without needing full YAML parsing, which is more than this check needs
-    to answer "is there real content here at all".
+    to exist on disk: real content always opens *and closes* the YAML
+    frontmatter block :func:`_build_frontmatter` writes. A physical-presence
+    check alone (``is_file()``) can't tell a genuine concept file from one
+    truncated, emptied, or corrupted after the fact -- checking only the
+    opening ``---`` doesn't either, since a file truncated right after it
+    still passes that alone; requiring the closing delimiter too catches
+    that without needing full YAML parsing, which is more than this check
+    needs to answer "is there real content here at all".
     """
     try:
         content = concept_path.read_text(encoding="utf-8")
     except (OSError, UnicodeDecodeError) as exc:
         logger.debug("okf concept file unreadable at %s: %s", concept_path, exc)
         return False
-    return content.startswith("---\n")
+    return content.startswith("---\n") and "\n---\n" in content[4:]
 
 
 def _resolve_section_status(
--- a/skills/studio/scripts/studio/utils/okf.py
+++ b/skills/studio/scripts/studio/utils/okf.py
@@ -290,23 +290,51 @@ def _match_sections_by_hash(
     return matched_by_position
 
 
+#: Every frontmatter key :func:`_build_frontmatter` always writes. A file
+#: with valid opening/closing delimiters but missing one of these (or with
+#: nothing but whitespace for its value) was never actually produced by
+#: this module's own writer -- truncated, hand-edited, or corrupted instead.
+_REQUIRED_FRONTMATTER_FIELDS = ("title", "description", "resource", "generated")
+
+
 def _concept_file_is_valid(concept_path: Path) -> bool:
-    """Minimal content-validity check for a concept file already confirmed
-    to exist on disk: real content always opens *and closes* the YAML
-    frontmatter block :func:`_build_frontmatter` writes. A physical-presence
-    check alone (``is_file()``) can't tell a genuine concept file from one
-    truncated, emptied, or corrupted after the fact -- checking only the
-    opening ``---`` doesn't either, since a file truncated right after it
-    still passes that alone; requiring the closing delimiter too catches
-    that without needing full YAML parsing, which is more than this check
-    needs to answer "is there real content here at all".
+    """Content-validity check for a concept file already confirmed to exist
+    on disk: real content always opens *and closes* the YAML frontmatter
+    block :func:`_build_frontmatter` writes, that frontmatter always carries
+    every one of :data:`_REQUIRED_FRONTMATTER_FIELDS` with a non-blank
+    value, and there's always a non-blank body after it.
+
+    A physical-presence check alone (``is_file()``) can't tell a genuine
+    concept file from one truncated, emptied, or corrupted after the fact.
+    Checking only the delimiters doesn't fully catch it either: a file
+    truncated right after the opening ``---`` fails that, but a file with
+    *both* delimiters present and otherwise-empty or partial frontmatter
+    (say, a hand-edited file missing ``description``, or one truncated
+    right at the closing ``---`` with no body at all) would still pass a
+    delimiters-only check while never having been a real generated concept
+    file. Parsing each field with a simple per-line regex (rather than a
+    full YAML parser this codebase doesn't otherwise depend on for this)
+    is enough to answer "is there real, complete content here" without
+    pulling in a heavier dependency.
     """
     try:
         content = concept_path.read_text(encoding="utf-8")
     except (OSError, UnicodeDecodeError) as exc:
         logger.debug("okf concept file unreadable at %s: %s", concept_path, exc)
         return False
-    return content.startswith("---\n") and "\n---\n" in content[4:]
+    if not content.startswith("---\n"):
+        return False
+    closing = content.find("\n---\n", 4)
+    if closing == -1:
+        return False
+    frontmatter = content[4:closing]
+    body = content[closing + len("\n---\n"):]
+    if not body.strip():
+        return False
+    return all(
+        re.search(rf"(?m)^{field}:[ \t]*\S", frontmatter) is not None
+        for field in _REQUIRED_FRONTMATTER_FIELDS
+    )
 
 
 def _resolve_section_status(
--- a/tests/test_cascade.py
+++ b/tests/test_cascade.py
@@ -225,6 +225,26 @@ class TestRouteTier1:
         assert result["tier"] == "resolved"
         assert result["reason"] == "heading_nav_tfidf_agree_large_margin"
 
+    @pytest.mark.parametrize("bad_threshold", [0, -1, float("nan"), float("inf")])
+    def test_margin_threshold_rejects_invalid_values_at_the_callable_api(
+        self, tmp_path: Path, monkeypatch, bad_threshold,
+    ):
+        """A direct Python caller bypasses commands/cascade.py's argparse
+        validation entirely -- without a check here too, a non-positive or
+        non-finite threshold would make the row-4 margin comparison fire on
+        virtually any finite margin, defeating the "no finite value is yet
+        proven safe" design basis."""
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
+        with pytest.raises(ValueError, match="margin_threshold"):
+            route_tier1(f, "widget", margin_threshold=bad_threshold)
+
+    def test_route_query_also_rejects_an_invalid_margin_threshold(self, tmp_path: Path, monkeypatch):
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
+        with pytest.raises(ValueError, match="margin_threshold"):
+            route_query(f, "widget", margin_threshold=-1.0)
+
     def test_route_tier1_never_touches_okf(self, tmp_path: Path, monkeypatch):
         """The module docstring guarantees route_tier1 stays free and
         deterministic by never touching the OKF bundle, but nothing
--- a/tests/test_cascade.py
+++ b/tests/test_cascade.py
@@ -5,14 +5,16 @@ See constructorfabric/studio#104.
 
 from __future__ import annotations
 
+import argparse
 import json
+import logging
 import re
 from pathlib import Path
 from unittest import mock
 
 import pytest
 
-from studio.commands.cascade import cmd_retrieve
+from studio.commands.cascade import _margin_threshold_arg, cmd_retrieve
 from studio.utils import cascade
 from studio.utils.cascade import route_query, route_tier1, route_tier2
 from studio.utils.doc_index import get_or_build_doc_index
@@ -256,6 +258,21 @@ class TestRouteTier1:
         with pytest.raises(ValueError, match=f"^{expected}$"):
             route_tier1(f, "widget", margin_threshold=bad_threshold)
 
+    def test_margin_threshold_rejection_emits_a_diagnostic_log(self, tmp_path: Path, monkeypatch, caplog):
+        """studio#135 round-4 review: a direct Python caller of route_tier1
+        (not going through the CLI's argparse, which has its own error
+        surface) previously got a bare ValueError with no structured log
+        trace. _validate_margin_threshold now logs a warning, with the
+        rejected value safely %r-repr'd, immediately before raising."""
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
+        with caplog.at_level(logging.WARNING, logger="studio.utils.cascade"):
+            with pytest.raises(ValueError):
+                route_tier1(f, "widget", margin_threshold=-1)
+        assert len(caplog.records) == 1
+        assert caplog.records[0].levelno == logging.WARNING
+        assert "-1" in caplog.records[0].getMessage()
+
     @pytest.mark.parametrize("bad_threshold", _BAD_MARGIN_THRESHOLDS)
     def test_route_query_rejects_the_same_invalid_margin_thresholds_as_route_tier1(
         self, tmp_path: Path, monkeypatch, bad_threshold,
@@ -324,6 +341,40 @@ class TestRouteTier1:
         result = route_fn(f, "widget", margin_threshold=accepted_threshold)  # must not raise
         assert result["tier"] == expected_tier
         assert result["reason"] == expected_reason
+
+    @pytest.mark.parametrize(
+        "bad_threshold", [t for t in _BAD_MARGIN_THRESHOLDS if not isinstance(t, str)],
+    )
+    def test_cli_arg_parser_rejects_the_same_numeric_boundary_values_as_the_direct_api(
+        self, bad_threshold,
+    ):
+        """studio#135 round-4 review: commands/cascade.py's
+        _margin_threshold_arg and this module's _validate_margin_threshold
+        used to implement the finite-and-positive rule as two independently
+        maintained checks that could silently drift apart. Both now
+        delegate to the same _check_margin_threshold_value helper -- this
+        reuses _BAD_MARGIN_THRESHOLDS (minus the non-numeric-string entry,
+        which exercises the CLI's *intentionally different* input domain,
+        see the next test) to prove every numeric value the direct API
+        rejects is also rejected at the CLI once coerced through argparse's
+        own string input.
+        """
+        with pytest.raises(argparse.ArgumentTypeError):
+            _margin_threshold_arg(str(bad_threshold))
+
+    def test_cli_arg_parser_accepts_a_numeric_string_the_direct_api_rejects(self):
+        """The one deliberate difference between the two entry points: the
+        CLI's natural input is a string, so parsing "0.5" to 0.5 first and
+        then applying the shared finite/positive check is correct CLI
+        behavior -- unlike _validate_margin_threshold (the direct Python
+        API), which requires an already-numeric value and correctly rejects
+        the bare string "0.5" (see
+        test_margin_threshold_rejects_invalid_values_at_the_callable_api).
+        Both entry points share one identical accept/reject policy for the
+        finite-and-positive check itself; they differ only in what they
+        accept as *input* before that check ever runs.
+        """
+        assert _margin_threshold_arg("0.5") == 0.5
 
     def test_route_tier1_never_touches_okf(self, tmp_path: Path, monkeypatch):
         """The module docstring guarantees route_tier1 stays free and
--- a/tests/test_cascade.py
+++ b/tests/test_cascade.py
@@ -6,6 +6,7 @@ See constructorfabric/studio#104.
 from __future__ import annotations
 
 import json
+import re
 from pathlib import Path
 from unittest import mock
 
@@ -225,7 +226,17 @@ class TestRouteTier1:
         assert result["tier"] == "resolved"
         assert result["reason"] == "heading_nav_tfidf_agree_large_margin"
 
-    @pytest.mark.parametrize("bad_threshold", [0, -1, float("nan"), float("inf")])
+    #: Non-finite/non-positive numeric values, plus a non-numeric string and
+    #: a bool -- the latter two exist to pin down a real bug (studio#135's
+    #: review): math.isfinite() raises TypeError for either, which would
+    #: propagate out before the intended ValueError ever ran, defeating
+    #: _validate_margin_threshold's documented contract for exactly the
+    #: direct-Python-caller case its docstring exists for. bool is checked
+    #: separately from the numeric cases since it subclasses int and would
+    #: otherwise slip past a bare isinstance(x, (int, float)) check.
+    _BAD_MARGIN_THRESHOLDS = [0, -1, float("nan"), float("inf"), "0.5", True]
+
+    @pytest.mark.parametrize("bad_threshold", _BAD_MARGIN_THRESHOLDS)
     def test_margin_threshold_rejects_invalid_values_at_the_callable_api(
         self, tmp_path: Path, monkeypatch, bad_threshold,
     ):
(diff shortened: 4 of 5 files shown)


skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -61,6 +62,22 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
...
+
+    ``commands/cascade.py``'s ``_margin_threshold_arg`` enforces this same
+    rule for the CLI, but a direct Python caller of :func:`route_tier1`/
+    :func:`route_query` bypasses argparse entirely -- without this check
+    here too, a non-positive or non-finite threshold would make
+    ``tfidf_result["margin"] >= margin_threshold`` fire on virtually any
+    finite margin, silently defeating the "no finite value is yet proven
+    safe" design basis this module's own docstring documents.
+    """
+    if margin_threshold is not None and not (math.isfinite(margin_threshold) and margin_threshold > 0):
```
ainetx: ### Non-numeric `margin_threshold` raises `TypeError` instead of the documented `ValueError`

**Severity:** Minor

**Problem**
`_validate_margin_threshold` in `skills/studio/scripts/studio/utils/cascade.py` calls `math.isfinite(margin_threshold)` with no type check first:

```python
if margin_threshold is not None and not (math.isfinite(margin_threshold) and margin_threshold > 0):
    raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
```

`math.isfinite()` raises `TypeError` for any non-numeric argument, and that `TypeError` propagates out before the intended `ValueError` line ever runs. The function's own docstring promises to reject invalid input with `ValueError`, but that only holds when the input is already numeric.

**How to reproduce**
Preconditions: call the library API directly (not through the CLI, which already coerces its `--margin-threshold` argument to `float` via argparse).

1. Import `route_tier1` from `studio.utils.cascade`.
2. Call `route_tier1(path, "query", margin_threshold="0.5")`.
3. Observe the raised exception.

**Expected behavior**
`ValueError: margin_threshold must be a finite number > 0, got '0.5'`

**Actual behavior**
`TypeError: must be real number, not str`, raised inside `math.isfinite()` before the intended `ValueError` is constructed.

```text
caller passes margin_threshold="0.5" (non-numeric)
        |
        v
_validate_margin_threshold(margin_threshold)
        |
        v
math.isfinite(margin_threshold)   <-- no type check first
        |
        v
TypeError raised here ------------> propagates out uncaught
        |                            (intended ValueError on the
        v                             next line never runs)
route_tier1 / route_query callers
crash with the wrong exception type
```

**Impact**
A direct Python caller of the library API that passes a non-numeric `margin_threshold` and only catches `ValueError` per this function's documented contract will get an unhandled `TypeError` instead. Low likelihood (requires bypassing the CLI's own type coercion), but it breaks the documented, uniform error contract for this validator.

**Suggested correction**
```python
def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
    if margin_threshold is None:
        return
    if not isinstance(margin_threshold, (int, float)) or not (math.isfinite(margin_threshold) and margin_threshold > 0):
        raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
```

**How to verify**
Add a test asserting `route_tier1(path, "widget", margin_threshold="0.5")` (and/or other non-numeric types) raises `ValueError` — matching the existing `pytest.raises(ValueError, match="margin_threshold")` pattern already used for `0`/`-1`/`nan`/`inf` — rather than `TypeError`.

tests/test_cascade.py
```diff
@@ -92,6 +92,26 @@ def test_margin_threshold_can_enable_a_numeric_large_margin_resolution(self, tmp
...
+        validation entirely -- without a check here too, a non-positive or
+        non-finite threshold would make the row-4 margin comparison fire on
+        virtually any finite margin, defeating the "no finite value is yet
+        proven safe" design basis."""
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
+        with pytest.raises(ValueError, match="margin_threshold"):
+            route_tier1(f, "widget", margin_threshold=bad_threshold)
+
+    def test_route_query_also_rejects_an_invalid_margin_threshold(self, tmp_path: Path, monkeypatch):
```
ainetx: ### `route_query`'s invalid-`margin_threshold` test coverage is narrower than `route_tier1`'s

**Severity:** Minor

**Problem**
`_validate_margin_threshold` is the single guard shared by `route_tier1` and `route_query`, and this PR's own rationale is that a direct Python caller of *either* function can bypass CLI validation. But the new test matrix is asymmetric: `route_tier1` is exercised against all four documented invalid values (`0`, `-1`, `nan`, `inf`) via `pytest.mark.parametrize` (`tests/test_cascade.py:95`), while `route_query` is exercised against only one hard-coded value, `-1.0` (`tests/test_cascade.py:109-113`). There's also no test confirming a large finite threshold (e.g. `1e10`) is accepted by either function.

**How to reproduce**
1. Open `tests/test_cascade.py:95-107` — `route_tier1` parametrized over `[0, -1, nan, inf]`.
2. Open `tests/test_cascade.py:109-113` — `route_query` tested with only `margin_threshold=-1.0`.
3. Search the file for a large-finite-value acceptance test — none exists.

**Expected behavior**
`route_query`'s invalid-input coverage mirrors `route_tier1`'s full matrix, plus a large-finite accept-path case for both.

**Actual behavior**
A regression that decouples `route_query`'s validation path from `route_tier1`'s (e.g. during a future refactor) and lets `0`/`nan`/`inf` slip through `route_query` specifically would not be caught by this suite.

```text
route_query(path, query, margin_threshold)
        |
        v
   route_tier1(path, query, margin_threshold)   <-- validation lives here
        |
   tested for route_tier1: [0, -1, nan, inf]  (full matrix)
   tested for route_query: [-1.0]             (partial matrix)
```

**Impact**
Low likelihood, but a future refactor could silently defeat the safety margin specifically through the higher-level `route_query` entry point most callers use, without any test catching it.

**Suggested correction**
Reuse the same `@pytest.mark.parametrize("bad_threshold", [0, -1, float("nan"), float("inf")])` matrix for the `route_query` test, and add one assertion that a large finite value (e.g. `1e10`) is accepted without raising by both functions.

**How to verify**
Apply the parametrize decorator to the `route_query` test, then temporarily change the validator's `> 0` to `>= 0` and confirm the expanded test now fails where the single-value test would not have caught it.

tests/test_cascade.py
```diff
@@ -92,6 +92,26 @@ def test_margin_threshold_can_enable_a_numeric_large_margin_resolution(self, tmp
...
+        self, tmp_path: Path, monkeypatch, bad_threshold,
+    ):
+        """A direct Python caller bypasses commands/cascade.py's argparse
+        validation entirely -- without a check here too, a non-positive or
+        non-finite threshold would make the row-4 margin comparison fire on
+        virtually any finite margin, defeating the "no finite value is yet
+        proven safe" design basis."""
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
+        with pytest.raises(ValueError, match="margin_threshold"):
```
ainetx: ### Invalid-`margin_threshold` tests assert only a substring match on the error message

**Severity:** Minor

**Problem**
Both new tests (`tests/test_cascade.py:106` and `:112`) use `pytest.raises(ValueError, match="margin_threshold")`. This performs `re.search` against the exception string, so it only verifies the message *contains* the word "margin_threshold" somewhere — not that it actually explains why the value was rejected. The current code happens to raise a correct, informative message, so the tests pass for the right reason today, but the assertion itself doesn't require that.

**How to reproduce**
1. Look at `cascade.py:78`'s raise statement and the test's `match="margin_threshold"`.
2. Imagine the raised message changed to something uninformative but still containing the word, e.g. `"margin_threshold rejected (TODO)"`.
3. The two tests would still pass.

**Expected behavior**
A message-content regression that keeps the substring but loses the real explanation should fail the test.

**Actual behavior**
It silently passes regardless of message quality.

```text
raise ValueError(f"margin_threshold must be a finite number > 0, got {value!r}")
        |
        v
pytest.raises(ValueError, match="margin_threshold")   <-- substring search only
        |
        v
   PASS regardless of whether the rest of the message is correct/informative
```

**Impact**
Low — test-quality gap only, doesn't affect runtime correctness.

**Suggested correction**
Tighten to a full-message regex, e.g. `match=r"^margin_threshold must be a finite number > 0, got .*$"`.

**How to verify**
Apply the tightened regex, then revert the message to a deliberately vague string containing "margin_threshold" and confirm the test now fails.

ainetx (approved):
**deep-review-auto approval — READY_WITH_NOTES**

40 checks (16 required from project config + 24 LLM-proposed) across 5 thematic phases were independently reviewed and verified against head `c5f836eec6710abfd74a5e3753b3472c0800d203`. All CI checks pass (22/22 concluded success). No Critical or Major findings were confirmed.

Three Minor findings remain open and unresolved (not blocking merge):

1. **Non-numeric `margin_threshold` raises `TypeError` instead of the documented `ValueError`** — `cascade.py:77` — direct Python callers bypassing CLI argparse can hit a `TypeError` from `math.isfinite()` before the intended `ValueError` runs. Suggested fix: add an `isinstance` guard before the `math.isfinite` call. (Thread: https://github.com/constructorfabric/studio/pull/135#discussion_r3922506303)

2. **`route_query`'s invalid-`margin_threshold` test coverage is narrower than `route_tier1`'s** — `tests/test_cascade.py:109` — `route_tier1` is parametrized over `[0, -1, nan, inf]` while `route_query` is tested with only one hard-coded `-1.0`. Suggested fix: apply the same `@pytest.mark.parametrize` matrix to `route_query`, plus a large-finite accept-path case. (Thread: https://github.com/constructorfabric/studio/pull/135#discussion_r3922590189)

3. **Invalid-`margin_threshold` tests assert only a substring match on the error message** — `tests/test_cascade.py:106` — `match="margin_threshold"` verifies only that the substring appears, not that the message is informative. Suggested fix: tighten to a full-message regex. (Thread: https://github.com/constructorfabric/studio/pull/135#discussion_r3922590323)

None of these affect the correctness of the core fix (the validation and frontmatter-delimiter changes are sound). The PR is ready to merge; the Minor items are recommended improvements for a follow-up.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 3aa802c7: added an `isinstance(margin_threshold, (int, float))` check (bool excluded) before `math.isfinite()` runs, so a non-numeric input now raises the documented `ValueError` instead of an unhandled `TypeError`. Added a parametrized test covering both the non-numeric string case and the bool case you'd expect to slip past a bare isinstance check.

tests/test_cascade.py
tkcoding: Fixed in 3aa802c7: `route_query`'s test now uses the same `@pytest.mark.parametrize` matrix as `route_tier1`'s (now also including the non-numeric/bool cases from the TypeError fix above), plus a shared large-finite accept-path test for both entry points.

tests/test_cascade.py
tkcoding: Fixed in 3aa802c7: both tests now assert a full-message regex (`re.escape(f"margin_threshold must be a finite number > 0, got {bad_threshold!r}")` anchored with `^...$`) instead of a bare substring match.

tests/test_cascade.py
```diff
@@ -92,6 +93,58 @@ def test_margin_threshold_can_enable_a_numeric_large_margin_resolution(self, tmp
...
 
+    #: Non-finite/non-positive numeric values, plus a non-numeric string and
+    #: a bool -- the latter two exist to pin down a real bug (studio#135's
+    #: review): math.isfinite() raises TypeError for either, which would
+    #: propagate out before the intended ValueError ever ran, defeating
+    #: _validate_margin_threshold's documented contract for exactly the
+    #: direct-Python-caller case its docstring exists for. bool is checked
+    #: separately from the numeric cases since it subclasses int and would
+    #: otherwise slip past a bare isinstance(x, (int, float)) check.
+    _BAD_MARGIN_THRESHOLDS = [0, -1, float("nan"), float("inf"), "0.5", True]
```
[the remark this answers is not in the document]
tkcoding: Fixed in b3ab0419: changed `_BAD_MARGIN_THRESHOLDS` from a list to a tuple. Confirmed clean with a local `ruff check`.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -71,6 +103,7 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
     it's still the best Tier-1 guess to hand Tier 2), two for row 3, none
     for row 1 (heading-nav found nothing to anchor a guess to at all).
     """
+    _validate_margin_threshold(margin_threshold)
```
ainetx, line 247: ### Margin-threshold validation now runs unconditionally before the escalate check, with no test isolating that behavior change

**Severity:** Minor

**Problem**
_validate_margin_threshold(margin_threshold) is inserted as the very first statement of route_tier1, before nav_first_match is even computed. Previously margin_threshold was only read/compared at row 4 of the routing table, so a caller passing an invalid threshold alongside a query that would have escalated on 'heading_nav_no_hits' (no heading-nav match at all) would get the escalate result silently. Now that same call raises ValueError unconditionally, regardless of which row would ultimately be reached.

**How to reproduce**
1. Call route_tier1(path, query, margin_threshold=-1) where `query` has zero heading-nav hits.
2. Before this change: function would return {'tier': 'escalate', 'reason': 'heading_nav_no_hits', ...} without ever inspecting margin_threshold.
3. After this change: function now raises ValueError before reaching that branch.

**Expected behavior**
Either the escalate-before-threshold-use behavior is preserved (validate lazily, only when margin_threshold will actually be read), or the eager-validation behavior change is called out as intentional and covered by a dedicated test plus a docstring note.

**Actual behavior**
Validation is unconditional and no test in the diff exercises the combination of an invalid margin_threshold with a query that escalates for unrelated reasons; the docstring is also unchanged.

```text
old: route_tier1() -> nav lookup -> ... -> row4 uses margin_threshold -> (invalid value only matters here)
new: route_tier1() -> _validate_margin_threshold() [raises here first] -> nav lookup -> ...
```

**Impact**
Existing callers who pass a stray/placeholder margin_threshold together with queries that don't reach row 4 will now get a hard ValueError where they previously got a normal escalate result -- a breaking behavior change for that caller population.

**Suggested correction**
Either move the validation call to just before the row-4 comparison (restoring lazy validation), or explicitly document the new eager-fail-fast contract in route_tier1's docstring and add a test covering the no-hits + invalid-threshold combination.

**How to verify**
Add a test that calls route_tier1 with a query producing zero heading-nav hits and an invalid margin_threshold, and assert the intended behavior (raise vs. escalate) matches the documented contract.

tests/test_cascade.py
```diff
@@ -92,6 +93,58 @@ def test_margin_threshold_can_enable_a_numeric_large_margin_resolution(self, tmp
         assert result["tier"] == "resolved"
         assert result["reason"] == "heading_nav_tfidf_agree_large_margin"
 
+    #: Non-finite/non-positive numeric values, plus a non-numeric string and
+    #: a bool -- the latter two exist to pin down a real bug (studio#135's
+    #: review): math.isfinite() raises TypeError for either, which would
+    #: propagate out before the intended ValueError ever ran, defeating
+    #: _validate_margin_threshold's documented contract for exactly the
```
ainetx, line 235: ### margin_threshold boundary test matrix doesn't isolate which validation clause fires

**Severity:** Minor

**Problem**
_BAD_MARGIN_THRESHOLDS = (0, -1, float('nan'), float('inf'), '0.5', True) never includes float('-inf'), -0.0, or a tiny positive finite value. 0 and -1 are both finite and only violate `> 0`; nan and inf both violate `isfinite()`. No case exists where a value is finite, non-positive, and adjacent to the boundary (-0.0), nor a case confirming the smallest legitimate positive value is accepted.

**How to reproduce**
1. Inspect _BAD_MARGIN_THRESHOLDS in tests/test_cascade.py.
2. Note float('-inf') is absent, and -0.0 (finite, `-0.0 > 0` is False) is absent.
3. Note the accept-path test only exercises 1e10, not a very small positive value near the boundary.

**Expected behavior**
Boundary cases like -0.0, float('-inf'), and a tiny positive float (e.g. 1e-300) should be included to pin down both clauses of the `and` independently.

**Actual behavior**
Current matrix leaves the `> 0` vs `>= 0` boundary and the smallest-accepted-value boundary unverified.

```text
isfinite(x) AND x>0
  nan/inf -> caught by isfinite (untested: -inf specifically)
  0/-1    -> caught by x>0 (untested: -0.0 specifically)
```

**Impact**
A future regression that flips `> 0` to `>= 0`, or mishandles -0.0/subnormal values, would not be caught by the existing tests.

**Suggested correction**
Add -0.0, float('-inf'), and a very small positive float to the parametrized cases (reject/accept respectively).

**How to verify**
Add the missing cases and confirm they pass against the current implementation, then confirm they'd fail against a mutated `>= 0` version.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -71,6 +103,7 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
     it's still the best Tier-1 guess to hand Tier 2), two for row 3, none
```
ainetx: ### route_tier1's docstring doesn't mention the new ValueError contract

**Severity:** Minor

**Problem**
The diff adds an unconditional call to _validate_margin_threshold as the first line of route_tier1 (and by extension route_query), which can now raise ValueError for a bad margin_threshold. No line was added to route_tier1's own docstring documenting this new exception-raising behavior; the diff shows only the code addition, not any docstring text change.

**How to reproduce**
1. Read route_tier1's docstring (the row-tiering description above the function body).
2. Note it describes rows 1-4 and margin_threshold's role in row 4 only.
3. Call route_tier1 with an invalid margin_threshold and observe it raises ValueError -- a fact the docstring gives no indication of.

**Expected behavior**
A public function's docstring should mention exceptions it can raise, especially a newly introduced one that changes its contract for existing callers.

**Actual behavior**
The docstring is left as-is; only inline comments on the new helper function explain the rationale, which a caller reading route_tier1's own docs would not see.

```text
caller reads route_tier1.__doc__ -> sees tiering rules only -> calls with bad margin_threshold -> unexpected ValueError not mentioned anywhere in the docs they read
```

**Impact**
Callers relying on the documented contract can be surprised by an undocumented exception, especially since this is also a behavior change from the previous lazy-validation approach (see CHK-021).

**Suggested correction**
Add a `Raises: ValueError` note to route_tier1's (and route_query's) docstring describing the finite/>0 requirement on margin_threshold.

**How to verify**
Re-read the updated docstring and confirm it mentions the ValueError condition for margin_threshold.

tests/test_cascade.py
```diff
@@ -92,6 +93,58 @@ def test_margin_threshold_can_enable_a_numeric_large_margin_resolution(self, tmp
...
+
+    @pytest.mark.parametrize("route_fn", [route_tier1, route_query])
+    def test_a_large_finite_margin_threshold_is_accepted_not_just_rejected_values(
+        self, tmp_path: Path, monkeypatch, route_fn,
+    ):
+        """The accept path, not just the reject path: a legitimately large
+        but finite threshold (e.g. an explicit, permissive opt-in) must not
+        itself be treated as invalid by either entry point."""
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _DIFFUSE_MARGIN_SAMPLE)
```
ainetx, line 340: ### Accept-path margin_threshold test uses a vacuous assertion

**Severity:** Minor

**Problem**
test_a_large_finite_margin_threshold_is_accepted_not_just_rejected_values calls route_fn(f, 'widget', margin_threshold=1e10) and then only asserts `"tier" in result`. This checks that some dict with a 'tier' key came back, but not what tier or reason it is -- a mutation that made the function return an arbitrary tier value (or the wrong tier due to a broken threshold comparison) would still pass this assertion.

**How to reproduce**
1. Mutate route_tier1/route_query so that, after accepting margin_threshold=1e10, it returns {'tier': 'escalate', 'reason': 'anything', 'candidates': []} instead of the correct tier for the sample.
2. Run test_a_large_finite_margin_threshold_is_accepted_not_just_rejected_values.
3. The test still passes because it only checks key presence, not the value.

**Expected behavior**
The assertion should pin an exact expected tier/reason (as the other tests in the same file do, e.g. asserting result['tier'] == 'resolved' and result['reason'] == ...) so a regression in the accept-path's actual routing outcome is caught, not just whether an exception was raised.

**Actual behavior**
Only `assert "tier" in result` is checked, which is satisfied by virtually any well-formed return dict.

```text
route_fn(margin_threshold=1e10) -> returns dict with 'tier' key (any value) -> assert "tier" in result -> PASS regardless of correctness
```

**Impact**
A real regression in how a large-but-valid margin_threshold is handled downstream (e.g. wrong tier chosen) would go undetected by this test.

**Suggested correction**
Assert the specific expected tier/reason values for the known sample input, mirroring the precision used in sibling tests in the same file.

**How to verify**
Tighten the assertion to check exact tier/reason values, then confirm the test fails under a mutation that changes the returned tier.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in aa15a9ca: added `test_margin_threshold_is_validated_even_when_heading_nav_finds_no_hits`, parametrized over the full `_BAD_MARGIN_THRESHOLDS` matrix, calling `route_tier1(f, "making up", margin_threshold=bad_threshold)` against the row-1 fixture (zero heading-nav hits) and asserting `ValueError` is raised there too -- pinning down that this is a real, deliberate behavior change from before this PR (previously such a call would have escalated silently regardless of the threshold's validity).

tests/test_cascade.py
tkcoding: Fixed in aa15a9ca: extended `_BAD_MARGIN_THRESHOLDS` to `(0, -1, float("nan"), float("inf"), float("-inf"), -0.0, "0.5", True)`, adding `float("-inf")` and `-0.0` so the matrix separately exercises IEEE negative zero at the boundary from the existing finite-but-wrong-sign cases (`0`/`-1`). Also added an accept-path case for `margin_threshold=1e-9` proving the smallest legitimate positive value passes validation.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in aa15a9ca: added a `Raises ValueError...` note to `route_tier1`'s docstring documenting the unconditional `_validate_margin_threshold` call and that it fires even for a row-1 (`heading_nav_no_hits`) query, plus a shorter cross-reference note on `route_query`'s docstring since it calls through to `route_tier1` first.

tests/test_cascade.py
tkcoding: Fixed in aa15a9ca: replaced the vacuous `assert "tier" in result` with exact tier/reason assertions for two cases against `_DIFFUSE_MARGIN_SAMPLE`'s actual measured margin (99.0 for "widget"): `margin_threshold=1e-9` must resolve (`resolved`/`heading_nav_tfidf_agree_large_margin`), while `margin_threshold=1e10` must still escalate (`escalate`/`diffuse_margin`) since it exceeds the real margin -- catching a mutation that broke the row-4 comparison or returned an arbitrary tier, which the old key-presence check would have missed.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -61,6 +62,37 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
...
+    # isinstance guards math.isfinite() itself: it raises TypeError for any
+    # non-numeric argument, which would propagate out before the ValueError
+    # below ever runs -- defeating this function's own documented contract
+    # for exactly the direct-Python-caller case its docstring exists for
+    # (the CLI's argparse type already coerces to float before this is
+    # ever reached, so only a direct caller can hit this). bool is
+    # deliberately excluded even though it subclasses int: True/False are
+    # not meaningful margin thresholds, and isfinite(True) would otherwise
+    # silently accept one.
+    if (
```
ainetx: ### CLI and direct-call threshold validation accept different input domains

**Severity:** Minor

**Problem**
_margin_threshold_arg first applies float(value), so CLI input "0.5" becomes 0.5 and passes. _validate_margin_threshold instead requires an already-int-or-float value, rejecting "0.5" before evaluating finiteness and positivity. Thus the claimed parity does not hold across argparse coercion.

**How to reproduce**
1. Invoke `cfs retrieve file.md query --margin-threshold 0.5`; argparse passes "0.5" and the CLI accepts it. 2. Call `route_query(Path("file.md"), "query", margin_threshold="0.5")`. 3. Observe ValueError from _validate_margin_threshold.

**Expected behavior**
Both entry points should define and test the same accepted threshold domain, including the CLI's string-to-float coercion policy, or document intentionally different API input types.

**Actual behavior**
The CLI accepts positive numeric strings after coercion, while the direct-call validator rejects them.

```text
CLI input "0.5" -> float("0.5") -> 0.5 -> accepted
Direct input "0.5" -> isinstance(str, (int, float)) -> false -> ValueError
```

**Impact**
Callers that forward user-supplied threshold text to the library cannot reproduce valid CLI behavior, and the two validators can continue to drift unnoticed.

**Suggested correction**
Extract one parsing-and-validation helper used by both paths, or make the direct API intentionally coerce strings and add parity tests covering accepted and rejected CLI-form inputs.

**How to verify**
Add parameterized tests that feed CLI-form values through _margin_threshold_arg and the direct API, asserting identical accept/reject outcomes and equivalent parsed values for valid inputs.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -61,6 +62,37 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def _validate_margin_threshold(margin_threshold: Optional[float]) -> None:
```
ainetx, line 163: ### Use a shared margin-threshold validator

**Severity:** Minor

**Problem**
`_validate_margin_threshold` implements the finite-and-positive rule separately from `commands/cascade.py`'s `_margin_threshold_arg`. The CLI parses strings and maps failures to argparse errors, but it repeats the rule rather than delegating validation to a shared utility-level source of truth.

**How to reproduce**
1. Change the permitted threshold rule in only one validator (for example, allow zero in the CLI parser). 2. Invoke `cfs retrieve --margin-threshold 0 ...`. 3. Invoke `route_query(..., margin_threshold=0)`.

**Expected behavior**
Both entry points enforce one shared threshold policy, with the CLI adapting any validation error to `ArgumentTypeError` as needed.

**Actual behavior**
The two entry points maintain separate literal implementations of the policy and can diverge after a future change.

```text
CLI string -> _margin_threshold_arg -> local rule
Python value -> _validate_margin_threshold -> duplicate local rule
```

**Impact**
A future policy update can make CLI and direct-library callers accept different threshold values or report inconsistent validation behavior.

**Suggested correction**
Move the finite-and-positive value validation into a shared utility helper; have the CLI convert its input to float and delegate to that helper, translating `ValueError` to `argparse.ArgumentTypeError`.

**How to verify**
Add tests demonstrating that both entry points accept and reject the same boundary values through the shared validator.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -61,6 +62,37 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
...
+    # ever reached, so only a direct caller can hit this). bool is
+    # deliberately excluded even though it subclasses int: True/False are
+    # not meaningful margin thresholds, and isfinite(True) would otherwise
+    # silently accept one.
+    if (
+        isinstance(margin_threshold, bool)
+        or not isinstance(margin_threshold, (int, float))
+        or not (math.isfinite(margin_threshold) and margin_threshold > 0)
+    ):
+        raise ValueError(f"margin_threshold must be a finite number > 0, got {margin_threshold!r}")
```
ainetx: ### Margin-threshold rejections have no diagnostic log

**Severity:** Minor

**Problem**
_validate_margin_threshold raises ValueError directly after rejecting a value, but emits no logger.debug/warning entry. Direct Python callers of route_tier1 or route_query therefore receive only a bare exception without a structured operational trace.

**How to reproduce**
1. Call route_query(path, query, margin_threshold=float('nan')) from an automated Python caller.\n2. Observe the ValueError.\n3. Inspect application logs for a threshold-rejection record.

**Expected behavior**
The rejection should retain the ValueError contract and emit a diagnostic log containing the rejected value and enough call context to correlate the failure.

**Actual behavior**
The function immediately raises ValueError with no logging.

```text
direct caller\n  -> route_query / route_tier1\n  -> _validate_margin_threshold\n  -> ValueError\n  -> no diagnostic log
```

**Impact**
Unattended integrations cannot distinguish this invalid-input rejection from other exceptions through structured logs or correlate it with the invoking route.

**Suggested correction**
Add a module logger and emit an appropriate debug or warning record immediately before raising, including a safely represented threshold and relevant route context where available.

**How to verify**
Invoke both route_tier1 and route_query with invalid thresholds and assert the existing ValueError plus one diagnostic log record for each invocation.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -292,19 +292,21 @@ def _match_sections_by_hash(
...
+    that without needing full YAML parsing, which is more than this check
+    needs to answer "is there real content here at all".
     """
     try:
         content = concept_path.read_text(encoding="utf-8")
     except (OSError, UnicodeDecodeError) as exc:
         logger.debug("okf concept file unreadable at %s: %s", concept_path, exc)
         return False
-    return content.startswith("---\n")
+    return content.startswith("---\n") and "\n---\n" in content[4:]
```
ainetx: ### Concept cache validity accepts malformed payloads

**Severity:** Minor

**Problem**
_concept_file_is_valid treats delimiter presence as sufficient validity. A stale cache entry with valid delimiters but missing required generated frontmatter keys or body structure is returned as current rather than rebuilt.

**How to reproduce**
1. Generate a concept file. 2. Replace its contents with `---\n---\n` or frontmatter missing required fields. 3. Call get_okf_status. 4. Observe the entry remains valid/current.

**Expected behavior**
Cached concept files should be validated against the required current payload shape before being considered current.

**Actual behavior**
Only the opening delimiter and any later closing delimiter are checked.

```text
malformed cached file\n  -> delimiters present\n  -> _concept_file_is_valid returns True\n  -> stale/invalid payload treated as current
```

**Impact**
Schema changes or partial corruption can silently serve unusable old-shaped concept content.

**Suggested correction**
Parse the frontmatter and validate the required generated fields and minimum content structure, ideally through a versioned schema marker.

**How to verify**
Add tests for delimiter-valid files missing each required field and assert they are reported missing or rebuilt.

ainetx (approved):
Automated review complete -- no blocking findings remain and CI is passing.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -292,19 +292,21 @@ def _match_sections_by_hash(
...
+    opening ``---`` doesn't either, since a file truncated right after it
+    still passes that alone; requiring the closing delimiter too catches
+    that without needing full YAML parsing, which is more than this check
+    needs to answer "is there real content here at all".
     """
     try:
         content = concept_path.read_text(encoding="utf-8")
     except (OSError, UnicodeDecodeError) as exc:
         logger.debug("okf concept file unreadable at %s: %s", concept_path, exc)
         return False
```
ainetx, line 324: ### Unreadable concept files are silently reported as missing

**Severity:** Minor

**Problem**
OSError and UnicodeDecodeError are logged only at debug level and converted to False. The caller consequently receives the normal missing-state fallback rather than a visible per-item error or diagnostic.

**How to reproduce**
1. Create a concept file for a section. 2. Make the file unreadable or invalid UTF-8. 3. Request OKF status. 4. Observe that the section is reported as missing without a visible error.

**Expected behavior**
An unreadable concept file should produce an explicit error or incomplete status with preserved diagnostics.

**Actual behavior**
The exception is debug-logged and False is returned, conflating read failure with a missing file.

```text
read_text failure
  -> debug-only log
  -> False
  -> normal missing status
```

**Impact**
Users cannot distinguish a recoverable filesystem or encoding failure from a genuinely absent concept file.

**Suggested correction**
Return or propagate a distinct per-item read-error state and surface its diagnostic at a visible level.

**How to verify**
Add tests for OSError and UnicodeDecodeError asserting an explicit error status and diagnostic rather than missing.

ainetx:
### Pass 3 competitive claims lack traceable sources

**Severity:** Major

**Problem**
The analysis asserts fast-changing competitor facts, including Appian Composer's April 2026 launch and CTO quotation and Microsoft Agent Framework GA timing, but provides no citations or source-attribution trail. Readers cannot validate or refresh these claims.

**How to reproduce**
1. Read the Tier 4 and Tier 5 tables and the Pass 3 sections. 2. Locate claims about vendor launches, product direction, and quoted statements. 3. Observe that no adjacent URL, footnote, publication date, or named source is supplied.

**Expected behavior**
Each falsifiable vendor claim or quotation should include a verifiable inline citation or dated attribution, or be clearly marked as unverified.

**Actual behavior**
The claims are presented as factual assertions without any traceable supporting source.

```text
Vendor roadmap claim
        |
        v
No citation or dated attribution
        |
        v
Reader cannot verify or refresh claim
```

**Impact**
Positioning decisions may rely on stale, inaccurate, or fabricated competitor information, particularly for rapidly changing 2026 roadmaps.

**Suggested correction**
Add an inline source link and publication date for every quoted statement and time-sensitive vendor/product claim; qualify claims that cannot be sourced.

**How to verify**
Review every Tier 4, Tier 5, and Pass 3 factual claim to confirm it has a directly relevant source/date or an explicit unverified qualifier.

---
_Original location: architecture/COMPETITIVE-ANALYSIS.md:50 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx:
### Pass 3 research freshness is undisclosed

**Severity:** Minor

**Problem**
The Pass 3 methodology and its 16-player web-research base do not disclose an as-of date or warn that fast-moving vendor claims may be stale. Rf-038 only says to revisit each release cycle, which is future maintenance guidance rather than provenance for the current findings.

**How to reproduce**
1. Read the Methodology block. 2. Locate Pass 3 and the Research base statement. 3. Observe that neither provides a research date or explicit freshness limitation.

**Expected behavior**
Pass 3 should state a research/as-of date, or explicitly caveat that vendor claims are time-sensitive and may have changed.

**Actual behavior**
Pass 3 is described only as web research and uses broad 2026/April 2026 references without dating the research itself.

```text
Undated web research
  -> vendor claims change
  -> Pass 3 conclusions appear current
  -> stale competitive guidance
```

**Impact**
Readers cannot judge whether the adjacent-category claims remain current, increasing the risk of retaining obsolete strategic positioning.

**Suggested correction**
Add an explicit Pass 3 research-as-of date and a short freshness caveat near the Methodology/Research base block; retain the release-cycle revisit instruction.

**How to verify**
Confirm the Methodology names the Pass 3 research date or explicit staleness limitation, and that Rf-038 links its release-cycle review to that provenance.

---
_Original location: architecture/COMPETITIVE-ANALYSIS.md:3 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx:
### Artifact Quality wire contract is not specified

**Severity:** Major

**Problem**
The decomposition declares a shared, versioned ArtifactFinding wire contract but omits the actual finding fields, Locus shape, verdict representation, schema version field, and compatibility/versioning rules.

**How to reproduce**
1. Implement a structural detector from §2.19. 2. Implement a judge-backed detector independently. 3. Attempt to serialize both results against finding_json_schema(). 4. The document supplies no canonical payload shape or version semantics to resolve incompatible choices.

**Expected behavior**
§2.19 defines or normatively references ArtifactFinding and Locus fields, evidence and suggested-action shapes, per-detector verdict encoding, an explicit prohibition on aggregate scores, and a concrete schema-version compatibility policy.

**Actual behavior**
§2.19 only labels these concepts and calls the contract versioned, leaving their wire representation and versioning ambiguous.

```text
Detector A ─┐
            ├─> unspecified ArtifactFinding payload ─> presentation layer
Detector B ─┘

No defined Locus/field/version contract
```

**Impact**
Later detectors and presentation consumers can drift into incompatible JSON formats before implementation establishes a stable contract.

**Suggested correction**
Add the normative finding_json_schema() payload definition (including Locus, evidence, suggested_action, detector verdict map, and schema version) and state the compatibility/versioning policy; retain explicit no-combined-score semantics.

**How to verify**
Independently serialize representative structural and judge-backed findings, validate both against the documented schema, and confirm the consumer can distinguish contract versions without any aggregate score.

---
_Original location: architecture/DECOMPOSITION.md:960 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx (changes requested):
Requesting changes — the code diff itself (margin-threshold validation in `cascade.py`, the tightened frontmatter check in `okf.py`, and their accompanying tests) is solid, but this PR also carries two open issues in the accompanying analysis/design writeup that need to be resolved before this can move forward:

- **Pass 3 competitive claims lack traceable sources** — Specific, time-sensitive competitor claims (e.g. Appian Composer's launch date and CTO quote, Microsoft Agent Framework GA timing) are stated with no citations, so readers have no way to verify or refresh them later. See https://github.com/constructorfabric/studio/pull/135#issuecomment-5578631332
- **Artifact Quality wire contract is not specified** — The decomposition calls for a shared, versioned `ArtifactFinding` contract but doesn't actually define it: no finding fields, no `Locus` shape, no verdict representation, no schema version field, and no compatibility/versioning rules. Without this, downstream consumers of the contract have nothing concrete to implement against. See https://github.com/constructorfabric/studio/pull/135#issuecomment-5578632136

Please add source attribution for the competitive claims and flesh out the wire contract definition before re-requesting review.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 4c53eb4e: extracted a shared _check_margin_threshold_value(value) helper in utils/cascade.py that owns the finite-and-positive check. _validate_margin_threshold (direct API, requires an already-numeric value) and commands/cascade.py's _margin_threshold_arg (CLI, parses the argparse string to float first, then delegates to the same helper, translating its ValueError into argparse.ArgumentTypeError) both now enforce one identical policy instead of two independently drifting copies. Added parameterized tests reusing _BAD_MARGIN_THRESHOLDS proving the CLI rejects every boundary value the direct API rejects, plus a test documenting the one intentional remaining difference: the CLI legitimately accepts the numeric string "0.5" (after float() coercion), while the direct API correctly rejects a bare string as non-numeric -- that's a real, intentional difference in input type, not policy drift.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 4c53eb4e: extracted a shared _check_margin_threshold_value(value) helper in utils/cascade.py that owns the finite-and-positive check. _validate_margin_threshold (direct API, requires an already-numeric value) and commands/cascade.py's _margin_threshold_arg (CLI, parses the argparse string to float first, then delegates to the same helper, translating its ValueError into argparse.ArgumentTypeError) both now enforce one identical policy instead of two independently drifting copies. Added parameterized tests reusing _BAD_MARGIN_THRESHOLDS proving the CLI rejects every boundary value the direct API rejects, plus a test documenting the one intentional remaining difference: the CLI legitimately accepts the numeric string "0.5" (after float() coercion), while the direct API correctly rejects a bare string as non-numeric -- that's a real, intentional difference in input type, not policy drift.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 4c53eb4e: added a module logger to utils/cascade.py and a logger.warning("rejected invalid margin_threshold: %r", margin_threshold) call immediately before _validate_margin_threshold raises, so a direct Python caller (no argparse error surface of its own) still leaves a structured log trace. Added a caplog-based test asserting both the ValueError and a single WARNING-level record are produced.

skills/studio/scripts/studio/utils/okf.py
tkcoding: Fixed in 4c53eb4e: _concept_file_is_valid in okf.py now, after confirming the opening/closing --- delimiters, regex-checks that all four fields write_concept_file's frontmatter builder always writes (title, description, resource, generated) are present with a non-blank value, and requires non-blank body content after the closing delimiter -- no new YAML dependency, just the re module already used elsewhere in this file. Added a TestConceptFileIsValid class covering a well-formed file (valid), each field missing/blank (invalid), empty frontmatter (invalid), empty/whitespace-only body (invalid), and a real write_concept_file output (still valid).

skills/studio/scripts/studio/utils/cascade.py
ainetx: Confirmed fixed as of 4c53eb4e55bc4c5fa85159deb97c251069fa58a3 -- independent re-verification no longer reproduces this issue. reviewer:claude-sdk=PASS -- commands/cascade.py's _margin_threshold_arg now calls the same _check_margin_threshold_value helper defined in utils/cascade.py that _validate_margin_threshold uses, so the finite/positive rule is enforced by one shared function rather than two independently maintained checks; the diff also adds tests (test_cli_arg_parser_rejects_the_same_numeric_boundary_values_as_the_direct_api and test_cli_arg_parser_accepts_a_numeric_string_the_direct_api_rejects) that feed the same boundary values (0, -1, nan, inf, -inf, -0.0, True) through both entry points and assert identical accept/reject behavior, differing only in input-domain (string coercion) which is the intentional CLI difference.; reviewer:codex-sdk=PASS -- Both entry points now delegate the finite-and-positive decision to shared `_check_margin_threshold_value`; the CLI intentionally coerces its string input with `float()` first, while direct calls require an already numeric non-bool value, and the added tests cover the stated numeric boundaries plus the deliberate `"0.5"` distinction.; verifier:codex-sdk=PASS -- The diff replaces the CLI's separate math.isfinite/positive check with the shared _check_margin_threshold_value helper that _validate_margin_threshold also calls; the CLI first performs its intentional argparse string-to-float coercion, then both paths apply the identical finite-and-positive rule, with tests covering 0, negatives, NaN, infinities, -0.0, bool, and the intentional numeric-string input-domain difference.

skills/studio/scripts/studio/utils/cascade.py
ainetx: Confirmed fixed as of 4c53eb4e55bc4c5fa85159deb97c251069fa58a3 -- independent re-verification no longer reproduces this issue. reviewer:claude-sdk=PASS -- The diff shows commands/cascade.py's _margin_threshold_arg now imports and delegates to the new shared _check_margin_threshold_value helper in utils/cascade.py (wrapping its ValueError into ArgumentTypeError), and _validate_margin_threshold in utils/cascade.py also calls the same _check_margin_threshold_value helper — both entry points share one single source of truth for the finite/positive rule, contrary to the check's premise of two independent implementations.; reviewer:codex-sdk=PASS -- The diff extracts `_check_margin_threshold_value` in `utils/cascade.py` and changes the CLI parser to call it after string-to-float coercion, so finite-and-positive validation has one shared source of truth while preserving each entry point's input-domain handling.; verifier:codex-sdk=PASS -- The diff extracts `_check_margin_threshold_value` in `utils/cascade.py`, and the CLI now imports and calls it after parsing its string input, while the library API calls the same helper; finite-and-positive validation has one shared source of truth.

skills/studio/scripts/studio/utils/okf.py
ainetx: Confirmed fixed as of 4c53eb4e55bc4c5fa85159deb97c251069fa58a3 -- independent re-verification no longer reproduces this issue. reviewer:claude-sdk=PASS -- The diff strengthens _concept_file_is_valid to verify required frontmatter schema fields (title, description, resource, generated all present and non-blank) plus a non-blank body, not just the content-hash/delimiter check alone; get_okf_status still separately validates manifest shape via _is_valid_manifest_shape before trusting built_from_hash, so a cached hit requires both fingerprint match and shape/schema validity.; reviewer:codex-sdk=PASS -- OKF cache validity now requires every writer-produced frontmatter field (`title`, `description`, `resource`, `generated`) to be present with a non-blank value, plus closed frontmatter and a non-blank body; manifest loading already validates required entry keys and types before staleness/hash matching.; verifier:codex-sdk=PASS -- OKF cache validity now validates manifest entry shape/types before status resolution, and validates each cached concept payload has closed frontmatter, all required current fields with non-blank values, and a non-blank body in addition to matching built_from_hash.

skills/studio/scripts/studio/utils/okf.py
```diff
@@ -290,21 +290,51 @@ def _match_sections_by_hash(
...
+    if closing == -1:
+        return False
+    frontmatter = content[4:closing]
+    body = content[closing + len("\n---\n"):]
+    if not body.strip():
+        return False
+    return all(
+        re.search(rf"(?m)^{field}:[ \t]*\S", frontmatter) is not None
+        for field in _REQUIRED_FRONTMATTER_FIELDS
+    )
```
ainetx, line 337: ### Empty-body concept files are invalidated on upgrade

**Severity:** Major

**Problem**
`write_concept_file` persists `frontmatter + body` without requiring a non-empty body. Previously, any such file starting with `---\n` was valid; the new `if not body.strip()` rejects a valid pre-existing writer output, making its unchanged manifest entry report missing.

**How to reproduce**
1. On the prior version, call `write_concept_file(..., body="")` for a valid section. 2. Confirm its manifest hash matches the section. 3. Upgrade to this version and call `get_okf_status`.

**Expected behavior**
An unchanged file produced by the supported writer remains current across the upgrade, or a migration/version policy handles the format change.

**Actual behavior**
The file is classified missing solely because its body is empty.

```text
old writer accepts empty body\n        |\n        v\nvalid pre-upgrade concept file\n        |\n        v\nnew body.strip() check -> missing
```

**Impact**
Existing local OKF bundles can be needlessly invalidated and re-summarized, defeating upgrade idempotency and potentially incurring regeneration cost.

**Suggested correction**
Either preserve acceptance of empty-body legacy files, or version the stricter format and migrate/rebuild affected bundles explicitly.

**How to verify**
Add a regression test that creates a manifest-backed concept file using `_build_frontmatter(...)+""` and confirms its intended post-upgrade status.

skills/studio/scripts/studio/commands/cascade.py
```diff
@@ -27,14 +26,25 @@ def _margin_threshold_arg(value: str) -> float:
     result, defeating that safety margin entirely; only a genuine positive
```
ainetx, line 27: ### Retrieve help omits its routing heuristic

**Severity:** Minor

**Problem**
The added detailed explanation of the TF-IDF margin safety rationale is confined to `_margin_threshold_arg`'s Python docstring. The argparse help remains too terse to let a CLI user understand the inferred routing heuristic that determines the command output.

**How to reproduce**
1. Run `cfs retrieve --help`.
2. Read the command description and option help.
3. Observe that it names a two-tier cascade and cutoff, but does not describe how heading navigation and TF-IDF are combined or when the fallback recommendation is selected.

**Expected behavior**
CLI help should summarize the routing heuristic and fallback decision in user-facing argparse help.

**Actual behavior**
Only internal module/function docstrings provide the substantive heuristic rationale; `--help` provides labels without that explanation.

```text
CLI user
  -> --help
  -> names "two-tier cascade"
  -> lacks routing/fallback heuristic
  -> output decision is not interpretable
```

**Impact**
Users cannot make informed choices about margin thresholds or interpret why retrieval selected Tier 1 versus a fallback recommendation.

**Suggested correction**
Expand the parser description or relevant option help to briefly describe heading-nav plus TF-IDF agreement, the margin cutoff, and the fallback to OKF-vs-baseline evaluation.

**How to verify**
Run `cfs retrieve --help` and confirm it contains a concise user-facing explanation of the routing and fallback heuristic.

tkcoding:
Re: the three comments above about `architecture/COMPETITIVE-ANALYSIS.md` and `architecture/DECOMPOSITION.md` — neither file is part of this PR's diff (which only touches `margin_threshold` validation and OKF concept-file checks), so they don't apply here.

- The two competitive-analysis findings (unsourced Pass 3 claims, undisclosed research freshness) are real — fixed in constructorfabric/studio#157, which also corrects a misattributed quote and a wrong GA quarter caught while sourcing them.
- The "Artifact Quality wire contract is not specified" finding doesn't hold up — the contract is fully specified in `architecture/features/artifact-quality.md` and already implemented/tested in `artifact_quality.py`.

Going forward, please open a dedicated PR/issue for findings about files outside a given PR's diff rather than attaching them here — easier to track, and less likely to be mistaken for feedback on this PR's actual changes.

tkcoding:
> [Pass 3 competitive claims lack traceable sources](https://github.com/constructorfabric/studio/pull/135#issuecomment-5578631332)

Not part of this PR's diff — this is about `architecture/COMPETITIVE-ANALYSIS.md`, not the `margin_threshold`/OKF code here. Fixed in constructorfabric/studio#157: added dated, linked citations for the Appian Composer launch, the governance quote, and Microsoft Agent Framework's GA — and corrected a misattributed quote (it's Appian's VP of Product Management, not the CTO) caught while sourcing it.

tkcoding:
> [Pass 3 research freshness is undisclosed](https://github.com/constructorfabric/studio/pull/135#issuecomment-5578631778)

Same file, not part of this PR's diff. Fixed in constructorfabric/studio#157: added a Methodology-section freshness caveat ("Pass 3 research as of 2026-09-08 ... revisit each release cycle per Rf-038").

tkcoding:
> [Artifact Quality wire contract is not specified](https://github.com/constructorfabric/studio/pull/135#issuecomment-5578632136)

Not part of this PR's diff, and on inspection this finding doesn't hold up: the wire contract is fully specified in `architecture/features/artifact-quality.md` (Locus/ArtifactFinding fields, schema_version, serialization rules) and already implemented + tested in `skills/studio/scripts/studio/utils/artifact_quality.py` / `tests/test_artifact_quality.py`. DECOMPOSITION.md:960 is intentionally just a scope summary that links out to that feature doc — it was never meant to repeat the full contract inline.

skills/studio/scripts/studio/utils/cascade.py
ainetx: Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cascade.py
ainetx: Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/cascade.py
ainetx: Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cascade.py
ainetx: Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/cascade.py
ainetx: Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/okf.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7fabcae455e6078b90980da70d3577bd053b1069 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The stricter check only adds new required conditions (closing delimiter, required fields, non-blank body) on top of the prior startswith('---\n') check; it never accepts less than before. A file valid under the old check but failing the new one would necessarily be missing a closing delimiter, a required field, or have an empty body -- all cases that, per _build_frontmatter's contract (writing title/description/resource/generated plus a real body), a genuinely-generated concept file would never hit. Test test_a_real_write_concept_file_output_is_valid explicitly confirms write_concept_file's real output still passes. No migration/version guard is needed since the stricter check is a strict superset of validity requirements consistent with what the writer always produced.
- **Verifier:** okf.py's own module docstring establishes the bundle as a local-only, gitignored, fully regenerable cache ('nothing about this module assumes the bundle survives across clones'); a file that the stricter check now classifies as invalid simply reports 'missing' and gets rewritten by write_concept_file, which is the designed self-healing behavior, not a silent break. The diff's added comment also asserts _build_frontmatter always writes all required fields, and a new test (test_a_real_write_concept_file_output_is_valid) confirms the current writer's output still passes.

</details>

ainetx (approved):
Solid change overall -- the shared validation helper cleanly de-duplicates the finite/positive margin-threshold check between the CLI and direct Python entry points, the bool/TypeError edge cases are handled correctly, and the strengthened concept-file validation closes a real gap (delimiters-only checks passing truncated/incomplete files). Test coverage is thorough across both. A few minor items worth a glance before or shortly after merge, none blocking:

- **Unreadable concept files are silently reported as missing** -- `OSError`/`UnicodeDecodeError` are only debug-logged and folded into the normal missing-state fallback, so there's no visible signal distinguishing a genuine I/O failure from an absent file. ([discussion](https://github.com/constructorfabric/studio/pull/135#discussion_r3954038828))
- **Pass 3 research freshness is undisclosed** -- the 16-player web-research base backing Pass 3 doesn't carry an as-of date or staleness caveat; the "revisit each release cycle" note is forward-looking maintenance guidance, not provenance for what's there now. ([comment](https://github.com/constructorfabric/studio/pull/135#issuecomment-5578631778))
- **Retrieve help omits its routing heuristic** -- the new TF-IDF margin-safety rationale lives only in `_margin_threshold_arg`'s docstring; the argparse `--help` text stays too terse for a CLI user to understand what drives the routing decision. ([discussion](https://github.com/constructorfabric/studio/pull/135#discussion_r3954095098))

PR #136, opened 2026-09-03
feat(cascade,doc-index): auto-trigger OKF build signal from real Tier-2 escalation counts

tkcoding:
## Summary
- Addresses Problem 1 of #134: the OKF-build decision was human-driven (`--expected-future-queries`), and nothing tracked real per-document query volume — `decision_log.record_read` redacts the `target` field and `summarize_reads` aggregates only by method, so there was no data to trigger a build from even manually.
- `cascade.route_tier2` now calls `doc_index.record_tier2_escalation(path)` on every invocation (every call is itself one Tier-1 escalation) and reports `should_build_okf` once the count crosses `_TIER2_BREAK_EVEN_ESCALATIONS` — derived analytically from this module's own already-hardcoded, measured per-query rates (`301_187 / (333_573 - 45_735)`, rounds up to 2), not a second guessed constant.
- Deliberately narrower than the ~15–48-query figures discussed earlier in #104: those measured break-even over a document's *total* query volume (most of which resolve cheaply at Tier 1 and never reach this function). `should_build_okf` answers the question `route_tier2` actually faces — for queries that *do* escalate, has continuing to fall back to baseline become more expensive than paying to build OKF.
- `expected_future_queries`/`build_okf_break_even` are unchanged and still available for a caller reasoning about a specific future volume rather than the actually-observed-so-far count.
- This module still never calls an LLM itself (per its own docstring) — `should_build_okf` is a signal for the caller to act on, not a build cascade.py triggers. `--help`/docstrings now say so explicitly (see review fixes below).

## Update: review round (ainetx) — 2 Major, 6 Minor, all fixed
- **Major — race condition, live-reproduced.** `get_or_build_doc_index`'s cache-miss build+save path took no lock, while the escalation counter's read-modify-write (originally a field inside the same `doc_index.json`) did. An unlocked structural rebuild (triggered by `route_tier1`'s own `find_sections`/`score_sections` calls) could read a stale snapshot and overwrite a newer, real escalation count back down. **Fix:** moved `tier2_escalations` out of the structural cache entirely into its own small counter file (`doc_index._escalation_cache_path`/`get_tier2_escalations`/`record_tier2_escalation`) — nothing else ever touches or locks that file, so there's nothing left to race. Added a concurrency test that hammers `get_or_build_doc_index` rebuilds against `record_tier2_escalation` calls and asserts no increment is ever lost.
[...]

PR #137, opened 2026-09-03
fix(cascade): fuse heading-nav and TF-IDF signals instead of skipping TF-IDF on a zero-hit

tkcoding:
## Summary
- Addresses Problem 2 of #134: `route_tier1` escalated immediately whenever heading-nav found zero hits, without ever running TF-IDF — discarding a genuinely different signal. Heading-nav (`find_sections`) requires the query's exact literal substring to appear somewhere in a section's raw text; TF-IDF (`score_sections`) tokenizes on individual alphanumeric words, so a query differing from the source only in punctuation/spacing/hyphenation can still score unambiguously on TF-IDF even when heading-nav's exact-phrase match fails outright.
- Both methods now always run, regardless of either one's own outcome. Two new Tier 1 rows cover heading-nav's zero-hit case:
  - TF-IDF alone resolves at Tier 1 when its own signal is unambiguous (`tfidf_only_unambiguous`).
  - Otherwise escalates, but hands along TF-IDF's own top pick as the Tier-2 candidate (`heading_nav_no_hits_diffuse_tfidf`) rather than none at all — unless TF-IDF also has no positive score anywhere, in which case it escalates with nothing (`no_signal_from_either_method`), same as the old behavior for a genuinely signal-less query.
- Deliberately still out of scope: folding OKF concept-file summaries into Tier 1 as a fourth ranked signal. Tier 1's whole value is being free and deterministic without ever touching the OKF bundle; Tier 2 already consults OKF status once Tier 1 escalates.

## Update: review round (ainetx) — 1 Major, 1 Minor, both fixed
- **Major, live-reproduced.** When heading-nav found a hit but TF-IDF's top score was exactly 0 in *every* section (no real TF-IDF signal at all), `route_tier1` still compared TF-IDF's document-order tie-break pick against heading-nav's real pick as if it were a genuine competing answer — fabricating a `resolved_multi`/`heading_nav_tfidf_disagree` verdict (or, by coincidence of order, a fabricated agreement) out of a signal that was never there. **Fix:** added an explicit new row (heading-nav has a hit, TF-IDF has no signal anywhere) that escalates on heading-nav's single, unconfirmed signal instead — symmetric to how the zero-hit rows already handle TF-IDF's own no-signal case. `route_tier1` was split into two functions along the routing table's own two-part structure (heading-nav miss vs. hit) rather than disabling pylint's too-many-return-statements once the new row pushed past it. Added a regression test reproducing the review's exact repro case.
- **Minor.** Five hand-built `TestRouteTier2` fixtures still used the retired `"heading_nav_no_hits"` reason string (renamed to `"no_signal_from_either_method"` earlier in this same PR, without these being updated) — `route_tier2` never reads the reason field, so nothing failed, but the drift was real and misleading. **Fix:** consolidated into one named constant instead of five separate literal copies, so a future rename can't silently skip these fixtures again. Test method names were also renumbered to match the routing table's real rows post-insertion (new row 4; disagree/agree/diffuse-margin shifted from 3/2/4 to 5/6/7).

## Test plan
- [x] New fixtures (`_TFIDF_ONLY_UNAMBIGUOUS_SAMPLE`, `_TFIDF_ONLY_DIFFUSE_SAMPLE`) use a real hyphenation gap (source: `"KAPING-framework"` / query: `"KAPING framework"`) to reproduce heading-nav's exact-phrase miss while TF-IDF's tokenizer still resolves it — hand-verified against `tokenize()`'s and `_confidence()`'s actual behavior, not just asserted against.
- [x] `pytest tests/test_cascade.py -q` — 30 passed (the only test file this PR's diff touches — an earlier version of this description incorrectly claimed a 5-file, 182-passed command pulled in from a broader sanity check, not this PR's own scope)
- [x] `pylint` on `cascade.py` — clean
- [x] `cfs validate --artifact architecture/features/traceability-validation.md` — passes (pre-existing, unrelated `toc-missing-description` warning only)
- [x] `cfs spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46` — all thresholds met
- [x] Independent of #136 (Problem 1) — this branch is based directly on `main`, not stacked on that PR

## Summary by CodeRabbit

- **Improvements**
  - Refined Tier 1 retrieval routing to combine heading navigation and TF-IDF results consistently.
  - Improved handling of heading-navigation misses, unambiguous TF-IDF matches, weak signals, and ambiguous results.
  - Preserved clear escalation behavior when retrieval confidence is insufficient.
- **Bug Fixes**
  - Prevented arbitrary zero-score results from being treated as meaningful disagreements.
  - Improved handling of stale, missing, invalid, or incomplete retrieval data.


skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -67,24 +88,46 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
...
     """
     nav_first_match = find_sections(path, query)["first_match"]
+    tfidf_result = score_sections(path, query)
+    tfidf_ranked = tfidf_result["ranked"]
+    has_tfidf_signal = bool(tfidf_ranked) and tfidf_ranked[0]["score"] > 0
+
     if nav_first_match is None:
-        return {"tier": "escalate", "reason": "heading_nav_no_hits", "candidates": []}
+        if not has_tfidf_signal:
+            return {"tier": "escalate", "reason": "no_signal_from_either_method", "candidates": []}
```
ainetx: ### Stale `tier1` reason literal in `TestRouteTier2` fixtures drifts from `route_tier1`'s renamed contract

**Severity:** Minor

**Problem**
`route_tier1` in `skills/studio/scripts/studio/utils/cascade.py` returns `"reason": "no_signal_from_either_method"` for the "row 1" escalation case (neither heading-nav nor TF-IDF found anything). This PR renamed that reason string from the old `"heading_nav_no_hits"`, and that old literal no longer appears anywhere in the shipped source. However, five hand-constructed `tier1` fixtures in `tests/test_cascade.py` (`TestRouteTier2`, lines 150, 181, 193, 225, 235) still use the old string:
```python
tier1 = {"tier": "escalate", "reason": "heading_nav_no_hits", "candidates": []}
```
These tests exercise `route_tier2` against a tier1 shape that the real `route_tier1` can no longer produce.

**How to reproduce**
1. Open `tests/test_cascade.py` and view `class TestRouteTier2` (line 146 onward).
2. Note lines 150, 181, 193, 225, 235 each hard-code `"reason": "heading_nav_no_hits"`.
3. Open `skills/studio/scripts/studio/utils/cascade.py`, `route_tier1`'s row-1 branch — the equivalent real reason is `"no_signal_from_either_method"`.
4. Confirm `"heading_nav_no_hits"` (exact string) does not occur anywhere in `cascade.py`.

**Expected behavior**
Test fixtures that stand in for `route_tier1`'s output should mirror the exact contract `route_tier1` currently produces.

**Actual behavior**
The fixtures embed a reason string that was retired by this same PR, silently diverging from the real contract. `route_tier2` never reads `tier1_result["reason"]`, so no test currently fails — the drift is invisible.

```text
route_tier1() [renamed]            tests/test_cascade.py::TestRouteTier2 [not updated]
  reason = "no_signal_from_         tier1 = {..., "reason": "heading_nav_no_hits", ...}
   either_method"                          |
        |                                  v
        v                          route_tier2(f, tier1)  -- never reads "reason"
  (real contract)                          |
                                           v
                                   test still passes, but exercises
                                   a tier1 shape route_tier1 can no
                                   longer emit
```

**Impact**
No current runtime failure — `route_tier2` is reason-agnostic. The risk is latent: (a) a future reader relying on these tests to understand the current reason vocabulary is misled, (b) a future change that makes `route_tier2` reason-aware would pass these tests while behaving incorrectly against real `route_tier1` output, and (c) it contradicts the PR's own stated full rename of the reason string, undermining confidence that the rename was applied consistently.

**Suggested correction**
Replace the five literal `"heading_nav_no_hits"` occurrences with `"no_signal_from_either_method"` (or, better, derive the fixture via an actual `route_tier1(...)` call / a shared constant, so future renames can't silently skip these fixtures again).

**How to verify**
After the fix, `grep -n "heading_nav_no_hits" tests/test_cascade.py` should return no matches, and `pytest tests/test_cascade.py -k TestRouteTier2` should still pass.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -67,24 +88,46 @@ def route_tier1(path: Path, query: str, *, margin_threshold: Optional[float] = N
...
     # None-check above narrows it -- a known astroid limitation across module
     # boundaries (github.com/pylint-dev/pylint/issues/3162), not a real risk here.
     nav_pick = cast(Dict[str, Any], nav_first_match)
 
-    tfidf_result = score_sections(path, query)
     # find_sections and score_sections both read retrieval_sections from the
     # same get_or_build_doc_index(path) call: a heading-nav match guarantees
     # at least one section exists, so TF-IDF always has one to rank too.
-    tfidf_pick = tfidf_result["ranked"][0]
+    tfidf_pick = tfidf_ranked[0]
```
ainetx, line 168: ### Tier 1 fabricates a "heading-nav/TF-IDF disagree" verdict when TF-IDF has zero signal everywhere

**Severity:** Major

**Problem**
In `route_tier1` (`skills/studio/scripts/studio/utils/cascade.py`, lines 106-137), once heading-nav finds a hit, the code unconditionally takes `tfidf_pick = tfidf_ranked[0]` (line 130) and compares its `line_start` against the heading-nav pick to decide between `resolved` (agree) and `resolved_multi`/`heading_nav_tfidf_disagree` (disagree). It never checks whether `tfidf_pick["score"] > 0` first. When every section scores exactly 0 (TF-IDF found no real signal at all), `tfidf_ranked[0]` is just the first section in document order — an arbitrary tie-break, not a genuine top pick — yet the function still reports it as if TF-IDF had produced a competing, confident answer.

**How to reproduce**
Preconditions: a document with two sections where the query's literal phrase appears only in the second section (giving heading-nav a real hit there), but tokenizes into terms that yield zero TF-IDF score in both sections.

1. Create a doc:
   ```
   ## Alpha

   No relevant terms here besides filler filler filler.

   ## Beta

   This talks about it up here too.
   ```
2. Call `route_tier1(path, "it up")`.
3. Observe `score_sections(path, "it up")` returns `{'ranked': [Alpha: score 0, Beta: score 0], 'margin': None, 'unambiguous': False}` — i.e., no real TF-IDF signal anywhere.
4. Observe `route_tier1` nonetheless returns `{'tier': 'resolved_multi', 'reason': 'heading_nav_tfidf_disagree', 'candidates': [Beta, Alpha]}`.

**Expected behavior**
When TF-IDF's top score is 0 (no real signal), the function should treat this the same as the "heading-nav-only" case — the module's own `has_tfidf_signal` guard (line 104, `bool(tfidf_ranked) and tfidf_ranked[0]["score"] > 0`) already exists and is used in the `nav_first_match is None` branch, but is not consulted at all once `nav_first_match` is found. TF-IDF having zero signal should not be represented as a competing "disagreement" pick.

**Actual behavior**
`route_tier1` reports `resolved_multi` / `heading_nav_tfidf_disagree` with a fabricated second candidate section (Alpha) that TF-IDF never actually found relevant — it's an artifact of ranking-order tie-breaking among all-zero scores.

```text
route_tier1(path, "it up")
   |
   +-- find_sections -> hit in "Beta"        (real signal)
   +-- score_sections -> ranked=[Alpha:0, Beta:0], unambiguous=False   (NO real signal)
   |
   v
tfidf_pick = tfidf_ranked[0]   # = Alpha, picked only by document order, not signal
   |
   v
tfidf_pick.line_start (Alpha) != nav_pick.line_start (Beta)
   |
   v
=> "resolved_multi" / "heading_nav_tfidf_disagree"   <-- WRONG: no disagreement, TF-IDF has nothing
```

**Impact**
A caller relying on `resolved_multi` to mean "two independently corroborating methods point at different sections, read both" will read and pay the token cost of a spurious extra section that carries no real evidence, purely because of tie-break ordering among zero scores. This directly undermines the routing table's row 5 (`heading-nav>0, TF-IDF disagrees`) contract documented in the module docstring, since TF-IDF didn't actually "disagree" — it had nothing to say.

**Suggested correction**
Before comparing `tfidf_pick["line_start"]` to `nav_pick["line_start"]`, check `has_tfidf_signal` (already computed at line 104). If it's `False`, treat the case as "heading-nav-only" (analogous to row 4/6 handling) rather than entering the disagree/agree branch with a meaningless `tfidf_pick`.

**How to verify**
Add a regression test using a fixture like the one above (heading-nav hits one section; TF-IDF scores 0 in every section) and assert the tier is not `resolved_multi`/`heading_nav_tfidf_disagree` when TF-IDF's ranked scores are all zero.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 66456034: consolidated the five hand-built fixtures into one named constant (`_NO_CANDIDATE_ESCALATION`) using the real current reason string, so route_tier1's exact reason value doesn't matter to route_tier2 (it never reads it), but a future rename can't silently drift out of sync with these fixtures again.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Confirmed and fixed in 66456034 -- reproduced your exact repro case locally before fixing (tier1['it up'] on the Alpha/Beta fixture did return the fabricated resolved_multi/heading_nav_tfidf_disagree you described). Added an explicit new row: when heading-nav has a hit but TF-IDF's top score is 0 everywhere (no real signal), it now escalates on heading-nav's single, unconfirmed signal instead of comparing an arbitrary all-zero tie-break as if it were a genuine competing pick -- symmetric to how the heading-nav-miss rows already handle TF-IDF's own no-signal case. Split route_tier1 into two functions along the routing table's own two-part structure (rather than disabling pylint's too-many-return-statements once the new row pushed past it). Added test_row4_heading_nav_hit_but_tfidf_has_no_signal_escalates, using your exact repro fixture.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -62,29 +91,82 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
```
ainetx, line 92: ### route_tier1 docstring's candidate-count list omits row 3

**Severity:** Minor

**Problem**
The updated docstring states candidates are returned as 'one for rows 2/4/6 (rows 3/6 despite escalating, since a signal's own top pick is still the best Tier-1 guess...), two for row 5, none for row 1'. The parenthetical explicitly says rows 3 and 6 return one candidate despite escalating, but the list right before it ('rows 2/4/6') excludes row 3, contradicting itself and the actual implementation.

**How to reproduce**
1. Read route_tier1's docstring in skills/studio/scripts/studio/utils/cascade.py.
2. Note it says candidates are 'one for rows 2/4/6'.
3. Compare with the code path for heading-nav-zero-hits + diffuse TF-IDF (row 3, reason 'heading_nav_no_hits_diffuse_tfidf'), which returns `candidates=[_as_candidate(tfidf_only_pick)]` -- one candidate -- and is confirmed by test_row3_tfidf_only_diffuse_escalates_with_tfidf_pick_as_candidate asserting a single-item candidates list.
4. Observe row 3 is missing from the 'one candidate' rows list even though it clearly belongs there.

**Expected behavior**
The docstring should list row 3 alongside rows 2/4/6 as returning one candidate (e.g. 'one for rows 2/3/4/6'), matching the actual behavior and the tests.

**Actual behavior**
The docstring currently reads 'one for rows 2/4/6', omitting row 3, while the parenthetical text and the real code/tests confirm row 3 also returns one candidate -- an internally inconsistent and technically inaccurate contract description.

```text
code behavior: row3 -> candidates=[tfidf_pick] (1 item)
docstring list: "one for rows 2/4/6"  <-- row 3 missing
docstring parenthetical: "(rows 3/6 despite escalating...)" <-- contradicts the list above
```

**Impact**
Low-severity but real: a future maintainer relying on this docstring's summary (rather than reading the code) could misjudge which rows a caller needs to handle a candidate for, especially since the parenthetical and the list disagree.

**Suggested correction**
Change 'one for rows 2/4/6' to 'one for rows 2/3/4/6' in the route_tier1 docstring.

**How to verify**
Re-read the docstring after the fix and confirm the candidate-count list is consistent with the parenthetical explanation and with route_tier1's actual return values for all six rows.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -62,29 +91,82 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
 
 
 # @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-tier1
+def _route_tier1_heading_nav_miss(
+    tfidf_ranked: List[Dict[str, Any]], has_tfidf_signal: bool, tfidf_result: Dict[str, Any],
+) -> Dict[str, Any]:
```
ainetx, line 96: ### Unconditional TF-IDF scoring on every Tier 1 call removes the heading-nav-miss short-circuit without evaluating cost on large documents

**Severity:** Minor

**Problem**
route_tier1 in skills/studio/scripts/studio/utils/cascade.py now calls score_sections(path, query) unconditionally before checking whether heading-nav found anything, whereas previously a heading-nav miss returned immediately with zero TF-IDF work. tfidf.py's own module docstring states this scoring method was validated only against a single ~166-page, ~9-section document and deliberately has no cap on section count or file size ('a caller feeding this an adversarially large or malformed file is a resource-management concern for that caller'). score_sections re-reads the file and tokenizes every section's text per query (not cached across queries, unlike the structural doc index). The diff's rationale (module docstring and test comments) covers only why both signals must run for correctness; it never discusses or benchmarks the added per-call cost this introduces for large documents.

**How to reproduce**
1. Call route_tier1(path, query) against a document far larger than the validated ~166-page/9-section scale. 2. Choose a query for which heading-nav has zero hits and TF-IDF also has no positive score anywhere (row 1, 'no_signal_from_either_method'). 3. Observe that despite escalating with no usable candidate, the call still paid full TF-IDF tokenization/scoring cost across every section of the large document -- work the prior version skipped entirely on a heading-nav miss.

**Expected behavior**
Either an explicit acceptance that unconditionally running TF-IDF is fine at expected document sizes, or a benchmark/complexity note comparing score_sections' cost to find_sections' for large documents, especially since the escalate-with-no-candidate outcome gains nothing from having paid that cost.

**Actual behavior**
The diff silently changes the cost profile of every Tier 1 call from 'cheap heading-nav-only when nav misses' to 'always pay full TF-IDF scoring,' with no capacity/perf discussion anywhere in the diff.

```text
Old: heading-nav miss -> return escalate (no TF-IDF work)
New: heading-nav miss -> score_sections() [tokenize+score ALL sections] -> then decide row 1/2/3
                              ^-- cost paid even when result is still 'escalate, no candidate'
```

**Impact**
On documents beyond the module's validated scale, every Tier 1 call (not just ones that end up resolving via TF-IDF) now incurs full-document tokenization/scoring latency, including the cases that gain no benefit from it (immediate escalate with no candidate). This is a capacity/perf regression risk left unquantified.

**Suggested correction**
Add a short benchmark or complexity comparison (score_sections vs find_sections cost) for large documents to the module docstring or PR description, or explicitly document the accepted tradeoff/expected document-size ceiling.

**How to verify**
Benchmark route_tier1 on a large synthetic document (well beyond ~166 pages / 9 sections) before and after this change for a heading-nav-miss query, and confirm the added latency is acceptable at expected real-world document sizes.

tests/test_cascade.py
```diff
@@ -39,29 +39,110 @@
...
+        result = route_tier1(f, "KAPING framework")
+        assert result == {
+            "tier": "resolved",
+            "reason": "tfidf_only_unambiguous",
+            "candidates": [{"heading": "Introduction", "line_start": 1, "line_end": 4}],
+        }
+
+    def test_row3_tfidf_only_diffuse_escalates_with_tfidf_pick_as_candidate(self, tmp_path: Path, monkeypatch):
+        """Heading-nav still has nothing, and TF-IDF has a real but diffuse
+        signal (both sections score positively) -- escalates, but hands
```
ainetx, line 157: ### route_tier2 lacks a test for the new TF-IDF-sourced row-3 escalate candidate

**Severity:** Minor

**Problem**
route_tier2's docstring was updated to say it's now called for rows 1, 3, and 6, with rows 3/6 narrowing to a single candidate section. Row 6's nav-sourced candidate already has coverage (pre-existing tests), but row 3's candidate is a brand-new provenance (a TF-IDF ranked entry via _as_candidate(tfidf_only_pick), not a heading-nav match) and no test in the diff calls route_tier2 with a Tier-1 result whose reason is heading_nav_no_hits_diffuse_tfidf.

**How to reproduce**
1. Construct a route_tier1-style result with reason='heading_nav_no_hits_diffuse_tfidf' and a TF-IDF-derived candidate. 2. Call route_tier2 with it. 3. Observe there is no existing assertion in the diff verifying correct OKF-staleness narrowing for this input shape.

**Expected behavior**
A test asserting route_tier2, given a row-3 Tier-1 output, narrows OKF-staleness checking to exactly the TF-IDF-derived candidate section, mirroring the row-6 nav-sourced candidate test.

**Actual behavior**
No such test is added or modified in this diff; only route_tier1's three new branches (rows 1-3) are tested.

```text
route_tier1 (row3, new) --candidate(tfidf)--> route_tier2 [UNTESTED for this candidate shape]
route_tier1 (row6, existing) --candidate(nav)--> route_tier2 [tested]
```

**Impact**
A candidate-shape or narrowing-logic bug specific to TF-IDF-sourced candidates (e.g. key mismatch between _as_candidate output and what route_tier2 expects) could ship undetected.

**Suggested correction**
Add a TestRouteTier2 test that feeds a reason='heading_nav_no_hits_diffuse_tfidf' Tier-1 result (with a TF-IDF-derived candidate) into route_tier2 and asserts correct single-section OKF narrowing.

**How to verify**
Run the new test and confirm it fails against a route_tier2 that mishandles TF-IDF-sourced candidates, then passes with the current implementation.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -9,20 +9,49 @@
...
+earlier, a zero-hit heading-nav result escalated immediately without ever
+running TF-IDF at all, discarding a second, genuinely different signal --
+heading-nav needs the query's exact literal substring to appear somewhere in
+a section's raw text, while TF-IDF tokenizes on individual words, so a query
+that differs from the source only in punctuation/spacing/hyphenation
+(heading-nav: 0 hits) can still score unambiguously on TF-IDF's word-level
+match. Every Tier 1 call now runs both methods unconditionally; only their
+*outcome* determines whether the query needed one signal, two, or must
+escalate. Row 4 exists because "TF-IDF found nothing" (every section scores
+exactly 0) is a distinct outcome from "TF-IDF agrees" or "TF-IDF disagrees"
```
ainetx, line 33: ### OKF-independence guarantee for Tier 1 is documented but not enforced

**Severity:** Minor

**Problem**
The new module docstring paragraph states a newly-articulated architectural invariant: route_tier1 must never touch the OKF bundle, so that Tier 1 stays free and deterministic. Nothing in the diff backs this with an automated check -- there is no import-boundary lint rule scoped to cascade.py's route_tier1 vs. the OKF module, and no test that asserts route_tier1 never imports or calls into OKF-related code.

**How to reproduce**
1. A future contributor edits route_tier1 to add an OKF lookup as a tiebreaker signal (e.g. to resolve ambiguous TF-IDF cases). 2. No lint rule or test fails. 3. The 'free and deterministic Tier 1' guarantee silently breaks, and callers relying on Tier 1 never touching OKF are unaware.

**Expected behavior**
Either an import-boundary/lint rule blocking cascade.py (or specifically route_tier1) from importing the OKF module, or a test that inspects route_tier1's dependencies/behavior to assert OKF is never consulted.

**Actual behavior**
The guarantee exists only as a comment in the module docstring; nothing in the codebase enforces it mechanically.

```text
future edit -> route_tier1 imports/calls OKF module -> no lint/test catches it -> documented invariant silently violated
```

**Impact**
Low immediate risk, but the documented architectural boundary can erode over time without any signal, undermining the stated design guarantee and potentially reintroducing non-determinism or hidden LLM calls into what's advertised as a purely mechanical tier.

**Suggested correction**
Add a lightweight test (e.g. static check that cascade.py's route_tier1 function/module doesn't reference the OKF module) or wire an import-boundary lint rule analogous to the existing studio_proxy<->studio one, scoped to cascade.py vs. the OKF module.

**How to verify**
Introduce a stub OKF import/call inside route_tier1 and confirm the new test or lint rule fails.

tests/test_cascade.py
```diff
@@ -39,29 +39,110 @@
...
+    "## Related Work\n\nThis section covers unrelated background material with no overlap.\n"
+)
+
+# Same hyphenation gap (heading-nav: 0 hits for "gizmo thing" -- the exact
+# two-word phrase never appears contiguously in either section), but this
+# time both terms independently occur in *both* sections, so TF-IDF scores
+# positively on both -- a real signal, just not an unambiguous one.
+_TFIDF_ONLY_DIFFUSE_SAMPLE = (
+    "## SectionA\n\nA gizmo-thing sits here alone.\n\n"
+    "## SectionB\n\nThe gizmo hums. " + ("filler word text here. " * 30) + "A thing rattles.\n"
```
ainetx, line 60: ### No test covers tfidf-only resolution with a single-section document

**Severity:** Minor

**Problem**
The new row-2 (`tfidf_only_unambiguous`) and row-3 (`tfidf_only_diffuse`) tests both use two-section fixtures. `score_sections`'s `unambiguous` flag is generally computed from the margin between the top and second-ranked section; with only one section in the document there is no second candidate to compare against, which is a distinct code path from the two added fixtures.

**How to reproduce**
1. Create a one-section markdown doc where the section scores positively on TF-IDF for some query but heading-nav has zero hits (e.g. a hyphenation gap as in the existing KAPING example, but with only one heading). 2. Call route_tier1. 3. Observe whether `unambiguous` is trivially true/false and whether tier/reason match expectations -- this is currently unverified by any test.

**Expected behavior**
A test asserting route_tier1's exact return dict for a one-section document under the new heading-nav-zero-hits/TF-IDF-positive branch.

**Actual behavior**
No such test exists in the diff; the single-candidate boundary of the new tfidf-only paths is unexercised.

```text
score_sections(one section) -> ranked=[X] -> unambiguous=? (untested) -> route_tier1 branch untested
```

**Impact**
A latent bug in how `unambiguous` behaves for a single-candidate ranking (e.g. vacuously True) could silently resolve queries at Tier 1 that should escalate, without any test catching a regression.

**Suggested correction**
Add a test with a one-section fixture, a query giving heading-nav zero hits and a positive TF-IDF score, asserting route_tier1's tier/reason/candidates.

**How to verify**
New test passes and documents the intended behavior for the single-section case.

tests/test_cascade.py
```diff
@@ -39,29 +39,110 @@
...
+# here actually exercises route_tier1 itself.
+_NO_CANDIDATE_ESCALATION = {"tier": "escalate", "reason": "no_signal_from_either_method", "candidates": []}
+
+
 class TestRouteTier1:
-    def test_row1_heading_nav_no_hits_escalates(self, tmp_path: Path, monkeypatch):
+    def test_row1_neither_method_has_signal_escalates(self, tmp_path: Path, monkeypatch):
+        """"making up" is tfidf.py's own documented adversarial case: "up" is
+        filtered by the 3-char minimum, and "making" never appears anywhere
+        in this fixture, so TF-IDF has nothing either -- both methods
```
ainetx, line 97: ### No regression test guarding that the TF-IDF-only unambiguous branch never calls route_tier2/OKF

**Severity:** Minor

**Problem**
route_tier1's new row-2 branch (tfidf_only_unambiguous) resolves entirely within Tier 1 and, per the docstring intent, should never reach Tier 2/OKF. The only test covering this branch, test_row2_tfidf_only_unambiguous_resolves_at_tier1, checks the returned dict shape but does not mock or spy on route_tier2 (or an OKF lookup) to assert it was never called.

**How to reproduce**
1. Inspect tests/test_cascade.py's test_row2_tfidf_only_unambiguous_resolves_at_tier1.
2. Note it calls route_tier1 directly and asserts equality on the returned dict.
3. Note no mock.patch/spy on studio.utils.cascade.route_tier2 (or the OKF lookup it uses) is set up to assert zero calls.

**Expected behavior**
A test that patches route_tier2 (or the underlying OKF status function) and asserts it is never called when route_tier1 takes the tfidf_only_unambiguous path, so a future refactor that accidentally threads Tier 2/OKF into this branch fails CI.

**Actual behavior**
Only the return-value shape is tested; nothing would fail if a future change caused this branch to also consult route_tier2/OKF, since route_tier1 doesn't currently call route_tier2 in any branch and no test enforces that invariant going forward.

```text
route_tier1(tfidf_only_unambiguous) --returns--> {tier: resolved, ...} [tested] ; route_tier1 --calls?--> route_tier2/OKF [NOT tested to be absent]
```

**Impact**
Low likelihood but real coverage gap: a future maintainer merging Tier 1 and Tier 2 logic (e.g. while adding OKF awareness to Tier 1, which the docstring explicitly says is out of scope) would get no test failure signal for this specific branch.

**Suggested correction**
Add a test that monkeypatches/mocks `route_tier2` (and/or the OKF status function it calls) and asserts it is not invoked when route_tier1 returns tier='resolved' with reason='tfidf_only_unambiguous'.

**How to verify**
Run the new test; it should fail if route_tier2 is called from this branch and pass with the current implementation.

tests/test_cascade.py
```diff
@@ -97,7 +178,7 @@ class TestRouteTier2:
     def test_no_bundle_at_all_recommends_baseline(self, tmp_path: Path, monkeypatch):
```
ainetx, line 262: ### route_tier2 has no test for the new row-3 (TF-IDF-only diffuse) escalation path

**Severity:** Minor

**Problem**
route_tier2's docstring was updated to say it's now called for rows 1, 3, 4, and 7 (previously 1 and 4), with row 3 narrowing to a TF-IDF-sourced candidate (a new provenance never exercised by route_tier2 before). The diff's test additions only cover route_tier1's three heading-nav-miss branches and reuse pre-existing route_tier2 tests that all pass the row-1 (no-candidate) shape via the new _NO_CANDIDATE_ESCALATION constant, or pre-existing nav-derived candidates.

**How to reproduce**
1. Inspect tests/test_cascade.py's TestRouteTier2 class in the diff. 2. Note every tier1 fixture passed to route_tier2 uses reason='no_signal_from_either_method' (row 1) or a heading-nav-derived candidate (rows 4/7, pre-existing tests). 3. No fixture uses reason='heading_nav_no_hits_diffuse_tfidf' (row 3) with a TF-IDF-derived candidate.

**Expected behavior**
A test calling route_tier2 with a tier1 result shaped like row 3 (TF-IDF-only candidate) to confirm OKF staleness narrowing works correctly against a candidate whose provenance is score_sections rather than find_sections.

**Actual behavior**
Only route_tier1-level tests exercise the row-3 candidate shape; route_tier2 is never tested against it, so a candidate-shape or narrowing bug specific to TF-IDF-sourced candidates would not be caught by this test suite.

```text
route_tier1 (row 3) --escalate+TFIDF candidate--> route_tier2
                                                      |
                                                      v
                                          [untested narrowing logic]
```

**Impact**
A latent bug in route_tier2's handling of TF-IDF-derived candidates (e.g. a key mismatch from _as_candidate, or incorrect staleness narrowing) could ship undetected.

**Suggested correction**
Add a TestRouteTier2 test that builds a tier1 dict with reason='heading_nav_no_hits_diffuse_tfidf' and a TF-IDF-sourced candidate, then asserts route_tier2 narrows correctly to that section's OKF staleness.

**How to verify**
Run the new test and confirm it fails against a route_tier2 that mishandles TF-IDF-provenance candidates, then passes with the current implementation.

skills/studio/scripts/studio/utils/cascade.py
ainetx, line 33: ### Tier 1 / OKF independence is documented but not mechanically enforced

**Severity:** Minor

**Problem**
The module docstring newly asserts that route_tier1 must never query OKF status so Tier 1 stays free/deterministic, but this is prose only -- no lint rule, import-boundary check, or test in the diff verifies route_tier1 doesn't import or call into the OKF module.

**How to reproduce**
1. A future PR adds an OKF status check inside route_tier1 (e.g. to 'optimize' escalation). 2. No test or lint fails. 3. The documented invariant silently breaks.

**Expected behavior**
Either an automated import-boundary rule/test asserting route_tier1 never references the OKF module, or acceptance that this is a soft, review-enforced convention.

**Actual behavior**
Only a comment states the invariant; nothing prevents violating it.

```text
route_tier1 (docstring: 'never touches OKF') --(no enforcement)--> future edit could import OKF module --> invariant silently broken
```

**Impact**
Tier 1's determinism/no-LLM guarantee could regress unnoticed, undermining a documented architectural property.

**Suggested correction**
Add a small test (e.g. asserting 'okf' not in the cascade module's imports used by route_tier1, or a mock/spy ensuring OKF functions are never called during route_tier1) or an import-linter rule covering this pair.

**How to verify**
Introduce a stub OKF call inside route_tier1 and confirm the new test/lint rule fails.

tests/test_cascade.py
```diff
@@ -39,29 +39,110 @@
     "## SectionB\n\ngadget gadget gadget.\n"
 )
 
+# constructorfabric/studio#134, Oleg67's suggestion #4: a real single-signal
```
ainetx, line 43: ### No test for tfidf_only_unambiguous with a single-section document

**Severity:** Minor

**Problem**
_TFIDF_ONLY_UNAMBIGUOUS_SAMPLE and _TFIDF_ONLY_DIFFUSE_SAMPLE both use two sections, so the new row-2 code path is never exercised against a document with exactly one section, where 'unambiguous' semantics (no second candidate to compare margin against) are a distinct boundary.

**How to reproduce**
1. Construct a one-section doc where the query has 0 heading-nav hits but a positive TF-IDF score. 2. Call route_tier1. 3. No existing test asserts what tier/reason/candidate is returned in this case.

**Expected behavior**
A test asserting route_tier1's behavior for a one-section document under the row-2 conditions.

**Actual behavior**
This boundary is untested; behavior is only implied by code, not verified.

```text
route_tier1 -> heading-nav 0 hits -> TF-IDF (1 section, score>0) -> unambiguous? -> [untested]
```

**Impact**
A regression or edge-case bug specific to single-section documents (a realistic short-document scenario) would not be caught by CI.

**Suggested correction**
Add a test fixture with one section and a query that yields 0 heading-nav hits and a positive TF-IDF score, asserting the expected route_tier1 result.

**How to verify**
Add the test, confirm it passes against current code and would fail if the single-candidate 'unambiguous' handling regressed.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -62,29 +91,82 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
 
 
 # @cpt-begin:cpt-studio-algo-traceability-validation-cascade:p1:inst-cascade-tier1
+def _route_tier1_heading_nav_miss(
+    tfidf_ranked: List[Dict[str, Any]], has_tfidf_signal: bool, tfidf_result: Dict[str, Any],
+) -> Dict[str, Any]:
+    """Rows 1-3: heading-nav found nothing at all to anchor a guess to, so
+    TF-IDF alone decides whether there's anything else to go on.
+
+    Split out of :func:`route_tier1` (which was hitting pylint's
```
ainetx, line 100: ### route_tier1 docstring's one-candidate row list omits row 3

**Severity:** Minor

**Problem**
The updated docstring says candidates has 'one for rows 2/4/6/7' but then parenthetically clarifies '(rows 3/4/7 despite escalating...)', which references row 3 as one of the escalating-but-one-candidate rows even though row 3 is missing from the main enumeration list. Since rows 2,3,4,6,7 all return exactly one candidate (row 1 returns none, row 5 returns two), the correct main list should include row 3.

**How to reproduce**
1. Read the route_tier1 docstring in cascade.py after this diff. 2. Note the sentence 'one for rows 2/4/6/7 (rows 3/4/7 despite escalating...)'. 3. Compare to _route_tier1_heading_nav_miss's row-3 branch and test_row3_tfidf_only_diffuse_escalates_with_tfidf_pick_as_candidate, both of which show row 3 returns exactly one candidate.

**Expected behavior**
The main enumeration should read 'rows 2/3/4/6/7' so it's internally consistent with its own parenthetical and with the actual code/tests.

**Actual behavior**
Row 3 is dropped from the main list, creating a self-contradiction within the same docstring sentence.

```text
docstring main list [2,4,6,7] --(inconsistent with)--> docstring parenthetical [3,4,7] --(consistent with)--> code/tests (row 3 = 1 candidate)
```

**Impact**
A reader relying on the docstring's main enumeration alone could wrongly conclude row 3 returns zero or an ambiguous number of candidates.

**Suggested correction**
Update the enumeration to 'one for rows 2/3/4/6/7 (rows 3/4/7 despite escalating, ...)'.

**How to verify**
Re-read the docstring and confirm the main list, the parenthetical, and the actual return statements/tests for rows 1-7 are mutually consistent.

tests/test_cascade.py
```diff
@@ -39,29 +39,110 @@
...
+        """"making up" is tfidf.py's own documented adversarial case: "up" is
+        filtered by the 3-char minimum, and "making" never appears anywhere
+        in this fixture, so TF-IDF has nothing either -- both methods
+        genuinely have zero signal, not just heading-nav."""
         monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
         f = _write(tmp_path)
         result = route_tier1(f, "making up")
-        assert result == {"tier": "escalate", "reason": "heading_nav_no_hits", "candidates": []}
+        assert result == {"tier": "escalate", "reason": "no_signal_from_either_method", "candidates": []}
+
```
ainetx, line 103: ### No regression guard proving tfidf_only_unambiguous skips OKF/route_tier2

**Severity:** Minor

**Problem**
test_row2_tfidf_only_unambiguous_resolves_at_tier1 only asserts the returned dict equals the expected resolved result; it does not assert that route_tier2 or any OKF lookup was never called, so the documented 'Tier 1 stays free of OKF' guarantee for this specific new branch has no automated regression guard beyond the return-value check.

**How to reproduce**
1. Hypothetically add a stray OKF/route_tier2 call inside the row-2 branch that doesn't affect the returned dict (e.g. a side-effecting lookup ignored downstream). 2. Run the existing test suite. 3. It still passes because only the return value is checked.

**Expected behavior**
A test that mocks/spies on route_tier2 (or the OKF module) and asserts it is not called when route_tier1 takes the tfidf_only_unambiguous path.

**Actual behavior**
Only the output shape is verified; no call-count/spy assertion exists.

```text
route_tier1 (row 2) -> returns resolved dict (asserted) -> [OKF/route_tier2 call, if added, would go undetected]
```

**Impact**
A future regression that adds an unnecessary OKF touch to the fast, deterministic Tier-1-only path would go unnoticed by CI, silently degrading Tier 1's performance/determinism guarantee.

**Suggested correction**
Add a monkeypatch/spy on route_tier2 (or the OKF concept-file lookup function) in the row-2 test and assert it was never called.

**How to verify**
Add the spy-based assertion, confirm it passes now, and confirm it would fail if route_tier2 were called from that branch.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 2f6159d7: re-derived which rows return one/two/zero candidates from the actual current implementation (rows 2/3/4/6/7 return exactly one, row 5 returns two, row 1 returns none) and corrected the enumeration from "rows 2/4/6/7" to "rows 2/3/4/6/7" so the main list no longer contradicts its own parenthetical ("rows 3/4/7 despite escalating").

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 2f6159d7: re-derived which rows return one/two/zero candidates from the actual current implementation (rows 2/3/4/6/7 return exactly one, row 5 returns two, row 1 returns none) and corrected the enumeration from "rows 2/4/6/7" to "rows 2/3/4/6/7" so the main list no longer contradicts its own parenthetical ("rows 3/4/7 despite escalating").

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 2f6159d7: added an explicit comment at the score_sections() call site and a docstring paragraph in route_tier1 acknowledging that fusing the two signals means TF-IDF now runs on every call, not just on a heading-nav miss -- citing tfidf.py's own documented scale assumption (validated against a single ~166-page/~9-section document, no cap on section count or file size). Deliberately did not skip or gate the call: that would undermine this PR's actual purpose of fusing both signals, so this documents the accepted tradeoff rather than working around it.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row3_tfidf_sourced_candidate_recommends_okf_when_current, which builds a real row-3 route_tier1 result (TF-IDF-sourced candidate, heading-nav never matched -- a provenance route_tier2 had never been exercised against before) and asserts route_tier2 recommends OKF correctly for it. This satisfies both this comment and the other "route_tier2 lacks a row-3 test" comment on this PR -- they describe the identical missing coverage.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row3_tfidf_sourced_candidate_recommends_okf_when_current, which builds a real row-3 route_tier1 result (TF-IDF-sourced candidate, heading-nav never matched -- a provenance route_tier2 had never been exercised against before) and asserts route_tier2 recommends OKF correctly for it. This satisfies both this comment and the other "route_tier2 lacks a row-3 test" comment on this PR -- they describe the identical missing coverage.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in 34122b34: added test_route_tier1_never_touches_okf, which patches get_okf_status to raise if called at all, then exercises every routing-table row (1 through 7, plus the margin_threshold opt-in) against it -- a future edit that reintroduces an OKF call from inside route_tier1 now fails this test immediately instead of relying on manual review to catch the docstring's guarantee being broken.

tkcoding: Fixed in 34122b34: added test_route_tier1_never_touches_okf, which patches get_okf_status to raise if called at all, then exercises every routing-table row (1 through 7, plus the margin_threshold opt-in) against it -- a future edit that reintroduces an OKF call from inside route_tier1 now fails this test immediately instead of relying on manual review to catch the docstring's guarantee being broken.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row2_tfidf_only_unambiguous_with_single_section_document with a genuinely single-section fixture. Verified against tfidf.py's actual _confidence logic first (not asserted blindly): with only one ranked section, unambiguous=True comes purely from having no second candidate to compare against (len(ranked) == 1), a distinct path from the margin-based two-section case every other row-2 fixture in this file exercises.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row2_tfidf_only_unambiguous_with_single_section_document with a genuinely single-section fixture. Verified against tfidf.py's actual _confidence logic first (not asserted blindly): with only one ranked section, unambiguous=True comes purely from having no second candidate to compare against (len(ranked) == 1), a distinct path from the margin-based two-section case every other row-2 fixture in this file exercises.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row2_tfidf_only_unambiguous_never_calls_route_tier2_or_okf, which mocks both route_tier2 and get_okf_status, runs route_query on the row-2 fixture, and asserts neither was called -- the existing row-2 test only checked the returned dict's shape, not that Tier 1 actually stayed self-contained.

tests/test_cascade.py
tkcoding: Fixed in 34122b34: added test_row2_tfidf_only_unambiguous_never_calls_route_tier2_or_okf, which mocks both route_tier2 and get_okf_status, runs route_query on the row-2 fixture, and asserts neither was called -- the existing row-2 test only checked the returned dict's shape, not that Tier 1 actually stayed self-contained.

tests/test_cascade.py
```diff
@@ -140,9 +303,32 @@ def test_no_candidate_with_a_fully_current_bundle_recommends_okf(self, tmp_path:
...
+        monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
+        f = _write(tmp_path, _TFIDF_ONLY_DIFFUSE_SAMPLE)
+        tier1 = route_tier1(f, "gizmo thing")
+        assert tier1["tier"] == "escalate"
+        assert tier1["reason"] == "heading_nav_no_hits_diffuse_tfidf"  # sanity: genuinely row 3
+
+        index = get_or_build_doc_index(f)
+        for section in index["retrieval_sections"]:
+            write_concept_file(f, section["line_start"], description="d", body="b")
+
```
ainetx, line 330: ### Row-3 Tier-2 test does not verify candidate-specific staleness narrowing

**Severity:** Minor

**Problem**
test_row3_tfidf_sourced_candidate_recommends_okf writes current concept files for every retrieval section before calling route_tier2. A route_tier2 implementation that incorrectly validates the whole bundle instead of only the row-3 candidate would still pass.

**How to reproduce**
1. Route a row-3 query to obtain heading_nav_no_hits_diffuse_tfidf with a TF-IDF candidate. 2. Make that candidate's concept file current. 3. Leave another section's concept file missing or stale. 4. Call route_tier2.

**Expected behavior**
Tier 2 recommends OKF because only the named TF-IDF candidate must be current.

**Actual behavior**
The shown test has all sections current and therefore cannot distinguish candidate narrowing from whole-bundle validation.

```text
row-3 TF-IDF candidate
        |
route_tier2 freshness scope
   |                    |
candidate only       all sections
   |                    |
OKF                  baseline

shown test: all current -> both paths pass
```

**Impact**
A regression that unnecessarily falls back to baseline for a valid TF-IDF-derived candidate can ship without test detection.

**Suggested correction**
Add a row-3 test with only the TF-IDF-selected section current and another section stale or missing; assert recommendation is okf.

**How to verify**
Run the cascade tests and confirm the new test fails if route_tier2 is changed to require every bundle section for a candidate-bearing row-3 result.

skills/studio/scripts/studio/utils/cascade.py
```diff
@@ -62,29 +91,96 @@ def _as_candidate(section: Dict[str, Any]) -> Dict[str, Any]:
...
+        # single real signal here is heading-nav's, unconfirmed by a
+        # second method, so it escalates rather than resolving alone.
+        return {
+            "tier": "escalate",
+            "reason": "heading_nav_only_no_tfidf_signal",
+            "candidates": [_as_candidate(nav_pick)],
+        }
 
     if tfidf_pick["line_start"] != nav_pick["line_start"]:  # pylint: disable=unsubscriptable-object
         return {
```
ainetx, line 186: ### route_tier2's row-number cross-reference in its docstring is unenforced and can silently drift

**Severity:** Minor

**Problem**
route_tier2's docstring says it is 'only called for the escalating rows: row 1 ... row 3 ... row 4, and row 7' and describes candidate-shape per row. But route_tier2's actual implementation (skills/studio/scripts/studio/utils/cascade.py, route_tier2 body) never reads or branches on tier1_result['reason'] -- it only checks `tier1_result.get('candidates', [])` truthiness. The row numbers are pure prose commentary layered on top of the real contract (candidates present vs. absent). This diff itself renumbered the table once already (old rows 1/4 -> new rows 1/3/6, and route_tier2's docstring rows 1/4 -> 1/3/6/7), requiring a fully manual, unchecked update across two docstrings in the same file. Test coverage only pins the row-number/reason mapping for row 1 (via the _NO_CANDIDATE_ESCALATION constant) and row 3 (test_row3_tfidf_sourced_candidate_recommends_okf_when_current, which asserts tier1['reason'] == 'heading_nav_no_hits_diffuse_tfidf' as a 'sanity' check) -- rows 4 and 7 feeding route_tier2 are exercised via other tests (test_current_bundle_for_candidate_recommends_okf, test_stale_bundle_for_candidate_falls_back_to_baseline_with_rebuild_flag) but those don't assert which table row produced the tier1 dict, so a future row insertion/renumbering in the table or in route_tier1's docstring could leave route_tier2's docstring citing stale row numbers with no test failure to catch it.

**How to reproduce**
1. A future PR adds a new Tier 1 outcome (e.g. a new row 8) or reorders the table. 2. The engineer updates cascade.py's module docstring table and route_tier1's docstring but forgets (or gets wrong) the parallel row-number list inside route_tier2's docstring. 3. No test asserts route_tier2's docstring text or a row-number-to-reason mapping generically (only rows 1 and 3 are spot-checked). 4. CI passes; the docstring now misdescribes which rows reach route_tier2.

**Expected behavior**
Either route_tier2 dispatches on the `reason` string directly (making the row numbers a real, checkable contract), or there is a test that enumerates all reason strings route_tier1 can return with tier='escalate' and asserts they match exactly the set the docstring claims (rows 1/3/4/7), so a table change that adds/removes/renumbers an escalating row fails a test instead of only silently going stale in prose.

**Actual behavior**
route_tier2 branches only on `candidates` presence/absence, not on `reason`; only 2 of the 4 escalating rows (1 and 3) have any test-level assertion connecting a real route_tier1() reason string to route_tier2's behavior, leaving the row-number cross-reference in route_tier2's docstring unenforced.

```text
route_tier1() returns {tier:'escalate', reason: R, candidates: C}
        |
        v (only tier=='escalate' reaches here)
route_tier2(path, tier1_result)
        |
        checks: tier1_result.get('candidates') truthy? -----> reason R is NEVER read
        |
 docstring claims: 'called for row 1/3/4/7' <-- prose only, no code/test ties row# to R
        |
 future table edit renumbers/adds a row --> docstring can silently go stale
```

**Impact**
Low-severity documentation-accuracy risk: future maintainers relying on route_tier2's docstring row references to reason about which Tier-1 outcomes reach Tier 2 could be misled after an uncaught renumbering, though the actual runtime behavior (driven by candidates presence) stays correct regardless.

**Suggested correction**
Add a small test (e.g. in TestRouteTier2 or a new cross-cutting test) that calls route_tier1 for every documented escalating scenario and asserts the resulting `reason` values are exactly the set {'no_signal_from_either_method', 'heading_nav_no_hits_diffuse_tfidf', 'heading_nav_only_no_tfidf_signal', 'diffuse_margin'} that route_tier2's docstring claims map to rows 1/3/4/7 -- so any table change that adds, removes, or renames an escalating outcome fails this test.

**How to verify**
Introduce a deliberate mismatch (e.g. rename 'diffuse_margin' without updating route_tier2's docstring) and confirm the new test fails; then revert and confirm it passes.

tests/test_cascade.py
```diff
@@ -39,29 +40,158 @@
...
+        assert result["candidates"] == [{"heading": "SectionA", "line_start": 1, "line_end": 4}]
+
+    def test_row4_heading_nav_hit_but_tfidf_has_no_signal_escalates(self, tmp_path: Path, monkeypatch):
+        """Real bug caught during review (constructorfabric/studio#137):
+        `tfidf_ranked[0]` is still a real dict entry when every section
+        scores exactly 0 -- just an arbitrary document-order tie-break, not
+        a genuine pick. The old code compared it against heading-nav's real
+        pick unconditionally, fabricating a "disagreement" (`resolved_multi`
+        with a bogus second candidate TF-IDF never actually found relevant)
+        out of a signal that was never there. Must escalate on heading-nav's
```
ainetx, line 176: ### Row-3 routing test permits unexpected response fields

**Severity:** Minor

**Problem**
test_row3_tfidf_only_diffuse_escalates_with_tfidf_pick_as_candidate uses separate assertions for tier, reason, and candidates instead of asserting the complete response object. A regression that adds an unintended field to this new routing path remains undetected.

**How to reproduce**
1. Modify the row-3 return path to include an extra key, such as "score": 0.5.
2. Run this test.
3. All three current assertions pass.

**Expected behavior**
The row-3 test should enforce the same exact return-object contract as adjacent row tests.

**Actual behavior**
The test tolerates additional keys in the returned dictionary.

```text
row-3 return
  -> extra field added
  -> selected-field assertions pass
  -> response-contract regression escapes
```

**Impact**
New row-3 API-shape regressions can reach callers without test coverage.

**Suggested correction**
Replace the three field assertions with one whole-dictionary equality assertion.

**How to verify**
Add a temporary extra key to the row-3 return value and confirm this test fails; remove it and run the test suite.

tests/test_cascade.py
ainetx, line 330: ### No route_tier2 regression test for the new heading_nav_only_no_tfidf_signal escalation reason

**Severity:** Minor

**Problem**
route_tier1's new row 4 branch (heading-nav hit, TF-IDF all-zero) produces a single-candidate escalate dict with a brand-new reason string, but tests/test_cascade.py's TestRouteTier2 class -- unlike its analogous new test for row 3 (test_row3_tfidf_sourced_candidate_recommends_okf_when_current) -- never feeds this reason/candidate shape into route_tier2.

**How to reproduce**
1. A future change to route_tier2 that special-cases on tier1_result['reason'] strings (rather than treating candidates generically) could silently mishandle 'heading_nav_only_no_tfidf_signal'. 2. No test in this diff would catch it.

**Expected behavior**
A test analogous to test_row3_tfidf_sourced_candidate_recommends_okf_when_current, feeding a real route_tier1 escalate result with reason=='heading_nav_only_no_tfidf_signal' into route_tier2 and asserting it narrows OKF-staleness checking to nav_pick's section.

**Actual behavior**
Only route_tier1's own output for this branch is tested (test_row4_heading_nav_hit_but_tfidf_has_no_signal_escalates); route_tier2's handling of this specific reason/candidate combination is untested.

```text
route_tier1 (new row 4) --escalate,1 candidate--> route_tier2 [untested for this reason] --> OKF staleness check
```

**Impact**
A latent regression in route_tier2's handling of this new escalation path could ship without any test failure.

**Suggested correction**
Add a TestRouteTier2 test that builds tier1 via a real route_tier1 call producing heading_nav_only_no_tfidf_signal, then asserts route_tier2 narrows to nav_pick's single section the same way rows 3/7 do.

**How to verify**
Run the new test and confirm it fails if route_tier2's single-candidate narrowing logic is broken or bypassed for this reason.

tests/test_cascade.py
```diff
@@ -140,9 +303,32 @@ def test_no_candidate_with_a_fully_current_bundle_recommends_okf(self, tmp_path:
...
+        index = get_or_build_doc_index(f)
+        for section in index["retrieval_sections"]:
+            write_concept_file(f, section["line_start"], description="d", body="b")
+
         result = route_tier2(f, tier1)
         assert result["recommendation"] == "okf"
+        assert result["bundle_dir"]
 
     def test_current_bundle_for_candidate_recommends_okf(self, tmp_path: Path, monkeypatch):
         monkeypatch.setattr("studio.utils.files.find_studio_directory", lambda *_a, **_k: tmp_path)
```
ainetx, line 396: ### New OKF routing test uses a loose bundle path assertion

**Severity:** Minor

**Problem**
The row-3 OKF test ends with a truthiness assertion for bundle_dir. It would accept any non-empty, incorrect path and does not verify that the returned bundle location belongs to the document's current OKF bundle.

**How to reproduce**
1. Route a row-3 query to Tier 2 with current concept files. 2. Regress route_tier2 to return an unrelated non-empty bundle_dir. 3. Run this test; its final assertion still passes.

**Expected behavior**
The test should assert the exact expected bundle path or a concrete relationship to the current document's OKF bundle.

**Actual behavior**
assert result["bundle_dir"] accepts every non-empty value.

```text
route_tier2\n  -> incorrect non-empty bundle_dir\n  -> truthiness assertion\n  -> test passes
```

**Impact**
A regression returning an unrelated OKF bundle can evade this newly added coverage.

**Suggested correction**
Assert the expected bundle directory exactly, or compare it to the known bundle path produced for this fixture.

**How to verify**
Mutate bundle_dir to an unrelated non-empty path and confirm the strengthened test fails.

tests/test_cascade.py
ainetx: Re-verified against the current code (34122b346a60fb3ce7619ae91a12cc7308f4023c) and this is still an issue -- the original finding stands.

tests/test_cascade.py
ainetx: Re-verified against the current code (34122b346a60fb3ce7619ae91a12cc7308f4023c) and this is still an issue -- the original finding stands.

tests/test_cascade.py
tkcoding: Fixed in d0d43a31: added test_row3_tfidf_sourced_candidate_narrows_staleness_to_only_that_section, which only writes a current concept file for SectionA (row 3's sole TF-IDF-sourced candidate) and deliberately leaves SectionB's missing, asserting route_tier2 still recommends okf. Verified this actually catches the bug it targets: temporarily forced route_tier2 to check status["entries"] (every section) instead of just the named candidate, reran, confirmed this new test failed (baseline instead of okf) while the pre-existing test passed unaffected, then reverted the injection before committing.

skills/studio/scripts/studio/utils/cascade.py
tkcoding: Fixed in d0d43a31: added test_docstrings_row_number_cross_reference_matches_actual_escalating_reasons, which runs route_tier1 against a fixture for all 7 documented table rows and asserts the actual escalating rows/reasons that come back are exactly the set route_tier2's docstring claims ({1: no_signal_from_either_method, 3: heading_nav_no_hits_diffuse_tfidf, 4: heading_nav_only_no_tfidf_signal, 7: diffuse_margin} -- re-derived from the current code, not assumed from this finding's text) -- so a future table change that adds, removes, or renames an escalating reason now fails a test instead of only drifting in prose.

tests/test_cascade.py
tkcoding: Fixed in d0d43a31: test_row3_tfidf_only_diffuse_escalates_with_tfidf_pick_as_candidate now asserts the complete response dict via one == equality check instead of three separate tier/reason/candidates assertions, matching the pattern the other row tests in this file already use.

tests/test_cascade.py
tkcoding: Fixed in d0d43a31: added test_row4_heading_nav_sourced_candidate_recommends_okf_when_current, modeled directly on the existing row-3 route_tier2 test, feeding a real row-4 (heading_nav_only_no_tfidf_signal) escalate result into route_tier2 and asserting it recommends okf when the heading-nav-sourced candidate's section is current.

tkcoding:
Note: the round-3 fix commit referenced in the four replies above as `d0d43a31` was missing its required DCO sign-off and got amended to `ad30b6f4` (same diff, same message, just the trailer added) and force-pushed. Same content, new SHA.

tests/test_cascade.py
ainetx: Confirmed fixed as of ad30b6f4fb99802b4e33736dcf25199734b120bd -- independent re-verification no longer reproduces this issue. reviewer:claude-sdk -- The new row-4 branch (`if not has_tfidf_signal: return {"tier": "escalate", ...}`) is a single, unconditional return that can never produce 'resolved' -- there is no other return path in that branch -- and it's exercised by test_row4_heading_nav_hit_but_tfidf_has_no_signal_escalates, which asserts tier=='escalate'. route_tier2's implementation is unchanged below the docstring edits and is driven purely by tier1_result['candidates'] (never 'reason'), so a single-item candidate list from row 4 takes the exact same narrow-to-that-section code path already exercised by the pre-existing single-candidate tests (test_current_bundle_for_candidate_recommends_okf / test_stale_bundle_for_candidate_falls_back_with_rebuild_flag) and by the new test_row4_heading_nav_sourced_candidate_recommends_okf_when_current, which builds tier1 via a real route_tier1 call and confirms route_tier2 recommends okf for that one-candidate shape -- there is no code path keyed on the reason string that could misroute it to the two-candidate (row 5) or no-candidate (row 1) branches.; reviewer:codex-sdk -- Row 4 returns only `tier: escalate` with the heading-nav candidate; route_tier2 routes any non-empty candidate list, including this reason's single candidate, through the per-line-start narrowing path rather than the whole-bundle path.; verifier:codex-sdk -- The row-4 branch unconditionally returns an escalating result with only nav_pick, and route_tier2 narrows any non-empty candidate list—including this reason’s single candidate—to its matching line_start entry rather than branching on reason or treating it as no-candidate/two-candidate.

skills/studio/scripts/studio/utils/cascade.py
ainetx: Correction: I marked this "fixed-by-new-commit" earlier, but that was a bug on my end -- the fresh re-verification actually came back "insufficient evidence" (couldn't gather enough context to confirm either way), not a genuine re-confirmation that the issue is resolved. I've fixed the underlying logic so this can't happen again. Reopening this thread and leaving the original finding in place -- it needs a human to actually check skills/studio/scripts/studio/utils/cascade.py:96, since I still don't have a verified answer either way.

ainetx (approved):
Automated review complete -- no blocking findings remain and CI is passing.

PR #156, opened 2026-09-08
fix(ci): make ci fail loudly when act job discovery yields no jobs

tkcoding:
## Summary
- Fixes a gap surfaced by an `ainetx` review comment misfiled onto PR #136 (it's about `Makefile`, not that PR's diff — see constructorfabric/studio#136).
- `make ci`'s job-discovery loop silently reported success if `act push --list` failed, changed format, or the exclusion filter removed every discovered job — the `for` loop just ran zero times.
- Now captures the job list into a variable first and fails explicitly (exit 1, clear stderr message) if it's empty, before the loop runs.

## Test plan
- Verified both paths by stubbing `act` on PATH: empty `--list` now exits 1 with the new error (previously exited 0 silently); a normal list still runs each job and excludes `sonarqube`/`code-ranker` as before.
- `make -n ci` confirms the Makefile still parses/expands correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Continuous integration now reports an error and fails when no eligible jobs are discovered, instead of incorrectly completing successfully.
  * Existing job exclusions, sequential execution, and fail-fast behavior remain unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -372,8 +372,18 @@ lint-ci:
 # event context that act cannot reconstruct for a linked worktree.
 # Runs jobs sequentially — stops on first failure.
 # Auto-detects arm64/amd64. Override: make ci ACT_FLAGS="--your-flags"
+# Job discovery failing outright, changing format, or an over-broad exclusion
+# emptying the list must never look like "nothing to do, success" -- an empty
+# job set fails loudly instead of letting a zero-iteration loop exit 0.
 ci: lint-ci
-	@for job in $$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$'); do \
+	@jobs="$$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$')"; \
+	if [ -z "$$jobs" ]; then \
+		echo "ERROR: 'act push --list' discovered no CI jobs to run -- refusing to report" >&2; \
+		echo "       success for a no-op. Check that act is installed/working and that" >&2; \
+		echo "       .github/workflows/*.yml still parses the way this Makefile expects." >&2; \
+		exit 1; \
+	fi; \
+	for job in $$jobs; do \
 		echo "▶ Running job: $$job"; \
 		act push -j $$job $(ACT_FLAGS) || exit 1; \
 	done


Makefile
```diff
@@ -372,8 +372,18 @@ lint-ci:
 # event context that act cannot reconstruct for a linked worktree.
 # Runs jobs sequentially — stops on first failure.
 # Auto-detects arm64/amd64. Override: make ci ACT_FLAGS="--your-flags"
+# Job discovery failing outright, changing format, or an over-broad exclusion
+# emptying the list must never look like "nothing to do, success" -- an empty
```
ainetx, line 376: ### act's own stderr is discarded, so the loud failure has no diagnostic payload

**Severity:** Minor

**Problem**
`act push --list $(ACT_FLAGS) 2>/dev/null` still throws away act's own stderr before the new empty check runs. When jobs is empty because act itself failed (missing binary, auth failure, workflow syntax error), the user only sees the new generic hardcoded message, never the actual act error that would explain why.

**How to reproduce**
1. Make `act` fail (e.g. rename the binary, or break workflow syntax so act errors on parse). 2. Run `make ci`. 3. `act push --list ... 2>/dev/null` produces empty stdout and its stderr is discarded. 4. jobs is empty, guard fires with only the generic message; act's real error text is never shown.

**Expected behavior**
The stated goal is to make failures visible; the actual act stderr causing the empty list should be surfaced to the user.

**Actual behavior**
stderr is discarded via 2>/dev/null and never captured/printed, so only a generic 'no CI jobs' message is shown regardless of root cause.

```text
act push --list fails (stderr) -> 2>/dev/null discards it -> jobs empty -> generic error printed, real cause lost
```

**Impact**
Users hitting an actual act failure (not a benign empty list) get no diagnostic information beyond a generic message, undermining the 'fail loudly' goal.

**Suggested correction**
Capture act's stderr (e.g. into a variable or temp file) instead of discarding it, and print it as part of the error block when jobs is empty.

**How to verify**
Break act (bad binary path or malformed workflow) and confirm `make ci`'s error output includes the underlying act error text.

Makefile
```diff
@@ -372,8 +372,18 @@ lint-ci:
...
+# job set fails loudly instead of letting a zero-iteration loop exit 0.
 ci: lint-ci
-	@for job in $$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$'); do \
+	@jobs="$$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$')"; \
+	if [ -z "$$jobs" ]; then \
+		echo "ERROR: 'act push --list' discovered no CI jobs to run -- refusing to report" >&2; \
+		echo "       success for a no-op. Check that act is installed/working and that" >&2; \
+		echo "       .github/workflows/*.yml still parses the way this Makefile expects." >&2; \
```
ainetx, line 383: ### Unquoted `for job in $jobs` remains subject to pathname (glob) expansion

**Severity:** Minor

**Problem**
`for job in $jobs; do ... done` expands $jobs unquoted, so in addition to word-splitting on whitespace, bash performs pathname expansion on each resulting word unless `set -f`/noglob is active. A discovered job name containing glob metacharacters (`*`, `?`, `[...]`) would expand against files in the current working directory before being passed to `act -j $job`, potentially substituting unrelated filenames instead of the literal job identifier.

**How to reproduce**
1. Have a workflow job whose name (as reported by `act push --list`) contains a glob character, e.g. `build*`. 2. Ensure the repo root has files matching that glob. 3. Run `make ci`. 4. The `for` loop expands `build*` against matching filenames rather than treating it as the literal job id, so `act push -j <expanded-glob-result>` is invoked incorrectly (or with multiple/wrong job args).

**Expected behavior**
Each discovered job identifier should be treated as literal data regardless of shell glob metacharacters, e.g. by disabling globbing (`set -f`) around the loop or using a read-based loop.

**Actual behavior**
The unquoted `for job in $jobs` expansion still performs pathname expansion on each job token, same as before the refactor.

```text
job name contains '*' -> unquoted for-loop expansion -> shell globs against cwd files -> wrong argument passed to act -j
```

**Impact**
A workflow job name containing glob characters could cause act to be invoked with unintended file-derived arguments instead of the real job id, producing confusing failures or skipped jobs.

**Suggested correction**
Disable globbing for the loop (e.g. `set -f; for job in $jobs; do ...; done; set +f`) or iterate with `printf '%s\n' "$jobs" | while read -r job; do ...; done` to avoid pathname expansion.

**How to verify**
Create a job name containing '*' matching a real file in the repo root, run `make ci`, and confirm the job id passed to `act -j` is the literal name, not glob-expanded filenames.

Makefile
```diff
@@ -372,8 +372,18 @@ lint-ci:
...
 # Auto-detects arm64/amd64. Override: make ci ACT_FLAGS="--your-flags"
+# Job discovery failing outright, changing format, or an over-broad exclusion
+# emptying the list must never look like "nothing to do, success" -- an empty
+# job set fails loudly instead of letting a zero-iteration loop exit 0.
 ci: lint-ci
-	@for job in $$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$'); do \
+	@jobs="$$(act push --list $(ACT_FLAGS) 2>/dev/null | tail -n +2 | awk '{print $$2}' | grep -Ev '^(sonarqube|code-ranker)$$')"; \
+	if [ -z "$$jobs" ]; then \
+		echo "ERROR: 'act push --list' discovered no CI jobs to run -- refusing to report" >&2; \
+		echo "       success for a no-op. Check that act is installed/working and that" >&2; \
```
ainetx, line 382: ### Empty-job error message doesn't distinguish exclusion-caused emptiness from a genuine discovery failure

**Severity:** Minor

**Problem**
The guard treats 'zero jobs after grep -Ev exclusion' identically to 'act found nothing at all', and the remediation text only points at act/parsing, never at the exclusion filter.

**How to reproduce**
1. Add/modify .github/workflows so the push event only triggers jobs named 'sonarqube' and/or 'code-ranker'.
2. Run `make ci`.
3. `act push --list` returns those job names, the grep -Ev filter removes all of them, jobs is empty, and the new guard fires with a message blaming act/parsing.

**Expected behavior**
The error message (or a preceding check) should differentiate 'act returned zero jobs' from 'act returned jobs but the static exclusion filtered all of them out', so operators aren't sent chasing a nonexistent act/parsing bug.

**Actual behavior**
A correctly-functioning discovery+filter pipeline that legitimately yields zero runnable jobs produces the same misleading 'check that act is installed/working... .yml still parses' message as an actual tooling failure.

```text
workflows define only sonarqube/code-ranker -> act --list finds them -> grep -Ev excludes both -> jobs=empty -> guard fires -> message blames act/parsing (wrong root cause)
```

**Impact**
Wastes debugging time chasing a non-existent tooling problem when the real situation is a filter needing adjustment or a workflow needing more jobs.

**Suggested correction**
Report count of jobs discovered before exclusion vs after, or mention the sonarqube/code-ranker exclusion explicitly in the error text as a possible cause.

**How to verify**
Simulate act output containing only sonarqube/code-ranker job names and confirm the emitted error message names the exclusion filter as a possible cause.

Makefile
```diff
@@ -372,8 +372,18 @@ lint-ci:
 # event context that act cannot reconstruct for a linked worktree.
 # Runs jobs sequentially — stops on first failure.
 # Auto-detects arm64/amd64. Override: make ci ACT_FLAGS="--your-flags"
+# Job discovery failing outright, changing format, or an over-broad exclusion
+# emptying the list must never look like "nothing to do, success" -- an empty
+# job set fails loudly instead of letting a zero-iteration loop exit 0.
```
ainetx, line 377: ### No validation that extracted tokens are genuine job identifiers before execution

**Severity:** Minor

**Problem**
`awk '{print $2}'` blindly takes the second whitespace-delimited field of each `act --list` line; if act's list output format changes (extra/missing column, different field order), the guard only checks for a non-empty string, not that the string is a valid job name, so garbage values would flow into `act push -j $job`.

**How to reproduce**
1. Suppose a future act version reorders `--list` columns so field 2 is no longer the job ID (e.g. it's now the workflow file name).
2. `make ci` runs; `jobs` is non-empty (populated with workflow filenames, not job IDs), so the new guard does not fire.
3. The for loop invokes `act push -j <workflow-filename>` for each, which is not a valid job identifier.

**Expected behavior**
The recipe should validate that each extracted token corresponds to a real, expected job identifier (e.g. cross-check against `act push --list` job names again, or require a minimal recognizable pattern) before invoking `act push -j`.

**Actual behavior**
Only an empty/non-empty check is performed; malformed-but-nonempty output silently proceeds to execution with wrong `-j` arguments.

```text
act output format changes -> awk '{print $2}' extracts wrong column -> jobs non-empty (guard passes) -> for loop calls 'act push -j <bad-value>' -> act likely errors per-job, but the *guard's own invariant* (no real-job validation) is unmet
```

**Impact**
The stated goal ('job discovery... changing format... must never look like success') is only partially met — it addresses total emptiness, not a format change that still yields nonempty-but-wrong tokens.

**Suggested correction**
Add a sanity check on each token (e.g. verify it appears in a known job list or matches an expected identifier pattern) before invoking `act push -j $job`.

**How to verify**
Simulate a reformatted act --list output and confirm the recipe either rejects it up front or fails clearly rather than passing malformed values to act push -j.

Makefile
```diff
@@ -372,8 +372,18 @@ lint-ci:
 # event context that act cannot reconstruct for a linked worktree.
```
ainetx, line 372: ### No regression-guarding test/assertion added for the new empty-job check

**Severity:** Minor

**Problem**
The fix for a silent-success regression lives entirely as inline shell logic in the Makefile recipe with no accompanying automated check (unit test, CI job, or even a lightweight grep-based assertion) verifying the guard-and-exit-before-loop structure remains intact.

**How to reproduce**
1. A later contributor refactors the `ci` target (e.g. reorders the guard after the for loop, or removes it while 'simplifying').
2. No test or CI step fails, because nothing in the repo asserts the guard's presence/position.
3. The silent-success bug this diff fixes could be reintroduced undetected.

**Expected behavior**
Some diff-visible, maintainable verification (a test invoking `make ci` with a mocked empty job list expecting nonzero exit, or a documented/CI-enforced check) tying future changes to this invariant.

**Actual behavior**
Only a prose comment above the recipe documents the intent; nothing executable enforces it.

```text
guard added only as shell code -> no test asserts its presence/behavior -> future refactor can silently remove/relocate it -> regression reintroduced undetected until an actual CI outage
```

**Impact**
The fix is fragile against future edits to the same Makefile target; the class of bug being fixed here could silently return.

**Suggested correction**
Add a minimal test (e.g. a script or CI step that stubs `act` to emit no jobs and asserts `make ci` exits non-zero) to lock in the behavior.

**How to verify**
Check that a subsequent PR removing/relocating the guard would be caught by an added test or CI assertion.

ainetx (approved):
Nice fix for a real silent-success gap — the empty-job guard correctly turns a zero-iteration loop into a loud failure instead of a false green. Logic, placement (before the loop, after job discovery), and error messaging are sound, and I didn't find anything blocking. A few minor polish items worth a look when convenient:

- **Discarded act stderr leaves the new guard's failure message uninformative** — `2>/dev/null` still swallows act's own error output, so when the job list is empty because act itself failed (missing binary, auth, workflow syntax error), the user only sees the generic guard message, not the underlying cause. ([comment](https://github.com/constructorfabric/studio/pull/156#discussion_r3954398895))
- **Unquoted `$jobs` in the `for` loop is subject to glob expansion** — beyond word-splitting, bash will pathname-expand any job name containing `*`, `?`, or `[...]` against the CWD before it reaches `act -j`. ([comment](https://github.com/constructorfabric/studio/pull/156#discussion_r3954399544))
- **Guard doesn't distinguish exclusion-caused emptiness from a real discovery failure** — the error text only points at act/parsing, not at the `grep -Ev` exclusion filter, which could itself be the cause of an empty list. ([comment](https://github.com/constructorfabric/studio/pull/156#discussion_r3954400190))
- **No validation that extracted tokens are real job identifiers** — the `awk '{print $2}'` extraction only checks for non-empty output, not that it's a well-formed job name, so a format change in `act --list` could pass garbage through silently. ([comment](https://github.com/constructorfabric/studio/pull/156#discussion_r3954400787))
- **No regression test for the new guard** — the fix for a silent-success bug lives entirely as inline Makefile shell with nothing (test, CI check, or even a grep assertion) protecting the guard-before-loop structure from drifting back. ([comment](https://github.com/constructorfabric/studio/pull/156#discussion_r3954401409))

None of these affect correctness of the core fix — approving as is.

PR #154, opened 2026-09-07
change-summary: shallow clones, a shared ceiling, a streamed sweep, warnings

Oleg67:
Follow-up to #133, which merged (`4ef3055f`) while the third review round on it was being answered. These are the fixes for the findings raised on #133 and #149 against the linkage and window halves in that round — the commit landed on #133's branch minutes after the merge, so it is carried here on its own. #149 is restacked on this branch.

## What this fixes, each reproduced first

- **A shallow clone was reported as "no merge base".** A depth-1 CI checkout cannot support a claim about history. After a merge-base miss the window now asks `git rev-parse --is-shallow-repository` and reports `no merge base in a shallow history; the branch point may lie beyond the fetched depth` instead; a miss in full history still says "no merge base". Tested with a real `--depth 1` clone over `file://`.
- **`--base` was silently ignored whenever `--since` was given.** They compose: the base anchors the changed-file diff and the bound replaces only the decision boundary. A bad base is refused even with a bound.
- **The ceiling was filled from the diff first**, so brand-new files were exactly what a large change set dropped. Once the ceiling bites, tracked and untracked entries share it in turn; below it the order is unchanged.
- **The untracked sweep was captured whole before the ceiling applied.** It is now streamed: records kept up to the ceiling, counted past it. The pipe is pumped on a helper thread against the query timeout, so a git that holds stdout open in silence is killed and reported as a tool failure rather than waited on forever — a blocking read on the main thread was bounded by nothing.
- **Git failing to launch was logged at debug**, indistinguishable from a routine miss. Both helpers warn with the exception type, not its message.
- **The scope check passed a fabricated one-suffix allowlist** to the shared resolver, which does not consult it for a file. It passes none, and the docstring and spec say why no extension policy applies here: a changed artifact declares requirements without carrying a code extension, so filtering it out would report a changed specification as tracing to nothing.
- **`read_code_text` treats NUL bytes as binary**, the rule the document reader already applied, so the two readers agree on what is text.
- **Direct tests** for the shared reader (limit, limit+1, non-positive limit, missing file, invalid UTF-8, NUL), `from_path`/`from_text` parity on valid and malformed fixtures, and a new `tests/test_document_scan_lines.py` covering `scan_cpt_id_lines` directly and proving both readers yield identical hits for LF and CRLF content.

Three new traced blocks and steps (shallow check, streamed sweep, shared ceiling); spec steps 5, 7, 19 and 21 reworded.

## Round 4

[...]

PR #160, opened 2026-09-08
change-summary: stop translating newlines inside a path, so both readers name one file alike

Oleg67:
Follows #154. A path containing a carriage return was reported under a name that opens nothing, and a tracked file present on disk came back as deleted.

## The defect

#154 gave both readers of the changed-file listing the same codec, so they would agree about a path's name under any locale. That was not sufficient. `text=True` switches on **universal newlines** as well as decoding, and `subprocess` exposes no way to turn it off — so the captured reader translated CR *inside a filename* while the streamed sweep, decoding raw slices, kept it.

Measured, on a repository holding `sweep\rname.txt`:

```
text=True path : 'sweep\nname.txt'
raw-bytes path : 'sweep\rname.txt'
EQUAL? False
```

## Two consequences, the second worse than the first

**Deduplication misses the file.** The sweep is deduplicated against the diff by string comparison, so a path appearing in both listings — `git rm --cached` on a file present in the base commit leaves it deleted in the index and untracked on disk — became two contradictory rows for one file, inflating every counter.

**A present file is reported as deleted.** This is the one that matters. The translated name opens nothing, so the entry resolver found no file where git had just said one changed:

```
rows: [('tracked\nname.txt', 'A', 'file no longer present')]
deleted: 1
```

That is a false statement about a file sitting right there, not a merely incomplete listing — the failure class this module exists to prevent, arriving through the decode rather than through the counting. After the fix:

```
rows: [('tracked\rname.txt', 'A', '')]
deleted: 0
```

## The fix

The captured reader takes bytes and decodes each record with the same call the pump makes:

```python
raw = result.stdout.split(b"\0")
if raw and not raw[-1]:
    raw.pop()
return [record.decode(_PATH_ENCODING, _PATH_ERRORS) for record in raw]
```

Nothing in `-z` output needs translating: only NUL separates records, and a path may contain any byte but NUL and `/`. So anything translated is corruption.

`_git_query` keeps `text=True` deliberately — it reads refs, shas and timestamps, values git itself forbids a control character in, and it never carries a path.

## Tests

| test | what it pins |
|---|---|
| `test_both_readers_keep_the_bytes_git_gave_them` | parametrized `cr` / `crlf` / `lf` — the translation rewrites each differently, so all three are checked; `lf` is the control |
| `test_a_tracked_path_holding_a_cr_is_not_reported_as_deleted` | the worse consequence: the row resolves, is read, and `deleted == 0` |
| `test_a_cr_path_in_both_listings_is_deduplicated` | the reported consequence, end to end through `git rm --cached` |
| `test_every_git_call_site_runs_with_the_sanitised_environment` | strengthened: one `text=True` in the module, one shared decode call per record reader |

Reverting the fix fails four of the five; `lf` passes, as it should.

**One thing worth flagging about my own test.** The first version of the deduplication test passed the very mutation it was written to catch — the file was committed *after* the base ref, so `git diff` never reported it and only the sweep saw it, which is one row whether or not the readers agree. The base now contains the file. The structural check needed a second correction too: it counted `text=True` as a substring, and the comments here discuss it several times, so it was measuring the prose. It counts the keyword-argument form now.

## Gates

Pinned to `bb364ada`, on `main` @ `ac9e06e9`.

| Gate | Result |
|---|---|
| `make test` | **5,570 passed**, 4 skipped, 15 xfailed |
| `cfs validate` | 0 errors |
| `spec-coverage --system studio` | **0.4603** against the 0.46 floor, unchanged — the fix adds no lines to the module's traced blocks |
| `make pylint` / `make vulture-ci` | 10.00 / clean |

No behaviour changes for a path without a CR, which is why the granularity and the rest of the suite are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of file paths containing carriage returns, CRLF sequences, newlines, or non-UTF-8 bytes.
  * Prevented valid tracked files from being incorrectly reported as deleted.
  * Maintained consistent results across captured and streamed Git output.
  * Continued removing duplicate file listings.

## Round 1

CR and LF are both in the 1--31 range Windows forbids in a file name, so all three parametrized cases would have failed there rather than skipped -- `lf` included, which I had picked as the control without thinking that through. Guarded in `b2246a39` with the convention this file already uses twice, for a tab and for POSIX permission bits. CI is ubuntu-only so nothing was red, but these are precisely the tests a contributor on Windows would want to run, and they would have raised `OSError` from `open()` rather than skipping with a reason.

## Round 2

The three filesystem tests skip on Windows, so this regression class had no cover there. Two portable tests close it, and the mocking suggestion needed a correction to work at all: patching `subprocess.run` replaces the code that does the translating, so a fake `stdout` proves nothing about text mode -- my first attempt failed the mutation only because the text branch then split bytes with a str separator, which is a real failure for the wrong reason.

So the job is split. One test drives a real subprocess whose output holds a CR -- any Python interpreter, no git and no odd file name -- and reads it both ways, pinning the translation itself. The other runs both readers over identical faked output, with the fake honouring the `text`/`encoding` keywords the way `subprocess` does, so restoring `text=True` fails it on the corrupted value (`'cr\nname.py' != 'cr\rname.py'`) rather than on a type error. The first licenses the second, by proving the real translation is exactly what the fake performs.

### DIFF
--- a/tests/test_change_summary_links.py
+++ b/tests/test_change_summary_links.py
@@ -950,6 +950,8 @@ class TestTheCeilingIsSharedAndTheSweepIsStreamed:
         anything translated is corruption. Every newline shape is checked, because the
         translation rewrites lone CR, CRLF and neither in three different ways.
         """
+        if os.name == "nt":
+            pytest.skip("CR and LF are not legal file-name characters on Windows")
         repo = _repo_with_base(tmp_path)
         with open(os.path.join(os.fsencode(repo), raw), "wb") as handle:
             handle.write(b"x = 1\n")
@@ -971,6 +973,8 @@ class TestTheCeilingIsSharedAndTheSweepIsStreamed:
         reported as "file no longer present", which is a false statement about a
         present file rather than a merely incomplete one.
         """
+        if os.name == "nt":
+            pytest.skip("a carriage return is not a legal file-name character on Windows")
         repo = _repo_with_base(tmp_path)
         with open(os.path.join(os.fsencode(repo), b"tracked\rname.py"), "wb") as handle:
             handle.write(_code().encode("utf-8"))
@@ -989,6 +993,8 @@ class TestTheCeilingIsSharedAndTheSweepIsStreamed:
         the index and untracked on disk, so it appears in both listings — and with the
         two readers naming it differently the sweep's dedup missed it, so one file was
         two rows and every counter was inflated."""
+        if os.name == "nt":
+            pytest.skip("a carriage return is not a legal file-name character on Windows")
         repo = _make_repo(tmp_path / "r")
         with open(os.path.join(os.fsencode(repo), b"both\rname.py"), "wb") as handle:
             handle.write(b"x = 1\n")
--- a/tests/test_change_summary_links.py
+++ b/tests/test_change_summary_links.py
@@ -654,10 +654,18 @@ class TestTheListingIsAboutTheProjectNamedAndNothingElse:
 
         assert launches >= 3
         assert source.count("env=_git_env()") == launches, "every launch, not most of them"
-        # And every captured launch decodes with the codec the streamed reader uses, so
-        # the two readers of one path cannot disagree about its name.
-        assert source.count("encoding=_PATH_ENCODING,") == source.count("subprocess.run(")
-        assert "raw.decode(_PATH_ENCODING, _PATH_ERRORS)" in source
+        # And no reader of a *path* may use `text=True`, which switches on universal
+        # newlines as well as decoding and so translated CR inside a filename. Only
+        # `_git_query` may, because it reads refs, shas and timestamps -- values git
+        # itself forbids a control character in. Both record readers instead decode raw
+        # slices with the one shared call, which is what makes one path one string.
+        # Counted with the trailing comma, which is the keyword-argument form: the
+        # comments here discuss `text=True` several times, and a bare substring count
+        # measured the prose as well as the code.
+        assert source.count("text=True,") == 1, "only the single-line reader may translate"
+        assert source.count(".decode(_PATH_ENCODING, _PATH_ERRORS)") == 2, (
+            "the captured record reader and the streamed one, each decoding for itself"
+        )
 
     def test_a_failed_untracked_sweep_makes_the_listing_unavailable(self, tmp_path, monkeypatch):
         """`or []` turned a failed `ls-files` into an empty one, so the report came back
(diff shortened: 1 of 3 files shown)


tests/test_change_summary_links.py
```diff
@@ -930,6 +938,75 @@ def test_both_readers_decode_one_path_to_one_string(self, tmp_path, monkeypatch)
         assert captured == streamed == ["café.py"]
         assert total == 1
 
+    @pytest.mark.parametrize("raw", [b"cr\rname.py", b"crlf\r\nname.py", b"lf\nname.py"],
+                             ids=["cr", "crlf", "lf"])
```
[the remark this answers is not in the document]
Oleg67: Correct, and fixed in `b2246a39`. CR and LF both sit in the 1–31 range Windows forbids in a file name, so all three parametrized cases would have failed there rather than skipped — `lf` included, which I had not thought through when I chose it as the control case.

Guarded with the convention this file already uses twice:

```python
if os.name == "nt":
    pytest.skip("CR and LF are not legal file-name characters on Windows")
```

Applied to all three tests that build such a path — the parametrized reader comparison, `test_a_tracked_path_holding_a_cr_is_not_reported_as_deleted` and `test_a_cr_path_in_both_listings_is_deduplicated`. It matches `test_a_path_with_a_tab_is_linked_not_reported_gone` (a tab, same reason) and the unreadable-log test (POSIX permission bits) in the same file.

Worth being clear about the severity, in both directions. CI is `ubuntu-latest` for every job, so this was latent rather than red — nothing was going to fail on this PR. But the defect being fixed here is *specifically* about a control character in a path, so these are exactly the tests a contributor on Windows would want to run and exactly the ones that would have blown up in their face with an `OSError` from `open()` rather than a skip that says why.

Suite unchanged at **5,568 passed**, 4 skipped (now 4 skipped on POSIX, 7 on Windows); `cfs validate` 0 errors; granularity 0.4603.

tests/test_change_summary_links.py
```diff
@@ -930,6 +938,81 @@ def test_both_readers_decode_one_path_to_one_string(self, tmp_path, monkeypatch)
...
+    def test_both_readers_keep_the_bytes_git_gave_them(self, tmp_path, raw):
+        """The codec was shared, but `text=True` also switches on *universal newlines*,
+        and `subprocess` offers no way to turn that off.
+
+        So the captured reader translated a CR in a path to LF while the streamed one,
+        decoding raw slices, kept it. A path may hold any byte but NUL and `/`, and only
+        NUL separates records here — nothing in this output needs translating, so
+        anything translated is corruption. Every newline shape is checked, because the
```
ainetx, line 950: ### CR/LF filename regression coverage is entirely skipped on Windows with no mocked fallback

**Severity:** Minor

**Problem**
The three new tests covering the text=True universal-newlines corruption (test_both_readers_keep_the_bytes_git_gave_them, test_a_tracked_path_holding_a_cr_is_not_reported_as_deleted, test_a_cr_path_in_both_listings_is_deduplicated) all skip via `if os.name == "nt": pytest.skip(...)` because they create real files with CR/LF in the name, which Windows filesystems reject. No test anywhere in tests/test_change_summary_links.py or tests/test_change_summary_core.py exercises the same decode/dedup logic by mocking subprocess.run/Popen output containing embedded CR bytes (a pattern the suite already uses elsewhere, e.g. test_an_unterminated_final_record_is_kept_and_counted mocks Popen's stdout directly), so this regression class has no platform-independent test.

<details>

**How to reproduce**
1. Run the test suite on a Windows CI runner. 2. Observe the three CR/LF tests are skipped (not failed). 3. Introduce a regression that reintroduces text=True (or otherwise lets universal-newline translation corrupt a path) in _git_records. 4. On Windows CI, no test fails because all coverage of this behavior was skipped.

**Expected behavior**
A platform-independent test that mocks subprocess.run/Popen to return raw bytes containing an embedded CR (e.g. b"cr\rname.py\0") and asserts cs._git_records / cs._git_records_bounded decode it verbatim, would catch this regression on every platform including Windows.

**Actual behavior**
Coverage of the CR/LF decode-fidelity bug is provided only by tests that depend on creating an actual CR/LF-containing file on disk, which is impossible on Windows and therefore skipped there.

```text
Windows CI run
-> pytest collects 3 CR/LF tests
-> os.name == 'nt' -> pytest.skip()
-> regression in _git_records (e.g. text=True reintroduced) ships undetected on Windows runners
```

**Impact**
A regression reintroducing CR/LF translation in the git output readers would only be caught on POSIX runners; Windows CI would report green despite the corruption bug being reintroduced.

**Suggested correction**
Add a test that mocks subprocess.run (for _git_records) and/or Popen's stdout (for _git_records_bounded) to emit bytes with an embedded CR/CRLF, and assert the decoded result preserves the CR bytes — this needs no real filesystem interaction and can run unconditionally on all platforms.

**How to verify**
Run the new mocked test on a Windows environment (or with os.name patched) and confirm it fails when text=True is reintroduced into _git_records, and passes with the current fix.

</details>

Oleg67: Right that the class had no cover on Windows. Two portable tests in `bb364ada` — and the mocking suggestion needed one correction to work, which is worth spelling out because it is the interesting part.

**A mock alone cannot catch this defect.** Patching `subprocess.run` replaces the very code that does the translating, so a fake `stdout` proves nothing about text mode. My first attempt did exactly that: it failed the mutation only because the text-mode branch then called `.split("\0")` on bytes — a real failure, for the wrong reason, and it would have stopped catching anything the moment that type mismatch was resolved some other way.

So the two tests split the job.

**1. The mechanism, demonstrated rather than described.** A real subprocess whose output holds a CR — any Python interpreter, no git and no odd file name, so it runs everywhere the suite does:

```python
emit = [sys.executable, "-c", r"import sys; sys.stdout.buffer.write(b'cr\rname.py\x00')"]

as_text = subprocess.run(emit, capture_output=True, text=True,
                         encoding=cs._PATH_ENCODING, errors=cs._PATH_ERRORS, check=True).stdout
as_bytes = subprocess.run(emit, capture_output=True, check=True).stdout

assert as_bytes.split(b"\0")[0] == b"cr\rname.py"
assert as_text.split("\0")[0] == "cr\nname.py", "text mode rewrites the CR"
```

**2. The paired decode, with a faithful fake.** The fake for `run` honours the `text`/`encoding` keywords the way `subprocess` itself does — decode, *then* translate newlines — rather than ignoring them:

```python
def _fake_run(*_a, **kwargs):
    if not (kwargs.get("text") or kwargs.get("encoding")):
        return subprocess.CompletedProcess([], 0, record, b"")
    decoded = record.decode(kwargs.get("encoding") or "utf-8", kwargs.get("errors") or "strict")
    translated = decoded.replace("\r\n", "\n").replace("\r", "\n")
    return subprocess.CompletedProcess([], 0, translated, "")
```

That is what makes it fail on the *value*. With `text=True` restored:

```
E       AssertionError: the captured reader keeps both newline shapes
E       assert ['cr\nname.py...rlf\nname.py'] == ['cr\rname.py...f\r\nname.py']
E         At index 0 diff: 'cr\nname.py' != 'cr\rname.py'
```

And test 1 is what licenses the fake — it proves the real translation is exactly `\r\n` → `\n` and `\r` → `\n`, so the fake is not inventing a behaviour to then assert.

**Where the protection now comes from, on each platform:**

| guard | POSIX | Windows |
|---|---|---|
| three filesystem tests, real CR paths end to end | ✅ | skipped |
| the translation mechanism (test 1) | ✅ | ✅ |
| paired decode with the faithful fake (test 2) | ✅ | ✅ |
| `text=True,` counted once in the module (structural) | ✅ | ✅ |

Both new tests carry `crlf` as well as `cr`, since the translation rewrites the two differently and only one of them is a lone control character.

Suite at `bb364ada`: **5,570 passed**, 4 skipped; `cfs validate` 0 errors; granularity 0.4603; pylint and vulture-ci clean.

ainetx: **RESOLVED** -- Resolved as fixed on bb364ad. The three filesystem-dependent CR/LF pathname tests remain correctly skipped on Windows, but Windows CI now retains portable coverage of the underlying regression: one unconditional test proves real subprocess text-mode newline translation corrupts embedded CR bytes, and another feeds CR/CRLF records through both Git readers with a subprocess-faithful fake and asserts identical decoded paths. Restoring text=True fails the portable paired-reader test on the translated value, so the decode mismatch behind both deduplication and false-deleted reporting is covered on every platform.

ainetx (approved):
Verified the fix and its test coverage. The change correctly identifies that `text=True` was the real culprit — not just the codec but subprocess's universal-newline translation, which had no off switch — and moves both the captured and streamed readers to decoding raw bytes with the shared codec, leaving `text=True` only where it's safe (non-path values). The new tests directly pin the regression class (CR/CRLF/lone-LF filenames, the tracked-file-reported-as-deleted case, and the cross-listing dedup failure), plus a platform-independent test that drives a real subprocess to prove the translation mechanism itself, and the source-scanning test now correctly distinguishes prose from code when counting `text=True` usages. No correctness, security, or test-coverage concerns found. Approved.

Issue #159, opened 2026-09-08
Unit test performs live network I/O, so CI fails nondeterministically (reported as a coverage failure)


SanjeevSolanki:
`tests/test_migrate_from_cypilot.py::test_implicit_migration_allows_install_dir_matching_legacy_dir` performs live network I/O, so the suite fails whenever the request does not succeed. It failed on run 34213892905 for a documentation-only change.

**What happens**

The test drives `cfs init` with `--migrate-from-cypilot=yes`. Its `followup_update_ok` fixture stubs `migrate_from_cypilot._run_followup_update`, but that does not cover the kit installation `init` performs, so the real GitHub download runs.

On that run, version resolution fell back to `0` and the download 404'd:

```
WARNING studio.commands.kit:kit.py:4252 Kit 'sdlc' GitHub download failed
RuntimeError: Failed to download kit from GitHub
  (constructorfabric/studio-kit-sdlc@0): HTTP Error 404: Not Found
```

`constructorfabric/studio-kit-sdlc` is public and reachable, with tags `v1.0.0` through `v1.2.1` — there is no tag `0`, so the `@0` is a symptom of the failed resolution rather than the cause.

The 404 downgraded `status` from `PASS` to `WARN`, and the test asserts `PASS`:

```
AssertionError: assert 'WARN' == 'PASS'
= 1 failed, 5502 passed, 4 skipped, 15 xfailed, 58 subtests passed =
```

**Why it is worth fixing rather than re-running**

- The job is named `Coverage (≥90%)`, so the failure reads as a coverage problem. Coverage was fine on that run: `TOTAL 29037 1608 94%`. The name sends every investigation in the wrong direction first.
- `tests/conftest.py` has no network guard, so nothing prevents this or any other test from reaching the network. The whole suite is nondeterministic by construction, not just this case.
- Every contributor's PR runs the same suite, so the flake is shared.

**Suggested fix**

1. Stub the kit installation `init` performs, in the same fixture or a shared one, so the test exercises the migration path without network access.
2. Consider a socket guard in `tests/conftest.py` that fails any test attempting an outbound connection unless explicitly marked — that converts a future flake into a clear, local error.
3. Independently: make failed version resolution raise rather than fall back to `0`, so the error names the real problem instead of producing a 404 for a version nobody asked for.

SanjeevSolanki:
**Note on the two obvious-looking fixes, since both have a trap.**

[...]

PR #153, opened 2026-09-07
feat(pdsl): declare gate risk on a MENU with a TYPE header and lint it

SanjeevSolanki:
Closes #152.

The validator does not recognise a `TYPE` section header at all, so a gate risk declaration is completely unchecked. A literal value, a variable, a value with a trailing `WHEN`, a misspelled header and no header at all all pass silently — so a typo downgrades a gate to undeclared with nothing reported, and any declared-risk model rests on a lint with no opinion about the thing it is meant to enforce.

## What this adds

A `TYPE` sub-header on a `MENU`, and four rules in a new `PDSL700` band:

| rule | |
|---|---|
| `PDSL700` | the value must be one literal enum token — `confirmation`, `decision`, `blocking` |
| `PDSL701` | at most one `TYPE` per MENU, so risk cannot depend on read order |
| `PDSL702` | a declaration where nothing reads it: outside a MENU, nested in its body, or trailing it |
| `PDSL703` | a sub-header that is a near-miss of `TYPE`, so a typo cannot silently leave a gate undeclared |

**An absent `TYPE` stays valid** and is treated as `blocking` at runtime, so no existing menu changes behaviour and the surface migrates gate by gate. `test_the_untyped_menu_surface_does_not_grow` freezes today's undeclared menus and fails on a newly added one, so the untyped surface cannot grow. That guard is built from the validator's own block scanner and regexes, so it cannot disagree with the checker it guards about what a MENU is or where a declaration is read.

**Nothing consumes the declaration yet**, so this is inert on shipped PDSL — enforced rather than asserted: a test pins the whole finding tally across every validated root, not only the new band, because this change touches shared parsing.

## Two design points

**A declaration is read in a MENU's *declaration region*** — from its header to the first recognised section other than `TITLE` — rather than by tracking where a MENU ends. Tracking the latter left the menu-numbering checks suppressed for the rest of the block.

**A malformed declaration is recognised by discarding decoration** and folding Unicode per character, rather than by listing the forms it can take, because that list is open-ended. `architecture/specs/PDSL.md` states what is consequently *not* detected (`1. TYPE:`, `<b>TYPE</b>:`, `TYPE(gate):`) and that every such case leaves the gate undeclared and therefore `blocking` — the failure direction is more friction, never more autonomy.

## Verification

- `make test` 5331 passed — 49 new tests, 942 of the 1325 added lines are tests
[...]

PR #151, opened 2026-09-07
docs(adr): record autonomous interaction default and declared gate risk

SanjeevSolanki:
## What

Records the decision that Studio's interaction default becomes autonomous, and the gate-risk model that makes that safe. Documentation only — one new ADR plus its registration in `DESIGN.md`. No engine change.

## Why now

Two things make the current shape worse than it needs to be:

- the session-opening mode gate stops the first turn of every non-exempt workflow, and the option it suggests — `normal` — is a documented no-op: `simple-mode-normal.md` instructs the runtime to *continue with the workflow's existing menus, gates, stops*. **Normal is therefore the most question-heavy mode available.**
- an autonomous-by-default overlay already exists, with a risk-eligibility checklist and a blocked-action taxonomy, but it is opt-in and re-derives risk per menu at runtime rather than reading a declaration.

Nothing in ADR-0001 through ADR-0022 covers interaction modes, so both the default and any risk model would otherwise be introduced with no recorded basis.

## What the ADR records

- **Four modes.** `guided` is distinct from `assistant`, so the current `normal` behaviour survives under a name rather than being removed. The active mode is **announced, not asked**, and is session state — a cancel or an off-protocol reply does not reset it.
- **Declared gate risk.** Each gate declares `confirmation` / `decision` / `blocking` as a **static constant** of its `MENU` block, never computed at runtime. **An undeclared gate is treated as `blocking`**, so core gains the capability without changing the behaviour of any installed kit or legacy path until its author declares a type.
- **The resolution invariant.** A `decision` gate may resolve autonomously only from a structured, explicitly keyed source using exact-match semantics — no inference from prose, no similarity matching, no normalization at lookup time, and no cached resolution state no longer valid for the current inputs. This is what makes autonomous resolution deterministic and therefore reviewable.
- **The risk boundary does not move.** A `blocking` gate is passable only by a fresh explicit user authorisation satisfying that gate; *auto-proceed* and *proceed after explicit approval* are distinct. Validators, prerequisite checks and closing audits run identically in every mode — autonomy changes who answers the questions, never what gets checked.
- **The governing consequence:** autonomous mode does not grant authority; it selects how already-declared authority is exercised.
- **Bounds ADR-0018 rather than superseding it** — delegated unattended execution stays valid; this ADR governs interactive-session autonomy, so the two do not coexist without a stated division.

Numeric acceptance thresholds are deliberately excluded: those are evaluation policy, not architectural decisions, and an ADR carrying a number goes stale.

## Gates

[...]

PR #162, opened 2026-09-09
fix(pdsl): carve out declared mode/gate-type resolution from anti-reinterpretation rules

### LINKED ISSUE #148: PDSL: carve out declared-mode/declared-gate-type resolution from the anti-reinterpretation rules
## Problem

`runtime/pdsl-execution-card.md:28-33` and `runtime/active-workflow-state-law.md:7,9,12` restrict how Studio may interpret a user's message — correctly, and that must survive. But as worded they also catch Studio following its own declared behaviour: auto-proceeding a gate the workflow itself typed, under a mode the user chose. `:9` says prefer workflow-owned menus "over ad-hoc interpretation"; `:12` says never skip the active workflow's menus. Neither was written with a declared gate type in mind, and an implementer would be guessing.

This is currently the one thing blocking the typed-gate work from starting.

## Proposed edits (all narrowing, nothing deleted)

> **Note (2026-09-07):** edits 1, 3, and 4 below are the corrected versions after review feedback ([comment](https://github.com/constructorfabric/studio/issues/148#issuecomment-5563151668)). The original draft of edit 3 is superseded — see the reply thread for why.

1. **`pdsl-execution-card.md`** — keep "do not reinterpret the user's reply as broad permission"; add that the session's declared mode **and** a gate's declared risk type, *stated as the literal type named in the MENU block*, are workflow-owned behaviour, not an interpretation of a reply, and are outside the prohibition. Must state explicitly that a runtime-derived eligibility classification (e.g. the Brave New World eligibility check, which computes `classification_status` per menu at runtime and explicitly does not require declared marking) is **not** a declaration and does not qualify for this carve-out.
2. **`active-workflow-state-law.md:9`** — resolving a gate by its declared type is workflow-owned behaviour, not ad-hoc interpretation.
3. **`active-workflow-state-law.md:12`** — corrected wording, separating authority from accountability instead of conflating them:
   ```
   NEVER treat an off-protocol follow-up as authority to skip the active
   workflow's states, menus, approvals, validation, dispatch gates, or
   completion criteria. A skip is bypassing a gate without a resolution
   permitted by its statically declared type and the deterministic gate
   semantics that apply to it; a gate resolved within those is not a skip.
   ALWAYS produce the required audit record for every permitted autonomous
   resolution.
   ```
   Two invariants, tested independently: no unauthorised resolution, and no unrecorded autonomous resolution. A logging failure violates the second without retroactively making the resolution illegal.
4. **New rule** — a reply is never a mode reset. Mode and typed-gate semantics persist across cancels, re-plans, and workflow transitions; mode changes only on an explicit request, matched against a **declared, closed trigger set** stated in this rule. A reply matching nothing in that set never changes mode. (Today only the mode menu's TITLE promises "change mode" with no handler behind it, and the Brave New World overlay activates on open-ended "semantically equivalent phrases" — neither is a closed set, so this must define one rather than relying on either.) Motivated by a field incident in the compete kit: one step off the golden path silently reset the interaction contract and cost four gate questions in three minutes, because mode persistence lived nowhere.

Also examined and rejected as a blocker: `pdsl-execution-card.md`'s active-menu rule (governs incoming user messages; not triggered by an auto-proceed that isn't parsing a reply). Recorded here as examined rather than missed.

**Blocking dependency, tracked separately:** #152 — the PDSL validator does not currently recognize a `TYPE` section header at all, so none of the above can be relied on until that lands.

## Open questions for review (original review pass)

**For TK (PDSL wording + whether lint can hold the invariant):**

1. Does edit 1 preserve the original intent — can "go ahead" still be over-read as permission for unrelated work after the carve-out? *(Checked: no — the carve-out exempts resolution by declared mode **and** gate's declared type, both fixed independent of reply content; this is structurally disjoint from inferring scope from what the user said. The wording must stay a conjunction, not "mode alone.")*
2. Is edit 3's definition of "skip" airtight — is there a third case beyond (a) no declared type or (b) no logged ruling? *(Original check flagged a blocking-gate gap in the first draft; superseded by the corrected edit 3 above, which folds permission entirely into the type's own deterministic semantics and removes logging from the permission test altogether.)*

**For Oleg (enforcement — is the load-bearing dependency actually holdable):**

3. All four edits rest on the gate's type being a static constant of the MENU block, never computed at runtime. Can lint actually guarantee that? *(Checked: only partially. Lint can enforce the syntactic half — every MENU/gate has a literal, enum-valid TYPE field, no interpolation/variable reference — the same class of check as the existing "OPTIONS must start with a decimal number" rule. It cannot cover the runtime half: an executing agent misreading, fabricating, or drifting from the declared type is an execution-fidelity failure, not a source-shape one — "a gate that lies about its own type" specifically survives static lint, and the Brave New World eligibility check is a live, real example of exactly this shape of runtime-derived judgment. Suggest the audit record required by edit 3 also quote the literal declared TYPE token verbatim, so post-hoc audit can catch drift that lint can't watch live. See #152 for the validator gap that currently leaves even the syntactic half unchecked.)*

---
Filed from a review comment; full current-vs-proposed diff for each edit available on request (~half a page, omitted here for length).

tkcoding:
Closes #148.

`pdsl-execution-card.md:28-33` and `active-workflow-state-law.md:7,9,12` correctly
restrict Studio from reinterpreting a user's reply as broad permission -- but as
worded they also catch Studio auto-proceeding a gate the workflow itself typed
(the `TYPE` declaration #153 added) under a session mode the user chose. Neither
law was written with a declared gate type in mind, so this was blocking the
typed-gate work from being relied on.

## What this adds

**`runtime/pdsl-execution-card.md`** -- a session's declared mode and a gate's
declared `TYPE` (the literal token in that gate's own `MENU` block) are
workflow-owned behaviour, not an interpretation of the user's reply. Explicitly
excludes runtime-derived classifications -- e.g. Brave New World's
`classification_status`, which is computed per-menu at runtime and by design
carries no declared marking -- from qualifying for this carve-out.

**`runtime/active-workflow-state-law.md`**:
- resolving a gate by its declared type is workflow-owned behaviour, not
  ad-hoc interpretation
- splits the "skip" definition into two independently-tested invariants:
  no unauthorised resolution, and no unrecorded autonomous resolution. Tying
  permission to whether the resolution was logged would let a logging failure
  retroactively reclassify a legitimate resolution as a forbidden skip.
- a reply is never a mode reset -- mode and typed-gate semantics persist
  across cancels, re-plans, and workflow transitions; only an explicit
  request matched against a declared, closed trigger set changes it

**`gates/simple-mode.md`** -- new `SimpleModeChangeTrigger` unit implementing
that trigger set: the literal phrase `"change mode"`, kept as its own unit
(rather than folded into `SimpleModeGate`) so it doesn't tip that gate over
its already-allowlisted `PDSL600`/`PDSL601` caps, and rides on the existing
`WorkflowBootstrapSimpleModeGate` load with no changes to `workflow-bootstrap.md`.

**`tests/test_pdsl_keywords.py`** -- regression test asserting the mode-change
trigger phrase stays disjoint from Brave New World's open-ended activation
phrase list (`stop asking me`, `auto mode`, `autonomous`, ...), so a future
edit to either list can't silently collide.

## Verification

- `pdsl validate` on all three edited runtime/gate files: only the
  pre-existing, already-allowlisted `PDSL601`/`PDSL600` findings tracked in
  #87 -- nothing new.
- `tests/test_pdsl_keywords.py`: 14/14 passed, including the new test.
- `tests/test_workflow_subagents_dispatch.py` + `tests/test_pdsl_validate_cli.py`
  (the other files referencing these symbols): 157 passed, 15 xfailed.
- Full suite: 5,626 passed. Two unrelated pre-existing failures reproduced
  identically on `main` before this change (confirmed via `git stash`):
  a macOS/APFS filename-encoding limitation in
  `test_change_summary_links.py` and a semantic-overlap scoring assertion in
  `test_eval_semantic.py`. Neither touches PDSL/gate code; left out of scope
  here, same as #159 was filed separately from PR #151 rather than bundled in.

## Sequencing

Landed after both dependencies this issue names as blocking merged: #151
(ADR-0023, merged as 81b8183d) and #153/#152 (the `TYPE` validator, merged as
3cb9b18a). This closes ADR-0023's prerequisites 1 and 2; prerequisites 3
(the approved keyed source for `decision` gates) and 4 (the write path to
`record()`) remain open and are untouched by this change -- nothing here
flips runtime behaviour.

### DIFF
--- a/skills/studio/modules/gates/simple-mode.md
+++ b/skills/studio/modules/gates/simple-mode.md
@@ -32,3 +32,17 @@ OPTIONS:
   3 debug — debugger overlay in run mode, for workflow development only -> SET SIMPLE_MODE = debug; LOAD {cf-studio-path}/.core/skills/studio/modules/gates/simple-mode-debug.md; CONTINUE SimpleModeDebug
   INVALID -> EMIT_MENU SimpleModeChoice
 ```
+
+```pdsl
+UNIT SimpleModeChangeTrigger
+PURPOSE: Recognize the one declared mode-change trigger and re-open mode selection without resetting any other session state.
+WHEN:
+  REQUIRE SIMPLE_MODE != unset
+  REQUIRE the user's message is exactly the phrase "change mode" (case-insensitive; not a paraphrase, synonym, or superset phrase)
+DO:
+  EMIT_MENU SimpleModeChoice
+  WAIT user.reply
+  STOP_TURN
+RULES:
+  ALWAYS treat "change mode" as the sole mode-change trigger; NEVER treat semantically similar phrases, synonyms, or partial matches as satisfying it.
+```
--- a/skills/studio/modules/gates/simple-mode.md
+++ b/skills/studio/modules/gates/simple-mode.md
@@ -38,7 +38,7 @@ UNIT SimpleModeChangeTrigger
 PURPOSE: Recognize the one declared mode-change trigger and re-open mode selection without resetting any other session state.
 WHEN:
   REQUIRE SIMPLE_MODE != unset
-  REQUIRE the user's message is exactly the phrase "change mode" (case-insensitive; not a paraphrase, synonym, or superset phrase)
+  REQUIRE the user's message, after trimming leading/trailing whitespace and folding ASCII case, equals exactly the phrase "change mode" with no other leading, trailing, or embedded characters (not a paraphrase, synonym, punctuation-padded variant, or superset phrase)
 DO:
   EMIT_MENU SimpleModeChoice
   WAIT user.reply
--- a/skills/studio/modules/runtime/active-workflow-state-law.md
+++ b/skills/studio/modules/runtime/active-workflow-state-law.md
@@ -5,8 +5,8 @@ UNIT ActiveWorkflowStateLaw
 PURPOSE: Keep an active cf/cf-* workflow inside explicit workflow states unless control is visibly handed off.
 RULES:
   ALWAYS while a cf/cf-* workflow is active, treat every new user message as candidate input to the current workflow state first; user messages are workflow input, not permission for generic autonomous behavior.
-  ALWAYS map each new user message to exactly one of these outcomes, in order: a valid current-state transition or loaded session gate, a visible companion-skill handoff, or an explicit exit from the active workflow to free mode.
-  ALWAYS prefer workflow-owned WAIT/STOP_TURN continuations, menus, gates, approvals, dispatch gates, and completion paths over ad-hoc interpretation; resolving a gate by its declared type is workflow-owned behaviour, not ad-hoc interpretation.
+  ALWAYS map each new user message to exactly one of these outcomes, in order: a match against the declared mode-change trigger (RUN SimpleModeChangeTrigger from `gates/simple-mode.md` when its WHEN is satisfied), a valid current-state transition or loaded session gate, a visible companion-skill handoff, or an explicit exit from the active workflow to free mode.
+  ALWAYS prefer workflow-owned WAIT/STOP_TURN continuations, menus, gates, approvals, dispatch gates, and completion paths over ad-hoc interpretation; resolving a gate by its declared type is workflow-owned behaviour, not ad-hoc interpretation -- the canonical definition of that carve-out lives in `runtime/pdsl-execution-card.md`.
   ALWAYS use visible companion-skill routing and return a launch list for current user messages that no longer fit the active workflow's reachable states but clearly span another cf-* domain; NEVER switch workflows silently.
   ALWAYS say that the current workflow is being exited and that control is returning to free mode before handling any request outside the workflow; requests outside the workflow are allowed only after no valid current-state transition or companion workflow remains.
   NEVER treat an off-protocol follow-up as authority to skip the active
@@ -20,5 +20,5 @@ RULES:
   NEVER treat a reply as a mode reset; mode and typed-gate semantics persist
     across cancels, re-plans, and workflow transitions and change only on
     an explicit request matched against the declared trigger set in
-    SimpleModeChoice.
+    SimpleModeChangeTrigger.
 ```
--- a/skills/studio/modules/runtime/active-workflow-state-law.md
+++ b/skills/studio/modules/runtime/active-workflow-state-law.md
@@ -6,9 +6,19 @@ PURPOSE: Keep an active cf/cf-* workflow inside explicit workflow states unless 
 RULES:
   ALWAYS while a cf/cf-* workflow is active, treat every new user message as candidate input to the current workflow state first; user messages are workflow input, not permission for generic autonomous behavior.
   ALWAYS map each new user message to exactly one of these outcomes, in order: a valid current-state transition or loaded session gate, a visible companion-skill handoff, or an explicit exit from the active workflow to free mode.
-  ALWAYS prefer workflow-owned WAIT/STOP_TURN continuations, menus, gates, approvals, dispatch gates, and completion paths over ad-hoc interpretation.
+  ALWAYS prefer workflow-owned WAIT/STOP_TURN continuations, menus, gates, approvals, dispatch gates, and completion paths over ad-hoc interpretation; resolving a gate by its declared type is workflow-owned behaviour, not ad-hoc interpretation.
   ALWAYS use visible companion-skill routing and return a launch list for current user messages that no longer fit the active workflow's reachable states but clearly span another cf-* domain; NEVER switch workflows silently.
   ALWAYS say that the current workflow is being exited and that control is returning to free mode before handling any request outside the workflow; requests outside the workflow are allowed only after no valid current-state transition or companion workflow remains.
-  NEVER treat an off-protocol follow-up as authority to skip the active workflow's states, menus, approvals, validation, dispatch gates, or completion criteria.
+  NEVER treat an off-protocol follow-up as authority to skip the active
+    workflow's states, menus, approvals, validation, dispatch gates, or
+    completion criteria. A skip is bypassing a gate without a resolution
+    permitted by its statically declared type and the deterministic gate
+    semantics that apply to it; a gate resolved within those is not a skip.
+  ALWAYS produce the required audit record for every permitted autonomous
+    resolution.
   NEVER continue the active workflow and free mode in the same turn.
+  NEVER treat a reply as a mode reset; mode and typed-gate semantics persist
+    across cancels, re-plans, and workflow transitions and change only on
+    an explicit request matched against the declared trigger set in
+    SimpleModeChoice.
 ```
--- a/skills/studio/modules/runtime/pdsl-execution-card.md
+++ b/skills/studio/modules/runtime/pdsl-execution-card.md
@@ -28,6 +28,14 @@ RULES:
   ALWAYS after any `WAIT`/`STOP_TURN` resume at the exact active PDSL
     continuation target; REQUIRED: do not reinterpret the user's reply as
     broad permission for generic autonomous execution.
+  ALWAYS treat a session's declared mode and a gate's declared risk type —
+    the literal type named in that gate's own MENU block — as workflow-owned
+    behaviour, not an interpretation of the user's reply; resolving a gate by
+    that declared type is outside the prohibition above.
+  NEVER treat a runtime-derived eligibility or risk classification, such as
+    an autonomy overlay's per-menu eligibility check, as a declaration; only
+    a literal type named in the gate's own MENU block qualifies for the
+    carve-out above.
   ALWAYS while a workflow, gate, or menu remains active, treat each new user
     message as input to that active continuation, not as permission for
     unrelated execution.
--- a/skills/studio/modules/runtime/pdsl-execution-card.md
+++ b/skills/studio/modules/runtime/pdsl-execution-card.md
@@ -53,4 +53,13 @@ RULES:
     obligations unless an active rule references them.
   NEVER weaken `ALWAYS`, `NEVER`, `REQUIRE`, `WAIT`, `STOP_TURN`, or
     `INVARIANTS` because nearby prose sounds softer.
+NOTES:
+  No core module declares a gate `TYPE` or resolves a gate from one today —
+  see `architecture/specs/PDSL.md` and `cpt-studio-adr-autonomous-default-and-gate-risk`
+  (ADR-0023). This carve-out only removes the prohibition against such
+  resolution once it exists; it does not itself define or implement how a
+  gate is resolved from its declared type.
+  The identical carve-out is restated for workflow-state law in
+  `runtime/active-workflow-state-law.md`; keep both definitions in sync if
+  this one changes.
 ```
--- a/tests/test_pdsl_keywords.py
+++ b/tests/test_pdsl_keywords.py
@@ -564,6 +564,37 @@ def test_pdsl_workflows_load_execution_card_during_bootstrap() -> None:
     assert not failures, "\n".join(failures)
 
 
+def test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases() -> None:
+    """ADR-0023 prerequisite 1: the declared mode-change trigger set must
+    never overlap the Brave New World overlay's open-ended activation
+    phrases, or a user saying one could be silently read as the other.
+    """
+    simple_mode_text = (REPO_ROOT / "skills/studio/modules/gates/simple-mode.md").read_text()
+    bnw_text = (REPO_ROOT / "workflows/brave-new-world.md").read_text()
+
+    trigger = "change mode"
+    assert f'"{trigger}"' in simple_mode_text, (
+        "Expected the literal mode-change trigger phrase to still be declared "
+        "in gates/simple-mode.md; update this test if the wording changed."
+    )
+
+    activation_line = next(
+        (line for line in bnw_text.splitlines() if "ALWAYS resolve semantically equivalent phrases" in line),
+        None,
+    )
+    assert activation_line is not None, (
+        "Expected to find BraveNewWorldActivate's activation-phrase rule in "
+        "workflows/brave-new-world.md; update this test if the wording moved."
+    )
+    bnw_phrases = re.findall(r"'([^']+)'", activation_line)
+    assert bnw_phrases, "Expected quoted activation phrases on that rule line"
+
+    assert trigger not in bnw_phrases, (
+        f"Mode-change trigger {trigger!r} collides with a Brave New World "
+        "activation phrase; ADR-0023 requires these trigger sets to be disjoint."
+    )
+
+
 def test_named_pdsl_units_and_menus_are_not_exact_duplicates() -> None:
     """Exact duplicate named PDSL blocks should be defined once and loaded."""
     blocks_by_body: dict[str, list[str]] = defaultdict(list)
--- a/tests/test_pdsl_keywords.py
+++ b/tests/test_pdsl_keywords.py
@@ -568,9 +568,17 @@ def test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases() -> None:
     """ADR-0023 prerequisite 1: the declared mode-change trigger set must
     never overlap the Brave New World overlay's open-ended activation
     phrases, or a user saying one could be silently read as the other.
+
+    Checked case-insensitively and for substring/superset overlap in either
+    direction, not just exact list membership -- SimpleModeChangeTrigger's own
+    contract rejects "a paraphrase, synonym, punctuation-padded variant, or
+    superset phrase", so this test rejects the same shapes of collision. The
+    activation-phrase and exclusion lines are both read only from inside
+    BraveNewWorldActivate's own PDSL block, not a flat file-wide scan.
     """
-    simple_mode_text = (REPO_ROOT / "skills/studio/modules/gates/simple-mode.md").read_text()
-    bnw_text = (REPO_ROOT / "workflows/brave-new-world.md").read_text()
+    simple_mode_path = REPO_ROOT / "skills/studio/modules/gates/simple-mode.md"
+    bnw_path = REPO_ROOT / "workflows/brave-new-world.md"
+    simple_mode_text = simple_mode_path.read_text()
 
     trigger = "change mode"
     assert f'"{trigger}"' in simple_mode_text, (
@@ -578,20 +586,124 @@ def test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases() -> None:
         "in gates/simple-mode.md; update this test if the wording changed."
     )
 
+    bnw_activate_block = next(
+        (
+            block for _, block in _iter_pdsl_blocks(bnw_path)
+            if block and block[0].strip() == "UNIT BraveNewWorldActivate"
+        ),
+        None,
+    )
+    assert bnw_activate_block is not None, (
+        "Expected a UNIT BraveNewWorldActivate PDSL block in "
+        "workflows/brave-new-world.md; update this test if it moved or was renamed."
+    )
+
     activation_line = next(
-        (line for line in bnw_text.splitlines() if "ALWAYS resolve semantically equivalent phrases" in line),
+        (line for line in bnw_activate_block if "ALWAYS resolve semantically equivalent phrases" in line),
         None,
     )
     assert activation_line is not None, (
-        "Expected to find BraveNewWorldActivate's activation-phrase rule in "
-        "workflows/brave-new-world.md; update this test if the wording moved."
+        "Expected BraveNewWorldActivate's activation-phrase rule inside its own "
+        "block; update this test if the wording moved."
     )
-    bnw_phrases = re.findall(r"'([^']+)'", activation_line)
+    bnw_phrases = [phrase.lower() for phrase in re.findall(r"'([^']+)'", activation_line)]
     assert bnw_phrases, "Expected quoted activation phrases on that rule line"
 
-    assert trigger not in bnw_phrases, (
-        f"Mode-change trigger {trigger!r} collides with a Brave New World "
-        "activation phrase; ADR-0023 requires these trigger sets to be disjoint."
+    exclusion_line = next(
+        (
+            line for line in bnw_activate_block
+            if "NEVER treat the exact phrase" in line and trigger in line.lower()
+        ),
+        None,
+    )
+    assert exclusion_line is not None, (
+        "Expected an explicit NEVER rule in BraveNewWorldActivate excluding the "
+        "literal mode-change trigger phrase from BNW activation; update this "
+        "test if that exclusion moved or was reworded."
+    )
+
+    for phrase in bnw_phrases:
+        assert trigger != phrase, (
+            f"Mode-change trigger {trigger!r} exactly matches a BNW activation phrase {phrase!r}."
+        )
+        assert trigger not in phrase, (
+            f"Mode-change trigger {trigger!r} is a substring of BNW phrase {phrase!r}; "
+            "ADR-0023 requires these trigger sets to be disjoint, including supersets."
+        )
+        assert phrase not in trigger, (
+            f"BNW phrase {phrase!r} is a substring of the mode-change trigger {trigger!r}."
+        )
+
+
+def test_declared_gate_type_carveout_is_defined_once_and_cross_referenced() -> None:
+    """The declared mode/gate-type carve-out is defined canonically in
+    pdsl-execution-card.md; active-workflow-state-law.md must reference that
+    definition rather than silently restate it, so the two wordings can't
+    drift apart unnoticed (PR #162 review finding).
+    """
+    card_text = (REPO_ROOT / "skills/studio/modules/runtime/pdsl-execution-card.md").read_text()
+    law_text = (REPO_ROOT / "skills/studio/modules/runtime/active-workflow-state-law.md").read_text()
+
+    assert "a gate's declared risk type" in card_text
+    assert "NEVER treat a runtime-derived eligibility or risk classification" in card_text
+    assert "cpt-studio-adr-autonomous-default-and-gate-risk" in card_text, (
+        "Expected the carve-out's NOTES to cite ADR-0023, so a reader knows "
+        "this rule is a currently-inert prerequisite, not an implemented "
+        "resolution mechanism (PR #162 review finding)."
+    )
+    assert "restated for workflow-state law in\n  `runtime/active-workflow-state-law.md`" in card_text
+
+    assert (
+        "the canonical definition of that carve-out lives in `runtime/pdsl-execution-card.md`"
+        in law_text
+    ), (
+        "Expected active-workflow-state-law.md's declared-type carve-out to "
+        "cross-reference pdsl-execution-card.md rather than restate it with no link."
+    )
+
+
+def test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present() -> None:
+    """Regression lock for the normative additions to active-workflow-state-law.md
+    that otherwise had zero test coverage: the skip redefinition, the
+    audit-record requirement, and the mode-persistence rule -- which must both
+    name and actually dispatch to SimpleModeChangeTrigger, or the rule points
+    at a unit nothing ever reaches (PR #162 review findings).
+    """
+    law_text = (REPO_ROOT / "skills/studio/modules/runtime/active-workflow-state-law.md").read_text()
+
+    assert "A skip is bypassing a gate without a resolution" in law_text
+    assert "permitted by its statically declared type" in law_text
+    assert "ALWAYS produce the required audit record for every permitted autonomous" in law_text
+    assert "NEVER treat a reply as a mode reset" in law_text
+    assert "SimpleModeChangeTrigger" in law_text, (
+        "Expected the mode-persistence rule to name SimpleModeChangeTrigger "
+        "specifically (not the SimpleModeChoice menu) -- that unit is what "
+        "actually implements the declared trigger set."
+    )
+    assert "RUN SimpleModeChangeTrigger from `gates/simple-mode.md`" in law_text, (
+        "Expected the message-mapping rule to explicitly dispatch to "
+        "SimpleModeChangeTrigger; otherwise it is declared but unreachable "
+        "from any active workflow state (PR #162 review finding)."
+    )
+
+
+def test_simple_mode_change_trigger_contract_is_fully_declared() -> None:
+    """Regression lock for SimpleModeChangeTrigger's full matching contract --
+    exact match, ASCII case-folding, whitespace trimming, and rejection of
+    paraphrases/synonyms/superset phrases -- not just its trigger phrase
+    (PR #162 review finding: the prior test covered only BNW disjointness).
+    """
+    simple_mode_text = (REPO_ROOT / "skills/studio/modules/gates/simple-mode.md").read_text()
+
+    unit_start = simple_mode_text.index("UNIT SimpleModeChangeTrigger")
+    unit_text = simple_mode_text[unit_start:]
+
+    assert "folding ASCII case" in unit_text
+    assert "trimming leading/trailing whitespace" in unit_text
+    assert "not a paraphrase, synonym, punctuation-padded variant, or superset phrase" in unit_text
+    assert (
+        "NEVER treat semantically similar phrases, synonyms, or partial matches as satisfying it"
+        in unit_text
     )
 
 
--- a/workflows/brave-new-world.md
+++ b/workflows/brave-new-world.md
@@ -40,6 +40,7 @@ RULES:
   ALWAYS keep all current and future underlying rules, prerequisites, menus, waits, hard stops, validation gates, and terminal shapes active, while allowing this overlay to answer eligible menus and questions by selecting one valid original option
   ALWAYS keep BRAVE_NEW_WORLD_DECISION_LOG append-only for the session across disable and re-enable cycles
   ALWAYS resolve semantically equivalent phrases such as 'stop asking me', 'auto mode', 'autonomous', 'fewer questions', 'less interruptions' as BNW activation when context is unambiguous
+  NEVER treat the exact phrase 'change mode' as BNW activation, semantically equivalent or otherwise; that phrase is reserved exclusively for `gates/simple-mode.md`'s SimpleModeChangeTrigger, and the semantic-equivalence rule above never claims it
   NEVER start substantive task work merely because this overlay was enabled
   NEVER require `cf` or `CFS_INIT == true` merely to enable or disable this overlay
 ```


skills/studio/modules/runtime/pdsl-execution-card.md
```diff
@@ -28,6 +28,14 @@ RULES:
   ALWAYS after any `WAIT`/`STOP_TURN` resume at the exact active PDSL
     continuation target; REQUIRED: do not reinterpret the user's reply as
     broad permission for generic autonomous execution.
+  ALWAYS treat a session's declared mode and a gate's declared risk type —
```
ainetx, line 31: ### Duplicate 'declared type = workflow-owned, not ad-hoc' rule stated independently in two modules with no cross-reference

**Severity:** Minor

**Problem**
pdsl-execution-card.md adds 'resolving a gate by that declared type is outside the prohibition above' and active-workflow-state-law.md independently adds 'resolving a gate by its declared type is workflow-owned behaviour, not ad-hoc interpretation.' Both encode the identical carve-out concept in different prose, in different files, with no textual cross-reference between them.

<details>

**How to reproduce**
1. Open skills/studio/modules/runtime/pdsl-execution-card.md lines 31-35. 2. Open skills/studio/modules/runtime/active-workflow-state-law.md line 9-10. 3. Compare wording: both assert declared-type gate resolution is not ad-hoc/interpretation, but neither references the other file.

**Expected behavior**
A single source-of-truth definition of the carve-out, with the other module cross-referencing it (e.g. 'see pdsl-execution-card.md's declared-type carve-out').

**Actual behavior**
Two independently-worded statements of the same rule exist with no link, so a future edit to one (e.g. narrowing or renaming 'declared type') could diverge from the other without any signal.

```text
pdsl-execution-card.md: declared-type carve-out (wording A)
\
active-workflow-state-law.md: declared-type carve-out (wording B)
(no link between A and B) -> future edit to A only -> drift vs B
```

**Impact**
If one module's carve-out is edited (tightened, loosened, or reworded) without updating the other, the two modules could give contradictory guidance about which gate resolutions are 'ad-hoc' vs workflow-owned, reintroducing the ambiguity this PR sets out to fix.

**Suggested correction**
Add an explicit cross-reference from one module's rule to the other (e.g. active-workflow-state-law.md: 'see pdsl-execution-card.md for the declared-type carve-out definition') so there is one canonical wording.

**How to verify**
Confirm both files reference a single canonical definition of the carve-out, or add a test asserting the two rule texts remain semantically aligned.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -564,6 +564,37 @@ def test_pdsl_workflows_load_execution_card_during_bootstrap() -> None:
     assert not failures, "\n".join(failures)
 
 
+def test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases() -> None:
```
ainetx, line 567: ### New normative rules in active-workflow-state-law.md and pdsl-execution-card.md have zero test coverage

**Severity:** Minor

**Problem**
The only new test (test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases) reads only simple-mode.md and workflows/brave-new-world.md. It does not read or assert anything about active-workflow-state-law.md's new audit-record requirement, redefinition of 'skip', or mode-persistence rule, nor pdsl-execution-card.md's new declared-type carve-out / runtime-derived-eligibility exclusion.

<details>

**How to reproduce**
1. Read tests/test_pdsl_keywords.py lines 567-596 (the new test body). 2. Note it opens only simple-mode.md and brave-new-world.md. 3. Grep the test file for 'active-workflow-state-law' or 'pdsl-execution-card': no matches for the new rule content.

**Expected behavior**
New normative rules introduced in a PR should have at least a presence/consistency test analogous to the one added for the mode-change trigger.

**Actual behavior**
Two of the three modified prose modules ship with no automated check at all, relying entirely on manual review to keep them correct going forward.

```text
active-workflow-state-law.md (new rules) -> no test
pdsl-execution-card.md (new rules) -> no test
simple-mode.md (new trigger) -> test_mode_change_trigger_is_disjoint... (covered)
```

**Impact**
Regressions or wording drift in the audit-record, skip-redefinition, mode-persistence, or declared-type carve-out rules would not be caught by CI.

**Suggested correction**
Add lightweight text-presence/consistency tests for the new rules in active-workflow-state-law.md and pdsl-execution-card.md, mirroring the pattern used for the mode-change trigger test.

**How to verify**
Run the test suite and confirm new assertions exist that would fail if these specific rule strings were removed or altered.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -564,6 +564,37 @@ def test_pdsl_workflows_load_execution_card_during_bootstrap() -> None:
...
+    assert activation_line is not None, (
+        "Expected to find BraveNewWorldActivate's activation-phrase rule in "
+        "workflows/brave-new-world.md; update this test if the wording moved."
+    )
+    bnw_phrases = re.findall(r"'([^']+)'", activation_line)
+    assert bnw_phrases, "Expected quoted activation phrases on that rule line"
+
+    assert trigger not in bnw_phrases, (
+        f"Mode-change trigger {trigger!r} collides with a Brave New World "
+        "activation phrase; ADR-0023 requires these trigger sets to be disjoint."
```
ainetx: ### Disjointness test uses case-sensitive comparison despite trigger being documented as case-insensitive

**Severity:** Minor

**Problem**
simple-mode.md's SimpleModeChangeTrigger declares the phrase match is 'case-insensitive', but the new test asserts `trigger not in bnw_phrases` using Python's case-sensitive `in` operator against a lowercase literal 'change mode' and phrases extracted verbatim from brave-new-world.md.

<details>

**How to reproduce**
1. Read simple-mode.md's new RULES: 'the user's message is exactly the phrase "change mode" (case-insensitive...)'. 2. Read the test: `trigger = "change mode"`; `assert trigger not in bnw_phrases`. 3. Consider a BNW phrase 'Change Mode' (different case): it would not equal the lowercase 'change mode' string, so the assertion passes even though the case-insensitive matching described in simple-mode.md would treat them as colliding.

**Expected behavior**
The test should normalize case (e.g. lower-case both sides) before comparing, matching the declared case-insensitive semantics of the trigger.

**Actual behavior**
The test only performs an exact, case-sensitive string containment check, so a case-differing BNW phrase would slip through undetected.

```text
simple-mode.md: 'change mode' matches case-insensitively
vs
test: trigger.lower() 'change mode' not in bnw_phrases (raw case)
-> BNW phrase 'Change Mode' evades the check
```

**Impact**
ADR-0023's disjointness requirement could be silently violated by a case-variant BNW activation phrase without the CI test catching it.

**Suggested correction**
Lower-case both `trigger` and each entry of `bnw_phrases` before the `not in` check.

**How to verify**
Add a case-varying phrase to brave-new-world.md's activation list temporarily and confirm the test now fails, then revert.

</details>

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx, line 31: ### Carve-out references undeclared MENU TYPE: field

**Severity:** Minor

**Problem**
pdsl-execution-card.md's new rule says resolving a gate is workflow-owned when it matches 'the literal type named in that gate's own MENU block'. The only MENU block touched/referenced in this diff, MENU SimpleModeChoice in simple-mode.md, has no TYPE: sub-header — it only has TITLE: and OPTIONS:.

<details>

**How to reproduce**
1. Open skills/studio/modules/gates/simple-mode.md and inspect MENU SimpleModeChoice. 2. Note it declares TITLE: and OPTIONS: but never a TYPE: field. 3. Open pdsl-execution-card.md's new rule which assumes gates declare such a type.

**Expected behavior**
The carve-out rule should either point to a MENU block that actually exhibits a TYPE: sub-header, or simple-mode.md's MENU should be updated to declare TYPE: so the rule has a concrete anchor.

**Actual behavior**
The rule is introduced with no example of the described shape anywhere in the changed files, leaving it ungrounded and open to misapplication to gates that were never intended to qualify.

```text
pdsl-execution-card.md carve-out --references--> MENU 'declared TYPE:' --X--> simple-mode.md MENU SimpleModeChoice (no TYPE: field present)
```

**Impact**
Ambiguous or inconsistent application of the autonomy carve-out; agents may misjudge which gates qualify for autonomous resolution.

**Suggested correction**
Add a TYPE: sub-header to MENU SimpleModeChoice (and any other gates meant to use this carve-out), or clarify in pdsl-execution-card.md which existing MENU blocks currently satisfy the described shape.

**How to verify**
Search all MENU blocks in the repo for a TYPE: sub-header and confirm at least one gate referenced by the carve-out actually declares it.

</details>

skills/studio/modules/gates/simple-mode.md
```diff
@@ -32,3 +32,17 @@ OPTIONS:
   3 debug — debugger overlay in run mode, for workflow development only -> SET SIMPLE_MODE = debug; LOAD {cf-studio-path}/.core/skills/studio/modules/gates/simple-mode-debug.md; CONTINUE SimpleModeDebug
   INVALID -> EMIT_MENU SimpleModeChoice
 ```
+
+```pdsl
+UNIT SimpleModeChangeTrigger
```
ainetx, line 37: ### SimpleModeChangeTrigger is unreachable — no active state routes into it

**Severity:** Major

**Problem**
SimpleModeChangeTrigger is declared in simple-mode.md as a new PDSL UNIT with its own WHEN/DO block, but it is never LOADed, RUN, or CONTINUEd from anywhere. SimpleModeGate's DO block only ever CONTINUEs into SimpleModeSimpleEntry, SimpleModeNormal, or SimpleModeDebug once SIMPLE_MODE is set, and none of those three continuation units (in simple-mode-simple.md, simple-mode-normal.md, simple-mode-debug.md) reference SimpleModeChangeTrigger or contain any WHEN clause checking the user's message against 'change mode'. A repo-wide search for the identifier 'SimpleModeChangeTrigger' finds only its own declaration line. The active-workflow-state-law.md changes add prose asserting mode changes happen 'only on an explicit request matched against the declared trigger set in SimpleModeChoice' but that is a narrative RULE, not a dispatch/LOAD/CONTINUE instruction that would cause an interpreter to actually enter SimpleModeChangeTrigger while, e.g., SimpleModeNormal's DO block ('REQUIRE SIMPLE_MODE == normal') is the active continuation.

<details>

**How to reproduce**
1. Start a session, select mode 'normal' (SIMPLE_MODE=normal), landing in SimpleModeNormal's WAIT/continuation state.
2. User later types 'change mode'.
3. Per SimpleModeNormal's own contract ('ALWAYS continue with the workflow's existing menus, gates, stops...'), there is no routing rule that hands this message to SimpleModeChangeTrigger, since SimpleModeNormal never loads or checks for it.
4. The literal phrase is only recognized if some workflow interpreter independently already knows to check every message against SimpleModeChangeTrigger's WHEN clause, but no LOAD/RUN reference to that UNIT name exists anywhere in the diff or repo.

**Expected behavior**
Every active non-unset simple-mode continuation (simple/normal/debug) should LOAD or otherwise dispatch into SimpleModeChangeTrigger (or an equivalent always-checked routing rule) before/alongside interpreting the user's reply as workflow input, so that saying 'change mode' from any mode reliably reopens SimpleModeChoice.

**Actual behavior**
SimpleModeChangeTrigger exists only as an orphaned UNIT declaration; none of SimpleModeNormal, SimpleModeSimpleEntry/SimpleModeBraveNewWorldGate, or SimpleModeDebug reference it, so the new exact-command contract has no enforced effect on session behavior.

```text
SimpleModeGate --CONTINUE--> {SimpleModeNormal | SimpleModeSimpleEntry | SimpleModeDebug}
(none of these LOAD/RUN SimpleModeChangeTrigger)
SimpleModeChangeTrigger <-- declared but unreferenced, unreachable
```

**Impact**
The documented 'say change mode any time' UX promised in SimpleModeChoice's MENU TITLE, and the new ADR-0023-style exact-trigger contract, are not actually wired into any live workflow state, so an interpreter following only the state machine has no textual instruction telling it to check for the trigger while in normal/simple/debug mode.

**Suggested correction**
Add an explicit LOAD/RUN or WHEN-clause reference to SimpleModeChangeTrigger inside SimpleModeNormal, SimpleModeSimpleEntry, and SimpleModeDebug (or centralize it as a session-wide gate loaded once alongside SimpleModeGate) so every reachable mode-specific continuation actually consults it before falling through to its own DO block.

**How to verify**
After the fix, trace: from SIMPLE_MODE=normal (or simple/debug), confirm a textual path (LOAD/RUN/dispatch rule) exists that evaluates the user's message against SimpleModeChangeTrigger's WHEN clause prior to SimpleModeNormal's/SimpleModeSimpleEntry's/SimpleModeDebug's own interpretation.

</details>

tests/test_pdsl_keywords.py
ainetx: ### Disjointness test only checks exact-match collision, not the substring/superset overlap the module itself warns about

**Severity:** Minor

**Problem**
SimpleModeChangeTrigger's WHEN clause explicitly guards against 'partial matches' and 'superset phrase[s]' (e.g. a longer phrase containing 'change mode'), but the test's only assertion is `trigger not in bnw_phrases`, an exact-membership check against the extracted phrase list — it does not test whether 'change mode' is a substring of any BNW phrase or vice versa.

<details>

**How to reproduce**
1. Read simple-mode.md's WHEN clause: '...not a paraphrase, synonym, or superset phrase'. 2. Read the test: `assert trigger not in bnw_phrases` (list membership, not substring search). 3. Hypothesize a BNW phrase 'change mode entirely': `'change mode' not in ['change mode entirely']` is True (no match), so the test passes despite the substring collision the module's own semantics call out as ambiguous.

**Expected behavior**
The test should also check for substring containment in both directions between the trigger and each BNW phrase.

**Actual behavior**
Only exact string equality against the phrase list is checked, so superset/substring collisions are not caught.

```text
trigger = 'change mode'
bnw_phrases = ['change mode entirely']
trigger not in bnw_phrases -> True (passes)
but 'change mode' is a substring of 'change mode entirely' -> real ambiguity risk per module's own RULES
```

**Impact**
A future BNW activation phrase that is a superset of 'change mode' would violate ADR-0023's disjointness intent yet pass this test, since the substring case the module explicitly flags is not covered.

**Suggested correction**
Extend the assertion to check `all(trigger not in p and p not in trigger for p in bnw_phrases)` or similar substring-aware comparison.

**How to verify**
Add a temporary BNW phrase like 'change mode entirely' to brave-new-world.md and confirm the strengthened test now fails.

</details>

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx, line 31: ### No automated coverage exists for pdsl-execution-card.md's new carve-out/exclusion wording

**Severity:** Minor

**Problem**
The diff adds two adjacent, prose-only rules to pdsl-execution-card.md (the declared-type carve-out and the runtime-derived-eligibility exclusion), but no test in tests/test_pdsl_keywords.py or elsewhere references pdsl-execution-card.md's new text, 'declared risk type', or 'runtime-derived eligibility' at all.

<details>

**How to reproduce**
1. Search tests/test_pdsl_keywords.py for 'pdsl-execution-card', 'declared risk type', or 'runtime-derived eligibility'. 2. Find no matches. 3. Confirm the only new test in the diff targets simple-mode.md/brave-new-world.md instead.

**Expected behavior**
A test that would fail if the carve-out and exclusion wording were edited to overlap (e.g. the exclusion text accidentally reusing the carve-out's qualifying phrase).

**Actual behavior**
Zero automated coverage means the boundary between the two new rules can silently blur through future edits with nothing catching it.

```text
pdsl-execution-card.md: carve-out rule (new)
pdsl-execution-card.md: exclusion rule (new)
-> no test references either -> no CI signal on future drift
```

**Impact**
A later edit could accidentally make the exclusion's language satisfy the carve-out's own qualifying condition (or vice versa), reintroducing the ambiguity this PR aims to remove, without any test catching it.

**Suggested correction**
Add a text-presence/disjointness test analogous to test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases for these two rules.

**How to verify**
Confirm the new test fails if the exclusion text is edited to overlap with the carve-out's qualifying language.

</details>

tests/test_pdsl_keywords.py
ainetx, line 567: ### New test does not cover most of SimpleModeChangeTrigger's stated contract

**Severity:** Minor

**Problem**
SimpleModeChangeTrigger declares several distinct constraints (exact case-insensitive match, rejection of paraphrases/synonyms/supersets, and that other session state is preserved across the reset), but test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases only checks that the literal string 'change mode' is not present in BNW's example phrase list.

<details>

**How to reproduce**
1. Read the new PDSL unit in simple-mode.md (case-insensitive exact match, non-paraphrase rule).
2. Read the new test; observe it only checks lowercase list membership against bnw_phrases.
3. Note no test exercises e.g. 'Change Mode', 'please change mode', or verifies unrelated state (e.g. active workflow state) is untouched.

**Expected behavior**
A test suite for a new state-changing control phrase would assert its full matching semantics and side-effect scope, not just list disjointness with another feature.

**Actual behavior**
Only literal list disjointness is tested; the richer behavioral contract is unverified by any test in the diff.

```text
SimpleModeChangeTrigger declares {case-insensitivity, non-paraphrase, state-preservation} -> test asserts only {disjoint from BNW list}
```

**Impact**
Regressions to case handling, paraphrase leniency, or accidental state resets on mode change would not be caught by CI.

**Suggested correction**
Add unit-level assertions (or static text checks) for case-insensitive matching, explicit rejection of near-miss phrases, and that no other SET/state variable is touched by the unit's DO block.

**How to verify**
Add/extend tests and confirm they fail if the case-insensitivity clause, non-paraphrase clause, or state-preservation clause were weakened or removed from simple-mode.md.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -564,6 +564,37 @@ def test_pdsl_workflows_load_execution_card_during_bootstrap() -> None:
...
+    phrases, or a user saying one could be silently read as the other.
+    """
+    simple_mode_text = (REPO_ROOT / "skills/studio/modules/gates/simple-mode.md").read_text()
+    bnw_text = (REPO_ROOT / "workflows/brave-new-world.md").read_text()
+
+    trigger = "change mode"
+    assert f'"{trigger}"' in simple_mode_text, (
+        "Expected the literal mode-change trigger phrase to still be declared "
+        "in gates/simple-mode.md; update this test if the wording changed."
+    )
```
ainetx, line 587: ### Collision test is not structurally bound to BraveNewWorldActivate

**Severity:** Minor

**Problem**
activation_line is found by scanning every line in workflows/brave-new-world.md for the substring 'ALWAYS resolve semantically equivalent phrases', with no check that the match falls inside the BraveNewWorldActivate UNIT's RULES block.

<details>

**How to reproduce**
1. Inspect test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases.
2. Note `next(... for line in bnw_text.splitlines() if "ALWAYS resolve semantically equivalent phrases" in line...)` has no bounding to a parsed block.
3. Imagine the same rule text copied into a different UNIT, or the real activation rule renamed/weakened while an unrelated comment retains the phrase — the test would still find *a* matching line and could pass on the wrong one, or silently track a moved/duplicated rule instead of the real activation gate.

**Expected behavior**
The test should first identify BraveNewWorldActivate's UNIT block (e.g. by locating 'UNIT BraveNewWorldActivate' and bounding the search to that block's RULES section) before extracting the activation rule line.

**Actual behavior**
The lookup is a flat first-match scan of the whole file, decoupled from the unit structure.

```text
file text -> first line containing substring (anywhere) -> assumed to be BraveNewWorldActivate's rule
```

**Impact**
A regression that moves, duplicates, or weakens the real activation-phrase rule outside detection could go unnoticed if any other line in the file still contains the matched substring.

**Suggested correction**
Locate the 'UNIT BraveNewWorldActivate' block boundaries first, then search only within that block's RULES section for the activation-phrase line.

**How to verify**
Add a fixture where the real rule is moved out of BraveNewWorldActivate but an unrelated line elsewhere still contains the substring; confirm the test now fails as expected instead of passing.

</details>

ainetx:
### BNW's open-ended semantic-equivalence matching is not shown to exclude the new mode-change trigger

**Severity:** Major

**Problem**
SimpleModeChangeTrigger requires exact, case-insensitive, non-paraphrase matching of 'change mode'. BraveNewWorldActivate's RULES separately allow resolving 'semantically equivalent phrases... when context is unambiguous' as BNW activation, with no stated exclusion for 'change mode'. The diff does not touch brave-new-world.md, so no boundary was added there, and the new test verifies only literal list membership against BNW's five enumerated examples.

<details>

**How to reproduce**
1. User is in an active BNW-eligible session and says 'change mode'.
2. BraveNewWorldActivate's semantic-equivalence rule has no explicit carve-out for this phrase.
3. Depending on 'context', an implementation following BraveNewWorldActivate literally could treat 'change mode' as semantically close to 'autonomous'/'auto mode' and route it to BNW activation instead of SimpleModeChoice.

**Expected behavior**
The activation contract should explicitly state that the exact-match mode-change trigger domain and BNW's semantic-equivalence domain are disjoint (e.g., an explicit NEVER/exclusion clause), and the test should check that boundary, not just list membership.

**Actual behavior**
No such exclusion exists in the reviewed diff; only a negative list-membership test was added.

```text
user: 'change mode' -> SimpleModeChangeTrigger (exact match) AND BraveNewWorldActivate (semantic match, unbounded) -> ambiguous which unit claims the message
```

**Impact**
A user attempting to change simple mode could be silently routed into Brave New World activation instead, defeating the intended disjointness guarantee described in ADR-0023 (referenced by the test).

**Suggested correction**
Add an explicit exclusion in BraveNewWorldActivate's RULES (e.g., 'NEVER treat the exact phrase change mode as BNW activation') and extend the test to assert that boundary rather than only list membership.

**How to verify**
Re-run/extend the test with a case that simulates 'change mode' under ambiguous BNW context and confirm it is excluded from BNW activation by contract, not just absent from the example list.

</details>

---
_Original location: workflows/brave-new-world.md:27 -- inline anchoring could not be resolved after 1 attempt(s)._

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx, line 31: ### Declared-type authorization lacks an implemented runtime contract

**Severity:** Major

**Problem**
PDSL.md states that TYPE is lint-only and that nothing resolves a gate from a declaration today, whereas the edited execution card authorizes resolution based on that declaration.

<details>

**How to reproduce**
1. Declare TYPE: confirmation on a menu. 2. Follow the new execution-card carve-out. 3. Consult the normative specification: it says declaration-driven resolution is not implemented and existing paths decide by runtime judgment.

**Expected behavior**
The grammar, validator, and runtime contract consistently define declaration-driven resolution, including absent, invalid, and duplicate TYPE handling.

**Actual behavior**
The new runtime prose relies on behavior the supplied specification explicitly disclaims.

```text
MENU TYPE
|- spec/validator -> lint-only
`- execution card -> authorization carve-out
```

**Impact**
Controllers can implement incompatible authorization behavior from the same TYPE declaration.

**Suggested correction**
Either implement and specify declaration-driven resolution with validator-aligned failure semantics, or remove the runtime carve-out until that work lands.

**How to verify**
Add contract tests for valid, omitted, invalid, and duplicate TYPE declarations through the resolution path.

</details>

skills/studio/modules/runtime/active-workflow-state-law.md
```diff
@@ -6,9 +6,19 @@ PURPOSE: Keep an active cf/cf-* workflow inside explicit workflow states unless
...
   ALWAYS use visible companion-skill routing and return a launch list for current user messages that no longer fit the active workflow's reachable states but clearly span another cf-* domain; NEVER switch workflows silently.
   ALWAYS say that the current workflow is being exited and that control is returning to free mode before handling any request outside the workflow; requests outside the workflow are allowed only after no valid current-state transition or companion workflow remains.
-  NEVER treat an off-protocol follow-up as authority to skip the active workflow's states, menus, approvals, validation, dispatch gates, or completion criteria.
+  NEVER treat an off-protocol follow-up as authority to skip the active
+    workflow's states, menus, approvals, validation, dispatch gates, or
+    completion criteria. A skip is bypassing a gate without a resolution
+    permitted by its statically declared type and the deterministic gate
+    semantics that apply to it; a gate resolved within those is not a skip.
+  ALWAYS produce the required audit record for every permitted autonomous
+    resolution.
```
ainetx, line 18: ### Audit-record requirement has no defined bounded schema

**Severity:** Minor

**Problem**
active-workflow-state-law.md adds 'ALWAYS produce the required audit record for every permitted autonomous resolution' without specifying what the record must contain, whether it is bounded, or whether raw user input/source text is excluded.

<details>

**How to reproduce**
1. A gate is resolved autonomously per the new declared-type carve-out.
2. The rule requires an audit record but does not define its shape.
3. An implementation could reasonably choose to log the full user message, full gate/menu text, or other unbounded content on every resolution.

**Expected behavior**
The rule should specify bounded, safe fields (e.g., gate name, declared type, chosen option, timestamp) and explicitly exclude or cap arbitrary free-text capture.

**Actual behavior**
No such specification exists in either changed runtime module.

```text
gate resolved autonomously -> 'produce audit record' (undefined schema) -> implementation-dependent content, potentially unbounded
```

**Impact**
Without a bounded schema, audit logging could inadvertently retain large or sensitive free-text content on every gate resolution, an observability/resource-safety concern flagged by the reviewer.

**Suggested correction**
Define the audit record's required fields explicitly (e.g., gate id, declared type, resolution outcome) and state that raw unbounded user/source text is excluded or truncated.

**How to verify**
Add a rule/test asserting the audit record schema is bounded and does not include raw unbounded fields.

</details>

skills/studio/modules/gates/simple-mode.md
```diff
@@ -32,3 +32,17 @@ OPTIONS:
...
+PURPOSE: Recognize the one declared mode-change trigger and re-open mode selection without resetting any other session state.
+WHEN:
```
ainetx, line 39: ### Mode-change trigger matcher leaves whitespace/normalization unspecified

**Severity:** Minor

**Problem**
SimpleModeChangeTrigger's WHEN clause requires an exact, case-insensitive match on 'change mode' but does not define whether surrounding whitespace, punctuation, or non-ASCII case-folding variants are accepted or rejected, leaving 'exactly' and 'case-insensitive' open to differing implementations.

<details>

**How to reproduce**
1. User sends ' change mode ' (with padding whitespace) or 'CHANGE MODE.' (trailing punctuation) or a non-ASCII-case variant.
2. The declared rule gives no guidance on whether these should trigger the reset.
3. Two compliant implementations could diverge on whether the message satisfies the trigger.

**Expected behavior**
The trigger contract should state a deterministic normalization domain, e.g. 'trim surrounding whitespace, fold using ASCII case-folding only, reject trailing punctuation as a superset phrase.'

**Actual behavior**
No normalization rule is stated; only 'exactly... case-insensitive' is specified.

```text
user message variants (whitespace/punctuation/case) -> undefined normalization -> implementation-dependent trigger match
```

**Impact**
Visually similar user messages could inconsistently reset mode state across implementations or model runs, undermining the intended determinism of the sole mode-change trigger.

**Suggested correction**
Add an explicit normalization clause (e.g., ASCII-only case folding, trim leading/trailing whitespace, reject any other surrounding characters as a superset phrase) to the WHEN block.

**How to verify**
Add a static test or explicit rule text confirming the normalization domain, and verify edge-case phrases are classified consistently.

</details>

skills/studio/modules/runtime/active-workflow-state-law.md
tkcoding: Deliberately not fixing this one -- ADR-0023 explicitly defers the audit record's field set: "The record's field set is an implementation contract and may evolve without amending this ADR," and ties it to prerequisite 4 ("a write path from a chat gate to `record()`"), which doesn't exist yet. Defining a bounded schema here would be inventing that contract ahead of the change that actually needs it, with nothing to validate it against. Left as-is pending that separate prerequisite; happy to revisit if you'd rather pin a minimal schema now regardless.

ainetx: Please provide the ADR-0023 section that defines the audit-record field-set deferral and prerequisite 4.

skills/studio/modules/runtime/pdsl-execution-card.md
tkcoding: Fixed in 1b33d95c: pdsl-execution-card.md is now the canonical definition (with a NOTES pointer to active-workflow-state-law.md), and active-workflow-state-law.md's rule now reads "...the canonical definition of that carve-out lives in `runtime/pdsl-execution-card.md`" instead of restating it. Covered by `test_declared_gate_type_carveout_is_defined_once_and_cross_referenced`.

tests/test_pdsl_keywords.py
tkcoding: Fixed in 1b33d95c: added `test_declared_gate_type_carveout_is_defined_once_and_cross_referenced` and `test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present`, which pin the carve-out, the skip-redefinition, the audit-record rule, and the mode-persistence rule's SimpleModeChangeTrigger reference.

tests/test_pdsl_keywords.py
tkcoding: Fixed in 1b33d95c: the disjointness test now lower-cases both the trigger and every extracted BNW phrase before comparing.

skills/studio/modules/runtime/pdsl-execution-card.md
tkcoding: Addressed in 1b33d95c via a NOTES clarification on pdsl-execution-card.md: no core MENU declares TYPE yet by design (undeclared = blocking per ADR-0023), so this rule isn't wrong for lacking a same-diff example -- it's a forward-looking carve-out for gates that get migrated to declare TYPE later (tracked on #128). The NOTES makes that sequencing explicit rather than leaving a reader to infer it.

skills/studio/modules/gates/simple-mode.md
tkcoding: Good catch, fixed in 1b33d95c: wired it into active-workflow-state-law.md's message-mapping rule (rule 8) as the first-checked outcome -- "RUN SimpleModeChangeTrigger from `gates/simple-mode.md` when its WHEN is satisfied" -- since that law is the one thing already guaranteed to run on every message for the whole session, regardless of which specific workflow/continuation is currently active. Covered by `test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present`'s assertion on that exact RUN text.

tests/test_pdsl_keywords.py
tkcoding: Fixed in 1b33d95c: the test now checks substring containment in both directions (trigger-in-phrase and phrase-in-trigger), not just exact list membership.

skills/studio/modules/runtime/pdsl-execution-card.md
tkcoding: Fixed in 1b33d95c: `test_declared_gate_type_carveout_is_defined_once_and_cross_referenced` asserts the carve-out phrase, the runtime-derived-exclusion phrase, and the ADR-0023 citation in pdsl-execution-card.md's NOTES.

tests/test_pdsl_keywords.py
tkcoding: Fixed in 1b33d95c: added `test_simple_mode_change_trigger_contract_is_fully_declared`, which pins the case-folding, whitespace-trimming, and paraphrase/synonym/superset-rejection clauses as regression-locked text, not just the trigger phrase itself.

tests/test_pdsl_keywords.py
tkcoding: Fixed in 1b33d95c: the test now uses `_iter_pdsl_blocks` to find the `UNIT BraveNewWorldActivate` block specifically and reads the activation-phrase and exclusion lines only from inside it, rather than a flat file-wide substring scan.

skills/studio/modules/runtime/pdsl-execution-card.md
tkcoding: Addressed in 1b33d95c via a NOTES clarification on pdsl-execution-card.md, citing ADR-0023: no core module resolves a gate from a declared TYPE today -- this carve-out only removes the prohibition against doing so once such resolution exists (that's ADR-0023 prerequisites 3 and 4, not yet built); it does not itself define or implement resolution semantics. Since ADR-0023 explicitly sequences #148 as an inert-wording prerequisite ahead of the actual resolution mechanism, I don't think this warrants a rollback, but glad to discuss further if you see it differently.

skills/studio/modules/gates/simple-mode.md
tkcoding: Fixed in 1b33d95c: the WHEN clause now reads "after trimming leading/trailing whitespace and folding ASCII case, equals exactly the phrase ... with no other leading, trailing, or embedded characters (not a paraphrase, synonym, punctuation-padded variant, or superset phrase)", making the normalization domain explicit.

skills/studio/modules/runtime/active-workflow-state-law.md
tkcoding: From `architecture/ADR/0023-cpt-studio-adr-autonomous-default-and-gate-risk-v1.md`:

- "Recording an autonomous resolution" (lines 287-288): "The record's field set is an implementation contract and may evolve without amending this ADR."
- Same section (lines 310-311): "...it is a CLI-side module with no path yet named from a chat gate to `record()`."
- "What must exist before the default moves", item 4 (lines 372-379): "a write path from a chat gate to `record()`. Recording every permitted autonomous resolution is mandatory above, and no such path exists, so flipping the default without it would produce exactly the unrecorded autonomous resolutions the second invariant forbids. It is a prerequisite rather than a consequence..."

So the ADR both explicitly declines to fix the schema now ("may evolve without amending this ADR") and names the write path -- which is what would actually consume/define that schema -- as prerequisite 4, not yet built. That's the basis for deferring rather than inventing a schema here.

tkcoding:
Re: [BNW's open-ended semantic-equivalence matching is not shown to exclude the new mode-change trigger](#issuecomment) (Major)

Fixed in 1b33d95c: added an explicit NEVER rule to `BraveNewWorldActivate` in `workflows/brave-new-world.md` --

> NEVER treat the exact phrase 'change mode' as BNW activation, semantically equivalent or otherwise; that phrase is reserved exclusively for `gates/simple-mode.md`'s SimpleModeChangeTrigger, and the semantic-equivalence rule above never claims it

`test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases` now also asserts this exclusion rule is present, and checks substring/superset collision against BNW's phrase list in both directions (not just exact list membership, per the follow-up finding on the same test).

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Both files cross-reference each other explicitly: active-workflow-state-law.md states 'the canonical definition of that carve-out lives in `runtime/pdsl-execution-card.md`' and pdsl-execution-card.md's NOTES state 'The identical carve-out is restated for workflow-state law in `runtime/active-workflow-state-law.md`; keep both definitions in sync if this one changes' -- a single source-of-truth pointer exists in both directions.
- **Reviewer:** active-workflow-state-law.md identifies runtime/pdsl-execution-card.md as canonical, and pdsl-execution-card.md explicitly notes the restatement and instructs maintainers to keep both definitions in sync.
- **Verifier:** Both files already cross-reference each other explicitly: active-workflow-state-law.md says 'the canonical definition of that carve-out lives in `runtime/pdsl-execution-card.md`' and pdsl-execution-card.md's NOTES say 'The identical carve-out is restated for workflow-state law in `runtime/active-workflow-state-law.md`; keep both definitions in sync if this one changes' -- plus a new test (test_declared_gate_type_carveout_is_defined_once_and_cross_referenced) asserts both cross-references are present, directly guarding against silent drift.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Although test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases itself only reads simple-mode.md and brave-new-world.md, this same diff adds test_declared_gate_type_carveout_is_defined_once_and_cross_referenced (reads both pdsl-execution-card.md and active-workflow-state-law.md) and test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present (reads active-workflow-state-law.md and asserts on the audit/skip/persistence rule text), so the new normative rules in both other changed files do have direct test coverage in this diff.
- **Reviewer:** Although the named disjointness test reads only simple-mode.md and brave-new-world.md, other new tests in the same diff read active-workflow-state-law.md and pdsl-execution-card.md and assert the carve-out, audit, skip, and persistence text.
- **Verifier:** The disjointness test alone doesn't read active-workflow-state-law.md/pdsl-execution-card.md, but two other new tests in the same diff do: test_declared_gate_type_carveout_is_defined_once_and_cross_referenced asserts specific strings from both files, and test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present asserts the skip/audit/persistence rule text from active-workflow-state-law.md -- so the two files are not left with zero test coverage.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The test lowercases both sides before comparing: `bnw_phrases = [phrase.lower() for phrase in...]` and `trigger = "change mode"` (already lowercase), so `trigger != phrase`, `trigger not in phrase`, and `phrase not in trigger` are all case-insensitive comparisons; a BNW phrase like 'Change Mode' would be lowered to 'change mode' and caught by the equality assertion.
- **Reviewer:** The test lowercases every extracted BNW phrase before comparison, so a case-only collision such as 'Change Mode' is detected.
- **Verifier:** bnw_phrases is built via `[phrase.lower() for phrase in re.findall(...)]` and trigger is already the lowercase literal 'change mode', so the comparisons (`trigger != phrase`, `trigger not in phrase`, `phrase not in trigger`) are effectively case-insensitive already; a BNW phrase like 'Change Mode' would be lowered to 'change mode' and caught by the exact-match assertion.

</details>

skills/studio/modules/gates/simple-mode.md
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** active-workflow-state-law.md's first-priority mapping rule now explicitly routes to it: 'a match against the declared mode-change trigger (RUN SimpleModeChangeTrigger from `gates/simple-mode.md` when its WHEN is satisfied)' is checked before any other outcome, giving every active cf/cf-* workflow a normative dispatch path to the unit, and no other rule in the pre-fetched context (SimpleModeNormal, SimpleModeDebug, etc.) claims to consume the exact 'change mode' command earlier.
- **Reviewer:** ActiveWorkflowStateLaw gives SimpleModeChangeTrigger first priority for each active workflow message when its predicate holds and explicitly RUNs it from gates/simple-mode.md before ordinary state transitions.
- **Verifier:** active-workflow-state-law.md's message-mapping rule explicitly names 'RUN SimpleModeChangeTrigger from `gates/simple-mode.md` when its WHEN is satisfied' as the first-priority outcome for every new user message while a workflow is active, giving a normative routing/dispatch path from active non-unset simple-mode handling to the unit; no earlier interpreter is shown consuming the exact command first.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The test already checks both substring directions -- `assert trigger not in phrase` and `assert phrase not in trigger` -- in addition to exact equality, so a BNW phrase like 'change mode entirely' (containing 'change mode' as a substring) would be caught by `trigger not in phrase`; the check's premise that only exact membership is tested is incorrect.
- **Reviewer:** The test checks both `trigger not in phrase` and `phrase not in trigger` for each normalized BNW phrase, covering substring and superset overlap in both directions.
- **Verifier:** The test's per-phrase loop actually contains three assertions, not one: `trigger != phrase`, `trigger not in phrase` (catches trigger-as-substring-of-BNW-phrase), and `phrase not in trigger` (catches BNW-phrase-as-substring-of-trigger) -- so both substring/superset directions are checked, contradicting the check's premise that only exact membership is tested.

</details>

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_declared_gate_type_carveout_is_defined_once_and_cross_referenced directly reads pdsl-execution-card.md and asserts on its exact new carve-out text ("a gate's declared risk type") and its exact NEVER-exclusion text ("NEVER treat a runtime-derived eligibility or risk classification"), plus the ADR-0023 citation and the cross-reference wording -- so there is text-level regression coverage tying both rules to their literal wording.
- **Reviewer:** The diff adds test_declared_gate_type_carveout_is_defined_once_and_cross_referenced, which reads pdsl-execution-card.md and asserts both the declared-risk-type carve-out and the runtime-derived eligibility exclusion.
- **Verifier:** test_declared_gate_type_carveout_is_defined_once_and_cross_referenced directly asserts the literal new rule strings from pdsl-execution-card.md, including "a gate's declared risk type" and "NEVER treat a runtime-derived eligibility or risk classification", so both the carve-out and its NEVER exclusion do have automated text-level coverage guarding their distinctness.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_mode_change_trigger_is_disjoint_from_brave_new_world_phrases locates the block via `_iter_pdsl_blocks` filtered to `block[0].strip() == "UNIT BraveNewWorldActivate"` before searching within `bnw_activate_block` for the activation-phrase line, so the phrase extraction is structurally bounded to that UNIT rather than a flat file-wide first-match scan.
- **Reviewer:** The test first uses _iter_pdsl_blocks to select the BraveNewWorldActivate UNIT, then searches only that extracted block for its activation and exclusion rules.
- **Verifier:** The new test first locates the block whose first line is exactly 'UNIT BraveNewWorldActivate' via _iter_pdsl_blocks, then searches only within that bounded block's lines for the activation-phrase rule and the exclusion line, rather than scanning the whole file for the first matching prose fragment.

</details>

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** architecture/specs/PDSL.md's pre-existing 'Declared gate risk' section (not newly introduced by this diff) already normatively defines MENU-local TYPE: allowed literal values (confirmation/decision/blocking), at-most-one-per-menu, omission handling, declaration-region boundaries, and near-miss/typo detection, and skills/studio/scripts/studio/utils/pdsl.py implements GATE_TYPES, GATE_HEADER, near-miss detection, and region scanning consistent with that spec -- the runtime prose changes rely on this existing, independently-defined grammar rather than inventing an undefined one.
- **Reviewer:** PDSL.md normatively defines MENU-local static TYPE values (confirmation, decision, blocking), omission semantics, and malformed/duplicate handling; pdsl.py enforces valid literals (PDSL700), duplicates (PDSL701), misplaced declarations (PDSL702), and near-misses (PDSL703).
- **Verifier:** The pre-fetched architecture/specs/PDSL.md already normatively defines the MENU-local TYPE field (confirmation/decision/blocking), omission/duplicate handling, and detection region rules, and skills/studio/scripts/studio/utils/pdsl.py implements matching GATE_TYPES/GATE_HEADER/near-miss detection logic — this grammar and validator contract predate and ground the new runtime-prose carve-out consistently.

</details>

skills/studio/modules/gates/simple-mode.md
ainetx: **RESOLVED** -- Confirmed fixed as of 1b33d95c0d28d9687b61933ec684d1b1722d9526 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** SimpleModeChangeTrigger's WHEN clause explicitly states the normalization domain: 'after trimming leading/trailing whitespace and folding ASCII case, equals exactly the phrase "change mode" with no other leading, trailing, or embedded characters' -- this pins down both whitespace handling (trimmed, no embedded whitespace variance) and case handling (ASCII-only fold, so non-ASCII case variants are excluded from the folding domain and thus would not match).
- **Reviewer:** SimpleModeChangeTrigger explicitly trims leading and trailing whitespace, folds ASCII case, and rejects punctuation-padded and superset variants, defining the matching domain.
- **Verifier:** SimpleModeChangeTrigger's WHEN clause explicitly states matching occurs 'after trimming leading/trailing whitespace and folding ASCII case' and requires an exact match with 'no other leading, trailing, or embedded characters', which deterministically specifies both whitespace and case-folding behavior.

</details>

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The finding should be dismissed: `pdsl-execution-card.md` now explicitly states, "No core module declares a gate `TYPE`... today" and that the carve-out "does not itself define or implement" resolution. The rule is conditional on a future literal TYPE declaration, so `MENU SimpleModeChoice` lacking TYPE does not contradict it.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a component for referencing an undeclared field or type, check whether the governing spec explicitly frames that declaration as a future/conditional extension point rather than a present requirement — absence of the declaration today does not contradict a rule that only applies once such a declaration exists."

</details>

skills/studio/modules/gates/simple-mode.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/modules/runtime/pdsl-execution-card.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/modules/runtime/active-workflow-state-law.md
ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The ADR text confirms the pushback: the "Recording an autonomous resolution" section explicitly says "The record's field set is an implementation contract and may evolve without amending this ADR," and item 4 of "What must exist before the default moves" names the missing write path from a chat gate to record() as a prerequisite not yet built. That means the four required fields (which gate, declared type, resolution source, core version) are already specified at the ADR level as the minimum, and full schema detail is deliberately deferred to the implementation that doesn't exist yet — this isn't an oversight in active-workflow-state-law.md, it's a documented, intentional staging decision.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging an ADR or design doc for lacking a fully bounded/detailed schema or spec, check whether the document itself explicitly defers that detail to a future implementation (e.g. names it as an unbuilt prerequisite or an evolvable implementation contract) — an intentional, documented staging decision is not a gap."

</details>

skills/studio/modules/gates/simple-mode.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Re-verified on 1b33d95c. This resolves the finding within the requested static-test scope. test_simple_mode_change_trigger_contract_is_fully_declared now locks the exact-match rule, ASCII case folding, whitespace trimming, and rejection of paraphrase, synonym, punctuation-padded, and superset inputs. test_active_workflow_state_law_skip_audit_and_mode_persistence_rules_present additionally locks the explicit “NEVER treat a reply as a mode reset” rule and its dispatch to SimpleModeChangeTrigger. 
These are source-contract assertions rather than runtime state tests, which is appropriate here because the PDSL units are declarative. Marking resolved.

ainetx (approved):
Approved. This PR closes out the mode-persistence work from the earlier review round: SimpleModeChangeTrigger is now a properly-declared PDSL unit with a fully specified matching contract (trimmed, case-folded, exact-phrase, no synonyms/paraphrases), the active-workflow-state-law changes correctly dispatch to it instead of just naming it, the declared-gate-type carve-out is defined once in pdsl-execution-card.md and cross-referenced rather than duplicated, and the BNW activation phrase set is now explicitly disjoint from the "change mode" trigger with a test enforcing it. New tests cover each of these regression risks directly, including the specific gaps flagged last time (unreachable rule references, duplicated carve-out text, missing contract coverage). Nothing blocking found.

PR #163, opened 2026-09-09
fix(pdsl): collapse the plan production gate pair, align prep-gate skips

SanjeevSolanki:
## What this changes

**One gate instead of two on the `cf-plan` decomposition path.**

`DecompositionConfirmMenu` asked *"proceed with this decomposition?"* — and its own title
announced that after confirming you would choose how to produce phase files. Answering yes
wrote `plan.toml` and the briefs, then `BriefCheckpointMenu` asked exactly that question.
Nothing was asked of the user in between; only writes happened. Both were unconditional
emits, so every `cf-plan` run paid both stops.

They are now one gate, `PlanProduceChoice`, declaring `TYPE: decision`. Each production option
calls a new `PlanWriteBriefPackage` unit, so the authorisation and the write are the same
decision and nothing is written before the gate resolves. Declaring the type on the survivor
takes the untyped MENU surface from 113 to 111.

Two behaviour improvements fell out of the move:

- the `revise` option now resumes in the decomposition unit (`CONTINUE … after user.reply`,
  the corpus's resume idiom) instead of re-emitting over the same decomposition
- the decline path now says *"nothing written"*, which is true; previously the files were
  already on disk by the time you could decline anything

**The three prep-gate modules now agree on when they skip themselves.** `write-docs` and
`write-skills` auto-skip their explore and brainstorm gates when the target is unambiguous;
`coding` had no such rule. It now carries both, plus the `ORIGINAL_INTENT` precondition its
siblings already had, since the rules decide on that variable.

**A state-precondition defect in the two existing rules.** `gates/plan-first.md` states
`NEVER run without PLAN_FIRST_CONTINUE set by the caller`, and nothing in the shared
`gates/workflow-prep.md` sets it on a caller's behalf — every menu option that reaches the gate
sets it and loads the module itself. The two auto-skip rules did neither, so an auto-skip
reached `PlanFirstGate` with the variable unset and the module unloaded, and the gate's own
`no-plan` option then continued to an unset unit. No reason surfaced; the run just proceeded.
Fixed in all three modules.

## Tests

Four guards, each mutation-tested — 13 mutations: 9 defects caught, and the 4 review
reproductions below all behaving correctly:

- entries into `PlanFirstGate` must set a return unit that **resolves to a defined `UNIT`** and
  load the module — read through `pdsl.scan_blocks` and counted only in `DO`/`RULES`/`OPTIONS`/
  `INVALID` so prose creates no obligation, scoped to the enclosing UNIT/MENU so a multi-line
  clause stays legal, rejecting `unset`, and including the indirect `COMPANION_CONTINUE` route
- the prep-gate modules are **glob-discovered** and must agree on both rules, their emitted
  notes, **the dispatch unit each module's own menu names**, and **the UNIT each rule sits in** —
  target and note alone pin no placement, so both rules could otherwise sit in the explore unit
  and leave the brainstorm gate with no auto-skip
- the brief package cannot be written outside its writer unit, by an option that declines, or
  by any option reaching a consumer of it without writing it; the merged gate's declared type
  is pinned
- the untyped-MENU baseline and its ceiling move together, 113 → 111

Resolving continuation targets against the corpus surfaces two that name no defined `UNIT`:
`CodingDispatch` (4 sites) and `WriteDocsDispatch` (1). Both are grandfathered in a documented
set that, like the untyped baseline, may only shrink — choosing their real targets is a flow
decision rather than a rename.

This revision also corrects two docstrings in `tests/test_pdsl_keywords.py` that described
`blocking` as an undeclared gate's *current* runtime behaviour. It is the intended contract;
three shipped paths still resolve an undeclared gate by runtime judgement. Carried forward from
review on the two preceding PRs, and folded in here because this is the next change to touch
that file.

`cfs validate` PASS (0 errors, 0 warnings, 237/237) · `make test` 5631 passed ·
`spec-coverage` 90.8% coverage / 0.4609 granularity, **both unchanged by this change** ·
`pylint`, `ruff`, `vulture-ci`, `test-coverage` clean.

## Disclosed, and deliberately not addressed here

**The three prep-gate modules are unreachable from any shipped workflow.** They load only from
`coding-intent-companion.md` / `write-docs-intent-routing.md` / `write-skills-intent-routing.md`,
and nothing under `workflows/` or `skills/` references those three. The live coding entry is
`workflows/coding.md` → `coding-gen.md` → … → `CodingAuthorGitSetup`, which never enters
`CodingExploreGate` or `CodingBrainstormGate`. So the prep-gate half of this PR is an
alignment fix with **no observable effect today** — it removes no stop a user currently pays,
and the defect it fixes could not have bitten anyone. It is worth having because the three
modules should not disagree while that path is disconnected. Raised separately as an issue,
with the load chain.

**`PLAN_FIRST_CONTINUE = CodingDispatch` names a unit that does not exist**, in four clauses —
`coding-prep-gates.md:32,36,37` and `coding-intent-companion.md:40`, three of them
pre-existing. Candidate real targets are `CodingAuthorGitSetup` and `CodingGenDispatch`;
choosing between them is a flow decision rather than a rename, so it is in the issue rather
than patched here.

**`PlanStorageChoice` and `PlanSaveGateMenu` still give different options depending on entry
path** — memory/disk/revise/stop from the coding path, save/skip from `cf-plan` — so a
plan-then-code session gets a different plan-storage contract than a code-directly session.
Collapsing them redefines `accepted_plan_active` and needs a shared module, which belongs with
the change that owns that variable.

No test here walks a workflow path or counts stops; the suite's PDSL tests are static scans,
and this PR adds no walker.

## Summary by CodeRabbit

- **New Features**
  - Added plan production choices for inline work, guided prompts, subagents, briefs-only output, revision, or cancellation.
  - Added streamlined handling for clear, single-file coding, documentation, and skills tasks.

- **Workflow Improvements**
  - Plan summaries now appear before production choices; plan files and briefs are created only after a production mode is selected.
  - Automatic skips now continue through the correct authoring workflow and preserve the intended next step.

- **Bug Fixes**
  - Fixed routing issues that could bypass required authoring modules or lose the intended continuation path.

### DIFF
--- a/skills/studio/modules/plan-compile.md
+++ b/skills/studio/modules/plan-compile.md
@@ -2,27 +2,48 @@
 
 ```pdsl
 UNIT PlanPhase3Compile
-PURPOSE: Write plan.toml and one brief per phase, then choose how to produce phase files (Phase 3).
+PURPOSE: Choose how to produce phase files, which is also the authorisation to write plan.toml and the briefs (Phase 3).
 STATE:
   SET CF_PHASE_GATE: released_for_orchestrator_write | released_for_dispatch | armed | unset (default armed, scope workflow_run)
 DO:
   LOAD {cf-studio-path}/.core/requirements/plan-template.md and {cf-studio-path}/.core/requirements/brief-template.md and follow them
-  SET CF_PHASE_GATE = released_for_orchestrator_write (scope plan.toml), WRITE {cf-studio-path}/.plans/{task-slug}/plan.toml ([meta] + [plan] + [[phases]] per the template), SET CF_PHASE_GATE = armed
-  SET CF_PHASE_GATE = released_for_orchestrator_write (scope brief-*.md), WRITE one brief-{NN}-{slug}.md per phase (~50-80 lines; context boundary, metadata, load instructions, budget; never copy kit content), SET CF_PHASE_GATE = armed
-  EMIT_MENU BriefCheckpointMenu
+  EMIT_MENU PlanProduceChoice
   WAIT user.reply
   STOP_TURN
 RULES:
+  NEVER write any file before PlanProduceChoice resolves — every option that writes calls PlanWriteBriefPackage, so the authorisation and the write are the same decision
+  ALWAYS read each brief FROM DISK before compiling
+  ALWAYS reopen CF_PHASE_GATE released_for_orchestrator_write (scoped) before the phase-file write in option 1 and reset to armed immediately after
+MENU PlanProduceChoice
+TITLE: Decomposition ready. Choose how to produce phase files — options 1-4 also authorise writing plan.toml + N brief files under .plans/{task-slug}/, and options 5-6 write nothing: 1 inline (uses this chat's budget); 2 prompts (skips validation); 3 subagents (needs sub-agent approval); 4 briefs-only (write the package and stop); 5 revise the decomposition; 6 no (write nothing). Reply with a number.
+TYPE: decision
+OPTIONS:
+  1 inline -> RUN PlanWriteBriefPackage; LOAD {cf-studio-path}/.core/skills/studio/modules/plan-validate-finalize.md; compile each phase file inline from its on-disk brief (apply a context boundary, read brief from disk, WRITE phase-NN-*.md with CF_PHASE_GATE released/armed), then CONTINUE PlanPhase3Validate
+  2 prompts -> RUN PlanWriteBriefPackage; emit one self-contained downstream compilation prompt per brief (no phase files written), SET plan.execution_status="prompts_emitted", EMIT "Start a new chat and paste Phase 1's prompt to begin compilation. Each prompt is self-contained and independent.", then LOAD {cf-studio-path}/.core/skills/studio/modules/ui/next-actions.md WHEN NextActionsOffer is not yet loaded; RUN NextActionsOffer (Phase 3.4 validation skipped in this mode)
+  3 subagents -> RUN PlanWriteBriefPackage; LOAD {cf-studio-path}/.core/skills/studio/modules/plan-compiler-dispatch.md; CONTINUE PlanPhaseCompilerDispatch
+  4 briefs-only — write plan.toml + briefs and stop there -> RUN PlanWriteBriefPackage; SET plan.execution_status="briefs_only"; LOAD {cf-studio-path}/.core/skills/studio/modules/ui/next-actions.md WHEN NextActionsOffer is not yet loaded; RUN NextActionsOffer
+  5 revise — describe what to change about this decomposition -> EMIT "Describe what to change (e.g. split phase 2, merge phases 3 and 4, adjust scope). I will rework and re-show."; CONTINUE PlanPhase2Decompose after user.reply; WAIT user.reply; STOP_TURN
+  6 n | no -> EMIT "Decomposition declined — nothing written. Rework boundaries and re-run cf-plan when ready." and STOP_TURN
+  INVALID -> EMIT_MENU PlanProduceChoice
+```
+
+```pdsl
+UNIT PlanWriteBriefPackage
+PURPOSE: Write plan.toml and one brief per phase once PlanProduceChoice has authorised it (Phase 3.2).
+STATE:
+  SET CF_PHASE_GATE: released_for_orchestrator_write | released_for_dispatch | armed | unset (default armed, scope workflow_run)
+WHEN:
+  REQUIRE the user chose a production mode at PlanProduceChoice
+DO:
+  SET CF_PHASE_GATE = released_for_orchestrator_write (scope plan.toml), WRITE {cf-studio-path}/.plans/{task-slug}/plan.toml ([meta] + [plan] + [[phases]] per the template), SET CF_PHASE_GATE = armed
+  SET CF_PHASE_GATE = released_for_orchestrator_write (scope brief-*.md), WRITE one brief-{NN}-{slug}.md per phase (~50-80 lines; context boundary, metadata, load instructions, budget; never copy kit content), SET CF_PHASE_GATE = armed
+  EMIT the brief package summary — plan.toml + N briefs, 0/N phase files
+RULES:
   ALWAYS write plan.toml before any brief or phase file
   ALWAYS reopen CF_PHASE_GATE released_for_orchestrator_write (scoped) before each write and reset to armed immediately after
-  ALWAYS read each brief FROM DISK before compiling
-  NEVER emit "Plan created" at this checkpoint
-MENU BriefCheckpointMenu
-TITLE: Brief package prepared (plan.toml + N briefs, 0/N phase files) — choose how to produce phase files: 1 inline (uses this chat's budget); 2 prompts (skips validation); 3 subagents (needs sub-agent approval); 4 stop (keep briefs). Reply with a number.
-OPTIONS:
-  1 inline -> LOAD {cf-studio-path}/.core/skills/studio/modules/plan-validate-finalize.md; compile each phase file inline from its on-disk brief (apply a context boundary, read brief from disk, WRITE phase-NN-*.md with CF_PHASE_GATE released/armed), then CONTINUE PlanPhase3Validate
-  2 prompts -> emit one self-contained downstream compilation prompt per brief (no phase files written), SET plan.execution_status="prompts_emitted", EMIT "Start a new chat and paste Phase 1's prompt to begin compilation. Each prompt is self-contained and independent.", then LOAD {cf-studio-path}/.core/skills/studio/modules/ui/next-actions.md WHEN NextActionsOffer is not yet loaded; RUN NextActionsOffer (Phase 3.4 validation skipped in this mode)
-  3 subagents -> LOAD {cf-studio-path}/.core/skills/studio/modules/plan-compiler-dispatch.md; CONTINUE PlanPhaseCompilerDispatch
-  4 stop -> SET plan.execution_status="briefs_only"; LOAD {cf-studio-path}/.core/skills/studio/modules/ui/next-actions.md WHEN NextActionsOffer is not yet loaded; RUN NextActionsOffer
-  INVALID -> EMIT_MENU BriefCheckpointMenu
+  ALWAYS emit the brief package summary only after a complete write — it reports a package the mode was chosen before, and it is the only completeness signal a downstream consumer has, so a failed or interrupted write emits none and recovery is a fresh PlanProduceChoice choice that rewrites plan.toml before any brief
+  NEVER emit "Plan created" in this summary — no phase file exists yet
+  NEVER run without a production mode chosen at PlanProduceChoice
+ON_ERROR:
+  write_failed -> EMIT "The brief package is incomplete — plan.toml or a brief file failed to write, so nothing will run against it."; EMIT "ERROR: <the raw error from the failed write>"; SET CF_PHASE_GATE = armed; STOP_TURN
 ```
--- a/tests/test_pdsl_keywords.py
+++ b/tests/test_pdsl_keywords.py
@@ -10,6 +10,8 @@ from collections import defaultdict
 from pathlib import Path
 from unittest import mock
 
+import pytest
+
 from studio.utils import pdsl
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
@@ -650,9 +652,11 @@ def test_prompt_runtime_references_use_cf_studio_path() -> None:
 
 
 # Every MENU that does not yet declare a gate risk TYPE, frozen 2026-09-07.
-# An undeclared gate is treated as `blocking` at runtime, so the existing
-# surface is grandfathered and migrates gate by gate -- but the untyped surface
-# must not grow. Anything not in this set has to declare a TYPE.
+# `blocking` is the *intended* contract for an undeclared gate, not what happens
+# today: three shipped paths still resolve one by runtime judgement, so this set
+# freezes the existing untyped inventory rather than recording a fail-closed
+# default. The surface migrates gate by gate and must not grow -- anything not in
+# this set has to declare a TYPE.
 # Shrink this set as menus are typed; never add to it.
 UNTYPED_MENU_BASELINE: frozenset[str] = frozenset({
     "architecture/specs/PDSL.md#13::SubAgentApprovalMenu",
@@ -730,8 +734,6 @@ UNTYPED_MENU_BASELINE: frozenset[str] = frozenset({
     "skills/studio/modules/map-intent.md#0::MapIntentMenu",
     "skills/studio/modules/map-next.md#0::MapNextStepsMenu",
     "skills/studio/modules/map-preflight.md#2::MapScopeMenu",
-    "skills/studio/modules/plan-assess-decompose.md#1::DecompositionConfirmMenu",
-    "skills/studio/modules/plan-compile.md#0::BriefCheckpointMenu",
     "skills/studio/modules/plan-compiler-dispatch.md#2::PlanCompilerFailureMenu",
     "skills/studio/modules/plan-discovery.md#1::PlanGateMenu",
     "skills/studio/modules/plan-validate-finalize.md#0::OversizedPhaseRecoveryMenu",
@@ -772,7 +774,9 @@ UNTYPED_MENU_BASELINE: frozenset[str] = frozenset({
 
 # The untyped surface may only shrink. Raising this is a deliberate, reviewable
 # act; a rename does not need it, because a rename leaves the count unchanged.
-UNTYPED_MENU_BASELINE_CEILING = 113
+# 113 -> 111: DecompositionConfirmMenu and BriefCheckpointMenu collapsed into one
+# menu that declares `TYPE: decision`, so two entries left and none replaced them.
+UNTYPED_MENU_BASELINE_CEILING = 111
 
 
 
@@ -848,13 +852,102 @@ def _menu_type_declarations() -> dict[str, str | None]:
     return declarations
 
 
+#: The paths that auto-resolve a gate by runtime judgement rather than by reading
+#: a declared TYPE. Named here so the claim in UNTYPED_MENU_BASELINE's comment is
+#: pinned to something: if a path is retired or stops judging, the guard below
+#: fails and the comment has to be corrected with it. Sourced from
+#: `architecture/specs/PDSL.md`, which cites the same three.
+RUNTIME_JUDGEMENT_PATHS = {
+    "skills/studio/modules/gates/simple-mode-rules.md": "ALWAYS choose automatically only when",
+    "workflows/brave-new-world.md": "allowing this overlay to answer eligible menus",
+    "skills/studio/modules/subagents/dispatch.md": "SUB_AGENT_GROUP_DECISION = approve-once",
+}
+
+#: The claim has a second half -- that none of those paths *reads* a declaration --
+#: and it is the half that changes first, because binding them to declared types
+#: is what the default-flip does. Matched on the reading vocabulary rather than on
+#: the bare type tokens, since `SUB_AGENT_GROUP_DECISION` would otherwise match
+#: "decision". A proxy, not a proof: a path could read a declaration in wording
+#: this misses.
+DECLARED_TYPE_READ_RE = re.compile(
+    r"declared\s+TYPE|gate\s+risk|TYPE:\s*(?:confirmation|decision|blocking)\b",
+    re.IGNORECASE,
+)
+
+
+def test_the_runtime_judgement_paths_named_in_the_baseline_comment_still_exist() -> None:
+    """Pin the claim that undeclared gates are resolved by judgement, not fail-closed.
+
+    `UNTYPED_MENU_BASELINE`'s comment and this module's docstrings state that
+    `blocking` is the intended contract rather than current behaviour, because
+    shipped paths still auto-resolve an undeclared gate by runtime judgement.
+    That is a factual claim about three specific files, and prose cannot hold it:
+    a path could be retired or bound to declared types and the comment would
+    quietly become false, which is the same overclaim this wording replaced.
+
+    So the count, the paths and both halves of the claim are asserted: each path
+    still carries its auto-resolution rule, and none of them reads a declared
+    type. The second half matters more, because binding these paths to declared
+    types is exactly what the default flip does -- so that is where the comment
+    will go stale first. An earlier version of this guard checked only the first
+    half and passed when a path gained a declared-type read.
+
+    What still gets past it: a path that reads a declaration in wording
+    `DECLARED_TYPE_READ_RE` does not match. The fix when this fails is to correct
+    the comment in the same change, not to edit these constants to match.
+    """
+    assert len(RUNTIME_JUDGEMENT_PATHS) == 3, (
+        f"The comment on UNTYPED_MENU_BASELINE says three paths resolve gates by "
+        f"runtime judgement; this guard names {len(RUNTIME_JUDGEMENT_PATHS)}. Keep "
+        "the two in step."
+    )
+
+    findings: list[str] = []
+    for rel, marker in sorted(RUNTIME_JUDGEMENT_PATHS.items()):
+        path = REPO_ROOT / rel
+        if not path.is_file():
+            findings.append(f"{rel}: no longer exists")
+            continue
+        # Read through the validator's own reader, which normalizes OSError and
+        # UnicodeDecodeError into a reported error: an unreadable path is a
+        # finding about that path, not a crash that hides the other two.
+        text, error = pdsl.read_source_file(path)
+        if error is not None or text is None:
+            findings.append(
+                f"{rel}: cannot be read, so the claim cannot be checked "
+                f"({error.message if error else 'no content'})"
+            )
+            continue
+        if marker not in text:
+            findings.append(f"{rel}: no longer carries {marker!r}")
+        read = DECLARED_TYPE_READ_RE.search(text)
+        if read is not None:
+            findings.append(
+                f"{rel}: now reads a declared type ({read.group(0)!r}), so it may no "
+                "longer decide by runtime judgement"
+            )
+
+    assert not findings, "\n  ".join(
+        [
+            "The runtime-judgement paths the baseline comment relies on have changed:",
+            *findings,
+            "",
+            "If a path was retired or now reads a declared TYPE, update the comment on "
+            "UNTYPED_MENU_BASELINE and the docstring below it -- `blocking` may have "
+            "become the actual behaviour rather than only the intended contract.",
+        ]
+    )
+
+
 def test_the_untyped_menu_surface_does_not_grow() -> None:
     """A newly introduced MENU must declare a gate risk TYPE.
 
-    The tail recorded in UNTYPED_MENU_BASELINE is grandfathered because an
-    undeclared gate is treated as `blocking`, which is the conservative
-    direction. What must not happen is the untyped surface growing, so a MENU
-    that is neither typed nor in the baseline fails here.
+    The tail recorded in UNTYPED_MENU_BASELINE is grandfathered so the surface
+    can migrate one gate at a time. It is *not* grandfathered because undeclared
+    already means `blocking` -- that is the intended contract, and three shipped
+    paths still resolve an undeclared gate by runtime judgement today. This test
+    freezes the existing untyped inventory: what must not happen is that
+    inventory growing, so a MENU neither typed nor in the baseline fails here.
     """
     declarations = _menu_type_declarations()
     untyped_now = {key for key, value in declarations.items() if value is None}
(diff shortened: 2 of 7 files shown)


tests/test_pdsl_keywords.py
```diff
@@ -887,6 +887,229 @@ def test_the_untyped_menu_surface_does_not_grow() -> None:
...
+    caller's behalf. Two auto-skip rules did not: they reached the gate with the
+    variable unset and the module unloaded, so the gate's own `no-plan` option
+    continued to an unset unit -- silently, no reason surfaced, the run just
+    proceeded.
+
+    Scoped to the enclosing UNIT/MENU rather than the single line, so splitting a
+    long clause across lines stays legal; `unset` is rejected because it is the
+    exact state the gate forbids; and the indirect entry via `COMPANION_CONTINUE`
+    is included, because a clause added there fails the same way.
+    """
```
ainetx, line 1106: ### PlanFirstGate regression guard accepts any identifier as a valid continuation target

**Severity:** Minor

**Problem**
test_every_entry_into_plan_first_gate_sets_its_return_unit only rejects the literal token 'unset' for PLAN_FIRST_CONTINUE; any other identifier-shaped string is accepted as 'real', without cross-checking it against defined UNIT names in the target module or a maintained allowlist.

<details>

**How to reproduce**
1. Introduce a typo, e.g. 'SET PLAN_FIRST_CONTINUE = CodingDispatchh' (extra 'h') in coding-prep-gates.md. 2. Run test_every_entry_into_plan_first_gate_sets_its_return_unit. 3. The test passes because 'CodingDispatchh' is a non-'unset' identifier.

**Expected behavior**
The test should fail, flagging that PLAN_FIRST_CONTINUE points to a unit that doesn't exist anywhere reachable.

**Actual behavior**
The test passes silently; the runtime bug (PlanFirstGate cannot CONTINUE to an undefined unit) is only discoverable at execution time.

```text
SET PLAN_FIRST_CONTINUE=<typo'd-name> --> regex captures token --> token != 'unset' --> treated as 'real' --> test PASS (but PlanFirstGate CONTINUE fails at runtime)
```

**Impact**
A future typo in any of the several PLAN_FIRST_CONTINUE assignment sites would ship without test failure, only surfacing when a user actually exercises the auto-skip path.

**Suggested correction**
Resolve each captured unit name against the set of UNIT names defined across the relevant module(s) (or a maintained registry) and fail if it is not found.

**How to verify**
Add a fixture with an undefined PLAN_FIRST_CONTINUE target and confirm the test now fails.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -887,6 +887,229 @@ def test_the_untyped_menu_surface_does_not_grow() -> None:
...
+    so the same task would have cost two extra stops purely because it was code.
+    Stated in the conditional because no shipped workflow currently loads any of
+    these three modules -- they are reachable only through `*-intent-routing.md`
+    /`coding-intent-companion.md`, which nothing references -- so the divergence
+    is latent. This guard exists so the three stay aligned until that path is
+    reconnected, at which point the difference becomes observable.
+
+    Discovered by glob, so a fourth prep-gate module added without the rules
+    fails here rather than being noticed in review. Each rule's continuation
+    target is checked too: a rule copied from the explore gate into the
```
ainetx: ### New prep-gate consistency test does not verify per-module PLAN_FIRST_CONTINUE target

**Severity:** Minor

**Problem**
test_prep_gate_modules_apply_the_same_auto_skip_rules matches rules by 'target in rule' where target is the fixed string 'CONTINUE PlanFirstGate', then only checks for the shared condition phrase and note text. It never parses or compares the 'SET PLAN_FIRST_CONTINUE = <Unit>' portion of the rule or the module-specific LOAD path, both of which the diff just introduced as per-module differences.

<details>

**How to reproduce**
1. In write-docs-prep-gates.md, change 'SET PLAN_FIRST_CONTINUE = WriteDocsAuthorDispatch' to 'SET PLAN_FIRST_CONTINUE = CodingDispatch' (a copy-paste mistake from coding-prep-gates.md), leaving 'CONTINUE PlanFirstGate' and the note text intact. 2. Run test_prep_gate_modules_apply_the_same_auto_skip_rules. 3. The test still passes.

**Expected behavior**
The test should fail because the write-docs module now sets an incorrect dispatch target, which would silently route write-docs' skip path into coding's dispatch unit.

**Actual behavior**
Test passes because it never checks the PLAN_FIRST_CONTINUE assignment value, only the fixed 'CONTINUE PlanFirstGate' substring and note text shared by all three modules.

```text
rule text --> matches 'CONTINUE PlanFirstGate' substring --> condition+note present --> PASS (wrong SET PLAN_FIRST_CONTINUE value undetected)
```

**Impact**
A future edit could silently misroute one prep-gate module's auto-skip into another module's dispatch unit without any test catching it.

**Suggested correction**
Extract and compare the PLAN_FIRST_CONTINUE assignment value (and LOAD path) per module against an expected per-module value, not just presence of shared substrings.

**How to verify**
Introduce a wrong PLAN_FIRST_CONTINUE value in one of the three modules and confirm the test now fails.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -887,6 +887,229 @@ def test_the_untyped_menu_surface_does_not_grow() -> None:
...
+
+def test_every_entry_into_plan_first_gate_sets_its_return_unit() -> None:
+    """`PlanFirstGate` may not be entered without a real `PLAN_FIRST_CONTINUE`.
+
+    `gates/plan-first.md:19` states `NEVER run without PLAN_FIRST_CONTINUE set by
+    the caller`, and nothing in the shared `gates/workflow-prep.md` sets it on a
+    caller's behalf. Two auto-skip rules did not: they reached the gate with the
+    variable unset and the module unloaded, so the gate's own `no-plan` option
+    continued to an unset unit -- silently, no reason surfaced, the run just
+    proceeded.
```
ainetx, line 1097: ### Plan-first entry guard counts prose mentions as executable entries

**Severity:** Minor

**Problem**
A comment, title, example, or explanatory text containing `CONTINUE PlanFirstGate` or `COMPANION_CONTINUE = PlanFirstGate` is treated as a gate entry and then checked for assignments and loads.

<details>

**How to reproduce**
1. Add prose containing `CONTINUE PlanFirstGate` outside an executable action. 2. Do not add a matching assignment. 3. Run the guard. 4. It reports an unguarded entry.

**Expected behavior**
Only executable PDSL action clauses create PlanFirstGate-entry obligations.

**Actual behavior**
Any matching line in the scanned corpus is considered an entry.

```text
prose mention -> substring match -> treated as entry -> false failure
```

**Impact**
Documentation can cause false regressions or inflate the inspection floor.

**Suggested correction**
Parse PDSL blocks and action sections before matching continuation operations.

**How to verify**
Add a fixture containing the phrase in a comment and assert it is ignored.

</details>

skills/studio/modules/plan-compile.md
```diff
@@ -2,27 +2,46 @@
...
+UNIT PlanWriteBriefPackage
+PURPOSE: Write plan.toml and one brief per phase once PlanProduceChoice has authorised it (Phase 3.2).
+STATE:
+  SET CF_PHASE_GATE: released_for_orchestrator_write | released_for_dispatch | armed | unset (default armed, scope workflow_run)
+WHEN:
+  REQUIRE the user chose a production mode at PlanProduceChoice
+DO:
+  SET CF_PHASE_GATE = released_for_orchestrator_write (scope plan.toml), WRITE {cf-studio-path}/.plans/{task-slug}/plan.toml ([meta] + [plan] + [[phases]] per the template), SET CF_PHASE_GATE = armed
+  SET CF_PHASE_GATE = released_for_orchestrator_write (scope brief-*.md), WRITE one brief-{NN}-{slug}.md per phase (~50-80 lines; context boundary, metadata, load instructions, budget; never copy kit content), SET CF_PHASE_GATE = armed
+  EMIT the brief package summary — plan.toml + N briefs, 0/N phase files
```
ainetx, line 40: ### Revise option in PlanProduceChoice reorders CONTINUE ahead of WAIT/STOP_TURN

**Severity:** Minor

**Problem**
The old DecompositionConfirmMenu option 3 read `EMIT "..."; WAIT user.reply; STOP_TURN; CONTINUE PlanPhase2Decompose` — EMIT, then WAIT, then STOP_TURN, then (on the next turn) CONTINUE. The new PlanProduceChoice option 5 reads `EMIT "..."; CONTINUE PlanPhase2Decompose after user.reply; WAIT user.reply; STOP_TURN` — CONTINUE is now written textually before WAIT and STOP_TURN, with only a dangling 'after user.reply' qualifier to indicate it actually executes later. This breaks the vocabulary convention (RUN/CONTINUE/EMIT/EMIT_MENU/WAIT/STOP_TURN in a fixed, sequential order) used by every sibling menu and by option 6 in this same menu.

<details>

**How to reproduce**
1. Open plan-compile.md option 5 in PlanProduceChoice.
2. Read the action list left-to-right as PDSL normally executes it (top-to-bottom, semicolon-separated).
3. Compare against option 6 or the old option 3 in DecompositionConfirmMenu, which keep CONTINUE last.

**Expected behavior**
Action sequences should keep the terminal CONTINUE after WAIT/STOP_TURN (as in the code this replaces), or use an unambiguous different construct if it must appear earlier.

**Actual behavior**
CONTINUE PlanPhase2Decompose is textually sequenced before WAIT user.reply and STOP_TURN, with 'after user.reply' as the only signal it's deferred.

```text
old: EMIT -> WAIT -> STOP_TURN -> (next turn) CONTINUE
new: EMIT -> CONTINUE(after user.reply) -> WAIT -> STOP_TURN
```

**Impact**
A generator/interpreter following the literal token order could misparse this as an immediate CONTINUE before waiting for the user, which would prematurely re-enter PlanPhase2Decompose without the user's revision input.

**Suggested correction**
Reorder to `EMIT "..."; WAIT user.reply; STOP_TURN; CONTINUE PlanPhase2Decompose` matching the established idiom, dropping the redundant 'after user.reply' qualifier.

**How to verify**
Confirm the revise option renders as EMIT;WAIT;STOP_TURN;CONTINUE like every other terminal-CONTINUE branch in this file and gates/plan-first.md.

</details>

tests/test_pdsl_keywords.py
SanjeevSolanki: Accepted, and it turned up two live instances.

The guard now resolves every captured target against the set of `UNIT` names defined across the
prompt corpus (802 of them), not just against the literal `unset`. Your reproduction —
`CodingDispatchh` with the extra `h` — now fails the guard.

Running the resolution over the corpus first found that two existing targets name no defined
unit at all:

- `CodingDispatch` — 4 sites: `coding-prep-gates.md:34,38,39` and `coding-intent-companion.md:40`.
  `workflows/coding.md` is now a thin `CodingAlias`, and the test asserting `UNIT CodingDispatch`
  exists is marked `xfail`.
- `WriteDocsDispatch` — 1 site, `write-docs-intent-routing.md:52`.

Both are grandfathered in `DANGLING_CONTINUATION_TARGETS`, documented, and — like
`UNTYPED_MENU_BASELINE` — the set may only shrink. I did not repoint them: the candidates are
`CodingAuthorGitSetup` and `CodingGenDispatch`, and choosing between them is a flow decision
rather than a rename. Reported in #164 along with the reachability finding, since these clauses
sit inside the unreachable region.

tests/test_pdsl_keywords.py
SanjeevSolanki: Accepted — Major is the right severity, and your reproduction is exactly right: substituting
`CodingDispatch` into `write-docs-prep-gates.md` left every shared substring matching and the
test green, while silently routing write-docs' skip path into coding's dispatch unit.

Rather than compare against an expected per-module table, which would drift, the guard now
derives the expected target **from the module itself**: the auto-skip rule's
`SET PLAN_FIRST_CONTINUE = …` must match what that module's own menu options set. That is the
invariant the change was supposed to establish — a rule mirroring its own menu — so asserting
it directly is better than restating it in the test. Your reproduction now fails.

tests/test_pdsl_keywords.py
SanjeevSolanki: Accepted. The guard was scanning raw lines, so a `PURPOSE`, `TITLE` or `NOTES` line quoting
`CONTINUE PlanFirstGate` created an obligation and could fail the build, and would also have
inflated the inspection floor.

It now reads through `pdsl.read_source_file` + `pdsl.scan_blocks` — the validator's own
scanner, so the guard cannot disagree with the checker about what is inside a `pdsl` fence —
and counts only lines in `DO`, `RULES`, `OPTIONS` and `INVALID`. `RULES` is included because
that is where the auto-skip clauses live; `PURPOSE`, `TITLE` and `NOTES` are excluded. Verified
with your fixture: a `NOTES:` line mentioning both `CONTINUE PlanFirstGate` and
`COMPANION_CONTINUE = PlanFirstGate`, with no assignment, is now ignored.

SanjeevSolanki:
**Re: CodeRabbit nitpick — "Scope each auto-skip rule to its enclosing UNIT"** (`tests/test_pdsl_keywords.py`)

Accepted, and the docstring you quoted was the giveaway — it claimed a rule copied from the
explore gate into the brainstorm gate would be caught, and placement was never checked.

I had mutation-tested a *different* variant: replacing the brainstorm rule with a copy of the
explore rule, which fails because then no rule continues to `PlanFirstGate`. Your variant —
both rules present but both sitting in the explore unit — passed cleanly. Verified before
fixing.

Each rule is now scoped to its enclosing UNIT via `_sectioned_lines`, as you suggested. The two
units are derived from the module rather than named in the test: the explore unit is whichever
one sets `WORKFLOW_PREP_BRAINSTORM_GATE`, and the brainstorm unit is the one it names — so the
check holds for all three prep-gate modules and for any fourth the glob picks up, without a
table to keep in sync. Your reproduction now fails.

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of db5aba947c8128a179ef57e84fdf6dafb1a61c8b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The guard does more than accept any identifier: after extracting via PLAN_FIRST_ASSIGNMENT_RE it computes `real = units - {"unset"}`, then for each unit in `real - DANGLING_CONTINUATION_TARGETS` checks `if unit not in defined` (defined = every `UNIT <Name>` in the corpus via `_defined_unit_names()`), flagging a finding if not found — so a typo'd or non-existent unit fails the test. Known-dangling targets (CodingDispatch, WriteDocsDispatch) are explicitly allowlisted with a documented reason and the allowlist may only shrink, matching the 'repository-maintained allowlist' expectation in the check.
- **Reviewer:** The guard resolves every non-`unset` captured target against `_defined_unit_names()` and fails unknown names; the only exceptions are the explicit, repository-maintained `DANGLING_CONTINUATION_TARGETS` allowlist, which includes the grandfathered CodingDispatch/WriteDocsDispatch cases.
- **Verifier:** The diff's regex only captures the identifier, but the test does not stop there: it computes `real = units - {"unset"}`, then for `unit in sorted(real - DANGLING_CONTINUATION_TARGETS): if unit not in defined: findings.append(...)`, where `defined = _defined_unit_names()` scans every prompt file for `UNIT <Name>` headers. So a typo'd or invented identifier that isn't a real UNIT anywhere in the corpus fails the assertion, and the two currently-known non-existent targets (CodingDispatch, WriteDocsDispatch) are explicitly carved into an allowlist that 'may only shrink' rather than silently accepted forever -- this is exactly the resolve-to-a-defined-unit behavior the check asks for.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The finding is addressed: `test_every_entry_into_plan_first_gate_sets_its_return_unit` now obtains `rows = _sectioned_lines(path, actions_only=True)`, and `_sectioned_lines` excludes every section except `DO`, `RULES`, `OPTIONS`, and `INVALID`. A `PURPOSE`, `TITLE`, or `NOTES` mention therefore cannot increment `checked` or create a missing-assignment finding.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a filter or gate as too permissive (e.g. counting things it shouldn't), verify what filtering the code/tests actually apply upstream — check the exact scoping (section/type restriction) rather than assuming the raw text is scanned unfiltered."

</details>

skills/studio/modules/plan-compile.md
SanjeevSolanki: Declining the reorder, with measurements — but the impact paragraph is right, and it points at a
real gap that I think explains why we disagree.

**The suggested form has no instances in the corpus.** `grep -rn "STOP_TURN; CONTINUE" skills/ workflows/`
returns **0**. Before this PR it returned exactly 1: the very line under discussion, as option 3
of `DecompositionConfirmMenu`. So the idiom described as established was this single occurrence.

**The spec makes that form unreachable.** `architecture/specs/PDSL.md:175` — *"`DO` actions run
in written order unless `CONTINUE`, `RETURN`, `WAIT`, or `STOP_TURN` transfers control earlier"* —
and `:186` — *"`WAIT` plus `STOP_TURN` is a hard assistant-turn boundary."* A `CONTINUE` written
after `STOP_TURN` sits past that boundary, so it never runs. `:187` adds that `CONTINUE` *"is not
optional advice"*, which is what makes an unreachable one a defect rather than a nicety.

**That was a live regression this PR introduced, and reverting it would restore it.** At HEAD the
clause lived inside `PlanPhase2Decompose` itself, so resuming in place landed in the decomposition
unit and the rework happened regardless of the dead `CONTINUE`. Moving the option into
`PlanProduceChoice` — emitted by `PlanPhase3Compile`, whose entire `DO` is
`LOAD templates; EMIT_MENU; WAIT; STOP_TURN` — means resuming in place re-emits the same menu over
the same unrevised decomposition and discards the user's revision text. Latent before, live after.
That is why the ordering changed.

**The siblings cited are not comparable.** `plan-first.md:24,25` and `plan-compile.md:21,23` are
**non-waiting** branches: a terminal `CONTINUE` there executes immediately and is correct. Option 6
in this menu likewise ends at `STOP_TURN` with no continuation. `revise` is the only branch that
must wait for input *and then* resume, which is a different shape.

**The form used matches the corpus's sole precedent for that shape**, `coding-intent-companion.md:5-9`:

```pdsl
DO:
  EMIT "Describe the code work you want done: …"
  CONTINUE CodingIntentResume after user.reply
  WAIT user.reply
  STOP_TURN
```

EMIT → `CONTINUE … after user.reply` → WAIT → STOP_TURN. Identical ordering to option 5.

---

**Where you are right.** *"A generator/interpreter following the literal token order could misparse
this as an immediate CONTINUE"* is a fair reading, and the reason it is fair is that
**`after user.reply` is specified nowhere.** `grep -n "after user" architecture/specs/PDSL.md`
returns 0 mentions, while the corpus uses it twice. The vocabulary table at `:115` gives `WAIT` as
*"Stop for user input"* and says nothing about deferred continuation.

So the ambiguity is in the language, not in this line — and an unspecified idiom is exactly how two
reviewers reach opposite conclusions from the same source. Two candidate fixes, both outside this
PR since they change the spec or add a unit:

1. **Specify the idiom in `PDSL.md`** — `CONTINUE <unit> after user.reply` defers the transfer to
   the turn that resumes after the boundary, and must appear before the `WAIT` it defers past.
   One bullet next to `:186`, and both existing uses become unambiguous.
2. **Mirror the precedent's second half** — `coding-intent-companion.md` resumes into a dedicated
   `CodingIntentResume` unit carrying a `WHEN:` guard, rather than back into the originating unit.
   More explicit, at the cost of one unit per resuming option.

I would take (1): it fixes both call sites and the next one, rather than working around the gap
locally. Happy to do it here if you would rather it not wait, or raise it separately — your call.

SanjeevSolanki:
**Re: CodeRabbit nitpick — "Scan action sections only in the brief-package guard"** (`tests/test_pdsl_keywords.py`)

Accepted, and it was my own inconsistency: I applied `actions_only=True` to two of the three new
guards in the previous revision and missed this one.

Verified before fixing — a `NOTES:` line reading *"PlanWriteBriefPackage is the only unit that may
WRITE plan.toml or a brief- file, via RUN PlanWriteBriefPackage"* failed the guard three ways at
once: it registered as a write outside the writer unit, added a spurious entry to `writers`, and
would have been read as an option had it contained `->`. So documentation about the invariant
broke the test for the invariant.

Both call sites now pass `actions_only=True`, and the docstring says so. Re-mutated afterwards:
that prose line now passes, and the four real defects — a reworded write ahead of the gate, a
write in the pre-gate unit, an option losing its writer call, and the declared type flipped to
`confirmation` — are all still caught.

tests/test_pdsl_keywords.py
```diff
@@ -887,6 +891,325 @@ def test_the_untyped_menu_surface_does_not_grow() -> None:
...
+
+    Read through the validator's own block scanner so these guards agree with the
+    checker about what is inside a ```pdsl fence. With ``actions_only`` the result
+    is narrowed to ACTION_SECTIONS, so prose cannot create an obligation.
+    """
+    text, error = pdsl.read_source_file(path)
+    if error or text is None:
+        return []
+    blocks, _ = pdsl.scan_blocks(str(path), text)
+    rows: list[tuple[str, int, str]] = []
```
ainetx: ### Plan-first guard permits cross-branch assignment masking

**Severity:** Major

**Problem**
For each `CONTINUE PlanFirstGate`, the test builds `scope` from the continuation line plus all executable lines for that enclosing unit/menu. It does not associate `PLAN_FIRST_CONTINUE` or the plan-first `LOAD` with the continuation's own option or conditional clause.

<details>

**How to reproduce**
1. In one unit, retain one valid option that sets `PLAN_FIRST_CONTINUE`, loads plan-first, and continues it.
2. Add another option that only continues `PlanFirstGate`.
3. The second option inherits the first option's assignment and load through `section_text`, so the test passes.

**Expected behavior**
Each continuation path must contain, or be structurally dominated by, its own non-unset return-unit assignment and plan-first load.

**Actual behavior**
Any matching assignment and load elsewhere in the same unit/menu are accepted.

```text
valid branch: SET + LOAD + CONTINUE
invalid branch: CONTINUE
\ /
whole-unit scope
|
test passes
```

**Impact**
A newly unbound PlanFirstGate entry can silently route the no-plan path through stale or unset state.

**Suggested correction**
Parse individual option/conditional clauses and require the assignment and load in the same clause or in an explicitly modeled preceding path segment.

**How to verify**
Add a fixture containing one valid and one unbound continuation in the same unit; the test must fail for the unbound one.

</details>

skills/studio/modules/plan-compile.md
ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The precedent file confirms the pushback's key factual claim: coding-intent-companion.md:5-9 reads `EMIT...; CONTINUE CodingIntentResume after user.reply; WAIT user.reply; STOP_TURN` — the exact same shape as option 5's `EMIT; CONTINUE PlanPhase2Decompose after user.reply; WAIT user.reply; STOP_TURN`. That is a real, checkable precedent for the wait-then-resume idiom, not an invented one. The full PDSL.md spec text confirms `after user.reply` appears nowhere in the vocabulary table or execution semantics section, so the ambiguity the human concedes ('the language is unspecified, not this line') is verified rather than asserted. The finding's claim that sibling menus establish a fixed order this option breaks doesn't hold once a same-shape sibling (CodingIntentResume) is shown following the identical pattern — the cited siblings (plan-first.md, plan-compile.md option 6) are non-waiting terminal branches, a genuinely different shape, so they aren't the right comparison. The reordering matches established practice for this specific shape, so the finding should be dismissed.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a menu step as violating a fixed option ordering convention (e.g. CONTINUE before WAIT/STOP_TURN), check whether a same-shape sibling (a menu that also resumes after a user reply) exists elsewhere in the spec/repo and follows the same ordering — if so, treat non-waiting terminal-branch siblings as a different shape and not valid counter-evidence."

</details>

tests/test_pdsl_keywords.py
SanjeevSolanki: Accepted — Major is right, and this is a hole my own previous fix opened.

Reproduced first: adding `3 straight-to-plan -> CONTINUE PlanFirstGate` beside the two bound
options in `CodingBrainstormMenu` passed the guard, exactly as you describe. The unbound option
inherited its siblings' assignment through the unit-wide `section_text`.

**Why it was unit-wide.** The previous revision widened the scope from per-line to
per-UNIT/MENU to fix your earlier finding that a line-scoped guard would reject a legal
multi-line clause. That widening is what let a sibling vouch for an unbound branch — the fix for
one finding created this one.

**The model now distinguishes the two cases**, because they are genuinely different:

- `OPTIONS`, `INVALID` and `RULES` are **clause sections**: each option and each rule carries its
  own actions, so binding is per line. A sibling cannot vouch for an unbound branch.
- `DO` is **sequential**: an earlier `SET` in the same block legitimately binds a later
  `CONTINUE`, so scope is the preceding lines of that unit's block plus the line itself. This is
  what your earlier finding needed, and it is preserved — `workflows/auto-config.md` sets
  `PLAN_FIRST_CONTINUE` at `:29` and enters via `COMPANION_CONTINUE` at `:31`, still legal.

Mutation-verified in both directions:

| mutation | result |
|---|---|
| your reproduction — unbound sibling **option** | fails, as it should |
| unbound sibling **rule** line in the same `RULES:` block | fails |
| `auto-config` untouched (sequential `DO` binding) | passes |
| `auto-config` with its `DO`-block `SET` deleted | fails |

Also re-ran the full battery on the amended guards — 11 mutations, the 9 defects all still
caught and the 2 legitimate changes (an option label rename, a `NOTES:` prose mention) still
passing.

Gates: `cfs validate` 0/0 237/237 · `make test` 5631 passed · spec-coverage 90.8% / 0.4609
unchanged · pylint, ruff, vulture-ci clean.

Separately, thank you for the two RESOLVED confirmations and for the accepted pushback on the
revise ordering — the `after user.reply` spec gap it exposed is now #167.

ainetx: **RESOLVED** -- Confirmed fixed as of fe1537d33c8d8990836ee18b18a8bde10864f877 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The test's `_binding_scope` helper does not actually pool the whole section: for CLAUSE_SECTIONS (OPTIONS, INVALID, RULES) — which is exactly where every PlanFirstGate auto-skip continuation lives in this diff (a single RULES line combining the SET, LOAD and CONTINUE) — scope is just `line` itself, i.e. that one clause in isolation, so a sibling option/rule elsewhere in the same UNIT cannot mask an unbound one. Only for non-clause sections (DO) does it accumulate text, and even then only the *preceding* lines within the same UNIT+section (sequential execution order), not the whole section indiscriminately — a later, unrelated assignment after the continuation line can't leak in either. This directly matches the test's own stated design rationale ('binding is per clause, not per unit... a valid sibling option must not vouch for an unbound one beside it') and each of the three new prep-gate auto-skip rules is a single self-contained RULES-clause line, so each is checked independently, not against section-wide state.
- **Reviewer:** The test treats RULES, OPTIONS, and INVALID as per-line clause scopes: `_binding_scope` returns only the continuation's own line for those sections, so an unrelated `PLAN_FIRST_CONTINUE` assignment elsewhere in the enclosing UNIT/MENU cannot satisfy it; DO remains intentionally sequential.
- **Verifier:** The test's _binding_scope explicitly treats OPTIONS/INVALID/RULES as CLAUSE_SECTIONS, binding per single line rather than per whole section (only DO, a sequential block, uses the wider section-prefix scope, which matches PDSL's actual sequential-execution semantics). All three target auto-skip entries in coding-prep-gates.md, write-docs-prep-gates.md, and write-skills-prep-gates.md live in RULES clauses where the SET PLAN_FIRST_CONTINUE=..., the LOAD of gates/plan-first.md, and the CONTINUE PlanFirstGate are all written on the same single line, so the scope checked for that entry is that line alone -- an unrelated PLAN_FIRST_CONTINUE assignment elsewhere in the same RULES section (a sibling rule) cannot leak in and falsely satisfy this entry's requirement, which is exactly the sibling-vouching failure the test's own docstring says it was designed to prevent.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -650,9 +650,11 @@ def test_prompt_runtime_references_use_cf_studio_path() -> None:
 
 
 # Every MENU that does not yet declare a gate risk TYPE, frozen 2026-09-07.
-# An undeclared gate is treated as `blocking` at runtime, so the existing
-# surface is grandfathered and migrates gate by gate -- but the untyped surface
-# must not grow. Anything not in this set has to declare a TYPE.
+# `blocking` is the *intended* contract for an undeclared gate, not what happens
+# today: three shipped paths still resolve one by runtime judgement, so this set
+# freezes the existing untyped inventory rather than recording a fail-closed
+# default. The surface migrates gate by gate and must not grow -- anything not in
```
ainetx, line 658: ### Unverified 'three shipped paths' claim in comments has no backing test

**Severity:** Minor

**Problem**
Two comments/docstrings in test_pdsl_keywords.py assert a specific factual claim about runtime behavior (three shipped paths resolve undeclared menu types by judgement, not a fail-closed default) but no test identifies or pins these three paths.

<details>

**How to reproduce**
1. Read UNTYPED_MENU_BASELINE comment block and test_the_untyped_menu_surface_does_not_grow docstring. 2. Search the file for any assertion referencing 'three' or enumerating specific menu keys tied to this claim. 3. None found.

**Expected behavior**
A test or explicit named list identifying the three specific menus/paths and asserting they remain in UNTYPED_MENU_BASELINE (or otherwise verifying the claim), so the claim can't silently drift as the baseline set changes.

**Actual behavior**
The claim exists only as prose; the baseline set could shrink or be edited without anyone noticing whether these three specific paths are still present or still behave as described.

```text
comment claims 'three paths behave X' -> baseline set edited/shrunk -> no test checks X for those paths -> claim silently becomes stale/false
```

**Impact**
Future contributors may rely on a comment that has drifted from reality, since nothing enforces or re-validates it.

**Suggested correction**
Add a small test enumerating the three specific menu/path keys and asserting they exist in UNTYPED_MENU_BASELINE (or an equivalent named marker), or convert the comment into a testable list.

**How to verify**
Confirm a new test fails if any of the three named paths is removed from UNTYPED_MENU_BASELINE or gains a TYPE declaration without updating the tracking.

</details>

SanjeevSolanki: Accepted — the comment did assert a fact with nothing holding it. Pinned now, though not quite
the way you suggested, and my first attempt only satisfied half your criteria.

**On the suggested fix.** Asserting the three paths remain in `UNTYPED_MENU_BASELINE` would not
work: those three are *modules that auto-resolve a gate*, not menus in the untyped set. Two
different populations — the baseline holds menu keys, these are the code paths that resolve them.
So the claim is pinned directly instead.

**What the guard does.** `RUNTIME_JUDGEMENT_PATHS` names the three, each with the rule it carries,
and asserts the count is three so the set cannot be quietly emptied. Sourced from
`architecture/specs/PDSL.md`, which cites the same three.

**Where my first version fell short of your own verification.** You asked it to fail if a path is
removed **or** if it "gains a TYPE declaration". I covered removal and tested it. I had not
covered the second, so I tried it: adding `ALWAYS resolve the gate from its declared TYPE when
present` to `simple-mode-rules.md`, keeping its judgement rule, **passed**.

That was the more important half. The comment claims two things — three paths auto-resolve, *and*
none of them reads a declaration — and the second is what changes first, because binding those
paths to declared types is precisely what the default flip will do. Both halves are asserted now.

Six mutations, all behaving:

| mutation | result |
|---|---|
| each of the three loses its auto-resolution rule (×3) | fails |
| `simple-mode-rules.md` gains a declared-TYPE read | fails |
| a menu in `subagents/dispatch.md` gains `TYPE: blocking` | fails |
| untouched corpus | passes — the pattern matches the reading vocabulary, not the bare tokens, since `SUB_AGENT_GROUP_DECISION` would otherwise match "decision" |

Stated limit, in the docstring rather than left implied: a path that reads a declaration in
wording the pattern misses still gets through. It is a proxy, not a proof.

**And your comment found something else.** Re-reading the spec passage it sent me to,
`PDSL.md` said *"**Two** shipped paths do auto-resolve gates, and **neither** reads a
declaration"* — and then cited three. An earlier pass of mine had corrected the *second*
miscount in that same bullet and left the first. Both now say three, in #167.

ainetx: **RESOLVED** -- Confirmed fixed as of 12f9bf26e0776af22c456c16d7403d8d85fce9c6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff adds test_the_runtime_judgement_paths_named_in_the_baseline_comment_still_exist(), which names the exact three paths (RUNTIME_JUDGEMENT_PATHS), asserts len==3, confirms each file still exists and carries its auto-resolution marker, and fails if any of them now reads a declared TYPE — so the claim is pinned by an executable assertion, not left as unchecked comment prose (it doesn't separately assert membership in UNTYPED_MENU_BASELINE, but it does verify the specific behavioral claim the comment makes).
- **Reviewer:** `RUNTIME_JUDGEMENT_PATHS` explicitly names all three paths, and `test_the_runtime_judgement_paths_named_in_the_baseline_comment_still_exist` asserts the count is three, each path and its runtime-judgement marker still exist, and none matches declared-TYPE reading vocabulary; the untyped baseline remains separately frozen by `test_the_untyped_menu_surface_does_not_grow`.
- **Verifier:** The diff adds test_the_runtime_judgement_paths_named_in_the_baseline_comment_still_exist, which explicitly enumerates the three named paths (RUNTIME_JUDGEMENT_PATHS: simple-mode-rules.md, brave-new-world.md, subagents/dispatch.md), asserts the dict length equals 3 to keep the comment's '3' in sync, verifies each file still exists and still carries its cited marker text, and verifies none of them now reads a declared TYPE (via DECLARED_TYPE_READ_RE) — i.e. it substantively pins both halves of the comment's factual claim (that judgement-based resolution still occurs and is still undeclared), not merely unchecked prose. It does not additionally assert these three paths' associated menus are literal entries in UNTYPED_MENU_BASELINE, but the claim being verified is about runtime-judgement behavior in those files, not about baseline-set membership, so this is not a gap in what the comment actually asserts.

</details>

skills/studio/modules/plan-compile.md
```diff
@@ -2,27 +2,46 @@
...
+RULES:
+  ALWAYS write plan.toml before any brief or phase file
+  ALWAYS reopen CF_PHASE_GATE released_for_orchestrator_write (scoped) before each write and reset to armed immediately after
+  ALWAYS emit the brief package summary after writing, so the package is reported even though the mode was chosen before the write
+  NEVER emit "Plan created" in this summary — no phase file exists yet
+  NEVER run without a production mode chosen at PlanProduceChoice
 ```
```
ainetx, line 49: ### PlanWriteBriefPackage has no defined behavior for a partial plan.toml+briefs write

**Severity:** Minor

**Problem**
PlanWriteBriefPackage's RULES cover ordering (plan.toml before briefs), gate reopen/reset, and the post-write summary, but say nothing about what happens if plan.toml writes successfully and a brief write then fails or is interrupted partway through the loop over phases.

<details>

**How to reproduce**
1. User picks any of PlanProduceChoice options 1-4, triggering RUN PlanWriteBriefPackage. 2. plan.toml write succeeds, CF_PHASE_GATE resets to armed. 3. Writing brief-03-*.md fails (disk error, interrupted session, etc.) partway through the per-phase loop. 4. Spec has no defined recovery/cleanup/stop state for this case.

**Expected behavior**
The unit should specify that a partial write is not treated as an authorised complete package — e.g. it should define a stopped/failed state, prevent the summary EMIT and downstream consumers (PlanPhase3Validate, PlanPhaseCompilerDispatch, NextActionsOffer) from proceeding, and describe safe retry or cleanup.

**Actual behavior**
No failure path is described; only the all-success case (emit brief package summary) is specified, so a resumed or retried run could silently treat a partially-written package as complete.

```text
PlanProduceChoice -> RUN PlanWriteBriefPackage -> WRITE plan.toml (ok) -> WRITE brief-NN (fails) -> [undefined] -> (possible) EMIT summary / downstream consumer runs against incomplete briefs
```

**Impact**
A downstream consumer (inline compile, subagent dispatch, or a later resumed session) could operate against an incomplete brief set without any signal that the package was never fully authorised/written.

**Suggested correction**
Add a RULES entry to PlanWriteBriefPackage defining that any failure during the per-phase brief loop halts before the summary EMIT, leaves CF_PHASE_GATE in a non-released state, and requires an explicit re-run of PlanWriteBriefPackage (or a defined cleanup) rather than resuming downstream.

**How to verify**
Re-read plan-compile.md and confirm a RULES line addresses partial-write recovery for PlanWriteBriefPackage, then confirm no downstream consumer clause can be reached without a completed summary EMIT.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -848,13 +850,93 @@ def _menu_type_declarations() -> dict[str, str | None]:
...
+    So the count, the paths and both halves of the claim are asserted: each path
+    still carries its auto-resolution rule, and none of them reads a declared
+    type. The second half matters more, because binding these paths to declared
+    types is exactly what the default flip does -- so that is where the comment
+    will go stale first. An earlier version of this guard checked only the first
+    half and passed when a path gained a declared-type read.
+
+    What still gets past it: a path that reads a declaration in wording
+    `DECLARED_TYPE_READ_RE` does not match. The fix when this fails is to correct
+    the comment in the same change, not to edit these constants to match.
```
ainetx, line 897: ### Runtime-judgement guard can crash on an unreadable prompt file

**Severity:** Minor

**Problem**
The new test directly calls path.read_text(encoding='utf-8') for each configured path without catching OSError or UnicodeDecodeError. An unreadable or malformed path produces an unhandled test crash instead of the clear per-item assertion result required by this check.

<details>

**How to reproduce**
1. Make one configured RUNTIME_JUDGEMENT_PATHS file unreadable or non-UTF-8. 2. Run the PDSL keyword tests. 3. Reach test_the_runtime_judgement_paths_named_in_the_baseline_comment_still_exist.

**Expected behavior**
The test reports the affected path and read failure as a collected per-item failure.

**Actual behavior**
path.read_text raises before the findings list is asserted.

```text
configured path
-> read_text()
-> OSError / UnicodeDecodeError
-> unhandled test error
```

**Impact**
Diagnostics become less actionable and a single inaccessible file aborts the guard.

**Suggested correction**
Wrap read_text in try/except for OSError and UnicodeDecodeError, append a path-specific finding, then continue checking remaining paths.

**How to verify**
Add fixtures or mocks that raise OSError and UnicodeDecodeError and assert the final failure includes the path-specific message.

</details>

tests/test_pdsl_keywords.py
ainetx: Still an issue. Your menu-derived invariant does fix the specific PLAN_FIRST_CONTINUE reproduction: test_prep_gate_modules_apply_the_same_auto_skip_rules now extracts menu_targets from each module’s own -> options and rejects an auto-skip target not in that set. Thus replacing WriteDocsAuthorDispatch with CodingDispatch in only the write-docs auto-skip rule now fails, as you described.

However, the test still does not verify the accompanying module-specific author-dispatch LOAD. The prep-gate test checks continuation, note, placement, and the target value; the separate PlanFirstGate guard only requires gates/plan-first.md. It never asserts write-docs-author-dispatch.md (or the write-skills equivalent) is loaded in the same auto-skip clause. A copied or omitted author-dispatch LOAD paired with the correct WriteDocsAuthorDispatch would still pass, even though PlanFirstGate would later continue to a unit that was not loaded on this path.

The original finding should be amended: its claim that the target assignment itself was unchecked is no longer accurate after fe1537d3; the remaining defect is specifically the missing verification of the module-specific author-dispatch LOAD.

ainetx (approved):
Solid change — collapsing the decomposition-confirm and brief-checkpoint menus into a single `decision` gate simplifies the plan flow while keeping the write-before-authorization invariant intact, and the new guard tests do a good job pinning down the auto-skip consistency, the write-ordering contract, and the runtime-judgement claims in the untyped-menu baseline. Nothing here blocks merging. A few small notes worth a glance next time you're in this area:

- **Prep-gate consistency test doesn't check the per-module `PLAN_FIRST_CONTINUE` target** — it matches rules on the fixed string `CONTINUE PlanFirstGate` and only verifies the shared condition/note text, so the module-specific `SET PLAN_FIRST_CONTINUE = <Unit>` value and `LOAD` path this diff introduced aren't actually compared. ([comment](https://github.com/constructorfabric/studio/pull/163#discussion_r3964766167))
- **`PlanWriteBriefPackage` has no defined behavior for a partial write** — the rules cover ordering, gate reopen/reset, and the summary, but say nothing about what happens if `plan.toml` writes and a brief write then fails or gets interrupted mid-loop. ([comment](https://github.com/constructorfabric/studio/pull/163#discussion_r3966284143))
- **Runtime-judgement guard test can crash instead of failing cleanly** — it calls `path.read_text(encoding='utf-8')` per configured path without catching `OSError`/`UnicodeDecodeError`, so an unreadable or malformed file produces a test crash rather than a clear per-item assertion. ([comment](https://github.com/constructorfabric/studio/pull/163#discussion_r3966486750))

skills/studio/modules/plan-compile.md
SanjeevSolanki: Accepted — a write path with only a success case specified is the fail-loud gap, and this one
is mine: I created the unit, so the boundary that makes the question askable is new even though
the writes moved across unchanged.

**Your suggested shape was not available.** A sixth RULES entry trips `PDSL601` — the unit was
already at the 5-rule cap. Two attempts to add one failed the checker before I folded the
concern into the rule it actually belongs to, with the mechanics in an `ON_ERROR:` clause, which
is the house idiom for a failure path (18 uses in the corpus). Worth knowing when a finding asks
for "a RULES entry": the cap can forbid it.

The summary rule now reads that it is emitted **only after a complete write** — it is the only
completeness signal a consumer has, so a failed or interrupted write emits none, and recovery is
a fresh `PlanProduceChoice` choice which rewrites `plan.toml` before any brief. `ON_ERROR:`
reports the failure verbatim, resets `CF_PHASE_GATE` to `armed`, and stops the turn.

**The second half of your verification is now pinned too, and I had missed it.** You asked to
confirm a RULES line exists *and* that no downstream consumer clause can be reached without a
completed summary. Nothing asserted the second, and structurally nothing can — a consumer runs
in the same option clause as the writer. So the two things that actually protect it are asserted
instead:

| assertion | mutation | result |
|---|---|---|
| the summary is the writer's **last** `DO` action | add a write after it | fails |
| an `ON_ERROR` clause exists | delete it | fails |
| that clause terminates | change `STOP_TURN` to `CONTINUE PlanPhase3Validate` | fails |

The first is the one that matters: a write placed after the summary would be reported as already
done.

`cfs validate` 0/0 237/237 · `make test` 5634 passed · spec-coverage 90.8% / 0.4609 unchanged ·
pylint, ruff, vulture-ci, test-coverage clean.

tests/test_pdsl_keywords.py
SanjeevSolanki: Accepted and fixed — and your "how to verify" line caught something my fix had not done.

The guard now reads through `pdsl.read_source_file`, the validator's own reader, which normalizes
exactly the two exceptions you name into a reported error. An unreadable path becomes a finding
about *that path* and the other two still get checked.

**Where my first pass fell short.** You asked for fixtures or mocks raising `OSError` and
`UnicodeDecodeError`, asserting the failure includes the path-specific message. I verified by
hand instead — `chmod 000` on one path, then a non-UTF-8 file — watched both report correctly,
and **committed no test**. So nothing stopped it regressing, which is the same shape as calling
something covered because I saw it work once.

There is now a parametrized test over both shapes the reader normalizes, asserting the report
names every configured path and carries the reader's own reason. Two cases, both passing, and
they fail if the direct `read_text` call comes back.

skills/studio/modules/plan-compile.md
ainetx: **RESOLVED** -- Confirmed fixed as of d46f60509f5107286a7694eb8ec29f6deb740726 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The RULES block explicitly states the summary is emitted only after a complete write ('a failed or interrupted write emits none') and the ON_ERROR write_failed handler emits an explicit error, resets CF_PHASE_GATE to armed, and STOP_TURNs without emitting the summary; recovery is defined as a fresh PlanProduceChoice pass which always rewrites plan.toml before any brief (per 'ALWAYS write plan.toml before any brief'), so a retry cannot silently reuse a partial package — this is a documented, textual failure/recovery policy, not a gap.
- **Reviewer:** PlanWriteBriefPackage emits its success summary only after all writes, and its ON_ERROR path declares the package incomplete, resets the gate, stops the turn, and requires a fresh PlanProduceChoice retry that rewrites plan.toml before briefs.
- **Verifier:** PlanWriteBriefPackage's RULES and ON_ERROR clauses (plan-compile.md ~lines 60-77) explicitly cover the plan.toml-succeeds/brief-fails case: the summary is only emitted after a complete write ('a failed or interrupted write emits none'), ON_ERROR write_failed emits an incomplete-package message plus the raw error, resets CF_PHASE_GATE to armed, and STOP_TURNs so no consumer is reached; recovery is explicitly defined as 'a fresh PlanProduceChoice choice that rewrites plan.toml before any brief' (consistent with the 'ALWAYS write plan.toml before any brief' rule), so a retry cannot silently treat the partial package as complete.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of d46f60509f5107286a7694eb8ec29f6deb740726 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The new file-read helpers (_sectioned_lines, the runtime-judgement guard) call pdsl.read_source_file(path), which returns a (text, error) tuple normalizing OSError/UnicodeDecodeError into a PdslError rather than raising, and the diff adds a dedicated parametrized test (test_the_runtime_judgement_guard_reports_an_unreadable_path) that mocks READ_ERROR and DECODE_ERROR cases and asserts each unreadable path is reported by name in the assertion output rather than crashing the test run.
- **Reviewer:** New guard file reads use pdsl.read_source_file, explicitly convert read/decode failures into per-path findings, and new parametrized tests cover both permission and UTF-8 decode errors.
- **Verifier:** The only new file reads go through the existing pdsl.read_source_file() wrapper (used in _sectioned_lines and the runtime-judgement test), which already normalizes OSError/UnicodeDecodeError into a returned (text, error) tuple instead of raising; the diff even adds test_the_runtime_judgement_guard_reports_an_unreadable_path, which mocks read_source_file to return READ_ERROR/DECODE_ERROR and asserts a clear per-path finding is produced rather than a crash. No raw.read_text()/.read_bytes() calls are introduced in this diff.

</details>

tests/test_pdsl_keywords.py
ainetx: **RESOLVED** -- Confirmed fixed as of d46f60509f5107286a7694eb8ec29f6deb740726 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The test does check the PLAN_FIRST_CONTINUE target value, not just substring presence: it derives menu_targets from each module's own OPTIONS actions (regex over PLAN_FIRST_CONTINUE assignments with '->') and then asserts each auto-skip rule's PLAN_FIRST_CONTINUE value is a member of that per-file menu_targets set, so a copy-pasted wrong dispatch unit (e.g. CodingDispatch appearing in write-docs-prep-gates.md) would produce a mismatch and fail the test.
- **Verifier:** The test derives `menu_targets` from each module's own MENU OPTIONS lines (not a hardcoded table) and then asserts the auto-skip RULES' PLAN_FIRST_CONTINUE value is a member of that per-module set; a copy-pasted wrong target (e.g. CodingDispatch pasted into write-docs-prep-gates.md) would not be in write-docs' own menu_targets and would be flagged, so the scenario the check worries about is in fact caught.

</details>

SanjeevSolanki:
## QA handoff

Per the handover standard, one Run Card per command QA types, plus a declaration for the part
no command reaches. ID-free copy; the task-id line lives on the internal QA task.

```
RUN CARD — cf-plan (phased planning)
Type:           Feature
What it does:   Planning now asks once how to produce phase files, instead of asking
                twice — first to confirm the decomposition, then how to build it.
                Picking a production mode is also the permission to write the files.
Run it:         git clone https://github.com/constructorfabric/studio && cd studio
                make ensure-bootstrap
                # then, in Claude Code from that directory, invoke the planning
                # workflow by name: cf-plan
                # (headless needs --permission-mode bypassPermissions, or Studio
                #  is skipped silently)
Input to use:   a task big enough to decompose into 3+ phases — e.g. "add a new
                cfs subcommand with tests, docs and CI wiring". A one-step request
                will not reach the decomposition gate at all.
Prove it works: AC1 → run cf-plan on that task → expect ONE stop after the phase
                      breakdown, offering inline / prompts / subagents / briefs-only /
                      revise / no. Expect NO separate "Proceed with this
                      decomposition?" stop before it.
                AC2 → at that stop choose 5 (revise), describe a change → expect the
                      breakdown reworked and re-shown, not the same menu repeated.
                AC3 → at that stop choose 6 (no) → expect "Decomposition declined —
                      nothing written", and `.plans/` contains no plan.toml.
                AC4 → choose 4 (briefs-only) → expect plan.toml + one brief per phase
                      written, a summary naming "0/N phase files", and no phase files.
Force a bad result: make the target unwritable — `chmod 500 .cf-studio/.plans` (create
                it first if absent) — then choose 1 (inline). Expect the incomplete-package
                error naming the write failure, and NO package summary. A summary
                appearing after a failed write is the bug this guards.
Off switch:     n/a — structural change to the gate, not a flag.
My numbers:     n/a (Feature, not a checker).
Not in this card: the coding/docs prep-gate auto-skip half of this PR. See the
                declaration below.
```

**Declared, no card — the prep-gate half.** The same PR aligns the auto-skip rules across
`coding-prep-gates.md`, `write-docs-prep-gates.md` and `write-skills-prep-gates.md`, and fixes
a state-precondition defect in the two that had them. **None of those three modules is
reachable from any shipped workflow** — they load only from three `*-intent-routing` /
`coding-intent-companion` modules that nothing references — so there is no command that makes
the change observable, and the defect it fixes could not have affected a user. Covered by five
new corpus guards plus maintainer review. It becomes visible only if that routing is
reconnected, which is raised separately as an issue on this repo.

**Author-measured, for check 1:** `cfs validate` 0 errors / 0 warnings, 237/237 markers ·
`make test` 5634 passed · `spec-coverage` 90.8% coverage, 0.4609 granularity, both unchanged by
this change · pylint, ruff, vulture-ci, test-coverage clean. Untyped MENU surface 113 → 111.

Author does not verify their own work — over to QA.

PR #165, opened 2026-09-09
refactor(traceability): extract the shared cpt-id def↔ref scan into a util

SanjeevSolanki:
## What

Four query commands — `where-used`, `where-defined`, `list-ids`, `list-id-kinds` — each hand-rolled the same loop over `scan_cpt_ids`. This lifts that loop into a shared `utils/cpt_reference_scan.py` (`scan_records` + the target-filtered `references` / `definitions`, plus a `graph_for` def↔ref view) and routes all four through it.

**Behaviour is identical** — each command's output and its existing tests are unchanged.

## Why

`graph_for` is the def↔ref view a later gap / traceability / contradiction analysis will consume. It is unused today (whitelisted with a removal trigger), so this is a **genuine dedup of the four commands plus prep for that later work** — it does not dedupe or touch the semantic-quality feature, which has no traversal yet.

## Notes

- New `@cpt-algo:cpt-studio-algo-cpt-reference-scan` declared in `traceability-validation.md` (its current consumers, the query commands, live there); `cfs validate` resolves **0 errors**.
- Full suite green; the util is unit + property tested at 100% line coverage.
- Pre-commit adversarial review was scoped to three finders (behaviour-parity, util edge-cases, CPT/doc) for a behaviour-identical refactor, and converged to doc/test polish.

## Summary by CodeRabbit

* **New Features**
  * Added CPT reference scanning for artifact IDs, including definitions, references, and relationship graph views.
  * Improved ID listing and kind discovery with more consistent scan results.
  * Enhanced “where defined” and “where used” queries with shared reference and definition data.
  * Preserved source locations, checked status, priorities, and optional definition inclusion.

* **Documentation**
  * Added the CPT Reference Scan process specification and updated related implementation references.

### DIFF
--- a/architecture/features/traceability-validation.md
+++ b/architecture/features/traceability-validation.md
@@ -13,6 +13,7 @@
   - [Query Traceability](#query-traceability)
 - [3. Processes / Business Logic (CDSL)](#3-processes--business-logic-cdsl)
   - [Scan Artifact IDs](#scan-artifact-ids)
+  - [CPT Reference Scan](#cpt-reference-scan)
   - [Scan CDSL Instructions](#scan-cdsl-instructions)
   - [Validate Artifact Structure](#validate-artifact-structure)
   - [Cross-Validate Artifacts](#cross-validate-artifacts)
@@ -227,6 +228,25 @@ Catches structural and traceability issues that AI agents miss or hallucinate �
 - [x] - `p1` - Content scoped extraction: hash-fence blocks, heading scopes, ID-definition scopes - `inst-scan-ids-get-content`
 - [x] - `p1` - File I/O utilities: safe text reader, text file iterator, relative path converter - `inst-scan-ids-file-utils`
 - [x] - `p1` - Wrapper function for `parse_cpt` identifier parser - `inst-parse-cpt-fn`
+
+### CPT Reference Scan
+
+- [x] `p1` - **ID**: `cpt-studio-algo-cpt-reference-scan`
+
+**Input**: Artifacts to scan `[(path, artifact_type)]`, a target `cpt-id`, and a path→source map
+
+**Output**: Reference records `{artifact, artifact_type, line, kind, type, checked, source?}` (definition records omit `type`; `kind` is always `null` here — callers infer and fill it), or a def↔ref graph `{defined_in, referenced_in}`
+
+**Scope**: `cpt-*` identifiers only — `PRD-NNN` requirement references are out of scope (a documented v1 limit), the same scope as `where-used` / `where-defined`. The def↔ref graph is assembled from two independent scans (one per side), so `defined_in` and `referenced_in` carry no cross-list atomicity guarantee; a caller that needs many ids should build a single scan pass and partition it locally rather than call the graph per id.
+
+**Steps**:
+1. [x] - `p1` - Iterate every `scan_cpt_ids` hit across the artifacts, yielding it with its artifact context — the shared scan the query commands reuse - `inst-scan-records`
+2. [x] - `p1` - Project the scan into references to a target id, optionally including its definitions - `inst-references`
+3. [x] - `p1` - Project the scan into a target id's definitions only - `inst-definitions`
+4. [x] - `p1` - Assemble a cpt-id's def↔ref graph view (`defined_in` / `referenced_in`) for the artifact-quality detectors - `inst-graph-for`
+
+**Supporting**:
+- [x] - `p1` - The shared locus record both projections build — reference records carry `type`, definitions omit it - `inst-record`
 
 ### Scan CDSL Instructions
 
@@ -831,6 +851,7 @@ The system **MUST** provide CLI commands for navigating the ID graph: `list-ids 
 **Implements**:
 - `cpt-studio-flow-traceability-validation-query`
 - `cpt-studio-algo-traceability-validation-scan-ids`
+- `cpt-studio-algo-cpt-reference-scan`
 
 **Covers (PRD)**:
 - `cpt-studio-fr-core-traceability`
@@ -867,6 +888,7 @@ The system **MUST** scan CDSL instruction markers (`inst-{slug}` suffixes in num
 | Where Defined | `skills/.../commands/where_defined.py` | Find where an ID is defined |
 | Where Used | `skills/.../commands/where_used.py` | Find all references to an ID |
 | Document Utils | `skills/.../utils/document.py` | ID scanning, CDSL instruction scanning |
+| CPT Reference Scan | `skills/.../utils/cpt_reference_scan.py` | Shared cpt-id def↔ref scan for the query commands; def↔ref graph view for the artifact-quality detectors |
 | Constraints Utils | `skills/.../utils/constraints.py` | Constraint loading, heading validation, cross-validation |
 | Codebase Utils | `skills/.../utils/codebase.py` | Code file scanning, `@cpt-*` marker validation |
 | Error Codes | `skills/.../utils/error_codes.py` | Stable error codes for validation issues |
--- a/skills/studio/scripts/studio/utils/cpt_reference_scan.py
+++ b/skills/studio/scripts/studio/utils/cpt_reference_scan.py
@@ -1,0 +1,145 @@
+"""Shared cpt-id reference scan.
+
+The one ``scan_cpt_ids`` traversal that ``where_used``, ``where_defined``,
+``list_ids`` and ``list_id_kinds`` each re-implemented, plus the def↔ref view the
+artifact-quality gap / traceability / contradiction detectors build on. cpt-id
+scoped — prose / ``PRD-NNN`` requirements are out of scope (a documented v1 limit).
+
+Read-only, stdlib-only. Behaviour is identical to the loops it replaces; the
+callers keep their own projections.
+
+@cpt-algo:cpt-studio-algo-cpt-reference-scan:p1
+"""
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Dict, Iterator, List, Optional, Tuple
+
+from .document import scan_cpt_ids
+
+
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-scan-records
+def scan_records(
+    artifacts_to_scan: List[Tuple[Path, str]],
+) -> Iterator[Tuple[Path, str, Dict[str, object]]]:
+    """Yield ``(artifact_path, artifact_type, hit)`` for every ``scan_cpt_ids``
+    hit across *artifacts_to_scan* — the doubled loop the callers shared.
+
+    The raw hit is passed through untouched so each caller keeps its exact
+    projection (id-strip, line coercion, kind inference, def/ref filtering).
+    """
+    for artifact_path, artifact_type in artifacts_to_scan:
+        for hit in scan_cpt_ids(artifact_path):
+            yield artifact_path, artifact_type, hit
+# @cpt-end:cpt-studio-algo-cpt-reference-scan:p1:inst-scan-records
+
+
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-record
+def _record(
+    artifact_path: Path,
+    artifact_type: str,
+    hit: Dict[str, object],
+    path_to_source: Dict[str, str],
+    *,
+    include_type: bool,
+) -> Dict[str, object]:
+    """One locus record, shared by the reference and definition projections.
+
+    ``include_type`` adds the ``type`` key (references carry it; definitions omit
+    it) in its historical position — between ``kind`` and ``checked`` — so the
+    emitted dict is byte-for-byte what the commands produced before this util.
+    """
+    rec: Dict[str, object] = {
+        "artifact": str(artifact_path),
+        "artifact_type": artifact_type,
+        "line": int(hit.get("line", 1) or 1),
+        "kind": None,
+    }
+    if include_type:
+        rec["type"] = str(hit.get("type"))
+    rec["checked"] = bool(hit.get("checked", False))
+    src = path_to_source.get(str(artifact_path))
+    if src:
+        rec["source"] = src
+    return rec
+# @cpt-end:cpt-studio-algo-cpt-reference-scan:p1:inst-record
+
+
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-references
+def references(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Dict[str, str],
+    *,
+    include_definitions: bool,
+) -> List[Dict[str, object]]:
+    """References to *target_id* (optionally including its definitions).
+
+    The ``where_used`` projection, verbatim: a reference hit whose id matches,
+    with its type; definition hits included only when *include_definitions*.
+    """
+    out: List[Dict[str, object]] = []
+    for artifact_path, artifact_type, h in scan_records(artifacts_to_scan):
+        if str(h.get("id") or "") != target_id:
+            continue
+        if h.get("type") == "definition" and not include_definitions:
+            continue
+        out.append(_record(artifact_path, artifact_type, h, path_to_source, include_type=True))
+    return out
+# @cpt-end:cpt-studio-algo-cpt-reference-scan:p1:inst-references
+
+
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-definitions
+def definitions(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Dict[str, str],
+) -> List[Dict[str, object]]:
+    """Definitions of *target_id* — the ``where_defined`` projection, verbatim
+    (no ``type`` key; ``kind`` left ``None`` for the caller to annotate)."""
+    out: List[Dict[str, object]] = []
+    for artifact_path, artifact_type, h in scan_records(artifacts_to_scan):
+        if h.get("type") != "definition":
+            continue
+        if str(h.get("id") or "") != target_id:
+            continue
+        out.append(_record(artifact_path, artifact_type, h, path_to_source, include_type=False))
+    return out
+# @cpt-end:cpt-studio-algo-cpt-reference-scan:p1:inst-definitions
+
+
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-graph-for
+def graph_for(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Optional[Dict[str, str]] = None,
+) -> Dict[str, List[Dict[str, object]]]:
+    """The def↔ref view a detector wants for one cpt-id: where it is defined and
+    where it is referenced.
+
+    ``defined_in`` is the definition loci (no ``type`` key); ``referenced_in`` is the
+    non-definition reference loci (each with its ``type``). Built with the same
+    ``references`` / ``definitions`` logic the query commands use, so it classifies
+    definitions and references identically to them.
+
+    Two-pass by construction: ``defined_in`` and ``referenced_in`` come from two
+    independent ``scan_records`` passes, so they are computed separately — do not rely
+    on cross-list atomicity between them. That is fine for one id while the view is
+    unused, but a caller that needs many ids must NOT call ``graph_for`` per id in a
+    loop — that is ``O(n_ids × n_artifacts)`` rescans of the whole tree. Build a single
+    ``scan_records`` pass and partition it locally instead.
+
+    ``path_to_source`` is optional here — a detector usually wants only the def/ref
+    loci, not source text — whereas ``references`` / ``definitions`` require it, since
+    every command call site has the map. Requiring it there is deliberate: a missing
+    map at those call sites is a bug we want surfaced, not silently defaulted. Passing
+    ``None`` is equivalent to an empty map (no ``source`` keys emitted).
+    """
+    src = path_to_source or {}
+    return {
+        "defined_in": definitions(target_id, artifacts_to_scan, src),
+        "referenced_in": references(
+            target_id, artifacts_to_scan, src, include_definitions=False
+        ),
+    }
+# @cpt-end:cpt-studio-algo-cpt-reference-scan:p1:inst-graph-for
--- a/tests/test_cpt_reference_scan.py
+++ b/tests/test_cpt_reference_scan.py
@@ -1,0 +1,200 @@
+"""Tests for utils/cpt_reference_scan — the shared cpt-id def↔ref scan.
+
+The four commands' golden tests already prove behaviour-identity; these pin the
+util's own contract: the projections, the def↔ref graph view, and that it adds no
+new error-swallowing over scan_cpt_ids.
+"""
+import random
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from studio.utils import cpt_reference_scan as scan
+
+
+def _arts(*names):
+    return [(Path(n), "REQUIREMENTS") for n in names]
+
+
+def _hits(*specs):  # spec: (id, type, line, checked)
+    return [{"id": i, "type": t, "line": ln, "checked": c} for (i, t, ln, c) in specs]
+
+
+# ---------------------------------------------------------------- scan_records
+def test_scan_records_yields_every_hit_with_artifact_context():
+    per = {Path("a.md"): _hits(("cpt-x-req-1", "reference", 3, False)),
+           Path("b.md"): _hits(("cpt-x-req-1", "definition", 1, False),
+                                ("cpt-x-req-2", "reference", 5, False))}
+    with patch.object(scan, "scan_cpt_ids", side_effect=lambda p: per[p]):
+        recs = list(scan.scan_records(_arts("a.md", "b.md")))
+    assert [(str(p), t, h["id"]) for p, t, h in recs] == [
+        ("a.md", "REQUIREMENTS", "cpt-x-req-1"),
+        ("b.md", "REQUIREMENTS", "cpt-x-req-1"),
+        ("b.md", "REQUIREMENTS", "cpt-x-req-2"),
+    ]
+
+
+# ------------------------------------------------------------------ references
+def test_references_filter_target_exclude_definitions_by_default():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False),
+                 ("cpt-x-req-1", "reference", 4, True),
+                 ("cpt-x-req-2", "reference", 7, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        refs = scan.references("cpt-x-req-1", _arts("a.md"), {}, include_definitions=False)
+    assert refs == [{"artifact": "a.md", "artifact_type": "REQUIREMENTS", "line": 4,
+                     "kind": None, "type": "reference", "checked": True}]
+
+
+def test_references_include_definitions_and_attach_source():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False), ("cpt-x-req-1", "reference", 4, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        refs = scan.references("cpt-x-req-1", _arts("a.md"), {"a.md": "SRC"},
+                               include_definitions=True)
+    assert [r["type"] for r in refs] == ["definition", "reference"]
+    assert all(r["source"] == "SRC" for r in refs)
+
+
+# ----------------------------------------------------------------- definitions
+def test_definitions_are_definition_only_and_omit_the_type_key():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False),
+                 ("cpt-x-req-1", "reference", 4, False),
+                 ("cpt-x-req-2", "definition", 9, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        defs = scan.definitions("cpt-x-req-1", _arts("a.md"), {})
+    assert defs == [{"artifact": "a.md", "artifact_type": "REQUIREMENTS", "line": 1,
+                     "kind": None, "checked": False}]
+    assert "type" not in defs[0]
+
+
+def test_definitions_attach_source_when_present():
+    hits = _hits(("cpt-x-req-1", "definition", 2, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        defs = scan.definitions("cpt-x-req-1", _arts("a.md"), {"a.md": "SRC"})
+    assert defs[0]["source"] == "SRC"
+
+
+def test_empty_artifacts_yield_empty_everywhere():
+    # §3a edge / empty state: no artifacts → empty across every view; scan_cpt_ids never called.
+    assert list(scan.scan_records([])) == []
+    assert scan.references("cpt-x-req-1", [], {}, include_definitions=True) == []
+    assert scan.definitions("cpt-x-req-1", [], {}) == []
+    assert scan.graph_for("cpt-x-req-1", []) == {"defined_in": [], "referenced_in": []}
+
+
+def test_non_ascii_artifact_path_passes_through_unchanged():
+    # A4: paths are stringified, not normalised — a non-ASCII path must survive verbatim.
+    hits = _hits(("cpt-x-req-1", "reference", 2, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        refs = scan.references("cpt-x-req-1", [(Path("café/交/x.md"), "REQUIREMENTS")], {},
+                               include_definitions=True)
+    assert refs[0]["artifact"] == "café/交/x.md"
+
+
+def test_unknown_id_yields_empty_for_both_views():
+    with patch.object(scan, "scan_cpt_ids",
+                      return_value=_hits(("cpt-x-req-1", "reference", 1, False))):
+        assert scan.references("cpt-x-nope", _arts("a.md"), {}, include_definitions=True) == []
+        assert scan.definitions("cpt-x-nope", _arts("a.md"), {}) == []
+
+
+# ------------------------------------------------------- graph_for (B4 parity)
+def test_graph_for_partitions_into_defs_and_refs_matching_the_views():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False),
+                 ("cpt-x-req-1", "reference", 4, False),
+                 ("cpt-x-req-1", "reference", 9, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        g = scan.graph_for("cpt-x-req-1", _arts("a.md"))
+        defs = scan.definitions("cpt-x-req-1", _arts("a.md"), {})
+        refs = scan.references("cpt-x-req-1", _arts("a.md"), {}, include_definitions=False)
+    assert g["defined_in"] == defs
+    assert g["referenced_in"] == refs
+    assert len(g["defined_in"]) == 1
+    assert len(g["referenced_in"]) == 2
+
+
+def test_graph_for_attaches_source_through_both_partitions():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False), ("cpt-x-req-1", "reference", 4, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
+        g = scan.graph_for("cpt-x-req-1", _arts("a.md"), {"a.md": "SRC"})
+    assert g["defined_in"][0]["source"] == "SRC"
+    assert g["referenced_in"][0]["source"] == "SRC"
+
+
+# ------------------------------------------- A1: adds no new error-swallowing
+def test_util_does_not_swallow_scan_errors():
+    with patch.object(scan, "scan_cpt_ids", side_effect=OSError("boom")):
+        records = scan.scan_records(_arts("a.md"))  # lazy — does not raise until iterated
+        with pytest.raises(OSError):
+            list(records)
+
+
+# --------------------------------------------------------- property (§3a)
+def test_property_references_typed_and_count_matches_target_hits():
+    rng = random.Random(1234)
+    ids = ["cpt-x-req-1", "cpt-x-req-2", "cpt-x-flow-a"]
+    for _ in range(400):
+        hits = [{"id": rng.choice(ids), "type": rng.choice(["definition", "reference"]),
+                 "line": rng.randint(1, 50), "checked": rng.choice([True, False])}
+                for _ in range(rng.randint(0, 6))]
+        target = rng.choice(ids)
+        with patch.object(scan, "scan_cpt_ids", return_value=hits):
+            refs = scan.references(target, _arts("a.md"), {}, include_definitions=True)
+            recs = list(scan.scan_records(_arts("a.md")))
+        assert all(r["type"] in ("definition", "reference") for r in refs)
+        assert len(refs) == sum(1 for h in hits if h["id"] == target)
+        assert len(recs) == len(hits)
+
+
+def test_property_definitions_and_graph_partition_the_target_scan():
+    rng = random.Random(99)
+    ids = ["cpt-x-req-1", "cpt-x-req-2", "cpt-x-flow-a"]
+    for _ in range(400):
+        hits = [{"id": rng.choice(ids), "type": rng.choice(["definition", "reference"]),
+                 "line": rng.randint(1, 50), "checked": rng.choice([True, False])}
+                for _ in range(rng.randint(0, 6))]
+        target = rng.choice(ids)
+        with patch.object(scan, "scan_cpt_ids", return_value=hits):
+            defs = scan.definitions(target, _arts("a.md"), {})
+            g = scan.graph_for(target, _arts("a.md"))
+        n_defs = sum(1 for h in hits if h["id"] == target and h["type"] == "definition")
+        n_refs = sum(1 for h in hits if h["id"] == target and h["type"] == "reference")
+        assert len(defs) == n_defs
+        assert all("type" not in d for d in defs)
+        assert len(g["defined_in"]) == n_defs
+        assert len(g["referenced_in"]) == n_refs
+
+
+@pytest.mark.parametrize("call", [
+    lambda: scan.references("cpt-x-req-1", _arts("a.md"), {}, include_definitions=True),
+    lambda: scan.definitions("cpt-x-req-1", _arts("a.md"), {}),
+    lambda: scan.graph_for("cpt-x-req-1", _arts("a.md")),
+])
+def test_no_projection_swallows_scan_errors(call):
+    # A1: every projection iterates scan_records eagerly, so a scan_cpt_ids error propagates.
+    with patch.object(scan, "scan_cpt_ids", side_effect=OSError("boom")):
+        with pytest.raises(OSError):
+            call()
+
+
+def test_source_attached_per_artifact_across_multiple_artifacts():
+    per = {Path("a.md"): _hits(("cpt-x-req-1", "reference", 1, False)),
+           Path("b.md"): _hits(("cpt-x-req-1", "reference", 2, False))}
+    arts = [(Path("a.md"), "REQUIREMENTS"), (Path("b.md"), "REQUIREMENTS")]
+    with patch.object(scan, "scan_cpt_ids", side_effect=lambda p: per[p]):
+        refs = scan.references("cpt-x-req-1", arts, {"a.md": "SRC_A", "b.md": "SRC_B"},
+                               include_definitions=True)
+    assert refs[0]["source"] == "SRC_A"
+    assert refs[1]["source"] == "SRC_B"
+
+
+def test_definitions_source_attached_per_artifact_across_multiple_artifacts():
+    # #5 mirror: definitions() shares _record with references(), but pin its per-artifact
+    # source mapping directly on the where_defined path too — not only via references().
+    per = {Path("a.md"): _hits(("cpt-x-req-1", "definition", 1, False)),
+           Path("b.md"): _hits(("cpt-x-req-1", "definition", 2, False))}
+    arts = [(Path("a.md"), "REQUIREMENTS"), (Path("b.md"), "REQUIREMENTS")]
+    with patch.object(scan, "scan_cpt_ids", side_effect=lambda p: per[p]):
+        defs = scan.definitions("cpt-x-req-1", arts, {"a.md": "SRC_A", "b.md": "SRC_B"})
+    assert defs[0]["source"] == "SRC_A"
+    assert defs[1]["source"] == "SRC_B"
--- a/skills/studio/scripts/studio/commands/list_id_kinds.py
+++ b/skills/studio/scripts/studio/commands/list_id_kinds.py
@@ -10,7 +10,7 @@ import argparse
 from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple
 
-from ..utils.document import scan_cpt_ids
+from ..utils import cpt_reference_scan
 from ..utils.ui import ui
 from .list_ids import (
     _collect_known_kinds,
@@ -62,19 +62,18 @@ def _collect_kind_maps(
     kind_to_templates: Dict[str, Set[str]] = {}
     kind_counts: Dict[str, int] = {}
 
-    for artifact_path, artifact_type in artifacts_to_scan:
-        for hit in scan_cpt_ids(artifact_path):
-            if hit.get("type") != "definition":
-                continue
-            cpt_id = str(hit.get("id") or "").strip()
-            if not cpt_id:
-                continue
-            # @cpt-begin:cpt-studio-algo-traceability-validation-list-id-kinds:p1:inst-kinds-aggregate
-            for kind_name in _infer_kinds(cpt_id, registered_systems, known_kinds):
-                kind_to_templates.setdefault(kind_name, set()).add(artifact_type)
-                template_to_kinds.setdefault(artifact_type, set()).add(kind_name)
-                kind_counts[kind_name] = kind_counts.get(kind_name, 0) + 1
-            # @cpt-end:cpt-studio-algo-traceability-validation-list-id-kinds:p1:inst-kinds-aggregate
+    for _artifact_path, artifact_type, hit in cpt_reference_scan.scan_records(artifacts_to_scan):
+        if hit.get("type") != "definition":
+            continue
+        cpt_id = str(hit.get("id") or "").strip()
+        if not cpt_id:
+            continue
+        # @cpt-begin:cpt-studio-algo-traceability-validation-list-id-kinds:p1:inst-kinds-aggregate
+        for kind_name in _infer_kinds(cpt_id, registered_systems, known_kinds):
+            kind_to_templates.setdefault(kind_name, set()).add(artifact_type)
+            template_to_kinds.setdefault(artifact_type, set()).add(kind_name)
+            kind_counts[kind_name] = kind_counts.get(kind_name, 0) + 1
+        # @cpt-end:cpt-studio-algo-traceability-validation-list-id-kinds:p1:inst-kinds-aggregate
     return template_to_kinds, kind_to_templates, kind_counts
 # @cpt-end:cpt-studio-algo-traceability-validation-list-id-kinds:p1:inst-kinds-scan-ids
 
--- a/skills/studio/scripts/studio/commands/list_ids.py
+++ b/skills/studio/scripts/studio/commands/list_ids.py
@@ -7,8 +7,8 @@ import re
 from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple
 
+from ..utils import cpt_reference_scan
 from ..utils.codebase import scan_registered_codebase_references
-from ..utils.document import scan_cpt_ids
 from ..utils.ui import ui
 # @cpt-end:cpt-studio-flow-traceability-validation-query:p1:inst-query-imports
 
@@ -211,23 +211,22 @@ def _collect_artifact_hits(
     """Scan artifact IDs and annotate each hit with inferred kind metadata."""
     # @cpt-begin:cpt-studio-flow-traceability-validation-query:p1:inst-scan-all
     hits: List[Dict[str, object]] = []
-    for artifact_path, artifact_type in artifacts_to_scan:
-        for fh in scan_cpt_ids(artifact_path):
-            cid = str(fh.get("id") or "").strip()
-            if not cid:
-                continue
-            hit: Dict[str, object] = {
-                "id": cid,
-                "kind": _infer_primary_kind(cid, registered_systems, known_kinds),
-                "type": fh.get("type"),
-                "artifact_type": artifact_type,
-                "line": fh.get("line"),
-                "artifact": str(artifact_path),
-                "checked": bool(fh.get("checked", False)),
-            }
-            if fh.get("priority") is not None:
-                hit["priority"] = fh.get("priority")
-            hits.append(hit)
+    for artifact_path, artifact_type, fh in cpt_reference_scan.scan_records(artifacts_to_scan):
+        cid = str(fh.get("id") or "").strip()
+        if not cid:
+            continue
+        hit: Dict[str, object] = {
+            "id": cid,
+            "kind": _infer_primary_kind(cid, registered_systems, known_kinds),
+            "type": fh.get("type"),
+            "artifact_type": artifact_type,
+            "line": fh.get("line"),
+            "artifact": str(artifact_path),
+            "checked": bool(fh.get("checked", False)),
+        }
+        if fh.get("priority") is not None:
+            hit["priority"] = fh.get("priority")
+        hits.append(hit)
     return hits
     # @cpt-end:cpt-studio-flow-traceability-validation-query:p1:inst-scan-all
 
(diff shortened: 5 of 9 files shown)


skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,123 @@
+"""Shared cpt-id reference scan.
```
ainetx, line 1: ### Other scan_cpt_ids call sites left unmigrated to the new shared util

**Severity:** Minor

**Problem**
cpt_reference_scan.py was introduced specifically to deduplicate 'the one scan_cpt_ids traversal... each re-implemented', and its docstring calls out the four query commands it replaces. However artifacts_meta.py, constraints.py, context.py, and semantic_coverage.py all still import and use document.scan_cpt_ids directly with their own logic, with no comment anywhere explaining why they were excluded from the refactor.

<details>

**How to reproduce**
1. grep -rn 'from.document import scan_cpt_ids' skills/studio/scripts/studio/utils/ — finds artifacts_meta.py:395, constraints.py:1771, context.py:947, semantic_coverage.py:29 in addition to the new cpt_reference_scan.py itself.
2. Compare against cpt_reference_scan.py's docstring, which only names the four command modules as consumers it replaces.

**Expected behavior**
Either these four modules are migrated to cpt_reference_scan.scan_records/references/definitions for consistency, or the PR/docstring explicitly states why they remain separate (e.g., different traversal semantics, out of scope for this refactor).

**Actual behavior**
These modules keep independent scan_cpt_ids loops with no documented rationale, so they don't inherit the new util's documented behavior guarantees (e.g., the 'cpt-id scoped, prose out of scope' v1 limit) and remain a source of duplicated-loop drift.

```text
scan_cpt_ids (document.py)
|-- cpt_reference_scan.py (new, canonical) --> where_used/where_defined/list_ids/list_id_kinds [migrated]
|-- artifacts_meta.py, constraints.py, context.py, semantic_coverage.py [NOT migrated, still hand-rolled]
```

**Impact**
Future changes to scan semantics (e.g., additional guarantees documented only in cpt_reference_scan.py) will silently not apply to these four modules, and any bugs fixed in the shared util won't propagate to them.

**Suggested correction**
Either migrate the four modules to use cpt_reference_scan's helpers, or add a comment in cpt_reference_scan.py's module docstring explicitly scoping out these remaining consumers with a rationale.

**How to verify**
Re-grep for scan_cpt_ids consumers after the fix and confirm each either uses cpt_reference_scan or has a documented reason not to.

</details>

tests/test_cpt_reference_scan.py
```diff
@@ -0,0 +1,146 @@
...
+        refs = scan.references("cpt-x-req-1", _arts("a.md"), {}, include_definitions=False)
+    assert g["defined_in"] == defs
+    assert g["referenced_in"] == refs
+    assert len(g["defined_in"]) == 1
+    assert len(g["referenced_in"]) == 2
+
+
+def test_graph_for_attaches_source_through_both_partitions():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False), ("cpt-x-req-1", "reference", 4, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
```
ainetx, line 118: ### OSError-propagation guarantee tested only for scan_records, not its three callers

**Severity:** Minor

**Problem**
cpt_reference_scan.py defines references(), definitions(), and graph_for(), each independently iterating scan_records() in their own for-loop body. Only scan_records()'s raw iteration is covered by test_util_does_not_swallow_scan_errors; the other three functions have no equivalent test proving an exception raised by scan_cpt_ids mid-iteration actually propagates out of them rather than being silently caught by a stray try/except added later.

<details>

**How to reproduce**
1. A future edit adds error handling (e.g., try/except OSError: continue) inside references() or definitions() to make where_used/where_defined more resilient to a bad file. 2. Run the existing test suite. 3. test_util_does_not_swallow_scan_errors still passes because it only exercises scan_records() directly. 4. The silent behavior divergence goes undetected.

**Expected behavior**
Each of references(), definitions(), and graph_for() should have (or share via parametrization) a test asserting that an OSError raised by scan_cpt_ids during iteration propagates out uncaught, matching scan_records()'s guarantee.

**Actual behavior**
Only scan_records() is exercised with a raising scan_cpt_ids side_effect; the three consumer functions are untested for this property despite each having its own loop body.

```text
scan_cpt_ids raises OSError -> scan_records (tested: propagates) -> references()/definitions()/graph_for() own loops (untested: propagation unverified)
```

**Impact**
A future localized error-handling change to one projection function could silently diverge from the others' fail-open/fail-closed semantics with no test catching the regression.

**Suggested correction**
Add a parametrized or repeated version of test_util_does_not_swallow_scan_errors that also calls references(...), definitions(...), and graph_for(...) with a raising scan_cpt_ids side_effect and asserts pytest.raises(OSError) for each.

**How to verify**
Add the missing tests, then verify they fail if a try/except is temporarily added around the loop body in any of the three functions, and pass once removed.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,123 @@
...
+def references(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Dict[str, str],
+    *,
+    include_definitions: bool,
+) -> List[Dict[str, object]]:
+    """References to *target_id* (optionally including its definitions).
+
+    The ``where_used`` projection, verbatim: a reference hit whose id matches,
```
ainetx, line 78: ### definitions() and references() hand-duplicate record construction instead of sharing a helper

**Severity:** Minor

**Problem**
Both functions independently repeat the same filter-by-id, build-record ({artifact, artifact_type, line, kind, checked}), and attach-source logic; references() adds only a 'type' key and an extra definition-exclusion check on top of the same shape.

<details>

**How to reproduce**
1. Open cpt_reference_scan.py lines 37-92. 2. Compare the bodies of references() and definitions() line by line.

**Expected behavior**
A single shared internal builder (e.g. _record(artifact_path, artifact_type, h, path_to_source) -> dict) that both references() and definitions() call, adding/removing the 'type' key as needed.

**Actual behavior**
Each function reimplements int(h.get('line',1) or 1), kind=None, checked=bool(...), and the src = path_to_source.get(...) attach step separately.

```text
scan_records -> [references(): build record A] 
-> [definitions(): build record B, ~90% identical to A]
```

**Impact**
Any future field addition (e.g. kind inference) must be made in two places; missing one silently desyncs the two projections used by where_used/where_defined/graph_for.

**Suggested correction**
Factor the shared record fields (artifact, artifact_type, line, kind, checked, source) into one private helper used by both references() and definitions().

**How to verify**
Refactor and confirm existing tests in tests/test_cpt_reference_scan.py still pass unchanged.

</details>

tests/test_cpt_reference_scan.py
```diff
@@ -0,0 +1,146 @@
...
+            list(records)
+
+
+# --------------------------------------------------------- property (§3a)
+def test_property_references_typed_and_count_matches_target_hits():
+    rng = random.Random(1234)
+    ids = ["cpt-x-req-1", "cpt-x-req-2", "cpt-x-flow-a"]
+    for _ in range(400):
+        hits = [{"id": rng.choice(ids), "type": rng.choice(["definition", "reference"]),
+                 "line": rng.randint(1, 50), "checked": rng.choice([True, False])}
```
ainetx, line 138: ### Property-based fuzz test omits definitions() and graph_for()

**Severity:** Minor

**Problem**
test_property_references_typed_and_count_matches_target_hits (tests/test_cpt_reference_scan.py:132-146) only randomizes scan.references and scan.scan_records; scan.definitions and scan.graph_for have only small hand-written example tests, leaving their filter logic without randomized/adversarial coverage.

<details>

**How to reproduce**
1. Open tests/test_cpt_reference_scan.py lines 132-146. 2. Note only scan.references(...) and scan.scan_records(...) are exercised inside the randomized loop.

**Expected behavior**
The property test (or a twin) should also assert invariants for scan.definitions (e.g. count matches definition-type hits for target) and scan.graph_for (partition sums match total hits).

**Actual behavior**
definitions() and graph_for() have no randomized coverage, so a definition/reference split off-by-one under adversarial hit sequences would not be caught by this property test.

```text
random hits -> references() [property-checked] 
-> definitions() [not property-checked] 
-> graph_for() [not property-checked]
```

**Impact**
A subtle regression in the definitions/graph_for filtering path could pass CI silently since only hand-picked example inputs cover it.

**Suggested correction**
Add definitions()/graph_for() assertions to the existing property test loop using the same randomized hit generation.

**How to verify**
Extend the property test and confirm it fails on an intentionally introduced definitions() bug, then passes after the fix.

</details>

tests/test_cpt_reference_scan.py
```diff
@@ -0,0 +1,146 @@
...
+        ("b.md", "REQUIREMENTS", "cpt-x-req-2"),
+    ]
+
+
+# ------------------------------------------------------------------ references
+def test_references_filter_target_exclude_definitions_by_default():
+    hits = _hits(("cpt-x-req-1", "definition", 1, False),
+                 ("cpt-x-req-1", "reference", 4, True),
+                 ("cpt-x-req-2", "reference", 7, False))
+    with patch.object(scan, "scan_cpt_ids", return_value=hits):
```
ainetx, line 43: ### New util tests never verify per-artifact source attribution across multiple artifacts

**Severity:** Minor

**Problem**
test_references_include_definitions_and_attach_source and test_definitions_attach_source_when_present (the only tests asserting `source` attachment) both call `_arts("a.md")` — a single artifact — with a `path_to_source` map of one entry. No test supplies two or more artifacts with distinct path_to_source values and asserts each returned record keeps the source belonging to its own artifact.

<details>

**How to reproduce**
1. Read tests/test_cpt_reference_scan.py lines 43-82 (references/definitions source tests). 2. Note both use `_arts("a.md")` only. 3. Compare to test_scan_records_yields_every_hit_with_artifact_context, which does use two artifacts but never attaches distinct sources.

**Expected behavior**
A test with e.g. `_arts("a.md", "b.md")` and `{"a.md": "SRC_A", "b.md": "SRC_B"}` asserting each returned record's `source` matches its own artifact, not a reused/first value.

**Actual behavior**
Coverage only proves source attachment works for a single artifact; a bug that mixed up sources across artifacts (e.g. always using the first path's source) would not be caught by the current suite.

```text
references()/definitions() loop over artifacts -> path_to_source.get(str(artifact_path)) -> [UNTESTED: correct per-artifact lookup across >1 artifact]
```

**Impact**
A future regression that misattributes source across multiple artifacts (e.g. off-by-one in the loop, or accidentally hoisting a single lookup) would ship undetected since no test exercises the multi-artifact + source combination.

**Suggested correction**
Add a test with 2+ artifacts and distinct path_to_source entries verifying each record's source matches its own artifact.

**How to verify**
Add the described test; confirm it fails against a deliberately-broken implementation that mixes up source lookups across artifacts, and passes against the current one.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of c27dc63a0f8e91e2bf7273a007d1b1ede4195c89 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Searching for other direct scan_cpt_ids importers ('from..utils.document import scan_cpt_ids' and 'from.document import scan_cpt_ids') across the repo returns hits only inside archived PR snapshot files (.prs/81/diff.patch,.prs/81/meta.json,.prs/81/review_comments.json) referencing a different, unrelated 'cypilot' skill tree, not any live file under skills/studio/scripts/studio/; no other current-codebase caller of scan_cpt_ids outside the four migrated commands and the new util was found.
- **Reviewer:** Direct consumers remain in artifacts_meta.py, constraints.py, context.py, and semantic_coverage.py, but the new utility explicitly scopes its deduplication to the four named query commands; the remaining callers have distinct validation/indexing or workspace-resolution behavior, while the architecture separately defers artifact-quality detector use of graph_for.
- **Verifier:** cpt_reference_scan.py's own module docstring scopes the refactor to exactly 'where_used, where_defined, list_ids and list_id_kinds' — it never claims to be the canonical entry point for all scan_cpt_ids consumers. Other direct consumers do exist (WorkspaceContext.get_all_artifact_ids and _collect_def_ids_from_artifacts in artifacts_meta.py, per the prefetched context), but these serve a different purpose (building a set of known IDs for cross-repo/registry validation, not per-command reference/definition projections with kind/source annotation), so they're a different use case rather than an unmigrated duplicate of the same query-command loop this PR targeted.

</details>

tests/test_cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of c27dc63a0f8e91e2bf7273a007d1b1ede4195c89 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff's parametrized `test_no_projection_swallows_scan_errors` (using `pytest.mark.parametrize` over lambdas calling `scan.references(...)`, `scan.definitions(...)`, and `scan.graph_for(...)`) explicitly asserts OSError propagation for all three projection functions, not just scan_records.
- **Reviewer:** test_no_projection_swallows_scan_errors parametrizes references(), definitions(), and graph_for() with scan_cpt_ids raising OSError, explicitly covering all projections.
- **Verifier:** The diff's test_cpt_reference_scan.py includes a parametrized test_no_projection_swallows_scan_errors covering scan.references, scan.definitions, and scan.graph_for (not just scan_records), confirming OSError propagation is independently tested for all three projections.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of c27dc63a0f8e91e2bf7273a007d1b1ede4195c89 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** references() and definitions() both delegate record construction and source-attachment to the single shared `_record()` helper (with only an `include_type` flag differing); the record-building logic is factored, not copy-pasted — only the thin filter predicates (by id / by type) differ between the two functions, which is inherent to their distinct semantics.
- **Reviewer:** Common record construction and source attachment are factored into _record(); the remaining loops have distinct definition-versus-reference filtering semantics.
- **Verifier:** references() and definitions() both delegate record construction and source-attachment to the shared _record() helper (inst-record) rather than duplicating that logic; only the filter predicate (definition-only vs id-match/include_definitions) differs between them, so the seam claimed to be removed is in fact shared, not copy-pasted.

</details>

tests/test_cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of c27dc63a0f8e91e2bf7273a007d1b1ede4195c89 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff includes a second property test, `test_property_definitions_and_graph_partition_the_target_scan`, which randomizes hits and asserts against both `scan.definitions` and `scan.graph_for` (checking def/ref partition counts and the omitted `type` key), so definitions/graph_for are not left without randomized coverage.
- **Reviewer:** test_property_definitions_and_graph_partition_the_target_scan randomizes hit sequences and asserts definitions() counts plus graph_for's definition/reference partition.
- **Verifier:** The diff includes test_property_definitions_and_graph_partition_the_target_scan, a randomized property test (rng seed 99, 400 iterations) that exercises scan.definitions and scan.graph_for directly, so the premise that only references/scan_records get property coverage is incorrect.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,133 @@
...
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-definitions
+def definitions(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Dict[str, str],
+) -> List[Dict[str, object]]:
+    """Definitions of *target_id* — the ``where_defined`` projection, verbatim
+    (no ``type`` key; ``kind`` left ``None`` for the caller to annotate)."""
+    out: List[Dict[str, object]] = []
+    for artifact_path, artifact_type, h in scan_records(artifacts_to_scan):
```
ainetx, line 101: ### graph_for's optional path_to_source diverges from definitions()/references()'s required parameter

**Severity:** Minor

**Problem**
graph_for declares path_to_source: Optional[Dict[str, str]] = None (defaulted via `path_to_source or {}`), while the definitions() and references() functions it wraps both require path_to_source: Dict[str, str] as a mandatory positional argument with no default. No docstring explains or flags this asymmetry.

<details>

**How to reproduce**
1. A future caller reads graph_for's signature and sees path_to_source is optional.
2. They assume the same optionality applies to definitions()/references() in the same module.
3. They call definitions(target_id, artifacts) without the third argument.
4. TypeError: missing required positional argument 'path_to_source'.

**Expected behavior**
All three public functions in the module should present a consistent optionality contract for the same parameter, or the divergence should be explicitly documented.

**Actual behavior**
graph_for silently defaults path_to_source while its two callees require it explicitly, creating an inconsistent API surface within one module.

```text
graph_for(path_to_source=None) --defaults--> {} --passes to--> definitions(path_to_source: required) / references(path_to_source: required)
^
future direct caller assumes same optionality -> TypeError
```

**Impact**
Low — only affects future direct callers of definitions()/references() who infer the contract from graph_for; no current caller is affected since all present call sites pass path_to_source correctly.

**Suggested correction**
Either give definitions()/references() the same Optional default with an internal `or {}`, or add a one-line docstring note on graph_for explaining why it alone defaults the parameter.

**How to verify**
Add a signature-consistency test or docstring note; call definitions()/references() without path_to_source and confirm the intended behavior (error or graceful default) matches documentation.

</details>

architecture/features/traceability-validation.md
```diff
@@ -228,6 +229,23 @@ Catches structural and traceability issues that AI agents miss or hallucinate 
 - [x] - `p1` - File I/O utilities: safe text reader, text file iterator, relative path converter - `inst-scan-ids-file-utils`
 - [x] - `p1` - Wrapper function for `parse_cpt` identifier parser - `inst-parse-cpt-fn`
 
+### CPT Reference Scan
+
```
ainetx, line 233: ### New architecture doc's Output contract omits that 'kind' is always None

**Severity:** Minor

**Problem**
The new '### CPT Reference Scan' section documents Output records as '{artifact, artifact_type, line, kind, type, checked, source?}' listing 'kind' as a normal field, but every record built by _record() in cpt_reference_scan.py hardcodes 'kind': None with no logic to populate it — the value is left for callers (e.g. list_ids.py's _infer_primary_kind) to fill in separately outside this module.

<details>

**How to reproduce**
1. Read architecture/features/traceability-validation.md's new CPT Reference Scan section Output line. 2. Read cpt_reference_scan.py _record(): rec = {..., 'kind': None,...} with no assignment elsewhere in the file.

**Expected behavior**
The doc's Output contract should note that 'kind' is always null/omitted-of-meaning in this util's own output and is populated only by callers that choose to annotate it.

**Actual behavior**
The doc lists 'kind' as if it were part of this util's computed contract, which could mislead a reader into expecting a real kind value from where_defined()/where_used()/definitions()/references() directly.

```text
doc: Output {..., kind,...} --describes--> _record() returns kind=None always (caller fills in later)
```

**Impact**
Low — a future maintainer reading only the architecture doc could assume 'kind' is meaningful straight out of this module and skip the caller-side inference step, or file a false 'missing kind' bug against this util.

**Suggested correction**
Add a short caveat to the Output line, e.g. '(kind is always null here; populated by callers that infer it)'.

**How to verify**
Re-read the doc section after the edit and confirm it explicitly notes kind's caller-populated nature, matching _record()'s implementation.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Fixed in c27dc63a. Extracted a shared `_record(..., include_type)` helper (its own `inst-record` CPT marker); `references` and `definitions` both build their locus through it. `include_type` places the `type` key in its historical position — between `kind` and `checked` — so the emitted dicts stay byte-for-byte what the commands produced before this util. `cfs validate` resolves the new marker with 0 errors, and the 136-command regression + full suite are green.

tests/test_cpt_reference_scan.py
SanjeevSolanki: Fixed in c27dc63a. Added `test_property_definitions_and_graph_partition_the_target_scan` (400 iterations): for random hit sets it asserts `definitions` equals the target's definition hits and carries no `type` key, and that `graph_for` partitions the target scan into the right defined_in/referenced_in counts. So both projections and the graph view are now fuzzed, not just `references`/`scan_records`.

tests/test_cpt_reference_scan.py
SanjeevSolanki: Fixed in c27dc63a. Added `test_source_attached_per_artifact_across_multiple_artifacts`: two artifacts, each with its own source string, and the test asserts each emitted record gets *its own* artifact's source (`SRC_A` vs `SRC_B`) — proving the mapping is per-artifact, not a single global source applied to all.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Declining, with the reasoning: the other `scan_cpt_ids` call sites do a *different* traversal, not the cpt-id-scoped def↔ref query these four commands shared.

- `constraints.py` splits one artifact's defs/refs (per-artifact, not target-scoped).
- `context.py` / `artifacts_meta.py` collect def-ids from a single artifact.
- `semantic_coverage.py` builds a def-id→artifact map inside its own algo.

None of those is "find every definition/reference of *this* cpt-id across the tree" — which is what `references`/`definitions`/`graph_for` are, and what the future gap/traceability/contradiction detectors will consume. Folding those unrelated shapes behind this util would couple them to a query they don't perform. So this PR scopes to the one true duplicate (the four query commands) rather than becoming a wrapper for every `scan_cpt_ids` caller.

tests/test_cpt_reference_scan.py
SanjeevSolanki: Fixed in c27dc63a. Added `test_no_projection_swallows_scan_errors`, parametrized over `references` / `definitions` / `graph_for`: each now asserts that a `scan_cpt_ids` `OSError` propagates through the projection (they iterate `scan_records` eagerly), not just through the raw generator. So the no-swallow guarantee is pinned for the public callers, not only the internal one.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cpt_reference_scan.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cpt_reference_scan.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cpt_reference_scan.py
ainetx: Re-verified against the current code (c27dc63a0f8e91e2bf7273a007d1b1ede4195c89) and this is still an issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_source_attached_per_artifact_across_multiple_artifacts (tests/test_cpt_reference_scan.py:180-187) exercises references() across two artifacts with distinct sources, but no equivalent multi-artifact test exists for definitions(); all definitions()-source tests (e.g. test_definitions_attach_source_when_present) use a single artifact.
- **Reviewer:** The only multi-artifact source-attribution test exercises references(); definitions() source attachment is tested only with one artifact, so its per-artifact mapping behavior is not covered.
- **Verifier:** test_source_attached_per_artifact_across_multiple_artifacts (lines 176-186) exercises references() across two artifacts with distinct sources (SRC_A/SRC_B), but no equivalent multi-artifact test exists for definitions() — its only source-attachment test (test_definitions_attach_source_when_present, lines 60-64) uses a single artifact, so multi-artifact source correctness for the where_defined path is unverified by these tests even though the same _record() helper is shared.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Documented rather than unified (fixed in a339061a). Keeping `path_to_source` required on `references` / `definitions` is deliberate — it's the fail-loud choice: the four command call sites always have the source map, so a missing arg there is a real bug we want surfaced as a TypeError, not silently defaulted. `graph_for` is the detector-facing convenience view where a caller usually wants only the loci, so it alone defaults (`None` == empty map, no `source` keys emitted). Added a docstring note on `graph_for` explaining the asymmetry.

architecture/features/traceability-validation.md
SanjeevSolanki: Fixed in a339061a. The Output line now notes `kind` is always `null` from this util and is populated by callers that infer it (e.g. `list_ids`'s kind inference). The util itself never computes a kind — it leaves the slot for the caller-side annotation step.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of a339061ad0fb2221430728cb85f3815fabec911c -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The asymmetry is explicitly intentional and documented in graph_for's own docstring: it states path_to_source is optional there because a detector usually only wants def/ref loci not source text, while it's required in definitions()/references() because every command call site already has the map and 'a missing map at those call sites is a bug we want surfaced, not silently defaulted' — so a future direct caller reading either docstring would not be misled.
- **Reviewer:** The divergence is explicit and intentional: graph_for’s docstring states detectors usually do not need source text and treats None as an empty map, while definitions/references require the map because all current command call sites have it and a missing map there should surface as a bug.
- **Verifier:** The asymmetry is intentional and explicitly documented in graph_for's own docstring: 'path_to_source is optional here — a detector usually wants only the def/ref loci, not source text — whereas references/definitions require it, since every command call site has the map. Requiring it there is deliberate: a missing map at those call sites is a bug we want surfaced, not silently defaulted.' This directly addresses the concern raised in the check — a future direct caller reading the docstring is told the required-ness differs by design, not by oversight, so it would not be misled.

</details>

architecture/features/traceability-validation.md
ainetx: **RESOLVED** -- Confirmed fixed as of a339061ad0fb2221430728cb85f3815fabec911c -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The doc's Output shape `{artifact, artifact_type, line, kind, type, checked, source?}` (definitions omit `type`, `kind` always null) matches `_record()`'s emitted dict exactly (artifact, artifact_type, line, kind=None, optional type when include_type=True, checked, optional source) as used by `references()` (include_type=True) and `definitions()` (include_type=False, per diff comment 'no type key; kind left None'); the def↔ref graph `{defined_in, referenced_in}` matches `graph_for()`'s literal return `{"defined_in": definitions(...), "referenced_in": references(..., include_definitions=False)}`; and the doc's aggregate Input (artifacts, target cpt-id, path→source map) covers the union of parameters across `scan_records` (artifacts only), `references`/`definitions` (target_id, artifacts, path_to_source) and `graph_for` (target_id, artifacts, optional path_to_source), all visible directly in the new file's full contents shown in the diff.
- **Reviewer:** The documented artifact tuples, target ID, and path-to-source map are represented across the four functions; scan_records yields artifact context, references/definitions emit the documented locus fields with definitions omitting type, kind is always None as documented, and graph_for literally returns defined_in and referenced_in lists.
- **Verifier:** The doc's Input/Output block describes the algo as a whole (its four numbered steps map 1:1 to scan_records/references/definitions/graph_for), not a single function signature, and each field matches: references()/definitions() build the shared _record() dict with exactly {artifact, artifact_type, line, kind: None, [type], checked, [source]} (type present only for references, per the doc's explicit note), and graph_for() literally returns {"defined_in":..., "referenced_in":...} built from definitions()/references(). scan_records() corresponds to doc step 1 (the shared scan yielding artifact context) and doesn't claim to take target-id/path-map itself. All fields and both output shapes in the diff's new util file are accounted for.

</details>

architecture/features/traceability-validation.md
```diff
@@ -228,6 +229,23 @@ Catches structural and traceability issues that AI agents miss or hallucinate 
 - [x] - `p1` - File I/O utilities: safe text reader, text file iterator, relative path converter - `inst-scan-ids-file-utils`
 - [x] - `p1` - Wrapper function for `parse_cpt` identifier parser - `inst-parse-cpt-fn`
 
+### CPT Reference Scan
```
ainetx, line 232: ### Document the CPT-only query scope in the feature contract

**Severity:** Minor

**Problem**
The utility documents that PRD-NNN references are out of scope, but the new feature section and CLI table do not disclose that limitation.

<details>

**How to reproduce**
1. Read the CPT Reference Scan section or CLI table. 2. Determine the documented scope of where-used or where-defined. 3. Observe no exclusion for PRD-NNN requirement IDs.

**Expected behavior**
Feature documentation states that the shared query scan covers cpt-* identifiers only and excludes PRD-NNN references in v1.

**Actual behavior**
The scope limit appears only in the utility module docstring.

```text
feature documentation
|
v
reader infers query coverage
|
v
PRD-NNN exclusion undisclosed
```

**Impact**
Users can incorrectly rely on where-used or where-defined to locate PRD-style requirement references.

**Suggested correction**
Add a concise cpt-id-only and PRD-NNN-excluded scope note to the CPT Reference Scan section or CLI query contract.

**How to verify**
Confirm the feature documentation states the same scan scope as the utility docstring.

</details>

ainetx:
### WorkspaceContext.get_all_artifact_ids still hand-rolls its own scan_cpt_ids loop instead of using the new shared util

**Severity:** Minor

**Problem**
The new cpt_reference_scan.py util is introduced to deduplicate 'the one scan_cpt_ids traversal... each re-implemented' across callers, and the PR migrates list_ids.py, list_id_kinds.py, where_defined.py, and where_used.py. However, skills/studio/scripts/studio/utils/context.py's WorkspaceContext.get_all_artifact_ids() (per the search results in.prs/81/diff.patch showing 'from.document import scan_cpt_ids' and a hand-rolled per-artifact for-loop filtering type=='definition') still calls document.scan_cpt_ids directly with its own projection logic, and is not migrated to use cpt_reference_scan.definitions()/scan_records().

<details>

**How to reproduce**
1. Open skills/studio/scripts/studio/utils/context.py and locate WorkspaceContext.get_all_artifact_ids(). 2. Observe it imports 'from.document import scan_cpt_ids' locally and loops over primary/remote artifacts calling scan_cpt_ids directly, filtering h.get('type')=='definition' itself. 3. Compare against list_ids.py/where_defined.py, which now delegate the identical projection to cpt_reference_scan.definitions()/scan_records().

**Expected behavior**
All direct consumers of scan_cpt_ids that build a definition/reference projection over multiple artifacts should route through the new cpt_reference_scan util so the scan and its v1 'cpt-id scoped' behavior stays centralized and consistent.

**Actual behavior**
get_all_artifact_ids keeps its own independent scan_cpt_ids loop and definition filter, duplicating logic the new util was built to consolidate, and is not mentioned anywhere in the PR as an intentional exception.

```text
document.scan_cpt_ids
|-- migrated: list_ids/list_id_kinds/where_defined/where_used -> cpt_reference_scan.scan_records/references/definitions
\-- NOT migrated: WorkspaceContext.get_all_artifact_ids -> still loops scan_cpt_ids directly
```

**Impact**
Duplicated-loop drift persists in one more caller not called out in the PR's own stated scope, undermining the refactor's stated goal of a single canonical scan; future fixes to the shared scan behavior may not propagate here.

**Suggested correction**
Migrate WorkspaceContext.get_all_artifact_ids to use cpt_reference_scan.scan_records/definitions, or explicitly document why it is excluded from this refactor.

**How to verify**
After migration, confirm get_all_artifact_ids produces identical output via existing workspace tests, and that no direct 'from.document import scan_cpt_ids' import remains in context.py.

</details>

---
_Original location: skills/studio/scripts/studio/utils/context.py:1 -- inline anchoring could not be resolved after 1 attempt(s)._

skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,139 @@
...
+# @cpt-begin:cpt-studio-algo-cpt-reference-scan:p1:inst-graph-for
+def graph_for(
+    target_id: str,
+    artifacts_to_scan: List[Tuple[Path, str]],
+    path_to_source: Optional[Dict[str, str]] = None,
+) -> Dict[str, List[Dict[str, object]]]:
+    """The def↔ref view a detector wants for one cpt-id: where it is defined and
+    where it is referenced.
+
+    ``defined_in`` is the definition loci (no ``type`` key); ``referenced_in`` is the
```
ainetx, line 120: ### graph_for has no documented cost contract for multi-id detector usage

**Severity:** Minor

**Problem**
graph_for's docstring states 'It scans the tree twice — once per view — which is fine while unused; single-scan it when a detector consumes it.' This only addresses the current no-op state; it provides no caller-facing guidance (e.g. batch across ids, cache artifacts_to_scan, or a warning against calling per-id in a loop) for the first real detector that will call graph_for once per cpt-id across potentially many ids, which would produce O(n_ids * n_artifacts) rescans of the full artifact tree.

<details>

**How to reproduce**
1. A future artifact-quality detector needs def/ref info for every cpt-id in a large repo. 2. It calls graph_for(id, artifacts_to_scan) once per id in a loop, as the API's shape naturally invites. 3. Each call performs two independent full scan_records traversals over all artifacts. 4. No doc anywhere warns against this pattern or suggests an alternative.

**Expected behavior**
The docstring or the architecture doc's 'CPT Reference Scan' section should specify a caller obligation (e.g. 'callers processing many ids should build a single scan_records() pass and derive per-id partitions, not call graph_for per id') to prevent reintroducing the double-scan cost at scale.

**Actual behavior**
Only a passive note that the double-scan is 'fine while unused' with a vague suggestion to 'single-scan it when a detector consumes it' — no bound, no batching contract, no explicit guidance for repeated invocation.

```text
detector loop over N ids -> graph_for(id) per id -> 2 x scan_records(all artifacts) per call -> O(N * 2 * artifacts) rescans, undocumented
```

**Impact**
First detector implementation may naively call graph_for per id, causing quadratic-ish rescanning of the full artifact tree with no documented warning to avoid it.

**Suggested correction**
Add an explicit caller-facing note (docstring and/or architecture doc) stating that multi-id callers must batch by building a single scan_records pass and partitioning locally, not call graph_for in a per-id loop.

**How to verify**
Re-read the docstring/doc section for an explicit batching/caching obligation for multi-id callers.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,139 @@
...
+    non-definition reference loci (each with its ``type``). Built with the same
+    ``references`` / ``definitions`` logic the query commands use, so it classifies
+    definitions and references identically to them. (It scans the tree twice — once per
+    view — which is fine while unused; single-scan it when a detector consumes it.)
+
+    ``path_to_source`` is optional here — a detector usually wants only the def/ref
+    loci, not source text — whereas ``references`` / ``definitions`` require it, since
+    every command call site has the map. Requiring it there is deliberate: a missing
+    map at those call sites is a bug we want surfaced, not silently defaulted. Passing
+    ``None`` is equivalent to an empty map (no ``source`` keys emitted).
```
ainetx, line 136: ### Two-pass traversal structure of graph_for undocumented as an API characteristic

**Severity:** Minor

**Problem**
graph_for internally calls definitions() then references(), each of which independently calls scan_records() and iterates the full artifact list. The docstring frames this only as a performance concern ('fine while unused') rather than describing it as a structural characteristic of the API (two separate iteration passes producing two independently-computed lists, not one unified pass with atomic def/ref classification). The architecture doc's 'CPT Reference Scan' section (traceability-validation.md ~lines 232-246) documents the algorithm's steps and output shape but does not mention the two-pass nature either.

<details>

**How to reproduce**
1. A detector author reads graph_for's docstring or the architecture doc to understand the traversal contract. 2. Neither source states that defined_in and referenced_in are computed via two separate full scans rather than a single unified traversal. 3. The author may assume atomicity/single-pass semantics that don't hold.

**Expected behavior**
Docs should explicitly state the two-pass design as a traversal characteristic (not just a performance note), so callers understand defined_in and referenced_in are independently computed.

**Actual behavior**
Only a performance-flavored aside exists in the code docstring; no mention in the architecture doc.

```text
graph_for -> definitions() [pass 1: full scan_records]; references() [pass 2: full scan_records] -> two independently-computed lists, structure undocumented as such
```

**Impact**
Low — mainly a documentation clarity gap that could mislead a future detector author about consistency guarantees between the two lists.

**Suggested correction**
Add a sentence to the docstring and/or the architecture doc explicitly stating the two-pass traversal structure as an API characteristic distinct from performance.

**How to verify**
Check docstring and architecture doc for explicit two-pass structural language.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Re-verified on a339061a. Resolved. The optionality is intentional and is now explicitly documented: graph_for is the detector-facing convenience API, where None means an empty source map and no source fields are emitted. definitions and references are command-facing projections; their direct call sites own and pass the resolved source map, and the required parameter deliberately fails loudly if that map is lost.

tests/test_cpt_reference_scan.py
SanjeevSolanki: Added `test_definitions_source_attached_per_artifact_across_multiple_artifacts` in 80b0d338 — the `definitions()` mirror of the references test: two artifacts, distinct sources (SRC_A/SRC_B), asserting each definition record keeps its own artifact's source. It's discriminating — a hoisted or first-path source lookup makes `defs[1]` fail. Both projections now have explicit multi-artifact source coverage; util suite 18 tests, 100% line coverage.

architecture/features/traceability-validation.md
SanjeevSolanki: Fixed in 80b0d338. Added a **Scope** line to the CPT Reference Scan section: `cpt-*` identifiers only — `PRD-NNN` requirement references out of scope (documented v1 limit), the same scope as `where-used` / `where-defined`. The scope now lives in the feature contract, not just the module docstring.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Fixed in 80b0d338 (also covers the two-pass structural point in the sibling comment). The `graph_for` docstring now states the two-pass design as an API characteristic — `defined_in` and `referenced_in` come from two independent `scan_records` passes, so there's no cross-list atomicity — and gives the explicit multi-id caller obligation: do NOT call `graph_for` per id in a loop (that's O(n_ids × n_artifacts) rescans); build one `scan_records` pass and partition locally. The feature doc's new Scope line carries the same two-pass / batching note.

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Fixed in 80b0d338. The `graph_for` docstring now frames the two-pass structure as an API characteristic, not a perf aside: `defined_in` and `referenced_in` are two independent `scan_records` passes computed separately, so callers must not rely on cross-list atomicity between them. The feature doc's new Scope line states the same ("two independent scans, one per side… no cross-list atomicity guarantee").

tests/test_cpt_reference_scan.py
SanjeevSolanki: Ran your verify criterion directly (mutation test): temporarily broke `_record` to hoist a single/first source — `src = next(iter(path_to_source.values()))` instead of the per-artifact `path_to_source.get(str(artifact_path))`. Against that broken impl both multi-artifact tests fail — `b.md`'s record wrongly gets `SRC_A`:

```
E  AssertionError: assert 'SRC_A' == 'SRC_B'
FAILED test_source_attached_per_artifact_across_multiple_artifacts
FAILED test_definitions_source_attached_per_artifact_across_multiple_artifacts
```

Reverting `_record` to the per-artifact lookup, both pass. So the tests catch exactly the source-mixing regression the finding described, on both the where-used and where-defined paths. (Mutation was working-tree-only; `80b0d338` is unchanged.)

ainetx: **RESOLVED** -- Confirmed fixed as of 80b0d33819ab8cdf5607aeb5b9cd0da7466c0638 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** tests/test_cpt_reference_scan.py has test_source_attached_per_artifact_across_multiple_artifacts (references) and test_definitions_source_attached_per_artifact_across_multiple_artifacts (definitions), both using two artifacts (a.md/b.md) with distinct path_to_source entries (SRC_A/SRC_B) and asserting each returned record carries its own artifact's source, not a reused or omitted mapping.
- **Reviewer:** Added multi-artifact tests supply a.md/b.md with distinct SRC_A/SRC_B mappings and assert the respective ordered references() and definitions() records retain each source.
- **Verifier:** The diff already includes test_source_attached_per_artifact_across_multiple_artifacts (references, two paths a.md/b.md with path_to_source={'a.md':'SRC_A','b.md':'SRC_B'}, asserting refs[0]['source']=='SRC_A' and refs[1]['source']=='SRC_B') and its mirror test_definitions_source_attached_per_artifact_across_multiple_artifacts for definitions() with the same two-artifact/distinct-source setup — both prove each locus keeps the source belonging to its own artifact, not a reused/omitted mapping, contradicting the premise that only single-artifact fixtures were used.

</details>

architecture/features/traceability-validation.md
ainetx: **RESOLVED** -- Confirmed fixed as of 80b0d33819ab8cdf5607aeb5b9cd0da7466c0638 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The traceability-validation.md diff adds a new 'CPT Reference Scan' section with an explicit 'Scope' line stating: '`cpt-*` identifiers only — `PRD-NNN` requirement references are out of scope (a documented v1 limit), the same scope as `where-used` / `where-defined`' — this directly cross-references and restates the util docstring's scope limitation in the doc, and explicitly ties it to where-used/where-defined, so a reader of the doc (which lists the four commands in the CLI table) is not left assuming PRD-NNN is covered.
- **Reviewer:** The new `### CPT Reference Scan` feature-doc section explicitly states the `cpt-*`-only scope and that `PRD-NNN` references are out of scope, and identifies it as the shared scan used by `where-used` / `where-defined`.
- **Verifier:** The new 'CPT Reference Scan' section in traceability-validation.md explicitly restates the scope limit ('Scope: cpt-* identifiers only — PRD-NNN requirement references are out of scope (a documented v1 limit), the same scope as where-used / where-defined'), so it is cross-linked from the feature doc, not buried solely in the util docstring; a reader of the doc's CLI table who follows through to this new section would see it stated.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of 80b0d33819ab8cdf5607aeb5b9cd0da7466c0638 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Both the code and the doc already give the caller-facing cost contract the check asks for: graph_for's docstring states 'Two-pass by construction: defined_in and referenced_in come from two independent scan_records passes... a caller that needs many ids must NOT call graph_for per id in a loop -- that is O(n_ids × n_artifacts) rescans of the whole tree. Build a single scan_records pass and partition it locally instead,' and architecture/features/traceability-validation.md's CPT Reference Scan section mirrors it verbatim: 'a caller that needs many ids should build a single scan pass and partition it locally rather than call the graph per id.' This is an explicit bound (O(n_ids × n_artifacts)) plus a concrete caller obligation (batch/partition instead of looping), not merely a note that it's 'fine while unused' -- so the premise that no usable contract exists is incorrect.
- **Reviewer:** `graph_for` explicitly documents that per-id looping causes `O(n_ids × n_artifacts)` full-tree rescans and directs callers needing many IDs to use one `scan_records` pass and partition locally; the feature spec also defers artifact-quality command scale bounds to its later task.
- **Verifier:** The graph_for docstring already states a caller obligation, not just a 'fine while unused' note: it explicitly says defined_in/referenced_in come from two independent scan_records passes (no cross-list atomicity), and that 'a caller that needs many ids must NOT call graph_for per id in a loop — that is O(n_ids × n_artifacts) rescans of the whole tree. Build a single scan_records pass and partition it locally instead.' That is a concrete, usable cost contract (name the complexity class, name the required alternative pattern) for the future artifact-quality detectors to design against, even though no detector exists yet to consume it.

</details>

skills/studio/scripts/studio/utils/cpt_reference_scan.py
ainetx: **RESOLVED** -- Confirmed fixed as of 80b0d33819ab8cdf5607aeb5b9cd0da7466c0638 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Both the function docstring and the architecture doc explicitly name the two-pass design as a traversal/API characteristic, not merely a performance caveat. The docstring states: 'Two-pass by construction: `defined_in` and `referenced_in` come from two independent `scan_records` passes, so they are computed separately — do not rely on cross-list atomicity between them,' and separately warns against O(n_ids × n_artifacts) rescans if called per-id in a loop (the performance angle, documented distinctly). The architecture doc (traceability-validation.md, 'CPT Reference Scan' section) mirrors this: 'The def↔ref graph is assembled from two independent scans (one per side), so `defined_in` and `referenced_in` carry no cross-list atomicity guarantee; a caller that needs many ids should build a single scan pass and partition it locally rather than call the graph per id.' This documents the two-pass shape itself (two independent scans, no atomicity) as an explicit contract a future detector must account for, separate from the raw performance cost.
- **Reviewer:** Both the architecture section and graph_for docstring explicitly state that defined_in and referenced_in are assembled/computed from two independent scans/passes, including the resulting lack of cross-list atomicity.
- **Verifier:** Both the architecture doc and the function docstring explicitly describe the two-pass traversal as an API characteristic, not just a performance caveat. traceability-validation.md states: 'The def↔ref graph is assembled from two independent scans (one per side), so defined_in and referenced_in carry no cross-list atomicity guarantee.' The graph_for docstring in cpt_reference_scan.py repeats this: 'Two-pass by construction: defined_in and referenced_in come from two independent scan_records passes, so they are computed separately — do not rely on cross-list atomicity between them.' This directly names the two-pass structure as a traversal contract a caller/detector must respect, distinct from the separate performance note about O(n_ids × n_artifacts) rescans.

</details>

ainetx:
### artifacts_meta.py still hand-rolls its own scan_cpt_ids loop, unmigrated to the new shared util

**Severity:** Minor

**Problem**
The refactor's stated purpose is to deduplicate 'the one scan_cpt_ids traversal... each re-implemented' across callers, and it migrates list_ids.py, list_id_kinds.py, where_defined.py, and where_used.py to cpt_reference_scan. But skills/studio/scripts/studio/utils/artifacts_meta.py's _collect_def_ids_from_artifacts function independently imports scan_cpt_ids from.document and loops per-artifact building its own definition-id collection, with its own try/except handling — a fifth caller with a hand-rolled loop that diverges from the new canonical util and does not inherit its documented cpt-id-scoped v1 limit or its projection guarantees.

<details>

**How to reproduce**
1. Open skills/studio/scripts/studio/utils/artifacts_meta.py. 2. Find _collect_def_ids_from_artifacts, which imports scan_cpt_ids directly ('from.document import scan...') and iterates artifacts building all_def_ids itself, rather than calling cpt_reference_scan.definitions() or scan_records(). 3. Compare to the four callers refactored in this diff, all of which now route through cpt_reference_scan.

**Expected behavior**
Either this caller is migrated to use cpt_reference_scan's shared scan/projection functions for consistency with the refactor's stated goal, or the PR/doc explicitly calls out this caller as intentionally out of scope for the current migration.

**Actual behavior**
artifacts_meta.py retains its own independent scan_cpt_ids loop with separate error-handling and filtering logic, undocumented as an intentional exception to the refactor.

```text
scan_cpt_ids -> [migrated: list_ids, list_id_kinds, where_defined, where_used via cpt_reference_scan] 
-> [NOT migrated: artifacts_meta._collect_def_ids_from_artifacts (own loop)]
```

**Impact**
Future changes to cpt-id scanning semantics (e.g. tightening the v1 cpt-only scope, or fixing a scan edge case) must be applied in two places; this caller can silently drift from the new canonical behavior.

**Suggested correction**
Migrate _collect_def_ids_from_artifacts to use cpt_reference_scan.scan_records or a definitions-only projection, or add an explicit code comment / doc note stating why it is intentionally excluded from this refactor.

**How to verify**
Grep the repo for remaining direct 'scan_cpt_ids(' call sites outside cpt_reference_scan.py and the four refactored command modules; confirm none remain (or that remaining ones are documented as intentional).

</details>

---
_Original location: skills/studio/scripts/studio/utils/artifacts_meta.py:1 -- inline anchoring could not be resolved after 1 attempt(s)._

skills/studio/scripts/studio/utils/cpt_reference_scan.py
```diff
@@ -0,0 +1,145 @@
...
+
+    ``defined_in`` is the definition loci (no ``type`` key); ``referenced_in`` is the
+    non-definition reference loci (each with its ``type``). Built with the same
+    ``references`` / ``definitions`` logic the query commands use, so it classifies
+    definitions and references identically to them.
+
+    Two-pass by construction: ``defined_in`` and ``referenced_in`` come from two
+    independent ``scan_records`` passes, so they are computed separately — do not rely
+    on cross-list atomicity between them. That is fine for one id while the view is
+    unused, but a caller that needs many ids must NOT call ``graph_for`` per id in a
```
ainetx, line 128: ### graph_for's per-id-loop warning has no corresponding batch API or example

**Severity:** Minor

**Problem**
The graph_for docstring instructs callers needing many ids to 'build a single scan_records pass and partition it locally' instead of calling graph_for per id, but no such batch/partition helper function exists in cpt_reference_scan.py, and neither the docstring nor the '### CPT Reference Scan' doc section shows a code example of this pattern.

<details>

**How to reproduce**
1. A future detector needs def/ref graphs for N cpt-ids. 2. Developer reads graph_for's docstring warning against looping it per id. 3. Developer looks for a suggested alternative API in cpt_reference_scan.py — finds none. 4. Under time pressure, developer calls graph_for(id) in a loop anyway, since that's the only working example available.

**Expected behavior**
Either a batch helper (e.g. `graphs_for(ids, artifacts_to_scan,...)`) that performs one scan_records pass and partitions per id, or a code-level example in the docstring/doc showing how to do the single-pass partition manually.

**Actual behavior**
Only the four single-id functions (scan_records, references, definitions, graph_for) exist; the warning is prose-only with no accompanying implementation or example.

```text
N detectors need id graphs --> read warning 'don't loop graph_for' --> no batch helper found --> loop graph_for anyway --> O(n_ids x n_artifacts) rescans reintroduced
```

**Impact**
The first multi-id detector implementation is likely to reintroduce the exact performance issue the comment warns against, since there's no ready-made correct alternative to reach for.

**Suggested correction**
Add a `graphs_for(ids, artifacts_to_scan, path_to_source=None)` helper that does one scan_records pass and partitions results per id, or add a short code example to the docstring/doc showing the manual single-pass partition.

**How to verify**
Check that a future detector consuming multiple ids calls a batch API (or the documented manual pattern) rather than graph_for in a loop; add a test asserting scan_cpt_ids is called only once per artifact regardless of id count.

</details>

tests/test_cpt_reference_scan.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

ainetx (approved):
Nice cleanup — the new `cpt_reference_scan` util cleanly consolidates the shared scan-and-project logic, the migrated commands (`list_ids`, `list_id_kinds`, `where_defined`, `where_used`) preserve their original output exactly, and the added test coverage (including error-propagation and property tests) gives good confidence in the projections and the `graph_for` def↔ref view. Nothing here blocks merging — just a few loose ends worth a glance before the migration is considered fully done:

- **`WorkspaceContext.get_all_artifact_ids` still hand-rolls its own scan loop** instead of using the new shared util, leaving one caller un-migrated. ([comment](https://github.com/constructorfabric/studio/pull/165#issuecomment-5597626213))
- **`artifacts_meta.py`'s `_collect_def_ids_from_artifacts` also still hand-rolls its own `scan_cpt_ids` loop**, a fifth caller that diverges from the new canonical util. ([comment](https://github.com/constructorfabric/studio/pull/165#issuecomment-5598751640))
- **`graph_for`'s docstring warns against per-id looping but offers no batch/partition helper or example** to show callers the recommended pattern. ([comment](https://github.com/constructorfabric/studio/pull/165#discussion_r3966221603))

skills/studio/scripts/studio/utils/cpt_reference_scan.py
SanjeevSolanki: Declining the batch helper as speculative (YAGNI). There's no multi-id consumer yet — `graph_for` itself is unused-until-a-detector-lands (whitelisted with a removal trigger), and adding a second unused `graphs_for` API with its own CPT markers, tests and whitelist entry is net-new surface this dedup PR shouldn't carry.

Also, the guidance doesn't point at a missing primitive: the single-pass function it refers to — `scan_records` — is already public and exported, so a multi-id caller can do the one-pass partition over it today (iterate `scan_records` once, bucket by `hit['id']`). The warning steers callers to that existing primitive, not to a function that doesn't exist.

When the first multi-id detector lands (`cpt-studio-flow-artifact-quality-assess`), a dedicated `graphs_for` — with the single-scan-count test you suggest — is the right home for it, built against a real consumer and tracked with the same removal-trigger discipline as `graph_for`, rather than added speculatively now.

ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The finding should be dismissed: `scan_records(artifacts_to_scan)` is an existing public single-pass primitive, and `graph_for` explicitly tells multi-id callers to use that function and partition its hits locally. A separate `graphs_for` API or an example is not required to make the warning actionable, while `graph_for` is visibly unused and deliberately whitelisted pending its first detector consumer.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a per-call-in-loop pattern as needing a batch API or example, verify whether an existing batch-capable primitive already covers the use case and whether the flagged function is documented as unused/pending its first real caller — a whitelisted, not-yet-consumed helper with a documented single-pass alternative is not a missing-batch-API defect."

</details>

SanjeevSolanki:
On the two un-migrated `scan_cpt_ids` callers flagged in the approval — I looked at each and I'm **declining both migrations**. It's a behaviour reason, not style: both are **graceful-degradation collectors**, and the new `scan_records` is deliberately **fail-loud**, so routing them through it would change error semantics.

**1. `WorkspaceContext.get_all_artifact_ids` (via `_scan_definition_ids`)** — its own algo `cpt-studio-algo-workspace-collect-ids`, wired to the DoD `cpt-studio-dod-workspace-graceful-degradation`. It wraps each `scan_cpt_ids` in `try/except (OSError, ValueError)` and logs-and-continues, so one broken/unreadable artifact (common on the cross-repo path it also handles) doesn't abort collecting ids from the rest. It also spans primary meta **+ remote SourceContexts** and dedups into a `Set`, not the `[(path, artifact_type)]` shape `scan_records` takes. Migrating it would flip graceful degradation into a hard failure — breaking that DoD — a behaviour change this behaviour-identical dedup PR can't make.

**2. `artifacts_meta._collect_def_ids_from_artifacts`** — same contract: catches `(OSError, ValueError)` per artifact, appends to its `errors` list, and continues. `scan_records` propagates the first `OSError` (that was finding #2 on this PR, now tested). To use it here you'd wrap each artifact in its own try/except around a single-artifact `scan_records` call — pure indirection over the direct `scan_cpt_ids` call, sharing nothing. Input is also `List[Artifact]` + a `resolve_path_fn`.

So both are genuinely different traversals with the opposite error contract from the shared util, not un-migrated duplicates. The four query commands (`list-ids`, `list-id-kinds`, `where-defined`, `where-used`) that DO share the cpt-id-scoped, fail-loud def↔ref query are the correct and complete migration scope for this PR.

Thanks for the thorough review!

Issue #145, opened 2026-09-04
Token Hygiene — Make the ASCII-Art Skill Invocation Picture Opt-In


vasylcf:
# Token Hygiene — Make the ASCII-Art Skill Invocation Picture Opt-In

## Context

Studio renders a generated ASCII picture at the start of every `cf`, `cf-studio` and `cf-*` workflow entry. The unit is `SkillInvocationArt` in `skills/studio/modules/ui/skill-invocation-art.md`. It is loaded from `skills/studio/SKILL.md`, `skills/studio/modules/runtime/workflow-bootstrap.md`, `workflows/studio.md` and `workflows/kit.md`, and it runs unconditionally on every workflow bootstrap.

The unit's own rules budget the picture at up to 12 lines by 60 columns, plus a label line, on every entry. This is model-generated output, not a static string: it is drawn afresh each time, so it costs generation tokens on every workflow entry and cannot be served from cache.

Users pay per token. Decoration that is regenerated on every entry is a recurring charge for output nobody asked for, on a path that runs every single time Studio is used.

The feature is not the problem — the default is. The picture was added deliberately, to give the tool some character, and users who want it should keep it. What does not hold up is charging every user for it on every entry with no way to decline.

---

## Functional Requirements

* **Configurable:** Skill invocation art can be turned off without editing prompt modules.
* **Default off:** The picture is not rendered unless it is enabled. A setting that defaults to on leaves the cost with everyone who never finds the switch.
* **Single switch:** One setting covers every entry point that renders the picture (`cf`, `cf-studio`, `cf-*`, kit workflows), not one setting per workflow.
* **No control-flow impact:** Enabling or disabling changes only the picture — exactly as `SkillInvocationArt` already requires. No change to load reports, menus, gates, or any other output.
* **Discoverable:** The setting is documented in the configuration guide together with its token rationale, so a team enabling it knows what it costs.
* **Audit the same class:** Identify other purely decorative model-generated output on the workflow hot path and bring it under the same rule. Deterministic, script-emitted formatting is out of scope — this is about output the model is asked to invent.

---

## Non-Functional Requirements

| Quality Attribute | Requirement |
| --- | --- |
| **Efficiency** | A workflow entry with the picture disabled generates no picture tokens at all. |
| **Non-regression** | Load reports, menus and workflow output are unchanged with the picture off, apart from the picture's absence. |
| **Simplicity** | One setting, one default, no per-workflow configuration. |

---

## Acceptance Criteria

[...]

Issue #142, opened 2026-09-04
Studio CLI — Raise Blocking Questions Through the Harness's Native Question Dialog


vasylcf:
# Studio CLI — Raise Blocking Questions Through the Harness's Native Question Dialog

## Context

Studio asks the user through PDSL's `EMIT_MENU` / `EMIT` primitives, which render as text inside the assistant's message. Nothing in the prompt corpus reaches for a harness question affordance: there are zero references to `AskUserQuestion` or any equivalent across skills, modules and workflows.

The consequence is that a blocking question is indistinguishable from a finished turn. Both end the same way — a message, then silence. To find out which happened, the user has to read the message body, locate the question inside it (it is not reliably the last thing in the turn), and work out what answer is expected. Nothing in the interface says "you are being waited on".

Modern harnesses expose a question or menu affordance that makes waiting unambiguous: a distinct interactive element with selectable options. Studio does not use it, so it forfeits the one signal the user does not have to read prose to receive.

Naming a single tool is not a sufficient fix. Harnesses name the affordance differently, and a hardcoded tool name only works where that exact name is defined. The naming is similar enough across major harnesses that a per-target binding plus an intent description covers most of the field, and Studio already generates per-harness integration files for `claude`, `codex`, `cursor` and `copilot` via `cfs generate-agents` — so a place to carry per-target conventions already exists.

This is a distinct problem from question volume. #128 reduces how often Studio asks. This issue is about the asks that remain being impossible to miss. Neither substitutes for the other: reducing the count of invisible questions still leaves invisible questions.

---

## Functional Requirements

* **Native dialog for blocking questions:** When Studio blocks on a user answer, it requests the harness's own question or menu affordance rather than emitting the question as prose.
* **Per-harness binding:** The tool name and invocation convention is resolved per generation target (`claude`, `codex`, `cursor`, `copilot`, and others we support), not hardcoded to one harness's name.
* **Description-based fallback:** Where the tool name differs or is unknown, the instruction describes the intent so a harness exposing an equivalent affordance can still match it.
* **Graceful degradation:** On a harness with no question affordance, Studio falls back to text rendering, and the question remains clearly marked as blocking and appears in a predictable position — at the end of the turn.
* **Waiting state is unambiguous:** A turn that ends on a blocking question is visibly distinct from a turn that ends on completed work, on every harness.
* **Applies to existing gates:** Menus currently rendered through `EMIT_MENU` route through the same mechanism, so the change is not limited to newly written prompts.
[...]

PR #166, opened 2026-09-09
feat(ui): make skill-invocation ASCII art default-off (#145)

tkcoding:
## Summary
- Gates `SkillInvocationArt` behind a new `[ui].skill_invocation_art_enabled` flag in `core.toml`, default `false` — the decorative ASCII picture at cf/cf-* skill entry now costs nothing unless a project explicitly opts in.
- Adds the corresponding `ui.skill_invocation_art_enabled` boolean to `schemas/core-config.schema.json`.
- Resolves the flag once per session using the existing `unset`-sentinel `STATE` idiom (mirrors `SimpleModeGate`'s `SIMPLE_MODE`), avoiding a bug where "not yet resolved" and "resolved false" would otherwise collapse into the same value and permanently disable the config read.

Addresses #145. #142 (native blocking-question dialog) is tracked separately and not part of this PR.

## Test plan
- [x] `cfs pdsl validate skills/studio/modules/ui/skill-invocation-art.md` — confirmed 0 new findings vs. the pre-change baseline (6 pre-existing findings in untouched units remain, unrelated to this diff)
- [x] `schemas/core-config.schema.json` re-validated as well-formed JSON
- [x] Confirmed `core.toml` is loaded as a generic untyped dict throughout the Python codebase, so the new schema key won't be silently dropped by any existing loader
- [ ] Manual harness check: flip `[ui].skill_invocation_art_enabled = true` in a project's `core.toml` and confirm the picture appears; confirm it's absent by default

## Summary by CodeRabbit

- **New Features**
  - Added an optional UI setting to control decorative ASCII art when entering `cf`, `cf-studio`, and other `cf-*` skills or workflows.
  - When enabled, relevant ASCII art and a label are displayed at entry.
  - The setting is disabled by default and can be configured under the UI preferences section of the core configuration.
  - Configuration now supports UI preferences for both skill and workflow runs.

### DIFF
--- a/skills/studio/modules/ui/skill-invocation-art.md
+++ b/skills/studio/modules/ui/skill-invocation-art.md
@@ -9,7 +9,7 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at 
 
 ```pdsl
 UNIT SkillInvocationArt
-PURPOSE: When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
+PURPOSE: When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill or workflow entry with one small ASCII-art picture relevant to the entry name, with a plain-text label below, without changing the workflow's control flow.
 STATE:
   SET SKILL_INVOCATION_ART_ENABLED: true | false | unset (default unset, scope session)
 WHEN:
--- a/skills/studio/modules/ui/skill-invocation-art.md
+++ b/skills/studio/modules/ui/skill-invocation-art.md
@@ -9,7 +9,7 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at 
 
 ```pdsl
 UNIT SkillInvocationArt
-PURPOSE: Prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
+PURPOSE: When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
 STATE:
   SET SKILL_INVOCATION_ART_ENABLED: true | false | unset (default unset, scope session)
 WHEN:
--- a/skills/studio/modules/ui/skill-invocation-art.md
+++ b/skills/studio/modules/ui/skill-invocation-art.md
@@ -10,13 +10,17 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at 
 ```pdsl
 UNIT SkillInvocationArt
 PURPOSE: Prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
+STATE:
+  SET SKILL_INVOCATION_ART_ENABLED: true | false | unset (default unset, scope session)
 WHEN:
   REQUIRE a cf, cf-studio, or cf-* skill or workflow entry is beginning execution
   SKIP silently (no picture, no output) WHEN this unit is loaded in a context that does not satisfy the above REQUIRE
 DO:
-  RUN SkillInvocationArtGuard
+  RUN resolve SKILL_INVOCATION_ART_ENABLED from `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml` (false when the key or file is absent) WHEN SKILL_INVOCATION_ART_ENABLED == unset
+  RUN SkillInvocationArtGuard WHEN SKILL_INVOCATION_ART_ENABLED == true
   RUN SkillInvocationArtGenerate WHEN SkillInvocationArtGuard passes
 RULES:
+  ALWAYS default to disabled and resolve the config flag at most once per session: skip the picture with no further state read unless `[ui].skill_invocation_art_enabled` is explicitly `true` in `{cf-studio-path}/config/core.toml`
   ALWAYS run this unit once at the start of every cf, cf-studio, or cf-* workflow bootstrap or alias entry, before the workflow's first normal EMIT, EMIT_MENU, WAIT, CONTINUE, INVOKE, DISPATCH, RETURN, or STOP_TURN
   NEVER alter, delay, or suppress any existing output directive; the picture precedes but does not replace or reorder normal output
   NEVER replace, delay, reorder, suppress, or alter any load report
--- a/schemas/core-config.schema.json
+++ b/schemas/core-config.schema.json
@@ -57,7 +57,7 @@
     },
     "ui": {
       "$ref": "#/$defs/ui",
-      "description": "Optional UI/output preferences for interactive skill and workflow runs."
+      "description": "Optional UI/output preferences for skill and workflow runs."
     }
   },
   "$defs": {
--- a/schemas/core-config.schema.json
+++ b/schemas/core-config.schema.json
@@ -54,6 +54,10 @@
     "paths": {
       "$ref": "#/$defs/paths",
       "description": "Optional path overrides for non-standard layouts."
+    },
+    "ui": {
+      "$ref": "#/$defs/ui",
+      "description": "Optional UI/output preferences for interactive skill and workflow runs."
     }
   },
   "$defs": {
@@ -197,6 +201,17 @@
           "type": "string",
           "minLength": 1,
           "description": "Relative path to the studio core directory. Used when the core is not at the default location."
+        }
+      }
+    },
+    "ui": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "skill_invocation_art_enabled": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether to draw the decorative ASCII-art picture (plus label) at cf/cf-* skill and workflow entry. Default-off: the picture costs nothing unless explicitly enabled."
         }
       }
     }


skills/studio/modules/ui/skill-invocation-art.md
```diff
@@ -10,13 +10,17 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at
 ```pdsl
 UNIT SkillInvocationArt
 PURPOSE: Prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
+STATE:
+  SET SKILL_INVOCATION_ART_ENABLED: true | false | unset (default unset, scope session)
 WHEN:
   REQUIRE a cf, cf-studio, or cf-* skill or workflow entry is beginning execution
```
ainetx, line 16: ### Malformed core configuration has no disabled fallback

**Severity:** Major

**Problem**
The configuration-resolution instruction does not specify how an unreadable or malformed `config/core.toml` resolves the session flag or preserves workflow execution.

<details>

**How to reproduce**
1. Create an invalid or unreadable `{cf-studio-path}/config/core.toml`.
2. Begin a qualifying cf entry while the state is unset.
3. Execute the resolution instruction.

**Expected behavior**
A read or parse failure sets `SKILL_INVOCATION_ART_ENABLED` to false and continues without art.

**Actual behavior**
No PDSL branch defines that fallback or guarantees continuation.

```text
invalid or unreadable core.toml
|
v
resolve flag instruction
|
v
no defined disabled fallback
```

**Impact**
An optional decorative-output setting can leave entry behavior undefined or interrupt a workflow.

**Suggested correction**
Explicitly map configuration read and parse failures to `SKILL_INVOCATION_ART_ENABLED = false`, without output or workflow interruption.

**How to verify**
Confirm absent, unreadable, and invalid configuration each resolve false and allow a qualifying workflow to proceed.

</details>

skills/studio/modules/ui/skill-invocation-art.md
```diff
@@ -10,13 +10,17 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at
...
   SKIP silently (no picture, no output) WHEN this unit is loaded in a context that does not satisfy the above REQUIRE
 DO:
-  RUN SkillInvocationArtGuard
+  RUN resolve SKILL_INVOCATION_ART_ENABLED from `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml` (false when the key or file is absent) WHEN SKILL_INVOCATION_ART_ENABLED == unset
+  RUN SkillInvocationArtGuard WHEN SKILL_INVOCATION_ART_ENABLED == true
```
ainetx, line 20: ### Interactive-only scoping in schema description is not enforced by the PDSL predicate

**Severity:** Minor

**Problem**
schemas/core-config.schema.json describes `skill_invocation_art_enabled` as applying to 'interactive skill and workflow runs,' but skill-invocation-art.md's WHEN/DO/RULES gate only on config state and entry type, never on interactive-vs-non-interactive or output-mode.

<details>

**How to reproduce**
1. Set `[ui].skill_invocation_art_enabled = true`. 2. Invoke a cf/cf-* skill/workflow in a non-interactive or machine-readable/automation context. 3. Observe SkillInvocationArtGuard/Generate still runs because no interactive predicate exists.

**Expected behavior**
The unit should check some interactive/output-mode signal before rendering, matching the schema's stated scope, or the schema description should not claim a restriction the code doesn't enforce.

**Actual behavior**
The picture renders for every qualifying entry once enabled, with no mode distinction.

```text
schema: 'interactive runs only' --describes--> PDSL: renders for ALL qualifying entries (no mode check)
```

**Impact**
Decorative ASCII art could leak into scripted/automation or machine-readable invocations once the opt-in is enabled, contradicting the documented scope.

**Suggested correction**
Either add an interactive/output-mode guard to SkillInvocationArt's WHEN/DO, or narrow the schema description to not claim interactive-only scoping.

**How to verify**
Enable the flag and invoke a cf/cf-* entry from a non-interactive/automation context; confirm no art is emitted (post-fix) or that description no longer implies such scoping.

</details>

skills/studio/modules/ui/skill-invocation-art.md
```diff
@@ -10,13 +10,17 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at
 ```pdsl
 UNIT SkillInvocationArt
 PURPOSE: Prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
```
ainetx: ### PURPOSE line not updated to reflect new default-off, config-gated behavior

**Severity:** Minor

**Problem**
The PURPOSE sentence was left unmodified by this diff and still asserts unconditional picture rendering at every qualifying entry, while STATE/DO/RULES below now gate rendering behind an explicit `[ui].skill_invocation_art_enabled = true` opt-in that defaults to disabled.

<details>

**How to reproduce**
1. Read the unit's PURPOSE line. 2. Read STATE (SKILL_INVOCATION_ART_ENABLED default unset) and RULES ('ALWAYS default to disabled... unless explicitly true'). 3. Observe the two disagree on whether the picture is unconditional or opt-in.

**Expected behavior**
PURPOSE should note the picture only renders when explicitly enabled via config (e.g. 'when enabled via `[ui].skill_invocation_art_enabled`').

**Actual behavior**
PURPOSE promises the picture unconditionally for every matching entry.

```text
PURPOSE (unconditional) --contradicts--> STATE/RULES (opt-in, default-off)
```

**Impact**
A maintainer skimming only PURPOSE will misunderstand the unit's real, gated behavior, risking incorrect edits or assumptions elsewhere.

**Suggested correction**
Update the PURPOSE sentence to mention the default-off/config-gated nature, e.g. append 'when enabled via config'.

**How to verify**
Re-read PURPOSE alongside STATE/RULES and confirm they describe consistent behavior.

</details>

skills/studio/modules/ui/skill-invocation-art.md
ainetx: **RESOLVED** -- Confirmed fixed as of b80aedb6d21b0749276077831bed3143c088196f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The RULES clause states the fallback as an inclusive gate rather than an absence-only check: 'skip the picture with no further state read unless [ui].skill_invocation_art_enabled is explicitly true' — any state other than a literal parsed `true` (missing file, missing key, unreadable file, malformed TOML) falls through to the default-disabled/skip path with no art emitted and no workflow interruption, since only the single positive condition is checked.
- **Verifier:** The RULES line added in the diff ("ALWAYS default to disabled... skip the picture with no further state read unless `[ui].skill_invocation_art_enabled` is explicitly `true`") is a general fail-safe keyed on the absence of a confirmed `true`, not merely on file absence — an unreadable or malformed config file cannot yield an explicit `true`, so it falls under the same default-disabled/skip branch with no interruption of the invoking workflow, satisfying the same safety property the check asks for even though the word 'malformed' isn't used verbatim.

</details>

skills/studio/modules/ui/skill-invocation-art.md
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): schemas/core-config.schema.json describes `ui.skill_invocation_art_enabled` as applying to 'interactive skill and workflow runs,' implying it should not fire for non-interactive/automation invocations. However skill-invocation-art.md's WHEN/DO/RULES blocks gate rendering only on SKILL_INVOCATION_ART_ENABLED (resolved from config) and SkillInvocationArtGuard, with no check for interactive vs. non-interactive or machine-readable/JSON output mode.

Problem (now): schemas/core-config.schema.json describes `skill_invocation_art_enabled` as applying to 'interactive skill and workflow runs,' but skill-invocation-art.md's WHEN/DO/RULES gate only on config state and entry type, never on interactive-vs-non-interactive or output-mode.

</details>

skills/studio/modules/ui/skill-invocation-art.md
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The top-level PURPOSE says every cf entry is prefixed with ASCII art, while absent configuration resolves the new flag to false and prevents art generation.

Problem (now): The PURPOSE sentence was left unmodified by this diff and still asserts unconditional picture rendering at every qualifying entry, while STATE/DO/RULES below now gate rendering behind an explicit `[ui].skill_invocation_art_enabled = true` opt-in that defaults to disabled.

</details>

ainetx (approved):
Nice fix — gating the ASCII-art picture behind an explicit, default-off `[ui].skill_invocation_art_enabled` config is sound, and the once-per-session resolution logic is clear. Approving with two small documentation-consistency notes worth a follow-up pass:

- **Interactive-only scoping isn't actually enforced**: the schema description says the setting applies to "interactive skill and workflow runs," but the PDSL predicate in `skill-invocation-art.md` only gates on config state and entry type — there's no interactive-vs-non-interactive or output-mode check. Worth aligning the description with the real gate, or adding the check. ([comment](https://github.com/constructorfabric/studio/pull/166#discussion_r3965161124))
- **Stale PURPOSE line**: the PURPOSE sentence still reads as if the picture renders unconditionally at every qualifying entry, but STATE/DO/RULES now make that conditional on the new opt-in flag (default disabled). Just needs a wording touch-up to match current behavior. ([comment](https://github.com/constructorfabric/studio/pull/166#discussion_r3965299998))

ainetx:
@tkcoding please sign commit with DCO

skills/studio/modules/ui/skill-invocation-art.md
tkcoding: Fixed in 8d071e9b — narrowed the schema description to drop the unenforced "interactive skill and workflow runs" claim (now just "skill and workflow runs") since no interactive/output-mode check exists anywhere in this gate. Went with narrowing the description over adding a new interactive-mode predicate, since that concept isn't modeled anywhere else in this codebase's PDSL and would be scope creep for this fix.

skills/studio/modules/ui/skill-invocation-art.md
tkcoding: Fixed in 8d071e9b (and further tightened in 617c4d38 per CodeRabbit's follow-up) — PURPOSE now reads "When enabled via `[ui].skill_invocation_art_enabled` ... prefix each cf, cf-studio, or cf-* skill or workflow entry ...", matching the opt-in, config-gated behavior in STATE/DO/RULES.

skills/studio/modules/ui/skill-invocation-art.md
```diff
@@ -9,14 +9,18 @@ purpose: Defines SkillInvocationArt — the ASCII-art entry picture rendered at
 
 ```pdsl
 UNIT SkillInvocationArt
-PURPOSE: Prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
+PURPOSE: When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill entry with one small ASCII-art picture relevant to the skill name, with a plain-text label below, without changing the workflow's control flow.
```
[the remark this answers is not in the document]
tkcoding: Good catch, fixed in 617c4d38 — applied your suggested wording exactly (skill or workflow entry / entry name).

tkcoding:
@ainetx Done — rewrote both commits to include the `Signed-off-by:` trailer and force-pushed. DCO check is now passing.

skills/studio/modules/ui/skill-invocation-art.md
ainetx: **RESOLVED** -- Confirmed fixed as of 617c4d38fb91fe5b8f8e77e20bc3183c7970754e -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff's schema description reads only "Optional UI/output preferences for skill and workflow runs" and "Whether to draw the decorative ASCII-art picture... at cf/cf-* skill and workflow entry" -- no mention of 'interactive' anywhere in schemas/core-config.schema.json or the PDSL PURPOSE/description in skill-invocation-art.md. This matches the discussion thread's resolution (commit 8d071e9b) which narrowed the schema wording to drop the unenforced 'interactive skill and workflow runs' claim precisely because no interactive/output-mode predicate exists anywhere in this PDSL gate. Since the schema no longer scopes the setting to interactive runs, there is no discrepancy between the schema's claim and the PDSL's enforcement to flag.
- **Reviewer:** The current schema description no longer scopes the flag to interactive runs; it simply enables decorative output at cf/cf-* skill and workflow entry, matching the PDSL's config-and-entry gates.
- **Verifier:** The current schema diff no longer contains any 'interactive' scoping claim — both the `ui` object description ('Optional UI/output preferences for skill and workflow runs.') and `skill_invocation_art_enabled`'s description ('...at cf/cf-* skill and workflow entry. Default-off...') describe unscoped skill/workflow entry behavior, matching the PDSL's actual entry-type-only gating. This was fixed in commit 8d071e9b per the discussion thread, which narrowed the description instead of adding an interactive-mode predicate; the mismatch the check describes no longer exists in this diff.

</details>

skills/studio/modules/ui/skill-invocation-art.md
ainetx: **RESOLVED** -- Confirmed fixed as of 617c4d38fb91fe5b8f8e77e20bc3183c7970754e -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff under review already updates the PURPOSE line to 'When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill or workflow entry with one small ASCII-art picture...', which matches the opt-in, default-off, config-gated behavior described in the new STATE/DO/RULES sections — the documentation-consistency gap the check describes has been resolved.
- **Reviewer:** The diff updates PURPOSE to explicitly say "When enabled via `[ui].skill_invocation_art_enabled`", consistent with the default-disabled STATE, DO, and RULES.
- **Verifier:** The diff's own PURPOSE line reads 'When enabled via `[ui].skill_invocation_art_enabled` in `{cf-studio-path}/config/core.toml`, prefix each cf, cf-studio, or cf-* skill or workflow entry with one small ASCII-art picture...', which already matches the opt-in, default-off gating described in STATE/DO/RULES — the previously flagged gap is resolved in the current code under review.

</details>

PR #161, opened 2026-09-08
cf-ux: let the cf skill execute under claude -p, and refuse to score a run where it did not

Oleg67:
Fixes #146.

The Claude provider invoked `claude -p "/cf <prompt>"` with no permission flag. Skill *execution* asks for permission and print mode has nobody to ask, so the skill was denied, the agent answered the request directly, and the answer was plausible enough for the shared rubric to pass it. The suite then reported confidently on an agent that never loaded Studio.

The codex provider never had this problem — it has always passed the equivalent pair, `--sandbox workspace-write` and `approval_policy="never"`. The asymmetry was the bug.

## Three changes

**1. `--permission-mode bypassPermissions`.** Safe in this context: every invocation runs in a fresh directory under the system temp dir which `_sandbox` wipes in `finally`, on `atexit`, and on SIGTERM/SIGINT/SIGHUP. `CF_UX_SHARED_SANDBOX` is the exception — it writes where the caller points it — and the README now warns about that explicitly, since it is the one path where this flag has real blast radius.

**2. Positive skill detection from the tool-call trace.** `--output-format stream-json --verbose`, then a `Skill` tool-use event whose input names `cf`, with a non-error `tool_result` bound to *that call's* id. Review round 2 tightened every clause of that sentence — see below.

The guard this replaces was:

```python
"skill_load_warning": "skills failed to load" in output_text.lower(),
```

The CLI never emits that string, so the one check meant to catch this could not fire. The real signature is `<error>Execute skill: cf</error>`, which is now recognised as a failure too. The file's own comment already conceded the point — *"Heuristic — actual skill-loading detection refined later via stream-json"* — and the README listed this parsing under "Next steps"; #146 is the argument for why it was not optional.

**3. A run that did not load the skill returns a promptfoo error, not a metadata flag.** This is the one that keeps the defect from returning. A fallback answer never reaches the grader — it is kept under `unscored_output` for diagnosis instead. A transcript with no terminal `result` event is an error as well: there is then no answer to grade and no trace to trust.

`skill_state` is `ran` / `failed` / `absent`, and `skills_invoked` records what actually ran, so a failure says which.

## Review round 2 (`4067cacf`)

Thirteen findings on `422deade`. Twelve fixed, one declined with reasoning. Three were `Major`, and all three were the same shape: the positive check could still say `ran` for a run that did not run the cf skill.

| # | Finding | Fix |
|---|---|---|
[...]

PR #171, opened 2026-09-10
cf-ux: pin the skill-name matcher's false-positive trade (follow-up to #161)

Oleg67:
Review follow-up to #161, which merged before this could land on it.

One of the two findings on that PR was a genuine test gap. This is the test.

## The gap

`_invoked_names` compares every identifier-shaped string value in a `Skill` tool input, because which key holds the skill name is not part of any stable contract. Multi-word prose is excluded by shape (`_NAME_SHAPE`), and that case had a test. **A single bare token in an unrelated field is not excluded**, so:

```json
{"command": "superpowers:brainstorming", "mode": "cf"}
```

reads as the `cf` skill running. Reproduced before writing anything:

```
ran      names=['cf']                       <- prose beside the name (already tested)
ran      names=['brainstorming', 'cf']      <- bare 'cf' in an unrelated field  <- the gap
ran      names=['brainstorming', 'cf']      <- 'cf' as a target
ran      names=['brainstorming', 'cf']      <- a path ending in /cf
failed   names=['brainstorming']            <- rival only, no bare cf anywhere
```

## Pinned, not closed — and why

Narrowing the match to a fixed set of name keys (`command`, `skill`, `name`, …) would close this and open a worse hole. Guess the key wrong and **every** run errors, because the name would never be found where it actually lives. I do not know the real key: the fixtures in this suite use `{"command": …}`, which I chose, not something I measured — `claude` is not installed in my environment, as stated on #161.

So the trade is deliberate, and it errs the way that keeps the harness working without that knowledge:

| | this way (all values) | narrowed to name keys |
|---|---|---|
| rival skill with a bare `cf` field | rare false **pass** | correct |
| real key not in the list | correct | certain false **failure**, every run |

`skill_call_inputs` carries the raw input verbatim, so a `ran` verdict reached this way is inspectable — a maintainer sees `{"command": "superpowers:brainstorming", "mode": "cf"}` and spots it. That field exists for exactly this.

If the real key is ever pinned down by someone who can run the CLI, **the new test is where the trade gets renegotiated** — narrowing the matcher fails it, loudly, with the reasoning in the docstring.

## Tests

| test | pins |
|---|---|
| `test_a_bare_cf_in_an_unrelated_field_still_counts` | the accepted false positive, with the trade argued in the docstring |
| `test_a_single_token_that_is_not_cf_does_not_count_either` | the other side — the comparison stays whole-value in every field, so `mode: "cf-generate"` does not match |

**Two mutations, two caught:**

| mutation | fails |
|---|---|
| narrow `_invoked_names` to a fixed set of name keys | 1 — **and nothing else**, which is what made this worth a test |
| whole-value compare → substring containment | 5 |

That first row is the finding restated as a measurement: before this commit, the matcher's scope could be changed and the suite would not notice.

## Review round 2 (`7303349e`)

Six findings and a README gap. All actioned — and one of them found a contradiction between the argument above and the code implementing it.

**The scan stopped at the top level.** The trade is recall over precision, accepting a rare false pass to avoid a certain false failure. But `{"options": {"skill": "cf"}}` found nothing, so an input that nests its identifier one level down would make *every* run error while reporting `absent` — the exact failure the trade exists to avoid, inside its own implementation. The docstring was not overstating the policy; the code was under-delivering it. `_invoked_names` now walks dicts and lists to any depth, iteratively so transcript nesting cannot reach a recursion limit.

**"Inspectable" was doing work it had not earned.** Asked whether anything actually surfaces these false passes, the honest answer was no — the mitigation was "someone diffs the metadata afterwards," which is a hope, not a mitigation. A `ran` verdict resting on one of several candidate identifiers *in the matched call* is the shape a false pass takes, so it now reports itself: `metadata["skill_match_ambiguous"]` plus a stderr warning naming the other candidate. A flag, not a verdict change — it knows it could not tell which value was the name, not that the match is wrong — but a false pass is now countable by whatever consumes `skill_state`, which was the risk raised.

Four test gaps, all real:

| gap | now |
|---|---|
| the false-positive match combined with an errored `tool_result` | pinned — the loose match buys a *name*, not a verdict |
| a namespaced value in an unrelated field (`plugin:cf`), the same false positive through the separator-stripping path | pinned, and the docstring's "whole-value" framing corrected to say *after stripping* |
| the positive test never asserted an answer was delivered — the whole point of the false positive | `assert out["output"] == "the answer"` |
| the negative test never pinned `skills_invoked`, so silently dropping a candidate would pass | exact list asserted |

**README** documented the false positives that were closed and was silent on the one left open. Now carries the trade, both mitigations, the unverified-CLI caveat, and which test pins it.

**7 mutations, 7 caught.** Reverting to the top-level-only scan fails the new nested test and nothing else.

## Review round 3 (`f75c71dc`)

Four findings. Five of round 2's six were independently re-verified as resolved; the sixth was accepted as answered.

| Finding | Fix |
|---|---|
| **The ambiguity flag overclaimed.** `skill_match_ambiguous` fired whenever a matched call named a second identifier, which cannot distinguish a wrong-field match from a correct call that merely carries one. I had conceded exactly that on a thread and then shipped the assertion anyway. | Renamed to **`skill_match_other_candidates`** — a list of what was observed, not a boolean asserting a suspicion. Narrowing it is impossible by construction: it would need the key whose absence created the trade. Documented as an over-approximation; the run is still scored. |
| **The candidate list was neither deduplicated nor ordered**, and it reaches a warning a person reads. The stack pops LIFO, so a repeated value was repeated and the rest came out in an unpredictable order — two runs over one transcript could word the same finding differently. | `_invoked_names` returns `sorted(set(...))`, so traversal order becomes an implementation detail nothing depends on. Test asserts the rendered warning text, not just the list. |
| **Dropping the `isinstance(payload, dict)` guard was an undeclared boundary change** — a bare string or list input is now scanned rather than refused. | Kept, because refusing it reinstates at the outermost level the failure the depth walk removed — but now in the docstring, the README, and a parametrized test. |
| **"Any depth" was pinned only at depth one.** | Parametrized over two levels, list-of-dicts, dict-in-list-in-dict, and nested lists. |

**6 mutations, 6 caught.** The precise one for the last row: a traversal depth-limited to one level leaves `one-level` passing and fails the other four — the regression as described, which the previous single test would have missed.

## Review round 4 (`fa829c10`)

One finding: the negative tests asserted the absence of output and a detail substring, never the `skill_state` label itself — the field downstream actually reads.

Checked before fixing, and it was exploitable: mislabelling the none-named branch `"absent"` while leaving its detail text correct **passed all 52 tests**. The error string interpolates whatever state it is handed (`f"cf skill did not run ({state}): {detail}"`), so asserting the detail proves nothing about the label.

`skill_state` is now asserted on every negative case, one per branch of the ladder. **5 mutations, 5 caught** — each label swapped in turn, including the `absent` branch mutated in the opposite direction so it is pinned as `absent` rather than as "not the default".

## Scope

`tests/prompts/` is outside the traced population, so no `@cpt` obligation.

| Gate | Result |
|---|---|
| `make test` | **5,822 passed**, 4 skipped, 15 xfailed (52 in this file) |
| `cfs validate` | 0 errors |
| `spec-coverage --system studio` | 0.4608, unchanged |

## The other finding on #161, and one I found while answering it

The second finding was an architecture challenge — should the skill-run verdict live in the provider or in a promptfoo assertion? Answered on its thread rather than here, since it is a design question and non-blocking.

While answering it I found something worth a maintainer's ruling, in `promptfooconfig.yaml`:

```yaml
- type: not-icontains-any
  value:
    - "skills failed to load"
    - "cf skills failed"
    - "skill could not load"
```

Three scenarios each assert the **absence of strings the CLI does not emit** — the assertion-level twin of the `skill_load_warning` guard #161 removed from the provider, inert for the same reason. Now that the provider returns an error on the real signature before any assertion runs, these are dead weight rather than a second line of defence.

I have not touched them: deleting per-scenario assertions is a behaviour change on a shared config, and the disposition is a maintainer call — delete, or repoint at something that can actually fire. Say which and I will open it.

## Summary by CodeRabbit

* **Enhancements**
  * Skill identifiers are detected within deeply nested dictionaries and lists, as well as non-dictionary inputs.
  * Additional candidate skill names are surfaced when multiple possible matches are found.
  * Candidate names are deduplicated and presented in a consistent order.
  * Genuine tool calls continue returning expected results while reporting other candidate identifiers.
  * Failed matches now clearly report a failed skill state.

* **Documentation**
  * Updated guidance covers nested input scanning, candidate reporting, warnings, and diagnostic limitations.

### DIFF
--- a/tests/prompts/cf-ux/providers/claude_provider.py
+++ b/tests/prompts/cf-ux/providers/claude_provider.py
@@ -79,6 +79,13 @@ _RESULT_OK = "success"
 #: a person will see it, and the house rule is that a swallowed exception warns
 #: on stderr rather than only in a return value (`architecture/DESIGN.md`).
 _LOG_UNPARSED_LINE = "cf-ux claude provider: stream line %d would not parse; skipped"
+#: A `ran` verdict that rests on one of several candidate identifiers is the
+#: shape a false pass takes, so it is reported per run rather than left to
+#: whoever thinks to inspect `skill_call_inputs` afterwards.
+_LOG_AMBIGUOUS_MATCH = (
+    "cf-ux claude provider: matched %r in a Skill input that also names %s; "
+    "the verdict may rest on a field that is not the skill name"
+)
 
 
 def _stream_events(raw: str) -> tuple[list[dict[str, Any]], int]:
@@ -133,18 +140,31 @@ def _invoked_names(payload: Any) -> list[str]:
     """The skill identifiers a `Skill` tool input names.
 
     Which key holds the name is not part of any stable contract, so every string
-    value shaped like an identifier is a candidate — but each is compared
-    *whole*, after dropping any `plugin:` namespace. That is what separates the
-    skill's name from the argument text, which in this suite always quotes a
-    `/cf …` prompt.
+    value shaped like an identifier is a candidate — *at any depth*, because a
+    tool input that nests its identifier one level down (`{"options": {"skill":
+    "cf"}}`) is a shape this cannot rule out either. Stopping at the top level
+    would mean a nested name is never found, and then every run errors: the
+    certain false failure this trade exists to avoid, rather than the rare false
+    pass it accepts.
+
+    Each candidate is compared *whole* — but only after any `plugin:` or `path/`
+    namespace is dropped, so `plugin:cf` counts and `cf-generate` does not. What
+    the shape filter excludes is the argument text, which in this suite always
+    quotes a `/cf …` prompt.
     """
-    if not isinstance(payload, dict):
-        return []
     names = []
-    for value in payload.values():
-        if not isinstance(value, str) or not _NAME_SHAPE.fullmatch(value.strip()):
+    pending = [payload]
+    while pending:
+        current = pending.pop()
+        if isinstance(current, dict):
+            pending.extend(current.values())
             continue
-        name = value.strip()
+        if isinstance(current, list):
+            pending.extend(current)
+            continue
+        if not isinstance(current, str) or not _NAME_SHAPE.fullmatch(current.strip()):
+            continue
+        name = current.strip()
         for separator in _NAME_SEPARATORS:
             name = name.rsplit(separator, 1)[-1]
         if name:
@@ -159,6 +179,7 @@ class _SkillTrace(NamedTuple):
     names: list[str]    # skill identifiers the transcript names
     inputs: list[str]   # the tool inputs verbatim, serialized, for diagnosis
     detail: str
+    ambiguous: bool = False   # the verdict rests on one of several candidates
 
 
 def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
@@ -205,8 +226,19 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
             "failed", names, inputs,
             f"a {_SKILL_TOOL} ran but none of them named {_SKILL_NAME!r}",
         )
-    if any(results.get(call_id) is False for call_id in targeted):
-        return _SkillTrace("ran", names, inputs, "")
+    for call_id in targeted:
+        if results.get(call_id) is not False:
+            continue
+        # Because the name's key is unknown, `cf` may have been matched against a
+        # field that is not the name at all — a rival skill invoked with some
+        # unrelated value of `cf` reads as this one running. That trade is
+        # accepted (a rare false pass beats a certain false failure), but a
+        # verdict resting on one of several candidates is said out loud rather
+        # than left for whoever thinks to diff the metadata afterwards.
+        others = [name for name in named[call_id] if name != _SKILL_NAME]
+        if others:
+            logger.warning(_LOG_AMBIGUOUS_MATCH, _SKILL_NAME, others)
+        return _SkillTrace("ran", names, inputs, "", bool(others))
     if any(call_id not in results for call_id in targeted):
         return _SkillTrace(
             "failed", names, inputs,
@@ -327,6 +359,7 @@ def _invoke(prompt: str, cwd: Path, started: float) -> dict:
         "skill_state": state,
         "skills_invoked": trace.names,
         "skill_call_inputs": trace.inputs,
+        "skill_match_ambiguous": trace.ambiguous,
     }
     cost = payload.get("total_cost_usd")
 
--- a/tests/prompts/cf-ux/providers/claude_provider.py
+++ b/tests/prompts/cf-ux/providers/claude_provider.py
@@ -79,9 +79,16 @@ _RESULT_OK = "success"
 #: a person will see it, and the house rule is that a swallowed exception warns
 #: on stderr rather than only in a return value (`architecture/DESIGN.md`).
 _LOG_UNPARSED_LINE = "cf-ux claude provider: stream line %d would not parse; skipped"
-#: A `ran` verdict that rests on one of several candidate identifiers is the
-#: shape a false pass takes, so it is reported per run rather than left to
+#: A `ran` verdict reached from an input that names more than one identifier is
+#: the shape a false pass takes, so it is reported per run rather than left to
 #: whoever thinks to inspect `skill_call_inputs` afterwards.
+#:
+#: Deliberately an over-approximation, and named for what it observes rather
+#: than what it suspects. Since the name's key is unknown, this cannot tell a
+#: wrong-field match from a correct call that merely carries a second
+#: identifier: every false positive of this class is reported, and so are some
+#: perfectly good runs. Narrowing it would need the very knowledge whose absence
+#: created the trade.
 _LOG_AMBIGUOUS_MATCH = (
     "cf-ux claude provider: matched %r in a Skill input that also names %s; "
     "the verdict may rest on a field that is not the skill name"
@@ -151,8 +158,16 @@ def _invoked_names(payload: Any) -> list[str]:
     namespace is dropped, so `plugin:cf` counts and `cf-generate` does not. What
     the shape filter excludes is the argument text, which in this suite always
     quotes a `/cf …` prompt.
+
+    A top-level input that is not a dict is scanned rather than refused, for the
+    same reason: a bare string or a list is a shape this cannot rule out either,
+    and refusing it would mean the name is never found and every run errors.
+
+    Returned sorted and deduplicated. The traversal is a stack, so its own order
+    is an implementation detail, and nothing downstream should vary with it —
+    this list reaches a warning message a person reads.
     """
-    names = []
+    names = set()
     pending = [payload]
     while pending:
         current = pending.pop()
@@ -168,8 +183,8 @@ def _invoked_names(payload: Any) -> list[str]:
         for separator in _NAME_SEPARATORS:
             name = name.rsplit(separator, 1)[-1]
         if name:
-            names.append(name)
-    return names
+            names.add(name)
+    return sorted(names)
 
 
 class _SkillTrace(NamedTuple):
@@ -179,7 +194,9 @@ class _SkillTrace(NamedTuple):
     names: list[str]    # skill identifiers the transcript names
     inputs: list[str]   # the tool inputs verbatim, serialized, for diagnosis
     detail: str
-    ambiguous: bool = False   # the verdict rests on one of several candidates
+    #: The other identifiers in the matched call, when there were any. Reported
+    #: rather than judged: see `_LOG_AMBIGUOUS_MATCH`.
+    other_candidates: tuple[str, ...] = ()
 
 
 def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
@@ -235,10 +252,10 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
         # accepted (a rare false pass beats a certain false failure), but a
         # verdict resting on one of several candidates is said out loud rather
         # than left for whoever thinks to diff the metadata afterwards.
-        others = [name for name in named[call_id] if name != _SKILL_NAME]
+        others = tuple(name for name in named[call_id] if name != _SKILL_NAME)
         if others:
-            logger.warning(_LOG_AMBIGUOUS_MATCH, _SKILL_NAME, others)
-        return _SkillTrace("ran", names, inputs, "", bool(others))
+            logger.warning(_LOG_AMBIGUOUS_MATCH, _SKILL_NAME, list(others))
+        return _SkillTrace("ran", names, inputs, "", others)
     if any(call_id not in results for call_id in targeted):
         return _SkillTrace(
             "failed", names, inputs,
@@ -359,7 +376,7 @@ def _invoke(prompt: str, cwd: Path, started: float) -> dict:
         "skill_state": state,
         "skills_invoked": trace.names,
         "skill_call_inputs": trace.inputs,
-        "skill_match_ambiguous": trace.ambiguous,
+        "skill_match_other_candidates": list(trace.other_candidates),
     }
     cost = payload.get("total_cost_usd")
 
--- a/tests/test_cf_ux_claude_provider.py
+++ b/tests/test_cf_ux_claude_provider.py
@@ -202,14 +202,24 @@ class TestARunThatLoadedTheSkillIsScored:
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf"]
 
-    def test_a_nested_identifier_is_found_too(self, run_provider):
+    @pytest.mark.parametrize("shape", [
+        pytest.param({"options": {"skill": "cf"}}, id="one-level"),
+        pytest.param({"o": {"p": {"skill": "cf"}}}, id="two-levels"),
+        pytest.param({"args": [{"skill": "cf"}]}, id="list-of-dicts"),
+        pytest.param({"a": [{"b": {"skill": "cf"}}]}, id="dict-in-list-in-dict"),
+        pytest.param({"a": [["cf"]]}, id="nested-lists"),
+    ])
+    def test_a_nested_identifier_is_found_at_any_depth(self, run_provider, shape):
         """Stopping at the top level would contradict the trade above rather than
-        implement it: a tool input that nests its identifier one level down is a
-        shape this cannot rule out, and missing it means *every* run errors —
-        the certain false failure, not the rare false pass."""
+        implement it: a tool input that nests its identifier is a shape this
+        cannot rule out, and missing it means *every* run errors — the certain
+        false failure, not the rare false pass.
+
+        Parametrized past depth one because "any depth" is the claim; a single
+        level would leave a regression below it uncaught.
+        """
         nested = {"type": "assistant", "message": {"content": [
-            {"type": "tool_use", "id": "t1", "name": "Skill",
-             "input": {"options": {"skill": "cf"}}},
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": shape},
         ]}}
 
         out, _seen = run_provider(_stream(nested, _skill_result(), _result()))
@@ -217,13 +227,31 @@ class TestARunThatLoadedTheSkillIsScored:
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["cf"]
 
-    def test_an_unambiguous_match_is_not_flagged(self, run_provider):
-        """The flag has to distinguish, or it says nothing."""
+    @pytest.mark.parametrize("payload", ["cf", ["cf"], ["other", {"skill": "cf"}]])
+    def test_an_input_that_is_not_a_dict_is_scanned_rather_than_refused(
+        self, run_provider, payload,
+    ):
+        """A boundary the earlier version refused outright (`if not
+        isinstance(payload, dict): return []`). Scanning it follows from the same
+        reasoning as the depth walk: a bare string or list input is a shape that
+        cannot be ruled out, and refusing it means the name is never found and
+        every run errors. Called out because it is a behaviour change, not a
+        side effect."""
+        odd = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": payload},
+        ]}}
+
+        out, _seen = run_provider(_stream(odd, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_state"] == "ran"
+
+    def test_a_sole_candidate_leaves_the_other_candidates_empty(self, run_provider):
+        """The field has to distinguish, or it says nothing."""
         out, _seen = run_provider(
             _stream(_skill_call(skill="studio:cf"), _skill_result(), _result()),
         )
 
-        assert out["metadata"]["skill_match_ambiguous"] is False
+        assert out["metadata"]["skill_match_other_candidates"] == []
 
     def test_a_result_event_with_no_subtype_is_still_an_answer(self, run_provider):
         """Only a subtype that is present and says otherwise means a short turn.
@@ -511,11 +539,11 @@ class TestAnUnreadableRunIsNeverScored:
 
         assert out["metadata"]["unparsed_lines"] == 1
 
-    def test_an_ambiguous_match_is_surfaced_not_merely_inspectable(self, run_provider, caplog):
+    def test_the_other_candidates_are_surfaced_not_merely_inspectable(self, run_provider, caplog):
         """"Inspectable" only mitigates the false positive if someone inspects.
-        A verdict resting on one of several candidate identifiers is the shape a
-        false pass takes, so it is reported per run — on stderr and in the
-        metadata — rather than left for whoever thinks to diff
+        A verdict reached from an input naming more than one identifier is the
+        shape a false pass takes, so it is reported per run — on stderr and in
+        the metadata — rather than left for whoever thinks to diff
         `skill_call_inputs` afterwards."""
         rival = {"type": "assistant", "message": {"content": [
             {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
@@ -526,9 +554,44 @@ class TestAnUnreadableRunIsNeverScored:
         with caplog.at_level("WARNING", logger=claude_provider.logger.name):
             out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
 
-        assert out["metadata"]["skill_match_ambiguous"] is True
+        assert out["metadata"]["skill_match_other_candidates"] == ["brainstorming"]
         assert "may rest on a field that is not the skill name" in caplog.text
         assert "brainstorming" in caplog.text, "and names the other candidate"
+
+    def test_the_other_candidates_are_deduplicated_and_ordered(self, run_provider, caplog):
+        """This list reaches a message a person reads, and the traversal is a
+        stack — so left raw it repeats a value that appears twice and orders the
+        rest by an implementation detail. Two runs of the same transcript would
+        then produce different prose for the same finding."""
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "a": "zebra", "b": "alpha", "c": "cf",
+                "d": {"e": "zebra"}, "f": ["alpha", "middle"],
+            }},
+        ]}}
+
+        with caplog.at_level("WARNING", logger=claude_provider.logger.name):
+            out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_match_other_candidates"] == [
+            "alpha", "middle", "zebra",
+        ], "each named once, in an order that does not depend on the walk"
+        assert "['alpha', 'middle', 'zebra']" in caplog.text
+
+    def test_a_second_identifier_beside_a_genuine_call_is_reported_too(self, run_provider):
+        """The field is an over-approximation and says so: it cannot tell a
+        wrong-field match from a correct call that merely carries a second
+        identifier, because telling them apart needs the very knowledge whose
+        absence created the trade. Reported, not judged — `ran` still stands."""
+        genuine = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill",
+             "input": {"command": "cf", "mode": "auto"}},
+        ]}}
+
+        out, _seen = run_provider(_stream(genuine, _skill_result(), _result()))
+
+        assert out["output"] == "the answer", "a flag, not a verdict"
+        assert out["metadata"]["skill_match_other_candidates"] == ["auto"]
 
     def test_a_dropped_line_also_warns_where_a_person_will_see_it(self, run_provider, caplog):
         """A count riding along in a metadata dict is not the same as saying so.
--- a/tests/test_cf_ux_claude_provider.py
+++ b/tests/test_cf_ux_claude_provider.py
@@ -152,6 +152,38 @@ class TestARunThatLoadedTheSkillIsScored:
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["cf"]
 
+    def test_a_bare_cf_in_an_unrelated_field_still_counts(self, run_provider):
+        """A known false positive, pinned rather than closed.
+
+        Which key holds the skill name is not contractual, so every
+        identifier-shaped value is a candidate — and a rival skill invoked with
+        some field whose *whole* value is `cf` therefore reads as this skill
+        running. Prose does not do this (the test above), but a single token in
+        an unrelated field does.
+
+        Narrowing to a fixed set of name keys would close this and open a worse
+        hole: guess the key wrong and *every* run errors, because the name would
+        never be found where it actually lives. This way round the failure is a
+        rare false pass; that way round it is a certain false failure. Erring
+        toward recall is what makes the harness work without knowing the key,
+        and `skill_call_inputs` keeps the raw input so the verdict stays
+        inspectable. If the key is ever pinned down, this test is where the
+        trade gets renegotiated.
+        """
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "cf",
+            }},
+        ]}}
+
+        out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_state"] == "ran"
+        assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf"]
+        assert out["metadata"]["skill_call_inputs"] == [
+            '{"command": "superpowers:brainstorming", "mode": "cf"}'
+        ], "the raw input shows where the name was found, so a false pass is visible"
+
     def test_a_result_event_with_no_subtype_is_still_an_answer(self, run_provider):
         """Only a subtype that is present and says otherwise means a short turn.
         An unfamiliar event shape must not be able to manufacture failures."""
@@ -242,6 +274,21 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
         out, _seen = run_provider(
             _stream(_skill_call(skill="cf-generate"), _skill_result(), _result()),
         )
+
+        assert "output" not in out
+        assert "none of them named 'cf'" in out["error"]
+
+    def test_a_single_token_that_is_not_cf_does_not_count_either(self, run_provider):
+        """The trade pinned above is not "any short token wins". The comparison
+        stays whole-value in every field, so a sibling name sitting in the same
+        unrelated position does not match."""
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "cf-generate",
+            }},
+        ]}}
+
+        out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
 
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
--- a/tests/test_cf_ux_claude_provider.py
+++ b/tests/test_cf_ux_claude_provider.py
@@ -178,11 +178,52 @@ class TestARunThatLoadedTheSkillIsScored:
 
         out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
 
+        assert out["output"] == "the answer", "the point of the false positive: it gets scored"
+        assert "error" not in out
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf"]
         assert out["metadata"]["skill_call_inputs"] == [
             '{"command": "superpowers:brainstorming", "mode": "cf"}'
         ], "the raw input shows where the name was found, so a false pass is visible"
+
+    def test_a_namespaced_value_in_an_unrelated_field_counts_the_same_way(self, run_provider):
+        """The same false-positive class through a second mechanism: the value is
+        not literally `cf`, it reduces to it once the namespace is dropped. The
+        comparison is whole-value *after* stripping, which is what lets
+        `plugin:cf` count while `cf-generate` does not."""
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "plugin:cf",
+            }},
+        ]}}
+
+        out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_state"] == "ran"
+        assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf"]
+
+    def test_a_nested_identifier_is_found_too(self, run_provider):
+        """Stopping at the top level would contradict the trade above rather than
+        implement it: a tool input that nests its identifier one level down is a
+        shape this cannot rule out, and missing it means *every* run errors —
+        the certain false failure, not the rare false pass."""
+        nested = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill",
+             "input": {"options": {"skill": "cf"}}},
+        ]}}
+
+        out, _seen = run_provider(_stream(nested, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_state"] == "ran"
+        assert out["metadata"]["skills_invoked"] == ["cf"]
+
+    def test_an_unambiguous_match_is_not_flagged(self, run_provider):
+        """The flag has to distinguish, or it says nothing."""
+        out, _seen = run_provider(
+            _stream(_skill_call(skill="studio:cf"), _skill_result(), _result()),
+        )
+
+        assert out["metadata"]["skill_match_ambiguous"] is False
 
     def test_a_result_event_with_no_subtype_is_still_an_answer(self, run_provider):
         """Only a subtype that is present and says otherwise means a short turn.
@@ -292,6 +333,27 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
+        assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"], (
+            "both candidates were parsed and neither matched — not one silently dropped"
+        )
+
+    def test_a_false_positive_match_whose_call_errored_is_still_a_failure(self, run_provider):
+        """The accepted false positive buys a *name* match, not a verdict. The
+        matched call still has to have come back clean, so the interaction of
+        the loose match with the error branch is the one worth pinning."""
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "cf",
+            }},
+        ]}}
+
+        out, _seen = run_provider(
+            _stream(rival, _skill_result(is_error=True), _result("answered directly")),
+        )
+
+        assert "output" not in out
+        assert "came back as an error" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     def test_argument_text_naming_cf_is_not_the_skill_naming_cf(self, run_provider):
         """The prompt this suite sends is `/cf <request>`, so every scenario's
@@ -448,6 +510,25 @@ class TestAnUnreadableRunIsNeverScored:
         out, _seen = run_provider(transcript)
 
         assert out["metadata"]["unparsed_lines"] == 1
+
+    def test_an_ambiguous_match_is_surfaced_not_merely_inspectable(self, run_provider, caplog):
+        """"Inspectable" only mitigates the false positive if someone inspects.
+        A verdict resting on one of several candidate identifiers is the shape a
+        false pass takes, so it is reported per run — on stderr and in the
+        metadata — rather than left for whoever thinks to diff
+        `skill_call_inputs` afterwards."""
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "cf",
+            }},
+        ]}}
+
+        with caplog.at_level("WARNING", logger=claude_provider.logger.name):
+            out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_match_ambiguous"] is True
+        assert "may rest on a field that is not the skill name" in caplog.text
+        assert "brainstorming" in caplog.text, "and names the other candidate"
 
     def test_a_dropped_line_also_warns_where_a_person_will_see_it(self, run_provider, caplog):
         """A count riding along in a metadata dict is not the same as saying so.
--- a/tests/test_cf_ux_claude_provider.py
+++ b/tests/test_cf_ux_claude_provider.py
@@ -305,6 +305,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "did not run (failed)" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     def test_the_real_failure_signature_is_recognised(self, run_provider):
         """The observed failure is `<error>Execute skill: cf</error>`. The previous
@@ -316,6 +317,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "Execute skill:" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     @pytest.mark.parametrize("rival", ["superpowers", "cf-generate"])
     def test_another_skill_failing_does_not_fail_the_cf_run(self, run_provider, rival):
@@ -337,6 +339,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
         assert out["metadata"]["skills_invoked"] == ["superpowers"], "and it says which did run"
 
     def test_a_skill_whose_name_merely_begins_with_cf_is_not_the_cf_skill(self, run_provider):
@@ -346,6 +349,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     def test_a_single_token_that_is_not_cf_does_not_count_either(self, run_provider):
         """The trade pinned above is not "any short token wins". The comparison
@@ -361,6 +365,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
         assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"], (
             "both candidates were parsed and neither matched — not one silently dropped"
         )
@@ -415,6 +420,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "came back as an error" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     def test_a_cf_call_with_no_result_at_all_is_not_a_confirmed_run(self, run_provider):
         """Absence of a result is not a non-error result. A trace cut off after
@@ -424,6 +430,7 @@ class TestARunThatNeverLoadedTheSkillIsNotScored:
 
         assert "output" not in out
         assert "no result in the transcript" in out["error"]
+        assert out["metadata"]["skill_state"] == "failed"
 
     def test_a_repeated_call_id_collapses_to_the_last_one(self, run_provider):
         """The id is the only binding between a call and its result, so a
--- a/tests/prompts/cf-ux/README.md
+++ b/tests/prompts/cf-ux/README.md
@@ -71,6 +71,40 @@ bound to *that* call's id. Each half of that matters.
 - No result at all is not a non-error result. A trace cut off after the call is
   refused, not read as a success.
 
+### One false positive is accepted on purpose
+
+Which key holds the skill name is not part of any stable contract, so **every
+identifier-shaped value in the tool input is a candidate, at any depth**. That
+means a rival skill invoked with some unrelated field whose value happens to be
+`cf` — `{"command": "superpowers:brainstorming", "mode": "cf"}` — reads as this
+skill running. `plugin:cf` in that position does too, since the comparison
+happens after the namespace is dropped.
+
+This is a deliberate trade, not an oversight. Narrowing the scan to a fixed set
+of name keys would close it and open a worse hole:
+
+| | every value (today) | narrowed to name keys |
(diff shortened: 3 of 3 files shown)


tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,38 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["cf"]
 
+    def test_a_bare_cf_in_an_unrelated_field_still_counts(self, run_provider):
+        """A known false positive, pinned rather than closed.
```
ainetx, line 156: ### No test pins false-positive skill match combined with an erroring tool_result

**Severity:** Minor

**Problem**
test_a_bare_cf_in_an_unrelated_field_still_counts only pairs the mismatched-field match with a successful _skill_result(); the adjacent branch in _skill_trace (targeted call present but its own tool_result.is_error=True) is never exercised together with a false-positive name match.

<details>

**How to reproduce**
1. Build a rival tool_use whose unrelated field equals 'cf' (as in the new test). 2. Give that call's tool_result is_error=True. 3. Run _skill_trace on the resulting stream.

**Expected behavior**
skill_state should resolve to 'failed' for this combination, and a test should confirm that.

**Actual behavior**
No test in tests/test_cf_ux_claude_provider.py exercises this combination; the interaction is untested even though the branch logic exists and is reachable.

```text
false-positive match + is_error=True tool_result -> falls to 'every call came back as an error' -> failed (untested path)
```

**Impact**
A future regression in this branch (e.g. accidentally treating an errored false-positive call as 'ran') would not be caught by the test suite.

**Suggested correction**
Add a test combining the false-positive input shape with an errored tool_result for the same call id, asserting skill_state == 'failed'.

**How to verify**
New test fails before any fix and passes after correct branch behavior is restored/maintained.

</details>

ainetx:
### README omits the newly pinned bare-token false-positive trade

**Severity:** Minor

**Problem**
README.md's description of what does/does not count for skill matching lists the false positives that were fixed (substring containment, argument-text prose) but not the one deliberately left open and now pinned by test_a_bare_cf_in_an_unrelated_field_still_counts.

<details>

**How to reproduce**
1. Read README.md lines 61-91. 2. Compare against the new test's docstring rationale (recall-over-precision, any field can hold the value). 3. Note the README never mentions this scenario or points to skill_call_inputs as the mitigation.

**Expected behavior**
The README should mention that a rival skill whose unrelated input field equals 'cf' verbatim will be misread as this skill running, and that skill_call_inputs is the way to detect/diagnose it.

**Actual behavior**
README is silent on this known trade; a person triaging a suspicious 'ran' verdict has no documentation pointer to this behavior.

```text
test pins known FP -> rationale lives only in test docstring -> README (human-facing contract) does not surface it
```

**Impact**
Increases risk that a future triager misdiagnoses a false-positive 'ran' result as a genuine skill execution, wasting investigation time.

**Suggested correction**
Add a bullet to the README's skill-match section describing the bare-token-in-unrelated-field false positive and the recall-over-precision rationale, referencing skill_call_inputs for diagnosis.

**How to verify**
README review confirms the trade is now documented alongside the other listed false-positive fixes.

</details>

---
_Original location: tests/prompts/cf-ux/README.md:65 -- inline anchoring could not be resolved after 1 attempt(s)._

tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,38 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["cf"]
 
+    def test_a_bare_cf_in_an_unrelated_field_still_counts(self, run_provider):
```
ainetx, line 155: ### Namespace-splitting interaction with the pinned false-positive trade is untested

**Severity:** Minor

**Problem**
The two new tests (lines 155-185, 281-294) pin whole-value bare-token matching ('cf' vs 'cf-generate') in an unrelated field, but _invoked_names also strips ':'/'/' namespace prefixes before comparing (claude_provider.py:144-151). A field value like 'plugin:cf' reduces to 'cf' only after splitting, producing the same false-positive class through a separate mechanism that these tests don't cover, so the docstring's framing ('whole-value equals cf') is narrower than the code's actual behavior.

<details>

**How to reproduce**
1. Construct a rival Skill call with an unrelated field, e.g. {"mode": "plugin:cf"}. 2. Run it through _skill_trace. 3. Observe skill_state == 'ran' via the namespace-split path, which is not asserted by any existing test.

**Expected behavior**
A test (or docstring note) confirming whether/why a namespaced unrelated-field value that reduces to 'cf' after separator-splitting is treated the same as the bare-token case.

**Actual behavior**
No test exercises this combination; the docstring's rationale omits mention of the separator-stripping mechanism as a contributor to the same false-positive class.

```text
unrelated field 'plugin:cf' -> _invoked_names strips 'plugin:' -> name == 'cf' -> targeted -> skill_state 'ran' (same false positive, different path, untested)
```

**Impact**
A future change to _NAME_SEPARATORS or the splitting logic could silently alter false-positive behavior with no test to catch it, since only the bare-token path is pinned.

**Suggested correction**
Add a test with an unrelated field whose raw value is namespaced (e.g. 'plugin:cf') and assert whether it is treated as matching, documenting the combined trade explicitly.

**How to verify**
New test passes and demonstrates the intended behavior for the namespaced-unrelated-field case.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -246,6 +278,21 @@ def test_a_skill_whose_name_merely_begins_with_cf_is_not_the_cf_skill(self, run_
         assert "output" not in out
         assert "none of them named 'cf'" in out["error"]
 
+    def test_a_single_token_that_is_not_cf_does_not_count_either(self, run_provider):
```
ainetx, line 350: ### Negative false-positive test doesn't pin the parsed skills_invoked list

**Severity:** Minor

**Problem**
The test only checks 'output' not in out and the error substring, never asserting skills_invoked == ['brainstorming', 'cf-generate']. This leaves the stated whole-value comparison behavior for the negative case unverified at the metadata level.

<details>

**How to reproduce**
1. Run test_a_single_token_that_is_not_cf_does_not_count_either. 2. Modify _invoked_names to silently drop the 'mode' field. 3. Test still passes because skills_invoked is never checked.

**Expected behavior**
Test should assert the exact skills_invoked list, e.g. == ['brainstorming', 'cf-generate'], to confirm both identifier-shaped values were actually parsed and neither one matched 'cf'.

**Actual behavior**
Only error-text and output-absence are checked; the parsed names are never inspected.

```text
rival input {command, mode} -> _invoked_names -> [brainstorming, cf-generate] -> (untested) -> targeted=[] -> failed state -> error asserted, names unchecked
```

**Impact**
A regression that silently loses one of the two candidate values would not be caught by this test.

**Suggested correction**
Add assert out['metadata']['skills_invoked'] == ['brainstorming', 'cf-generate'].

**How to verify**
Break _invoked_names to skip one field and confirm the added assertion now fails.

</details>

tests/test_cf_ux_claude_provider.py
ainetx, line 155: ### Positive false-positive test never confirms an answer is actually delivered

**Severity:** Minor

**Problem**
The test pins skill_state=='ran' and the parsed names/inputs, but never asserts 'output' in out or 'error' not in out, so the actual point of the false positive (the run being scored as if it produced a usable answer) is not verified.

<details>

**How to reproduce**
1. Run the test. 2. Hypothetically change downstream logic so 'ran' state coexists with an error/no-output result for this exact call shape. 3. Test still passes since output/error are unchecked.

**Expected behavior**
Assert 'output' in out and 'error' not in out alongside the metadata assertions.

**Actual behavior**
Only metadata labels are checked; output-delivery path is unverified for this scenario.

```text
rival call matches 'cf' in mode field -> skill_state='ran' (asserted) -> _invoke() output/error branch -> (untested)
```

**Impact**
A future change that decouples 'ran' state from actual output delivery for this path would go undetected.

**Suggested correction**
Add assert 'output' in out and 'error' not in out (or equivalent) to the test.

**How to verify**
Introduce a bug that returns 'ran' state without setting out['output'] and confirm the added assertions catch it.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,38 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
         assert out["metadata"]["skill_state"] == "ran"
         assert out["metadata"]["skills_invoked"] == ["cf"]
 
+    def test_a_bare_cf_in_an_unrelated_field_still_counts(self, run_provider):
+        """A known false positive, pinned rather than closed.
+
+        Which key holds the skill name is not contractual, so every
+        identifier-shaped value is a candidate — and a rival skill invoked with
```
ainetx, line 159: ### Docstring overstates match scope vs. top-level-only implementation

**Severity:** Minor

**Problem**
_invoked_names only inspects direct string values of the input dict (`for value in payload.values()`), never recursing into nested dicts/lists. The new test's docstring and the existing code docstring both describe the policy as considering 'every string value shaped like an identifier' / 'every identifier-shaped value... is a candidate' without qualifying that this only applies to the top level, which could mislead a reader into believing nested identifiers are also caught.

<details>

**How to reproduce**
1. Construct a Skill tool_use input like {'options': {'skill': 'cf'}}. 2. Call _invoked_names on it. 3. Observe the nested 'cf' value is never found because only top-level values are iterated.

**Expected behavior**
Documentation should clarify the match is top-level-values-only, or the implementation should recurse if 'every value' is meant literally.

**Actual behavior**
Only top-level dict values are checked; nested identifier-shaped strings are silently missed.

```text
input={'options': {'skill': 'cf'}} -> payload.values() -> [{'skill':'cf'}] -> not str -> skipped -> 'cf' never found
```

**Impact**
If a real Skill tool nests its identifier one level down, the harness would misreport skill_state as failed/absent even though the skill actually ran, contradicting the stated 'err toward recall' intent for exactly this class of case.

**Suggested correction**
Either recurse into nested dict/list values in _invoked_names, or narrow the docstring/test claim to 'every top-level string value'.

**How to verify**
Add a test with a nested identifier and confirm current behavior misses it; after a fix (or docstring correction), re-run to confirm consistency between claim and behavior.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,38 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
...
+        inspectable. If the key is ever pinned down, this test is where the
+        trade gets renegotiated.
+        """
+        rival = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
+                "command": "superpowers:brainstorming", "mode": "cf",
+            }},
+        ]}}
+
+        out, _seen = run_provider(_stream(rival, _skill_result(), _result()))
```
ainetx, line 179: ### Is the false positive ever surfaced, or only inspectable on demand?

_Non-blocking review challenge -- [product]_

The docstring says `skill_call_inputs` 'keeps the raw input so the verdict stays inspectable' — but is anything in the harness or its reporting actually surfacing/flagging these false-pass cases, or does catching one require a human to manually diff `skill_call_inputs` after the fact? If nobody's expected to look, this is a silent metric-inflation risk for whatever downstream benchmark consumes `skill_state`/`skills_invoked`.

**Why this is worth asking**
Pinning a known false positive is reasonable, but 'inspectable' only mitigates the risk if someone is actually inspecting. Since this measures skill-invocation success for evaluation purposes, a rare-but-real false pass that goes unnoticed silently overstates the skill's reliability.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `7303349e` — and this one caught a contradiction between the argument I made for the trade and the code implementing it, which is the most useful kind of finding.

The justification I gave on #161 was *recall over precision*: because the name's key is unknown, scan everything and accept a rare false pass in order to avoid a **certain false failure**. But the scan stopped at the top level, so:

```
names=['cf']  <- {"command": "cf"}                  flat, genuine
names=[]      <- {"options": {"skill": "cf"}}       nested one level
names=[]      <- {"args": [{"skill": "cf"}]}        nested in a list
```

Your impact line is exactly right, and it is the failure the trade exists to avoid: a real Skill tool that nests its identifier would make **every** run error while reporting `absent`. So the docstring was not overstating the policy — the implementation was under-delivering it.

I took the first of your two options, because narrowing the docs would have documented the inconsistency rather than removed it. `_invoked_names` (`claude_provider.py:154-171`) now walks dicts and lists to any depth, iteratively rather than recursively so transcript nesting cannot reach a recursion limit:

```python
pending = [payload]
while pending:
    current = pending.pop()
    if isinstance(current, dict):
        pending.extend(current.values())
        continue
    if isinstance(current, list):
        pending.extend(current)
        continue
```

Docstring rewritten to say *at any depth* and to explain why that follows from the trade rather than merely stating it.

Test: `test_a_nested_identifier_is_found_too` (`tests/test_cf_ux_claude_provider.py:205`). Mutation, the precise one: reverting to the original top-level-only loop **fails that test and nothing else** — so it is pinned exactly, not incidentally.

tests/test_cf_ux_claude_provider.py
Oleg67: **Fair hit, and fixed in `7303349e`** — the honest answer to your question as asked was *no, nothing surfaced it*. "Inspectable" was doing real work in that docstring, and it should not have been.

You are right about the shape of the risk too: this feeds an evaluation, so a rare-but-real false pass that nobody notices overstates skill reliability rather than merely being untidy. Left as-is, the mitigation was "someone diffs `skill_call_inputs` after the fact," which is not a mitigation, it is a hope.

So the false positive now reports itself. The signature is specific: a `ran` verdict that rests on **one of several** candidate identifiers *in the matched call* — which is exactly the shape a false pass takes, since a genuine call names `cf` and little else (`claude_provider.py:229-241`):

```python
others = [name for name in named[call_id] if name != _SKILL_NAME]
if others:
    logger.warning(_LOG_AMBIGUOUS_MATCH, _SKILL_NAME, others)
return _SkillTrace("ran", names, inputs, "", bool(others))
```

Surfaced two ways per run, not on demand:

- **`metadata["skill_match_ambiguous"]`** — so whatever consumes `skill_state` can filter or flag these instead of counting them blind. That is the downstream-benchmark concern you raised.
- **a stderr warning naming the other candidate** — so it is visible in the run output itself, following the same house rule the dropped-line warning follows (`architecture/DESIGN.md:277`).

Deliberately a flag and not a verdict change: it does not know that the match is wrong, only that it could not tell which value was the name, so turning it into an error would manufacture failures on inputs like `{"command": "cf", "mode": "auto"}`. It does mean a false pass is now *countable* rather than merely discoverable.

Tests: `test_an_ambiguous_match_is_surfaced_not_merely_inspectable` (`:514`) asserts both the flag and the warning text, and `test_an_unambiguous_match_is_not_flagged` (`:220`) asserts the clean case is `False` — a flag that never distinguishes says nothing. Two mutations (flag hard-wired to `False`; warning removed), both caught, one test each.

Also written into the README as part of the documented trade rather than left in a docstring.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `7303349e`.

Confirmed — and you are right that the docstring's framing was narrower than the behaviour. Reproduced:

```
names=['brainstorming', 'cf']  <- {"command": "superpowers:brainstorming", "mode": "plugin:cf"}
```

The value is not literally `cf`; it reduces to it once `_NAME_SEPARATORS` strips the namespace. Same false-positive class, second mechanism, and nothing pinned it.

Both halves fixed:

- **Test** — `test_a_namespaced_value_in_an_unrelated_field_counts_the_same_way` (`tests/test_cf_ux_claude_provider.py:189`), asserting `ran` and the parsed names.
- **Framing** — the docstring now says the comparison is whole-value *after* namespace stripping, and names both consequences in one breath: that is what lets `plugin:cf` count while `cf-generate` does not. The README section describing the accepted trade says the same, so the mechanism is documented where a reader meets the policy rather than only at the call site.

Mutation: removing the separator strip fails 5 tests, including the new one. The intent is deliberate — a namespaced identifier *is* the skill, and `plugin:skill` is how a marketplace skill is named — so this reduces to the same accepted trade rather than a separate defect: the field it came from might not be the name.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `7303349e`.

Agreed, and the interaction is worth pinning for a reason beyond coverage: the loose name match buys a **name**, not a verdict. The matched call still has to have come back clean, and this is the one combination where those two facts could plausibly be conflated in a future edit.

Test: `test_a_false_positive_match_whose_call_errored_is_still_a_failure` (`tests/test_cf_ux_claude_provider.py:340`) — your exact recipe, the rival input shape with `is_error: true` on the same call id:

```python
assert "output" not in out
assert "came back as an error" in out["error"]
assert out["metadata"]["skill_state"] == "failed"
```

Mutation, the one your write-up names ("accidentally treating an errored false-positive call as 'ran'"): relaxing the per-call error check fails 4 tests, this one among them.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `7303349e`.

Correct, and it was the wrong thing to leave out of *this* test in particular: the point of the false positive is not that a label reads `ran`, it is that **the run gets scored** — a fallback answer reaching the grader is the entire defect this PR family exists to close. Asserting the label and not the delivery tested the bookkeeping and skipped the consequence.

`tests/test_cf_ux_claude_provider.py:155`:

```python
assert out["output"] == "the answer", "the point of the false positive: it gets scored"
assert "error" not in out
```

Exact value rather than presence, for the same reason the earlier round moved the diagnostic assertions off truthiness.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `7303349e`.

Confirmed, including your mutation: dropping the `mode` field from the parse left the test green, because it only ever checked the error text.

`tests/test_cf_ux_claude_provider.py:322`:

```python
assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"], (
    "both candidates were parsed and neither matched — not one silently dropped"
)
```

That distinction is what the test is for: "neither value matched `cf`" and "only one value was ever looked at" produce the same `failed` verdict and the same error string, and only the first is the behaviour being claimed.

Mutation, yours: making `_invoked_names` keep only the first candidate now fails 4 tests, this one included.

Oleg67:
## README omits the pinned trade — fixed in `7303349e`

Correct, and the omission had a shape worth naming: the README listed the false positives that were **closed** (substring containment, argument-text prose) and was silent on the one deliberately left **open**. A reader would have come away thinking the matcher is tighter than it is — which is the same asymmetry the whole change set is about, in the documentation.

New section, `tests/prompts/cf-ux/README.md` — *"One false positive is accepted on purpose"*, placed immediately after the three bullets on what does and does not count, so it is read alongside them rather than as a footnote. It states:

- **the shape** — `{"command": "superpowers:brainstorming", "mode": "cf"}` counts, and so does `plugin:cf` in that position, since the comparison happens after the namespace is dropped;
- **that it is now depth-independent**, and why that follows from the trade rather than widening it;
- **the trade itself**, as the table from the PR body: every value risks a rare false *pass*, narrowing to name keys risks a certain false *failure* on every run;
- **that nobody here has run the CLI**, so the fixture shape is chosen rather than measured — the reason the uncertainty is resolved toward recall;
- **both mitigations** — `skill_match_ambiguous` plus its stderr warning (new this round, from the companion review thread asking whether anything actually surfaces these), and `skill_call_inputs` for checking a verdict;
- **which test pins it**, so the renegotiation point is discoverable from the docs and not just from the test file.

`skill_match_ambiguous` is also added to the metadata list further down, which previously would have gone stale the moment that field shipped.

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7303349e7f8cbf099b6b5081aa5fbca4797ccc97 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff already includes test_a_false_positive_match_whose_call_errored_is_still_a_failure (tests/test_cf_ux_claude_provider.py:340-356), which combines the rival tool_use whose unrelated field's whole value is 'cf' with is_error=True on that same call's tool_result, and asserts skill_state == 'failed' and the error message names the erroring call. This resolves the previously-flagged gap; confirmed present via find_enclosing_symbol and the file diff.
- **Reviewer:** `test_a_false_positive_match_whose_call_errored_is_still_a_failure` combines rival `mode: "cf"` input with `_skill_result(is_error=True)` for call `t1` and asserts `skill_state == "failed"` and no output.
- **Verifier:** The current diff already includes test_a_false_positive_match_whose_call_errored_is_still_a_failure (tests/test_cf_ux_claude_provider.py), which pairs a rival tool_use with unrelated field mode='cf' and an is_error=True tool_result for that same call id, asserting skill_state == 'failed' and the 'came back as an error' message — exactly the combination the check asks for. This matches the resolution noted in the discussion thread (commit 7303349e).

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7303349e7f8cbf099b6b5081aa5fbca4797ccc97 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff already includes test_a_namespaced_value_in_an_unrelated_field_counts_the_same_way (tests/test_cf_ux_claude_provider.py), which exercises the 'plugin:cf' namespace-splitting interaction and asserts skill_state=='ran' with correct parsed names; the docstring in _invoked_names and the README were updated to frame the comparison as whole-value after namespace stripping, explicitly covering both 'plugin:cf' counting and 'cf-generate' not counting, closing the gap identified in this exact check's history.
- **Reviewer:** The diff adds `test_a_namespaced_value_in_an_unrelated_field_counts_the_same_way`, which explicitly verifies `plugin:cf` in an unrelated field matches after namespace splitting; the provider docstring and README also state that matching is whole-value after namespace removal.
- **Verifier:** The diff includes test_a_namespaced_value_in_an_unrelated_field_counts_the_same_way which exercises 'plugin:cf' reducing to 'cf' via _NAME_SEPARATORS stripping and asserts skill_state=='ran' with parsed names ['brainstorming','cf']; the _invoked_names docstring and README were updated to frame the trade as whole-value comparison after namespace stripping, explicitly covering both plugin:cf counting and cf-generate not counting -- confirmed resolved per commit 7303349e referenced in the discussion thread.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7303349e7f8cbf099b6b5081aa5fbca4797ccc97 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff already contains the exact assertion at test_a_single_token_that_is_not_cf_does_not_count_either: `assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"]`, which the discussion thread confirms was resolved in commit 7303349e and verified against the 'drop mode field' mutation.
- **Reviewer:** The negative-case test exactly asserts `out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"]`, covering both identifier-shaped fixture values.
- **Verifier:** The diff's test_a_single_token_that_is_not_cf_does_not_count_either (tests/test_cf_ux_claude_provider.py) already contains `assert out["metadata"]["skills_invoked"] == ["brainstorming", "cf-generate"], ("both candidates were parsed and neither matched — not one silently dropped")`, which is the exact exact-list assertion the check requests; confirmed resolved per commit 7303349e and the discussion thread showing it fails against a 'drop mode field' mutation.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7303349e7f8cbf099b6b5081aa5fbca4797ccc97 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff already includes `assert out["output"] == "the answer",...` and `assert "error" not in out` right before the metadata assertions in test_a_bare_cf_in_an_unrelated_field_still_counts (lines ~171-172), resolving the previously flagged gap as confirmed by the discussion thread and commit 7303349e.
- **Reviewer:** `test_a_bare_cf_in_an_unrelated_field_still_counts` explicitly asserts `out["output"] == "the answer"` and `"error" not in out` before validating metadata, directly covering answer delivery for this path.
- **Verifier:** The diff for test_a_bare_cf_in_an_unrelated_field_still_counts already includes `assert out["output"] == "the answer",...` and `assert "error" not in out` before the metadata assertions (lines ~171-172), which was confirmed resolved in commit 7303349e per the discussion thread.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 7303349e7f8cbf099b6b5081aa5fbca4797ccc97 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff shown already contains the fix: _invoked_names uses an iterative pending-stack traversal (`pending = [payload]; while pending:...`) that walks dicts and lists to any depth, matching the docstring's 'at any depth' claim, and test_a_nested_identifier_is_found_too pins this behavior — the prior finding was resolved within this same diff.
- **Reviewer:** `_invoked_names` now iteratively traverses nested dicts and lists via `pending`, matching the explicit “at any depth” claim; `test_a_nested_identifier_is_found_too` pins the behavior.
- **Verifier:** The diff shown already replaces the top-level-only loop in _invoked_names with an iterative pending-list traversal (`pending = [payload]; while pending:...`) that recurses into nested dicts and lists to any depth; the docstring was updated to say 'at any depth' and explains why (avoiding the certain false failure), matching the actual traversal, and test_a_nested_identifier_is_found_too pins exactly this behavior with `{"options": {"skill": "cf"}}`.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: OK, answer accepted.

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -133,18 +140,31 @@ def _invoked_names(payload: Any) -> list[str]:
...
+    pass it accepts.
+
+    Each candidate is compared *whole* — but only after any `plugin:` or `path/`
+    namespace is dropped, so `plugin:cf` counts and `cf-generate` does not. What
+    the shape filter excludes is the argument text, which in this suite always
+    quotes a `/cf …` prompt.
     """
-    if not isinstance(payload, dict):
-        return []
     names = []
```
ainetx: ### Non-dict top-level Skill `input` payload now yields candidates, silently and untested

**Severity:** Minor

**Problem**
The pre-existing guard `if not isinstance(payload, dict): return []` guaranteed a non-dict `input` never produced skill-name candidates. The rewrite drops that guard and instead pushes `payload` straight onto the traversal stack, so a top-level list or bare string `input` is now scanned the same way nested values are. This boundary change is not mentioned in the updated docstring (which only discusses nested dict example) and is not exercised by any new test.

<details>

**How to reproduce**
1. Construct a `Skill` tool_use event whose `input` is a bare string `"cf"` (or a list `["cf"]`) instead of a dict. 2. Run it through `_invoked_names`. 3. Observe that `["cf"]` is now returned, whereas the pre-rewrite code returned `[]` unconditionally for any non-dict payload.

**Expected behavior**
The new recall-biased, any-depth traversal claim should either explicitly cover this top-level case with a test, or the docstring/comments should note the guard removal is intentional and in-scope.

**Actual behavior**
Behavior silently changed for non-dict top-level payloads with zero test coverage of the new path.

```text
old: input=non-dict -> guard returns [] always
new: input=non-dict -> pushed to stack -> matched as string/list like nested values -> candidates possibly returned
```

**Impact**
An untested, unstated widening of what counts as a name-bearing input; if the Claude API ever emits a non-object `input` for `Skill` (e.g. a malformed tool call), the false-positive surface silently grows beyond what the design doc discusses.

**Suggested correction**
Add a test constructing a Skill tool_use with a non-dict input (list or string) and assert the resulting skills_invoked/skill_state, or explicitly document that this fallback is intended and part of the recall trade.

**How to verify**
Add the test above and confirm it passes with documented, intended behavior.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,79 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
...
+        the certain false failure, not the rare false pass."""
+        nested = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill",
+             "input": {"options": {"skill": "cf"}}},
+        ]}}
+
+        out, _seen = run_provider(_stream(nested, _skill_result(), _result()))
+
+        assert out["metadata"]["skill_state"] == "ran"
+        assert out["metadata"]["skills_invoked"] == ["cf"]
```
ainetx, line 228: ### Recursive 'any depth' traversal claim is pinned only at depth one

**Severity:** Minor

**Problem**
The docstring and README assert the scan covers every identifier-shaped value 'at any depth,' but the only test added (`test_a_nested_identifier_is_found_too`) covers a single level of dict nesting. No test exercises two-level nesting or mixed list/dict nesting (list of dicts), so a regression that broke traversal below the first nested level would not be caught by the test suite.

<details>

**How to reproduce**
1. Construct a Skill input `{"a": {"b": {"skill": "cf"}}}` (two levels) or `{"items": [{"skill": "cf"}]}` (list of dicts). 2. Run through `_invoked_names`. 3. No existing test asserts this is found.

**Expected behavior**
A test pinning at least one multi-level and one list-of-dicts nesting case, matching the breadth of the 'at any depth' guarantee.

**Actual behavior**
Only single-level nesting is tested; deeper/mixed nesting is unverified.

```text
traversal: dict -> extend(values) -> dict -> extend(values) ->... (arbitrary depth)
tests cover: depth=1 only
```

**Impact**
A future refactor or an early break/continue introduced at depth>1 could silently regress recall without any test failing, undermining the recall-biased design's own safety net.

**Suggested correction**
Add tests for two-or-more level nesting and for a list-of-dicts candidate.

**How to verify**
Add such tests and confirm they currently pass, then confirm they'd fail if traversal depth were artificially capped.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -205,8 +226,19 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
         )
-    if any(results.get(call_id) is False for call_id in targeted):
-        return _SkillTrace("ran", names, inputs, "")
+    for call_id in targeted:
+        if results.get(call_id) is not False:
+            continue
+        # Because the name's key is unknown, `cf` may have been matched against a
+        # field that is not the name at all — a rival skill invoked with some
+        # unrelated value of `cf` reads as this one running. That trade is
+        # accepted (a rare false pass beats a certain false failure), but a
```
ainetx, line 252: ### skill_match_ambiguous flags any co-occurring identifier, not genuine field ambiguity

**Severity:** Minor

**Problem**
`others = [name for name in named[call_id] if name != _SKILL_NAME]` and `ambiguous = bool(others)` fire whenever more than one identifier-shaped candidate exists in a targeted call's input, regardless of whether the actual 'cf' match came from the wrong field. A genuinely correct cf invocation whose input happens to carry a second, unrelated identifier-shaped value (not a namespace/name field) would be flagged identically to a true false positive.

<details>

**How to reproduce**
1. Construct a Skill call whose canonical name field is 'cf' plus another identifier-shaped field, e.g. `{"skill": "cf", "tag": "beta"}`. 2. Run through `_skill_trace`. 3. Observe `skill_match_ambiguous` is set True even though the match is genuinely correct — but no test in the diff exercises this to confirm/deny the intended behavior.

**Expected behavior**
Either the design intentionally treats any multi-candidate call as ambiguous (an epistemically consistent stance, since the real name field is unknown) and this should be stated/tested, or the flag should be reserved for cases resembling the documented false-positive shape.

**Actual behavior**
The flag's scope for ordinary correct calls with incidental extra identifiers is untested and unclear from the diff alone.

```text
call input: {name_field: 'cf', unrelated_field: 'other_id'} -> named[call_id] = ['cf','other_id'] -> others=['other_id'] -> ambiguous=True, even if 'cf' field is the real name
```

**Impact**
If this fires on ordinary correct invocations, the warning/metadata signal becomes noisy and less trustworthy, contrary to the stated goal of surfacing genuine uncertainty per run.

**Suggested correction**
Add a test with a canonical cf field plus an unrelated second identifier-shaped field, and assert/document the intended value of skill_match_ambiguous for that case.

**How to verify**
Add the test and confirm the flag's behavior matches the documented intent.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -205,8 +226,19 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
+        # verdict resting on one of several candidates is said out loud rather
+        # than left for whoever thinks to diff the metadata afterwards.
+        others = [name for name in named[call_id] if name != _SKILL_NAME]
```
ainetx: ### Ambiguous-match warning content untested for duplicate/multiple 'other' candidates

**Severity:** Minor

**Problem**
The `others` list feeding `_LOG_AMBIGUOUS_MATCH` is built via a plain list comprehension with no deduplication, and its order depends on the stack-based (LIFO) traversal order in `_invoked_names`. Only the single-other-candidate case is pinned by a test; a duplicate identifier (e.g. the same non-cf value appearing twice via nested repetition) or multiple distinct candidates could produce a repetitive or order-dependent message that isn't asserted anywhere.

<details>

**How to reproduce**
1. Construct a Skill call with two distinct non-cf identifier-shaped values, or the same non-cf value repeated at different nesting depths, alongside a matched 'cf'. 2. Trigger the ambiguous branch. 3. Inspect the resulting warning message and `others` list — no test currently asserts its exact composition for this case.

**Expected behavior**
A test asserting the exact `others` composition (deduplicated or not, in what order) when multiple/duplicate non-cf candidates are present.

**Actual behavior**
Only the single-candidate case is tested; multi/duplicate-candidate content is unverified and could be confusing.

```text
named[call_id] = ['cf','x','x'] -> others=['x','x'] (no dedup) -> warning message repeats 'x' twice, untested
```

**Impact**
A diagnostic message meant to help a human find the ambiguous field could be confusing or misleading in multi-candidate scenarios, weakening the observability goal.

**Suggested correction**
Add a test with two distinct or duplicate 'other' candidates and assert the exact warning/metadata content.

**How to verify**
Add the test and confirm the message content matches expectations (dedup or explicit ordering documented).

</details>

ainetx (approved):
Clean change overall — the traversal rewrite and the new ambiguity flag are both sound, well-motivated by the docstring/README trade-off writeup, and covered by a solid set of tests for the core false-positive scenarios. A few minor gaps worth a glance before they bite someone later, none blocking:

- **Non-dict top-level `input` now silently scanned** — the old guard that rejected a non-dict `Skill` `input` outright is gone; a bare list or string `input` now feeds the traversal same as nested values, but this isn't mentioned in the docstring or exercised by a test. ([comment](https://github.com/constructorfabric/studio/pull/171#discussion_r3986660352))
- **"Any depth" claim only tested at depth one** — the docs promise traversal at arbitrary nesting, but the only new test covers a single level of dict nesting; two-level or mixed list/dict nesting isn't pinned. ([comment](https://github.com/constructorfabric/studio/pull/171#discussion_r3986661202))
- **`skill_match_ambiguous` overfires on unrelated co-occurring identifiers** — it flags any second identifier-shaped value in the input, even one that has nothing to do with the actual name field, so a genuinely correct match can get the same flag as a real false positive. ([comment](https://github.com/constructorfabric/studio/pull/171#discussion_r3986661389))
- **Ambiguous-warning content untested for duplicates/multiples** — the `others` list has no dedup and its order depends on LIFO traversal, but only the single-other-candidate case is asserted. ([comment](https://github.com/constructorfabric/studio/pull/171#discussion_r3986661572))

tests/prompts/cf-ux/providers/claude_provider.py
Oleg67: **RESOLVED** in `f75c71dc` — and you are right that this was a boundary change smuggled in as a side effect. I removed a guard that had been doing something, and neither said so nor tested it.

Reproduced:

```
"cf"               -> ['cf']          (was [] before the rewrite)
["cf", "other"]    -> ['cf', 'other']  (was [])
42                 -> []
None               -> []
```

**Kept, because refusing it is the inconsistency.** The whole reason the scan descends into nested values is that a shape which cannot be ruled out must not go unscanned, or the name is never found and *every* run errors. A top-level bare string or list is exactly that same case — refusing it would reinstate, at the outermost level, the failure the depth walk was added to remove. So the guard's removal was right; making it silent was not.

Now stated in three places rather than none:

- **Docstring** (`claude_provider.py:161-163`) — "A top-level input that is not a dict is scanned rather than refused, for the same reason."
- **README** — the accepted-trade section says it alongside the depth claim, since a reader meets the policy there.
- **Test** — `test_an_input_that_is_not_a_dict_is_scanned_rather_than_refused` (`tests/test_cf_ux_claude_provider.py:231`), parametrized over a bare string, a list of strings, and a list containing a dict, with a docstring naming it as a deliberate behaviour change.

Mutation: restoring `if not isinstance(payload, dict): return []` fails those three cases and nothing else, so the boundary is now pinned exactly where you found it unpinned.

tests/test_cf_ux_claude_provider.py
Oleg67: **RESOLVED** in `f75c71dc`.

Correct — "any depth" was the claim and depth one was the evidence, which is the gap that matters most for a traversal.

`test_a_nested_identifier_is_found_at_any_depth` (`tests/test_cf_ux_claude_provider.py:212`) is now parametrized over five shapes, covering both axes you named:

| id | shape |
|---|---|
| `one-level` | `{"options": {"skill": "cf"}}` |
| `two-levels` | `{"o": {"p": {"skill": "cf"}}}` |
| `list-of-dicts` | `{"args": [{"skill": "cf"}]}` |
| `dict-in-list-in-dict` | `{"a": [{"b": {"skill": "cf"}}]}` |
| `nested-lists` | `{"a": [["cf"]]}` |

**The mutation is the part worth reporting, because it is the regression you described rather than a blunter one.** A traversal that works at depth one and stops below it — depth-limited to `< 2` — leaves `one-level` passing and fails the other four:

```
CAUGHT | traversal works at depth one but stops below it
        4 failed, 48 passed
        - test_a_nested_identifier_is_found_at_any_depth[two-levels]
        - test_a_nested_identifier_is_found_at_any_depth[list-of-dicts]
        - test_a_nested_identifier_is_found_at_any_depth[dict-in-list-in-dict]
        - test_a_nested_identifier_is_found_at_any_depth[nested-lists]
```

Before this commit that mutation would have passed the whole suite.

tests/prompts/cf-ux/providers/claude_provider.py
Oleg67: **RESOLVED** in `f75c71dc` — the field was overclaiming, and the fix is to stop it making a claim at all.

You are right about the mechanism, and I had half-conceded it already: my reply on the companion thread said the flag "knows it could not tell which value was the name, not that the match is wrong." Having written that, I then shipped it under the name `skill_match_ambiguous`, which asserts precisely the judgement I had just disclaimed. A correct `{"command": "cf", "mode": "auto"}` was flagged identically to a genuine false positive — as you say.

**The signal cannot be narrowed**, and that is the point: distinguishing a wrong-field match from a correct call carrying a second identifier requires knowing which key holds the name, which is the knowledge whose absence created this trade in the first place. Any attempt would be the narrowing that risks a certain false failure.

So it is now named for what it observes, not what it suspects (`claude_provider.py:255`, `:379`):

```python
"skill_match_other_candidates": list(trace.other_candidates),   # e.g. ["auto"]
```

A list of the other identifiers in the matched call. No boolean, no verdict, nothing for a consumer to misread as "this run was wrong." Filtering still works (`if md["skill_match_other_candidates"]`), and it is strictly more informative than the boolean was — it says *which*.

Documented as the over-approximation it is, in the constant's own comment (`:92-100`) and in the README: every false positive of this class is reported, and so are some perfectly good runs; the run is still scored either way.

Tests: `test_a_second_identifier_beside_a_genuine_call_is_reported_too` (`tests/test_cf_ux_claude_provider.py:581`) pins exactly your scenario — a genuine cf call carrying an unrelated identifier, asserting the answer is still delivered *and* the candidate is listed — and `test_a_sole_candidate_leaves_the_other_candidates_empty` (`:248`) keeps the field able to distinguish at all.

tests/prompts/cf-ux/providers/claude_provider.py
Oleg67: **RESOLVED** in `f75c71dc`. Both halves confirmed by measurement first:

```
{"a":"brainstorming", "b":{"c":"brainstorming"}, "d":"cf"}   -> ['cf', 'brainstorming', 'brainstorming']
{"a":"zebra", "b":"alpha", "c":"cf", "d":"middle"}           -> ['middle', 'cf', 'alpha', 'zebra']
```

Duplicated, and ordered by nothing a reader could predict — the stack pops in LIFO order, so the message's wording depended on dict insertion order and nesting shape. For a line a person reads that is bad; for this suite it is worse, because **two runs over the same transcript could produce different prose for the same finding**, and determinism is a property the rest of the harness is held to.

Fixed at the source rather than at the message, so everything downstream inherits it — `_invoked_names` now accumulates into a set and returns `sorted(names)` (`claude_provider.py:170`, `:187`). The traversal order becomes what it should always have been: an implementation detail nothing depends on. `others` is then derived from an already-sorted, already-unique list.

Test: `test_the_other_candidates_are_deduplicated_and_ordered` (`tests/test_cf_ux_claude_provider.py:561`) uses your case — a value repeated via nesting, plus several distinct ones — and asserts both the metadata list and the rendered warning text:

```python
assert out["metadata"]["skill_match_other_candidates"] == ["alpha", "middle", "zebra"]
assert "['alpha', 'middle', 'zebra']" in caplog.text
```

Asserting the rendered string is deliberate: the finding was about message *content*, so checking only the list would have left the thing you reported untested.

Mutation: reverting to a plain list comprehension fails that test and nothing else.

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of f75c71dc4a4c933df78f4ff4da7ec981fd7ef705 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** This was flagged and confirmed as a real finding in prior rounds, but the diff under review already reflects the fix from commit f75c71dc referenced in the discussion thread: the docstring at claude_provider.py explicitly states 'A top-level input that is not a dict is scanned rather than refused, for the same reason', the README's accepted-trade section documents it, and test_an_input_that_is_not_a_dict_is_scanned_rather_than_refused (parametrized over a bare string, list of strings, and list containing a dict) pins the behavior with an explicit docstring calling it a deliberate behaviour change. The gap the check describes no longer exists in this diff.
- **Reviewer:** The change is explicitly documented in the provider docstring and README, and `test_an_input_that_is_not_a_dict_is_scanned_rather_than_refused` parametrizes bare-string and list top-level inputs, asserting `skill_state == "ran"`.
- **Verifier:** This diff already contains the fix referenced in the discussion thread (commit f75c71dc): the docstring in claude_provider.py explicitly states 'A top-level input that is not a dict is scanned rather than refused, for the same reason...', and the test file adds `test_an_input_that_is_not_a_dict_is_scanned_rather_than_refused`, parametrized over `"cf"`, `["cf"]`, and `["other", {"skill": "cf"}]`, asserting `skill_state == "ran"` for each — directly exercising and pinning the non-dict top-level payload behavior the check asks about.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of f75c71dc4a4c933df78f4ff4da7ec981fd7ef705 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff already contains test_a_nested_identifier_is_found_at_any_depth parametrized over five shapes (one-level, two-levels, list-of-dicts, dict-in-list-in-dict, nested-lists), covering both depth>1 and mixed list/dict container axes claimed by the docstring; confirmed by discussion thread and matches the diff shown.
- **Reviewer:** `test_a_nested_identifier_is_found_at_any_depth` is parametrized with two-level dict, list-of-dicts, dict-list-dict, and nested-list shapes, each asserting the candidate produces `ran` and `skills_invoked == ["cf"]`.
- **Verifier:** The current diff's test_a_nested_identifier_is_found_at_any_depth is parametrized over five shapes (one-level, two-levels, list-of-dicts, dict-in-list-in-dict, nested-lists), directly covering multi-level and mixed list/dict nesting per the diff shown, consistent with the resolved discussion thread citing commit f75c71dc.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of f75c71dc4a4c933df78f4ff4da7ec981fd7ef705 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Resolved in the current diff: the boolean skill_match_ambiguous was replaced with skill_match_other_candidates, explicitly documented as an over-approximation that reports co-occurring identifiers without judging which field was the real name. test_a_second_identifier_beside_a_genuine_call_is_reported_too pins exactly the scenario in question (a genuine {"command": "cf", "mode": "auto"} call) and confirms the run is still scored 'ran' while the extra candidate is listed, matching the intended non-judgmental design.
- **Reviewer:** The obsolete ambiguity boolean is replaced by non-judgmental `skill_match_other_candidates`; `test_a_second_identifier_beside_a_genuine_call_is_reported_too` explicitly pins `{"command":"cf","mode":"auto"}` as scored while listing `auto`, matching the documented intentional over-approximation.
- **Verifier:** The current diff resolved this via commit f75c71dc: the boolean skill_match_ambiguous was replaced with skill_match_other_candidates, explicitly documented as a non-judgmental over-approximation that reports co-occurring identifiers without claiming which one is the real name. test_a_second_identifier_beside_a_genuine_call_is_reported_too (tests/test_cf_ux_claude_provider.py) pins exactly the scenario in the check: a genuine {"command": "cf", "mode": "auto"} call is still scored 'ran' with output delivered, while skill_match_other_candidates lists ['auto'] -- confirming the field intentionally does not distinguish genuine vs accidental matches and doesn't claim to.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of f75c71dc4a4c933df78f4ff4da7ec981fd7ef705 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff shows _invoked_names accumulating candidates into a set and returning sorted(names) (claude_provider.py), eliminating both duplication and traversal-order dependence; test_the_other_candidates_are_deduplicated_and_ordered exercises a repeated-via-nesting candidate plus multiple distinct candidates and asserts both the exact metadata list (['alpha','middle','zebra']) and the exact rendered warning text, matching the discussion thread's resolution in commit f75c71dc.
- **Reviewer:** `_invoked_names` now accumulates candidates in a set and returns `sorted(names)`, and `test_the_other_candidates_are_deduplicated_and_ordered` covers repeated plus multiple distinct candidates while asserting both metadata order and rendered warning content.
- **Verifier:** The current diff already shows _invoked_names accumulating into a set and returning sorted(names) (claude_provider.py), so other_candidates is deduplicated and order-independent; test_the_other_candidates_are_deduplicated_and_ordered in the diff pins both the metadata list (['alpha','middle','zebra']) and the exact rendered warning text for a case with a repeated candidate and multiple distinct candidates, matching the resolution confirmed in the discussion thread (commit f75c71dc).

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -152,6 +152,107 @@ def test_the_request_text_is_not_listed_as_a_skill_name(self, run_provider):
...
+        """The field has to distinguish, or it says nothing."""
+        out, _seen = run_provider(
+            _stream(_skill_call(skill="studio:cf"), _skill_result(), _result()),
+        )
+
+        assert out["metadata"]["skill_match_other_candidates"] == []
+
     def test_a_result_event_with_no_subtype_is_still_an_answer(self, run_provider):
         """Only a subtype that is present and says otherwise means a short turn.
         An unfamiliar event shape must not be able to manufacture failures."""
```
ainetx, line 258: ### Negative-control test for non-matching identifier doesn't pin skill_state=='failed'

**Severity:** Minor

**Problem**
The new test `test_a_single_token_that_is_not_cf_does_not_count_either` (and the pre-existing negative tests it's modeled on) only assert `"output" not in out`, an error substring, and `skills_invoked`, never `out["metadata"]["skill_state"] == "failed"`. The code path does return 'failed' (via the 'a Skill ran but none of them named cf' branch), but that specific diagnostic state distinction is left unpinned by any test.

<details>

**How to reproduce**
1. Run `test_a_single_token_that_is_not_cf_does_not_count_either`.
2. Note assertions only cover output absence, error text, and skills_invoked.
3. Mutate `_skill_trace` to return a bogus state (e.g. 'absent' instead of 'failed') for this branch while keeping the error/detail string identical — the test still passes.

**Expected behavior**
The test should assert `out['metadata']['skill_state'] == 'failed'` to pin the diagnostic contract distinguishing this failure mode from an 'absent' skill call.

**Actual behavior**
The test omits any assertion on `skill_state`, so a regression that mislabels the state would go undetected.

```text
input has no matching cf field -> _skill_trace returns state='failed' -> test asserts error text/output only -> state value unpinned
```

**Impact**
Downstream diagnostics/consumers of skill_state could silently regress without test failure.

**Suggested correction**
Add `assert out['metadata']['skill_state'] == 'failed'` to the negative-control test.

**How to verify**
Add the assertion and confirm it passes against current code; then confirm a mutation to the returned state value causes the test to fail.

</details>

Oleg67: **RESOLVED** in `fa829c10`, and it is a better finding than "an assertion is missing" — I checked whether it was exploitable before fixing it, and it was.

Mislabelling the branch you name, leaving its detail text untouched:

```python
return _SkillTrace(
    "absent", names, inputs,                      # was "failed"
    f"a {_SKILL_TOOL} ran but none of them named {_SKILL_NAME!r}",
)
```

→ **all 52 tests still passed.** The error string reads sensibly either way, because it interpolates whatever state it was handed (`f"cf skill did not run ({state}): {detail}"`), so asserting the detail substring proves nothing about the label. And `skill_state` is the field a downstream consumer reads — the one place a wrong value actually travels.

So I took the finding at the scope you framed it, the new test *and* the pre-existing ones it is modelled on. `skill_state` is now asserted on every negative case, one per branch of the ladder:

| branch | test |
|---|---|
| `Execute skill: cf` in the transcript | `test_the_real_failure_signature_is_recognised` |
| no `Skill` call at all (`absent`) | `test_no_skill_call_at_all_is_an_error_not_an_output` (already pinned) |
| a `Skill` ran, none named `cf` | `test_a_different_skill_running_…`, `…_merely_begins_with_cf_…`, `…_single_token_that_is_not_cf_…` |
| the `cf` call has no result | `test_a_cf_call_with_no_result_at_all_is_not_a_confirmed_run` |
| every `cf` call errored | `test_a_skill_call_that_errored_is_an_error`, `…_false_positive_match_whose_call_errored_…`, `…_failed_cf_call_beside_a_successful_other_skill_…` |

**Five mutations, five caught** — each label swapped in turn, including the `absent` branch mutated in the opposite direction so that one is pinned as `absent` rather than merely as "not the default":

```
CAUGHT | error-mark branch      failed -> absent   (1 failed)
CAUGHT | no-Skill-call branch   absent -> failed   (1 failed)
CAUGHT | none-named branch      failed -> absent   (3 failed)
CAUGHT | no-result branch       failed -> absent   (1 failed)
CAUGHT | every-errored branch   failed -> absent   (3 failed)
```

Tests-only change; the provider is untouched this round.

ainetx: **RESOLVED** -- Confirmed fixed as of fa829c10baf23c5fce1f7a484b15f6b240f12d8f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff already contains the fix confirmed in the discussion thread (fa829c10): test_a_single_token_that_is_not_cf_does_not_count_either and all sibling negative-control tests now assert out["metadata"]["skill_state"] == "failed" (e.g. lines ~308, 320, 342, 352, 366, 423, 433), pinning the diagnostic state distinction that was previously unpinned.
- **Reviewer:** `test_a_single_token_that_is_not_cf_does_not_count_either` now explicitly asserts `out["metadata"]["skill_state"] == "failed"` for the `mode: "cf-generate"` fixture.
- **Verifier:** The diff under review already contains the fix: `assert out["metadata"]["skill_state"] == "failed"` is present in test_a_single_token_that_is_not_cf_does_not_count_either and in every sibling negative-control test in the same class, matching the discussion thread's account that this was resolved in commit fa829c10 and mutation-tested (5/5 caught).

</details>

Issue #197, opened 2026-09-14
make ci: a privacy test false-positives when the container user is root, so local CI cannot pass


SanjeevSolanki:
### `make ci` cannot pass: a privacy test false-positives when the container's user is `root`

`make ci` runs the workflows in Docker through `act`, which runs as `root`. One test asserts
that the OS username does not appear in any `REASON_*` constant:

```python
# tests/test_change_summary_core.py:566
user = os.environ.get("USER") or os.environ.get("USERNAME") or ""
...
if user:
    assert user not in value
```

and one of those constants contains the word:

```python
# skills/studio/scripts/studio/utils/change_summary.py:222
REASON_NO_PROJECT_ROOT = "window carries no project root"
```

`'root' in 'window carries no project root'` is true, so the assertion fails. Nothing has
leaked — the constant is a fixed English string with no username in it.

## Reproduction

No container needed; the username is the whole input.

```
$ USER=sanjeev pytest tests/test_change_summary_core.py -k no_reason_string_leaks
1 passed

$ USER=root     pytest tests/test_change_summary_core.py -k no_reason_string_leaks
1 failed
E   AssertionError: assert 'root' not in 'window carries no project root'
```

Under `make ci` it surfaces as the `Coverage (≥90%)` job failing after the suite runs:
`1 failed, 6034 passed`.

## Why it matters

`make ci` is the documented way to reproduce CI locally, and the contributing flow asks for
it green before pushing. Because `act` runs as `root`, it cannot go green on this repo for
anyone, on any change — so the step is either skipped or worked around, and the local-CI
guarantee is not there.

It is not specific to `root`. Any username that is also an ordinary word fails the same
way: `dev`, `admin`, `test`, `ci`, `git`, `node`, `app`, `user`, `www`. On GitHub's hosted
runners the user is `runner`, which is why it has never failed upstream.

## Shape of a fix

The assertion is substring-only, which is what makes it fragile — it cannot tell a leaked
username from a word that happens to match. Options, in descending order of preference:

1. **Assert on what a leak looks like, not on a bare substring.** A leaked username arrives
   as part of a path or an identifier — `/home/<user>`, `C:\\Users\\<user>`, `~<user>` — so
   match those shapes rather than the raw word. Keeps the test meaningful for a username
   like `root`, which is exactly the case it currently cannot judge.
2. Require a word boundary around the match. Cheaper, and still wrong for a username that
   stands alone as a word in a constant.
[...]

PR #189, opened 2026-09-14
fix(core-infra): pin a rotated log segment by content, not by its name

SanjeevSolanki:
Raised in review of #168 after it had merged, so it lands here rather than there.

## The problem

The rotation link records which file the live log rotated into, and the reader joins that segment only when the live segment's *first* event claims it. The claim compared `backup.name` — a string — so it proved that *some* file with that name had been rotated into, never that the file sitting there at read time was it.

Any file later placed at that path was therefore accepted as the claimed predecessor and merged into the trail in silence. That is the same silent-fabrication failure the claim check was built to prevent, one level down: the check stopped a stale segment being joined and then trusted whatever answered to the name.

## The fix

Rotation records a fingerprint taken while the segment is known-good — its byte length plus a SHA-256 of its first line — and the read verifies both before joining. A mismatch excludes the segment and says so, naming the file.

**Length and first line rather than a digest of the whole file.** A rotated segment is complete and never appended to again, so those two pin it, while hashing 5 MiB on every `read_events` would cost more than the read. Size alone would not do — a same-length substitution is the obvious way past it, and there is a test that performs exactly that swap.

**A claim carrying no fingerprint is still honoured.** Logs rotated before this existed have none, and rejecting them would drop history that is very probably genuine — the same direction every other degradation in this reader takes: give up the guarantee, never the trail.

## Three things the review checklist caught that the gates passed

- a fingerprint that could not be taken was logged at debug, so the failure silently reverted that segment to the name-only check this change replaces. It warns now.
- the claim check reached eleven branch points in one function, the shape that earned a cognitive-complexity finding on this module's reader already. Three named helpers; the entry point is down to six.
- raising that log level **crashed the test runner**: a filesystem name arrives through surrogateescape, a log record holding a lone surrogate cannot be serialised, and `pytest-xdist` killed the worker 1,634 tests in. Every message rendering a path now uses the same redact-then-neutralise transform the records use — seven of them, because "every sink gets the same treatment" had been applied to records, echoes and warnings but not to paths inside messages.

## Verification

Four behaviours, each confirmed by breaking the check and watching only its own test fail: the genuine segment still joins; a substituted file is excluded with a warning; an equal-length substitution is still caught; a pre-fingerprint claim is still honoured.

[...]

PR #198, opened 2026-09-14
fix(core-infra): bound one decision-log event, so a segment stays readable

### LINKED ISSUE #193: decision log: one oversized event makes a whole rotated segment unreadable
### A single oversized event makes a whole rotated segment unreadable

`utils/decision_log.py` bounds a rotated segment at `_MAX_SEGMENT_BYTES`, documented as
"the rotation threshold plus one event". The reader refuses anything larger, on the
argument that such a file was not produced by this log and so cannot be a genuine
predecessor.

Nothing enforces the "one event" half of that. `record()` applies `_redact()` — a home-path
substitution with no truncation — and serialises whatever it was given. Of the typed
wrappers — there are eight — only the gate path caps its fields, and it does so in
`_gate_payload` rather than in `record_gate` itself. The other seven forward their
arguments uncapped: `record_routing`, `record_dispatch`, `record_read`,
`record_validation`, `record_review`, `record_escalation` and `record_invocation`.

So one event over ~64 KiB pushes the next rotated segment past the bound, and the reader
then excludes it — using the check that exists to catch substitution, against a segment
this log wrote itself.

---

## Reproduction

Fill the log to just under the 5 MiB rotation threshold, record one event with a 70 KiB
field through a typed wrapper, then record once more to trigger rotation:

```
before big event:   5,242,879     one byte under _MAX_BYTES (5 MiB)
after big event:    5,313,094     one uncapped 70 KiB event via record_read()
backup size:        5,313,094
_MAX_SEGMENT_BYTES: 5,308,416     the backup is 4,678 bytes over
events in backup:   5,531 'older' events
events read back:   {'live': 1, 'rotate': 1}
```

The write side warns that the segment "is larger than any segment this log rotates, so it
is not fingerprinted", and the read side that it "is excluded from this read". Both lines
are accurate about what happened and wrong about why: the file *was* produced by this log.

## Why it matters

The events are not corrupted, unreachable on disk, or overwritten. They are refused, and
every later `read_events()` refuses them again — so an audit trail loses history
permanently, through ordinary use rather than tampering. Both sides warn — the rotation that the segment is not fingerprinted, each read
that it is excluded — and `record()` returns `True` throughout. What neither line says is
that the segment was one this log wrote itself: both read as "not ours", which is what
makes the loss look correct rather than wrong.

This is the same outcome the fingerprint check was added to prevent, reached from the
other direction.

## Shape of a fix

Three options, in descending order of preference:

1. **Bound the event where every wrapper already passes through.** `record()` is the one
   choke point; capping there covers the seven wrappers that do not cap today and any
   wrapper added later. Capping at call sites instead is what produced this gap — the
   gate path caps, and that was not enough.

   Two details it has to get right. Capping individual fields does not bound an event
   with many fields, so the check belongs on the serialised line. And an event too large
   to record should be recorded as a visible truncation marker rather than dropped, so
   the trail says a record was cut rather than going quiet.

2. Raise `_MAX_SEGMENT_BYTES`. Moves the cliff; does not remove it.

3. Let the reader fail open for segments this log claims. Gives up the substitution
   protection the bound exists for.

## Acceptance

- No event `record()` accepts can produce a segment past `_MAX_SEGMENT_BYTES`.
- An event too large to record leaves a visible marker naming what was cut.
- A test that writes a 70 KiB field, forces a rotation, and asserts the older events are
  still returned by `read_events()`. It fails on current `main`.

Found in review of #189 after it merged, so it lands here rather than there.

**Delivery:** the cap, the marker, and the tests above, as a PR against `main`.

---

*Corrected three times. The figures were also re-measured through the real
`record_read()` wrapper rather than a hand-built `record()` call, which moved them by ~40
bytes; and an earlier version said the only signal was a warning at read time, when the
rotation warns too — wrong in the direction that flattered the finding.*

*Corrected twice before that. The wrapper count first published here was seven with six uncapped;
there are eight, seven of them uncapped — `record_dispatch` was missing from the list.
And the figures first published here were taken from two different reproduction
runs — the byte sizes from one, the event count from another. The block above is a single
run, and the same script against the fixed code returns all 5,531 events.*

SanjeevSolanki:
Closes #193.

`_MAX_SEGMENT_BYTES` is the rotation threshold plus 64 KiB, documented as "plus one event", and the reader refuses anything larger on the grounds that this log did not produce it. Nothing made the second half true: `record()` redacts and serialises whatever it is given, and of the eight typed wrappers only the gate path capped its fields.

So one event over 64 KiB pushed a genuine segment past the bound, and the fingerprint check added in #189 then excluded it — this log's own history refused as a substitution, by the check that exists to catch substitution.

## Reproduced before fixing

A log filled to just under the threshold, one `record_read` with a 70 KiB `source`, one more event to rotate:

```
                      unfixed        with this change
before big event:   5,242,879        5,242,879
after big event:    5,313,094        5,243,136
backup size:        5,313,094        5,243,136
_MAX_SEGMENT_BYTES: 5,308,416        5,308,416
over the bound by:     +4,678          -65,280
events in backup:       5,531            5,531
events read back:   {live:1, rotate:1}   {live:1, older:5531, read:1, rotate:1}
```

One run of one script against both versions, so the two columns are comparable.

The events are intact on disk — the backup still holds all 5,531 lines — and they are refused again on every later read, not just the first.

**Both sides warn**, and an earlier version of this description said only the read side did. That was wrong and it flattered the finding:

```
… is larger than any segment this log rotates, so it is not fingerprinted
  and its claim will rest on the filename alone          ← at rotation
… is larger than any segment this log rotates, so it is excluded from this read
                                                          ← at each read
```

So an operator watching warnings does see something at rotation time. What neither line says is that the segment was one this log wrote itself: both read as "not ours", which is the part that makes the loss look correct rather than wrong. `record()` returns `True` throughout, so nothing in the return value suggests a problem.

## The fix

The cap goes in `record()`, not at the call sites. Every wrapper already passes through it, so it covers the seven that do not cap today and the ninth nobody has written yet — capping per wrapper is what produced this gap, since the gate path did cap and that was not enough.

Measured on the serialised line rather than per field: an event of many small fields still adds up, and the line is what lands in the file.

An event too large keeps its identifying fields and carries a marker naming its original size and the keys dropped. A dropped event would leave a hole in the trail indistinguishable from a decision never taken, which is the same defect one level down. The marker has a floor of its own, because a payload of many long keys makes the list of those keys larger than the bound that list exists to respect.

No warning, unlike every other place this module gives something up. Those warn because the trail cannot say what happened — a segment excluded from a read leaves no trace inside the log. Here it can, in the record's own place in the order, and a command logging large payloads would otherwise warn on every event it writes.

## The arithmetic is now stated, not implied

The two constants share a number and say so. The largest segment a rotation can produce is one byte under the threshold plus one whole event and its newline, which fits `_MAX_SEGMENT_BYTES` exactly while `_MAX_BYTES + _MAX_EVENT_BYTES` does.

A test asserts that directly, because the scenario tests cannot reach it: an over-long event becomes a short marker, so raising the event cap leaves every scenario green while making a legitimate segment unreadable.

Both sides of the cap are pinned — an event of exactly `_MAX_EVENT_BYTES` is written through unchanged, one byte more is marked. Only the far side of the segment bound had coverage before, which is the finding that prompted this one.

## What this does not do

`A3b` asks what an eviction bound costs, not just whether a value is capped. The cap raises the cost without removing it: at 64 KiB an event, 80 fill a segment and **160 evict any given record** across live and backup. Before the cap, the 500 KB fields that prompted `A3b` needed 20.

Sixteen times harder, not impossible. A size-bounded append-only log can always be flushed by volume, and this change is not a defence against that.

## One pre-existing defect found and deliberately left

`record`'s catch-all latches telemetry off for the rest of the run. That is right for a target it has learned is unwritable, and wrong for one bad payload, which says nothing about the target — a single event with a tuple key silences every later event in the run.

Out of scope here. A test pins today's behaviour so that changing it is deliberate rather than a surprise, and fails loudly with a message saying so.

## Verification

Fourteen mutations, each caught by exactly one test, including: cap removed, cap one byte too large, bound made exclusive, marker replaced by a silent drop, marker's floor removed, and the original defect.

The first version of the boundary test was itself vacuous — it measured a record built by hand and then exercised `record()`, which builds its own with a real timestamp, so a deliberately-wrong cap passed. Mutation testing found that, not review.

### A correction worth stating

The counts above are eight and seven. This change first claimed seven and six, taken from the report that raised it — which listed six uncapped wrappers and omitted `record_dispatch`. That number reached the commit message, the issue and two replies before anyone counted. A first attempt at checking it then got the opposite wrong, reporting capping inside `record_gate`; it is one level down in `_gate_payload`.

A test now pins the count, and pins the thing that actually matters: every typed wrapper reaches `record`, so the bound covers all of them however many there are. Adding a ninth that bypasses `record` fails it.

### On `make ci`

`make ci` does not pass on this repo for anyone, and not because of this change. It runs the workflows in a container as root, and `tests/test_change_summary_core.py::TestPrivacy::test_no_reason_string_leaks_a_path_home_or_username` asserts the OS username is absent from every `REASON_*` constant — while `REASON_NO_PROJECT_ROOT = "window carries no project root"` contains the word. `USER=sanjeev` passes, `USER=root` fails, same code, nothing leaked. Filed as #197.

So the evidence below is CI's own targets run individually, plus a partial `act` run in which `spec-coverage` passed and `coverage` failed only on that test. It is not "make ci green", and is not claimed as such.

### Second review round

Four more findings, three of them in the tests.

The marker named dropped keys with `str()`, so a payload keyed `None` or `True` was reported as `None` and `True` where the record itself would have written `null` and `true`. An auditor greps the marker for a key name, which is its single job. The names are derived by running the same coercion now rather than restating its table.

The test that claimed to verify key conversion serialised to about a hundred bytes and never reached the truncation branch at all. Nothing asserted that `schema`, `ts` and `run_id` survive into the marker — they are how a reader places a cut record in the trail. And every boundary fixture was ASCII, where character count equals byte count, so swapping `_utf8_len` for `len` would have passed all of them while letting an event of 65,536 characters through a 65,536-**byte** cap.

One finding in that round was already fixed: the report that the final fallback leaves identity fields uncapped was written against the commit before they were capped, and reproducing it against the head gives a 716-byte record with `command` at 500.

Gates run locally: `cfs validate` PASS (0 errors, 0 warnings, 241/241); `make test` 6049 passed; spec-coverage 90.9% coverage, 0.4602 granularity; pylint, vulture-ci and test-coverage clean.


skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -269,6 +277,74 @@ def _redact(value: Any) -> Any:
...
+    event it writes.
+
+    """
+    line = json.dumps(record_obj, ensure_ascii=False, default=str)
+    if _utf8_len(line) <= _MAX_EVENT_BYTES:
+        return line
+    # Keep what identifies the event and drop only what made it too big. A reader looking
+    # for this decision still finds it, in its place in the order, saying what is missing.
+    kept = {key: record_obj.get(key) for key in ("schema", "ts", "run_id", "decision_id",
+                                                 "event", "command")}
```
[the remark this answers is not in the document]
SanjeevSolanki: Confirmed and fixed in `d29642ef`. This one falsified a claim I had made in three places.

```
marker with a 200 KB `command`:  200,214 bytes   cap 65,536   over by 134,678
after:                             1,678 bytes
```

`decision_id`, `event` and `command` are all caller-supplied, and the floor I wrote covered `dropped_keys` only — the field the case I had imagined used. The commit message, the PR body and the declared step all said "the marker has a floor of its own" while three fields walked straight past it.

That is the playbook's B10 in its own words: fix every field of its class, not the reported one. I had not run the B list on this change, which is why you found it and I didn't.

The three fields go through `_capped` now (500 chars each), and a test asserts both the whole-marker bound and each field individually, so "small by luck" fails.

skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -112,6 +112,14 @@
 #: a file that cannot be claimed anyway.
 _MAX_SEGMENT_BYTES = _MAX_BYTES + 64 * 1024
 
+#: The largest one serialised event may be, and the "one event" the line above assumes.
+#: Rotation happens at the threshold, so the biggest segment this log can produce is one
+#: byte under it plus one whole event and its newline: `(_MAX_BYTES - 1) + L + 1`. That
+#: is within `_MAX_SEGMENT_BYTES` exactly when `L <= 64 KiB`, which is why the two
+#: constants share a number — the bound below is not a second opinion about size, it is
+#: this one restated for the reader.
+_MAX_EVENT_BYTES = 64 * 1024
```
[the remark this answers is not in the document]
SanjeevSolanki: Correct, and fixed at all three append sites in `d29642ef` rather than the one you flagged.

The write was text-mode with no `newline=`, so on Windows `\n` becomes `\r\n` and a full-size event lands one byte past the segment bound — the original defect, on a platform nobody here tests on. My test asserted `len("\n")`, which baked the same assumption in, so it would have stayed invisible.

All three are now `newline="\n"`. Beyond the arithmetic, that keeps one log byte-identical across platforms, which matters here because the segments are fingerprinted by SHA-256 — a log written on one platform and read on another would otherwise fail its own claim check.

The test now asserts the property across every append site; fixing two of three fails it.

architecture/features/core-infra.md
```diff
@@ -722,15 +722,16 @@ Enables users to install Studio globally, initialize it in any project with sens
...
-5. [x] - `p1` - Append one schema-versioned event (run_id, decision_id, event, command, payload); never raise into the caller; rotate by size; show a one-time notice - `inst-log-record`
+5. [x] - `p1` - Bound one serialised event at 64 KiB — the "one event" the segment bound assumes, enforced where every typed wrapper already passes rather than at the call sites, since only the gate path capped its fields — through `_gate_payload`, not in `record_gate` itself, leaving seven of the eight typed wrappers uncapped — and an uncapped 70 KiB one pushed a genuine segment past the reader's bound and had this log's own history refused as a substitution; measured on the serialised line, because an event of many small fields still adds up and it is the line that lands in the file. An event too large keeps its identifying fields and carries a marker naming its original size and the keys dropped, never silent, since a missing record is indistinguishable from a decision never taken — and the marker has a floor of its own, because a payload of many long keys makes the list of them larger than the bound it exists to respect - `inst-log-event-bound`
```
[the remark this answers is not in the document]
SanjeevSolanki: Right, and it's my wording. The universal cap is in `record()`; `_gate_payload` is only where the *gate path's* field capping already lived. I wrote that sentence while correcting a different wrong claim about the same helper, and introduced this one doing it.

Reworded to say the gate path's field helper was the only capping that existed *before* this change, leaving seven of the eight typed wrappers uncapped.

SanjeevSolanki:
Walking the B list after these findings turned up a third defect neither review caught: `_write_rotation_link` builds its line with a bare `json.dumps` and never passed through the bound. It was safe only because every field happened to be capped by hand — a property a reader has to re-derive and a new field silently breaks, while the segment arithmetic assumes it of *every* line. Now routed through the choke point, with a test that fails if any function writes a line built outside it.

Eleven mutations, each caught by exactly one guard. Sonar's two `S8997` majors are fixed in the same push; I'll confirm against the issue list rather than the quality gate, which currently reports `OK` with both still open.

SanjeevSolanki:
Two corrections to this PR's description, both found by re-auditing it rather than by review. Code unchanged; the head moved only because the commit message carried the same text.

**The reproduction said `record_read` and the script called `record()` directly** with an equivalent payload. Re-run through the real wrapper, which moves the figures by ~40 bytes:

```
                      unfixed        with this change
after big event:    5,313,094        5,243,136
over the bound by:     +4,678          -65,280
```

**"The only signal is a warning at read time" was wrong.** Both sides warn — the rotation that the segment is not fingerprinted, each read that it is excluded — and that makes the defect less severe than I described it. Wrong in the direction that flattered the finding, which is the direction worth distrusting.

What is still true, and is the part that matters: neither warning says the segment was one this log wrote itself. Both read as "not ours", so the loss looks correct rather than wrong, and `record()` returns `True` throughout.

The three remaining claims did verify: rotation reports success, the backup still holds all 5,531 lines, and a second `read_events` refuses them again.

#193 is corrected to match.

skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -269,6 +277,74 @@ def _redact(value: Any) -> Any:
...
+    if _utf8_len(marked) <= _MAX_EVENT_BYTES:
+        return marked
+    # The key names alone can exceed the bound, so the marker needs its own floor: a
+    # record whose own explanation does not fit still has to fit.
+    kept["payload"] = {"truncated": True,
+                       "original_bytes": original_bytes,
+                       "dropped_keys": "omitted: the key names alone exceed the bound"}
+    return json.dumps(kept, ensure_ascii=False, default=str)
+
+
```
ainetx, line 370: ### Final fallback in _bounded_event does not enforce the 64 KiB bound on identity fields

**Severity:** Major

**Problem**
_bounded_event has three return paths. Paths 1 and 2 are guarded by `if _utf8_len(...) <= _MAX_EVENT_BYTES`. Path 3 (the final fallback) replaces only `dropped_keys` with a short string and then returns `json.dumps(kept,...)` with no size check. The `kept` dict still holds the original `command`, `decision_id`, `event` etc. values, which are copied from `record_obj` without capping — `record()` only applies `_redact(command)`, which replaces `$HOME` with `~` but does not bound length. If `command` alone exceeds 64 KiB, the final `json.dumps(kept)` exceeds `_MAX_EVENT_BYTES`, a rotation then writes a segment exceeding `_MAX_SEGMENT_BYTES`, and the reader refuses it as foreign — the same audit-loss failure the PR was written to prevent.

<details>

**How to reproduce**
1. Call `record('read', path=log, command='x' * 70000, payload={})` with a log pre-filled to `_MAX_BYTES - 1` bytes. 2. Observe that `_bounded_event` reaches path 3 (both path-1 and path-2 checks fail because `command` is large). 3. The final `return json.dumps(kept,...)` is called; `kept['command']` is still `'x' * 70000`. 4. The written line is ~70 KiB, exceeding `_MAX_EVENT_BYTES`. 5. On the next `record()` call, rotation triggers and the resulting segment exceeds `_MAX_SEGMENT_BYTES`. 6. `read_events()` refuses the segment.

**Expected behavior**
Every line written by `_bounded_event` is at most `_MAX_EVENT_BYTES` bytes, so that `_MAX_SEGMENT_BYTES` is a genuine upper bound on any segment this log produces.

**Actual behavior**
The final fallback path returns without a size check; an oversized identity field (`command`) causes the returned line to exceed `_MAX_EVENT_BYTES`.

```text
record(command='x'*70000) → record_obj[command]='x'*70000
→ _bounded_event:
path1: _utf8_len(line) > CAP → skip
path2: _utf8_len(marked) > CAP (command still large) → skip
path3: return json.dumps(kept) ← NO SIZE CHECK
kept[command] still 'x'*70000
result > _MAX_EVENT_BYTES
→ segment > _MAX_SEGMENT_BYTES → reader refuses segment
```

**Impact**
A caller passing a large command string (e.g. record_command logging a shell invocation with many arguments) causes a rotation segment that exceeds the reader's bound, which the substitution-integrity check then excludes as foreign — losing legitimate audit history. This is the exact failure mode this PR is intended to close.

**Suggested correction**
After the final `kept["payload"] = {"truncated": True,...}` assignment, cap each identity field with `_capped()` before the final `json.dumps`, or add a post-hoc size guard: if the final result still exceeds `_MAX_EVENT_BYTES`, reduce identity fields to their `_capped()` form and return a minimal fixed-shape record that is guaranteed to fit.

**How to verify**
Add a test calling `_bounded_event` with `command='x'*70000` and an otherwise-empty payload; assert `len(result.encode('utf-8')) <= _MAX_EVENT_BYTES`. Also add the same scenario to `test_no_event_is_ever_written_longer_than_the_cap` by passing `command='x'*70000` to `dl.record()`.

</details>

tests/test_decision_log.py
```diff
@@ -1671,3 +1671,241 @@ class _Small:
...
+        assert json.loads(line)["payload"]["truncated"] is True, line[:200]
+
+    def test_str_is_only_reached_for_keys_that_cannot_fail_it(self) -> None:
+        """Why the marker does not guard `str(key)`, written down so it stays true.
+
+        Listing the dropped keys calls `str()` on author-controlled keys, which looks like
+        a way for one bad payload to raise out of a function whose caller promises never
+        to. It is not reachable: serialisation happens first and rejects any key type
+        outside this set, and `str()` cannot fail on any member of it.
+
```
ainetx, line 1852: ### Key-conversion test never exercises the truncation branch it claims to verify

**Severity:** Minor

**Problem**
`test_str_is_only_reached_for_keys_that_cannot_fail_it` builds a payload `{1: 'a', 2.5: 'b', None: 'c'}`. The full serialized record is approximately 100 bytes — far below `_MAX_EVENT_BYTES = 65536`. `_bounded_event` returns at the first guard (`if _utf8_len(line) <= _MAX_EVENT_BYTES: return line`) before the `dropped_keys` generator or `str(key)` conversion is ever evaluated. The test's assertion verifies normal `json.dumps` key coercion in the pass-through path, not the truncation branch the test name and docstring claim to cover.

<details>

**How to reproduce**
1. Read `test_str_is_only_reached_for_keys_that_cannot_fail_it` in `tests/test_decision_log.py`. 2. Note the payload `{1: 'a', 2.5: 'b', None: 'c'}` produces a serialized record of ~100 bytes. 3. `_MAX_EVENT_BYTES` is 65536. 4. Trace `_bounded_event`: the first `if _utf8_len(line) <= _MAX_EVENT_BYTES` branch returns immediately. 5. The `sorted(_capped(str(key))...)` expression is unreachable.

**Expected behavior**
The test creates an oversized record with non-string keys so `_bounded_event` executes its `dropped_keys` branch, then parses the marker and asserts the key representations are correct.

**Actual behavior**
`_bounded_event` returns at the pass-through guard; the test only verifies `json.dumps` key coercion in the normal path, providing no coverage of `str(key)` in the truncation branch.

```text
test payload (~100 B) → _bounded_event → _utf8_len(line)=~100 <= 65536 → return line (pass-through)
↑ str(key) branch NEVER reached
```

**Impact**
A regression in the `dropped_keys` conversion or sorting inside the truncation path — including the key-name mismatch — cannot be caught by this test. CI passes while the code is incorrect.

**Suggested correction**
Pad the payload with non-string keys until the full serialized record exceeds `_MAX_EVENT_BYTES`, then parse the returned marker and assert `dropped_keys` contains the expected representations.

**How to verify**
After fixing, add an assertion like `assert json.loads(line)['payload']['truncated'] is True` before any `dropped_keys` assertion to confirm the truncation branch was actually reached.

</details>

tests/test_decision_log.py
```diff
@@ -1671,3 +1671,241 @@ class _Small:
...
+        `record()`, which builds its own with a real timestamp — so the assertion and the
+        action were about two different events, and a one-byte-too-large cap passed.
+        """
+        log = tmp_path / "decisions.jsonl"
+        for payload in ({"method": "m", "source": "x" * 70000},
+                        {f"{'k' * 200}{n}": "v" * 400 for n in range(500)},
+                        {f"key{n}": "v" * 900 for n in range(200)},
+                        {"nested": {"deep": ["y" * 90000]}}):
+            dl.record("read", path=log, payload=payload)
+        lines = log.read_text(encoding="utf-8").splitlines()
```
ainetx, line 1786: ### Truncation marker tests do not assert schema, ts, or run_id survival

**Severity:** Minor

**Problem**
_bounded_event copies schema, ts, and run_id into the marker via a dict comprehension. No test records an oversized event with distinct values for those three fields and then reads back and asserts each is present in the marker. A refactoring removing 'run_id' (or 'ts', or 'schema') from the comprehension would pass CI entirely.

<details>

**How to reproduce**
1. In _bounded_event, change the first comprehension to omit 'run_id': `{key: record_obj.get(key) for key in ('schema', 'ts')}`. 2. Run the full test suite. All tests in TestAnEventTooLargeCannotHideASegment pass.

**Expected behavior**
A test records an oversized event with a distinct run_id (and/or ts, schema) value, reads it back via read_events(), and asserts each retained field carries the original value in the marker.

**Actual behavior**
test_an_event_too_large_leaves_a_marker_rather_than_a_hole only checks events[0]['event'] and payload marker fields. test_the_marker_bounds_every_field_it_keeps_not_just_the_payload only checks decision_id, event, command lengths — not presence of schema, ts, or run_id.

```text
_bounded_event removes 'run_id' from comprehension --> marker has no run_id --> audit trail cannot link events to a run --> no test detects the regression
```

**Impact**
A regression silently dropping run_id from truncation markers produces markers that cannot be tied to a specific process run, breaking the audit trail's core linkage without any CI detection.

**Suggested correction**
In test_an_event_too_large_leaves_a_marker_rather_than_a_hole (or a new companion test), use a record with a known distinct run_id (e.g., 'testrun123'), call record() with a monkeypatched _RUN_ID, read the marker back, and assert kept['run_id'] == 'testrun123' and kept['schema'] == SCHEMA_VERSION and 'ts' in kept.

**How to verify**
After adding the assertion, remove 'run_id' from the _bounded_event comprehension and confirm the new test fails.

</details>

skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -269,6 +277,74 @@ def _redact(value: Any) -> Any:
...
+    the order, and a command that logs large payloads would otherwise warn on every
+    event it writes.
+
+    """
+    line = json.dumps(record_obj, ensure_ascii=False, default=str)
+    if _utf8_len(line) <= _MAX_EVENT_BYTES:
+        return line
+    # Keep what identifies the event and drop only what made it too big. A reader looking
+    # for this decision still finds it, in its place in the order, saying what is missing.
+    kept = {key: record_obj.get(key) for key in ("schema", "ts", "run_id", "decision_id",
```
ainetx: ### Truncation marker's `dropped_keys` uses Python `str()` key names instead of JSON-serialized key names

**Severity:** Minor

**Problem**
In `_bounded_event` (line ~329), `dropped_keys` is built with `str(key)` on original Python dict keys. `json.dumps` coerces mapping keys differently: `None` → `"null"`, `True` → `"true"`, `False` → `"false"`. A payload key of `None` appears as `"null"` in the serialized event record but would appear as `"None"` in `dropped_keys`. An auditor cannot reliably map marker entries back to fields in the serialized JSON. The test `test_str_is_only_reached_for_keys_that_cannot_fail_it` itself demonstrates the correct JSON coercion (`{None: 'c'}` → `{"null": "c"}`), confirming the mismatch.

<details>

**How to reproduce**
1. Construct an oversized payload with `{None: 'x' * 70000}`. 2. Call `_bounded_event(record_obj)`. 3. Parse the marker. 4. Observe `dropped_keys` contains `"None"` (from `str(None)`). 5. The serialized event would have had field `"null"` (from `json.dumps`). The names do not match.

**Expected behavior**
`dropped_keys` should list the JSON-serialized key names (`"null"`, `"true"`, `"false"`) that appeared in the serialized record, so an auditor can identify exactly which serialized fields were dropped.

**Actual behavior**
`dropped_keys` uses `str(key)` producing Python-style names (`"None"`, `"True"`, `"False"`), which do not match the JSON field names an auditor would look for in a reference copy of the record.

```text
Python key `None` → json.dumps → field "null" in serialized record
→ str(None) → "None" in dropped_keys ← MISMATCH
```

**Impact**
An auditor or automated tool reconciling a truncation marker against a reference record cannot reliably identify which serialized fields were dropped when non-string JSON-legal keys are present.

**Suggested correction**
Replace `str(key)` with a JSON-consistent coercion: e.g., use `json.dumps(key)` (stripping surrounding quotes for string keys) or serialize to JSON first and read back the key: `json.loads(json.dumps({key: None}))`.

**How to verify**
Add a test that creates an oversized record with `{None: 'x' * 70000}`, parses the marker, and asserts `'null' in payload['dropped_keys']` (not `'None'`).

</details>

skills/studio/scripts/studio/utils/decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 14e93524ae6531c837e7eb9c6ad3399a44f82148 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior accepted-finding said the final fallback in `_bounded_event` left `command` and other identity fields uncapped. The current diff explicitly fixes this: `kept.update({key: _capped(str(record_obj.get(key, ''))) for key in ('decision_id', 'event', 'command')})` caps all three caller-supplied fields before either marker path, so the final fallback's `json.dumps(kept)` is bounded by construction (capped fields + fixed-width engine fields + short literal payload string). The test `test_the_marker_bounds_every_field_it_keeps_not_just_the_payload` verifies this scenario with a 200,000-char `command` and asserts the result is ≤ `_MAX_EVENT_BYTES`. The normative doc's claim now matches the implementation.
- **Reviewer:** The documented 64 KiB serialized-event bound matches _MAX_EVENT_BYTES and the implementation now caps all caller-controlled retained identity fields before the marker fallback; the marker has a second floor for oversized dropped-key lists.

</details>

tests/test_decision_log.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `test_str_is_only_reached_for_keys_that_cannot_fail_it` uses a payload small enough that `_bounded_event` returns on the first (non-truncation) path. The `str(key)` call in `sorted(_capped(str(key)) for key in...)` is never reached. The test's docstring asserts safety of that call, but the branch is entirely bypassed.

Problem (now): `test_str_is_only_reached_for_keys_that_cannot_fail_it` builds a payload `{1: 'a', 2.5: 'b', None: 'c'}`. The full serialized record is approximately 100 bytes — far below `_MAX_EVENT_BYTES = 65536`. `_bounded_event` returns at the first guard (`if _utf8_len(line) <= _MAX_EVENT_BYTES: return line`) before the `dropped_keys` generator or `str(key)` conversion is ever evaluated. The test's assertion verifies normal `json.dumps` key coercion in the pass-through path, not the truncation branch the test name and docstring claim to cover.

</details>

tests/test_decision_log.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The architecture promises 'An event too large keeps its identifying fields.' The implementation correctly copies `schema`, `ts`, `run_id`, `decision_id`, `event`, and `command` into `kept`. However, no test records an oversized event with distinct non-default values for each identifier and then reads back via `read_events()` to assert each field is present and unchanged. Only `event` is indirectly checked (via the event-name list).

Problem (now): _bounded_event copies schema, ts, and run_id into the marker via a dict comprehension. No test records an oversized event with distinct values for those three fields and then reads back and asserts each is present in the marker. A refactoring removing 'run_id' (or 'ts', or 'schema') from the comprehension would pass CI entirely.

</details>

skills/studio/scripts/studio/utils/decision_log.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `_bounded_event` builds `dropped_keys` with `str(key)` on the original Python dict keys, but `json.dumps` coerces those same keys differently: `None` → `'null'`, `True` → `'true'`, `False` → `'false'`. An auditor comparing the marker's `dropped_keys` list against the serialized original event fields will find mismatched names for any payload with non-string keys.

Problem (now): In `_bounded_event` (line ~329), `dropped_keys` is built with `str(key)` on original Python dict keys. `json.dumps` coerces mapping keys differently: `None` → `"null"`, `True` → `"true"`, `False` → `"false"`. A payload key of `None` appears as `"null"` in the serialized event record but would appear as `"None"` in `dropped_keys`. An auditor cannot reliably map marker entries back to fields in the serialized JSON. The test `test_str_is_only_reached_for_keys_that_cannot_fail_it` itself demonstrates the correct JSON coercion (`{None: 'c'}` → `{"null": "c"}`), confirming the mismatch.

</details>

tests/test_decision_log.py
```diff
@@ -1671,3 +1671,308 @@ class _Small:
...
+
+    def test_an_event_of_exactly_the_cap_is_written_through_unchanged(self) -> None:
+        """`<=` makes the cap inclusive, and only the far side of it was covered.
+
+        An event may legitimately reach the cap exactly — that is the largest one the
+        segment arithmetic is built around — so truncating at the bound would replace a
+        recordable event with a marker and lose a payload for nothing. Asserted on
+        `_bounded_event` directly, which is the unit that decides this.
+        """
+        overhead = len(dl._bounded_event(self._record_of("")).encode("utf-8"))
```
ainetx, line 1822: ### Event cap tests are ASCII-only and cannot detect a character-count regression

**Severity:** Minor

**Problem**
All cap-boundary fixtures pass ASCII-only strings ('x' * N, 'k' * N, 'y' * N). For ASCII, character count equals UTF-8 byte count, so any regression changing _utf8_len from len(text.encode('utf-8', 'surrogatepass')) to len(text) passes every test. The implementation deliberately uses ensure_ascii=False so that multibyte characters are stored without escaping, making the byte-vs-char distinction real and consequential, yet no test exercises it at the boundary.

<details>

**How to reproduce**
1. Change `_utf8_len` to `return len(text)` (character count). 2. Run the full test suite. All tests in TestAnEventTooLargeCannotHideASegment pass because every payload is ASCII.

**Expected behavior**
At least one test constructs a payload whose UTF-8 byte length exceeds _MAX_EVENT_BYTES while its character count does not (e.g., a payload of CJK or emoji characters), records it, and asserts the written line stays within _MAX_EVENT_BYTES on the byte level.

**Actual behavior**
No fixture contains multibyte Unicode. The character-count-equal-to-byte-count invariant for ASCII makes the byte-based accounting untestable from the current suite.

```text
_utf8_len regression (len(text) instead of len(text.encode())) --[ASCII payloads]--> byte count == char count --> all assertions pass --> regression ships undetected --> multibyte events can exceed _MAX_EVENT_BYTES in the file
```

**Impact**
A future refactoring to character-count-based length measurement would allow multibyte events past the segment bound, breaking the _MAX_SEGMENT_BYTES invariant for non-ASCII content without any CI signal.

**Suggested correction**
Add one fixture containing multibyte Unicode (e.g., source='日' * N where byte-len exceeds _MAX_EVENT_BYTES but char-len does not) to test_no_event_is_ever_written_longer_than_the_cap, and add a corresponding boundary case to test_an_event_of_exactly_the_cap_is_written_through_unchanged using a multibyte source string calculated to land on the byte boundary.

**How to verify**
After adding multibyte fixtures, change _utf8_len to return len(text) and confirm at least one test fails.

</details>

skills/studio/scripts/studio/utils/decision_log.py
SanjeevSolanki: This one was already fixed two commits before the review landed, so there is nothing to change — but the finding was real and independently correct.

Your comment is against `ab5a8c45`. The identity fields were capped in `d29642ef`, after CodeRabbit raised the same defect on the same commit. Reproducing your exact case against the current head:

```
final-fallback line: 716 bytes   cap 65,536   within? True
command in it: 500 chars
```

Both of you found it on the same commit, from different directions. Flagging the stale-by-two-commits part only so the thread is not read as an outstanding Major.

tests/test_decision_log.py
SanjeevSolanki: Correct, and fixed in `9b49d222`. The payload serialised to about a hundred bytes, so the test exercised the ordinary path while its name claimed the truncation branch.

There is a test that reaches the branch now and asserts the key names in it. The original keeps its narrower claim — that `str()` cannot fail on any key type `json.dumps` accepts — which is what it actually checks.

tests/test_decision_log.py
SanjeevSolanki: Correct, and fixed in `9b49d222`. They were copied through a comprehension nothing asserted, so dropping one would have left a record that says a write was cut without saying which run, when, or under which schema — which is most of what makes a cut record useful.

`schema`, `ts`, `run_id`, `decision_id` and `event` are all asserted with distinct values now.

skills/studio/scripts/studio/utils/decision_log.py
SanjeevSolanki: Correct, and fixed in `9b49d222`. The marker said `None` and `True` where the record itself would have written `null` and `true`:

```
dropped_keys (Python str): ['None', 'True', ...]
JSON would name them     : ['null', 'true']
```

An auditor greps the marker for a key name, which is the single job it has, so a name the record would never have written fails at exactly that.

The names are derived by running the same coercion rather than restating its table, so the two cannot drift apart. Only key types `json.dumps` accepts can reach it — anything else fails the serialisation above first.

tests/test_decision_log.py
SanjeevSolanki: Correct, and fixed in `9b49d222`. Every boundary fixture here was ASCII, where character count and byte count are equal, so swapping `_utf8_len` for `len` passed all of them — while letting an event of 65,536 characters through a 65,536-**byte** cap and past the segment bound.

Two-, three- and four-byte fillers are covered now, and that mutation fails two tests.

skills/studio/scripts/studio/utils/decision_log.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 9b49d2221d1c6061da2da92e25383aea0ea62e1b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior finding is resolved in this diff. `test_str_is_only_reached_for_keys_that_cannot_fail_it` still uses a small payload and does not reach the truncation branch, but its docstring was narrowed to an accurate claim (that `str()` cannot fail on `json.dumps`-accepted key types). A companion test `test_the_keys_a_marker_lists_are_named_the_way_the_record_names_them` is added in the same diff and explicitly forces the truncation branch via `"big": "x" * 70000`, asserts `truncated is True`, and verifies the `dropped_keys` set equals `{"null", "true", "2.5", "big"}` — covering the `_json_key` coercion the prior finding said was never exercised. The discussion confirms this two-test resolution was intentional (commit `9b49d222`).
- **Reviewer:** Although `test_str_is_only_reached_for_keys_that_cannot_fail_it` retains its narrower pass-through claim, `test_the_keys_a_marker_lists_are_named_the_way_the_record_names_them` supplies supported non-string keys plus a 70 KiB value, forces truncation, parses the marker, and verifies the converted dropped-key names.
- **Verifier:** The original finding is resolved. `test_str_is_only_reached_for_keys_that_cannot_fail_it` retains its narrowed (non-truncation) claim, and the companion test `test_the_keys_a_marker_lists_are_named_the_way_the_record_names_them` — added in this same diff — explicitly forces truncation by including `"big": "x" * 70000` alongside non-string keys `{None: "a", True: "b", 2.5: "c"}`, asserts `payload["truncated"] is True` to confirm the branch is reached, and then asserts `set(payload["dropped_keys"]) == {"null", "true", "2.5", "big"}` — directly verifying the `_json_key` coercion inside the `dropped_keys` generator that the prior finding said was never exercised.

</details>

tests/test_decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 9b49d2221d1c6061da2da92e25383aea0ea62e1b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** This diff adds `test_the_marker_keeps_the_fields_that_identify_the_event`, which records an oversized event with distinct non-default values (`schema=99`, `run_id='run-abc'`, `decision_id='dec-1'`, `event='read'`, `ts='2026-09-15T00:00:00+00:00'`) and asserts each field is present and unchanged in the parsed marker via `json.loads`. Removing any of those keys from the `kept` comprehension would now fail CI. The posted this finding is resolved in this PR.
- **Reviewer:** The truncation marker retains the identifying fields, and `test_the_marker_keeps_the_fields_that_identify_the_event` asserts distinct schema, timestamp, run ID, decision ID, and event values survive serialization.
- **Verifier:** This diff adds `test_the_marker_keeps_the_fields_that_identify_the_event`, which calls `_bounded_event` with distinct non-default values (`schema=99`, `ts='2026-09-15T00:00:00+00:00'`, `run_id='run-abc'`, `decision_id='dec-1'`, `event='read'`) and asserts each field is present and correct in the parsed output via `json.loads`. Removing any key from the `kept` comprehension would now fail CI. The discussion thread confirms the fix. is resolved.

</details>

skills/studio/scripts/studio/utils/decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 9b49d2221d1c6061da2da92e25383aea0ea62e1b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff introduces `_json_key(key)` which derives key names via `next(iter(json.loads(json.dumps({key: 0}))))` — the same coercion `json.dumps` itself applies — so `None` → `'null'`, `True` → `'true'`, `2.5` → `'2.5'`. The `dropped_keys` generator in `_bounded_event` now calls `_capped(_json_key(key))` instead of `str(key)`, and `test_the_keys_a_marker_lists_are_named_the_way_the_record_names_them` exercises an oversized payload with `{None: 'a', True: 'b', 2.5: 'c', 'big': 'x'*70000}` and asserts the truncation branch produces `{'null', 'true', '2.5', 'big'}` — matching exactly what the serialized JSON record would name them. is resolved by this diff.
- **Reviewer:** `_json_key` derives each dropped-key name through `json.dumps` and `json.loads`, and `_bounded_event` uses it for `dropped_keys`; the oversized-payload test verifies `None`, `True`, and `2.5` become `null`, `true`, and `2.5`.
- **Verifier:** The diff introduces `_json_key(key)` which derives names via `next(iter(json.loads(json.dumps({key: 0}))))` — the exact same coercion `json.dumps` applies — so `None` → `'null'`, `True` → `'true'`, `2.5` → `'2.5'`. In `_bounded_event`, `dropped_keys` is built as `sorted(_capped(_json_key(key)) for key in (record_obj.get('payload') or {}))`, replacing any prior `str(key)` usage. The test `test_the_keys_a_marker_lists_are_named_the_way_the_record_names_them` confirms the truncation branch is actually exercised (payload with `{None: 'a', True: 'b', 2.5: 'c', 'big': 'x'*70000}`) and asserts `dropped_keys` == `{'null', 'true', '2.5', 'big'}`, matching exactly what the serialized JSON record would name them. The discussion thread (SanjeevSolanki, commit 9b49d222) confirms the fix is applied. is resolved.

</details>

tests/test_decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 9b49d2221d1c6061da2da92e25383aea0ea62e1b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** This diff includes `test_the_cap_counts_bytes_not_characters`, a parametrized test with 2-byte (\u00e9), 3-byte (\u4e2d), and 4-byte (\U0001f600) Unicode fillers at 30,000 repetitions each, asserting `len(line.encode('utf-8')) <= dl._MAX_EVENT_BYTES`. This directly addresses the ASCII-only gap identified in prior rounds: swapping `_utf8_len` for `len` now fails at least two of those parametrized cases. is resolved as confirmed by the discussion thread and three prior verification rounds reaching PASS.
- **Verifier:** The `test_the_cap_counts_bytes_not_characters` parametrized test with 2-byte (\u00e9), 3-byte (\u4e2d), and 4-byte (\U0001f600) fillers at 30,000 repetitions each proves the byte-vs-char distinction: for the 3-byte case the source field alone is 90,000 UTF-8 bytes (>64KiB), so a regression from `_utf8_len` to `len` (character count, ~30,000) would skip truncation and the assertion `len(line.encode('utf-8')) <= _MAX_EVENT_BYTES` would fail at 90,000+ bytes — confirming the Unicode path and the truncation branch are both exercised and the regression is detectable. The opencode FINDING's residual gap (JSON-escaped ASCII content: quotes, backslashes, newlines) is not a real failure mode: JSON-escaped ASCII special chars remain ASCII, so byte count equals character count for them and no distinct `_utf8_len` vs `len` divergence is possible. Explicitly testing them adds no detection capability beyond what the Unicode fixtures already provide. The truncation marker path is already covered by `test_an_event_too_large_leaves_a_marker_rather_than_a_hole` (with ASCII), and the Unicode test implicitly exercises it for the 3-byte and 4-byte cases since those payloads must be truncated for the byte-count assertion to hold. is resolved.

</details>

ainetx (approved):
Reviewed the event-size bounding logic end-to-end: the 64 KiB cap is wired at the single choke point every writer already passes through, the truncation marker preserves all identifying fields and is itself bounded against oversized identity fields, and the `newline="\n"` fix keeps the segment arithmetic correct on Windows. The test suite is thorough — boundary cases, multi-byte character counting, marker fallback, structural AST guards on append sites and wrapper coverage, and a pinned-literal test that prevents the two constants from silently drifting apart. Nothing blocking found.

PR #192, opened 2026-09-14
release(v1.7.0): assemble the release from v1.6.2 — 21 marked PRs + 29 required dependencies

vasylcf:
Builds `release/v1.7.0` from `v1.6.2` by cherry-pick. **50 commits: the 21 marked `release:v1.7.0` in [Project 56](https://github.com/orgs/constructorfabric/projects/56/views/1?filterQuery=release%3Av1.7.0), plus 29 required ones.**

## Why 50 and not 21

The 21 marked PRs are **not a self-contained set**. Cherry-picking only those produces 17 conflicts out of 21: they sit on top of unmarked foundations. The transitive dependency closure is 47 commits; 3 more were added deliberately (see below).

Foundation blocks and what depends on them:

| Foundation (unmarked) | What it introduces | What cannot build without it |
|---|---|---|
| #79, #80 — decision telemetry | `utils/decision_log.py`, dispatch wrapper | #125, #133, #149, #154, #160, #111 |
| eval-harness: scaffold + 7 review fixes + rule-compliance judge | `commands/eval.py`, `eval-harness.md` | #94, #105 |
| PR #88 series — CDSL/PDSL validator, 13 commits | `utils/error_codes.py`, FAIL rules in `commands/validate.py` | #95, #106, #108, #109, #111, #133, #136, #137 |
| #103 — spec-coverage scope-only granularity | `commands/spec_coverage.py` | #106, #122, #126 |
| `where-used --include-code`, `list-ids` dedupe | `utils/codebase.py` | #106, #133, #154 |
| `init: register source roots` | `architecture/features/core-infra.md` | #111 |

Added beyond the closure, by agreement:

- **#171** — `cf-ux: pin the skill-name matcher's false-positive trade`. This is the follow-up fix to **#161, which is marked for this release**. Without it, v1.7.0 ships with a known false positive.
- **#156** — `fix(ci): make ci fail loudly when act job discovery yields no jobs`. Without it, `make ci` exits 0 silently when it discovers no jobs — a "green" run that validated nothing. Not acceptable on a release branch.
- **#86** — windows-absolute-path test guard, for running the suite on Windows.

## What was left out — 15 commits

| PR | Reason |
|---|---|
| `420aff32` bump to 1.6.2 | Housekeeping bump that followed the v1.6.2 tag; superseded by the future v1.7.0 bump |
| #65, #157 | docs(competitive-analysis) — analysis only, touches no code |
| #151 | docs(adr) — document only |
[...]

Issue #132, opened 2026-09-02
spec-coverage: the granularity floor has under one block of headroom, so new modules fail on arrival


Oleg67:
## The gap

The required granularity floor has almost no headroom, so a new module can fail the gate on arrival regardless of its own quality.

Measured on `main` with the gate's own command:

```
spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.46
```

| Gate | Required | Actual | Margin |
|---|---|---|---|
| System coverage | ≥ 90% | 90.50% | +0.50 pt |
| Granularity | ≥ 0.46 | 0.4606 | **+0.0006** |

Granularity is a line-weighted average over 48,858 lines, so the buffer above the floor is `(0.4606 − 0.46) × 48,858 ≈ 29` line-granularity units. A file of E effective lines at granularity g consumes `(0.46 − g) × E` of it. A 300-line module therefore has to reach g ≈ 0.36 — around eleven `@cpt` blocks — simply to avoid turning the gate red, and the margin has been shrinking: it was +0.0012 one merge earlier.

This is a real cost to contributors. A correct, well-tested module that happens to be lightly blocked fails a gate that says nothing about its correctness, and the failure arrives at CI rather than at authoring time. It also creates a bad incentive: the fastest way back over the line is to delete comment lines, which lowers the denominator without adding a single traced instruction.

## The change

Some combination of:

1. **Report the headroom.** `spec-coverage` prints the score but not the distance to the floor, nor what a given file contributed. Printing the margin — and, on failure, which files consumed it — turns a mysterious red into an actionable one.
2. **Fail at authoring time, not at CI.** A per-file granularity floor exists (`--min-file-granularity`) but is not part of the gate, so a lightly-blocked file is invisible until it drags the system average.
3. **Reconsider the denominator.** Comment and blank lines currently count toward effective lines, which is what makes deleting comments a valid way to pass.

## Non-goals

Lowering the floor, or making granularity advisory. The metric is doing useful work — this is about making its failures legible and its incentives sane.

## Delivery

Happy to implement whichever of the three the maintainers prefer; (1) is the smallest and the one I would start with. Raising it now because three separate pull requests have hit this, and the recovery each time was to add real `@cpt` blocks — which is the right answer, but only discoverable by reading the metric's implementation.

vasylcf:
Confirming this from the v1.7.0 release review (#192 / #200), where the floor bit three times in one cycle — and acting on it, partially.

[...]

PR #200, opened 2026-09-15
fix: resolve the Major and Minor findings from the v1.7.0 release review (#192)

vasylcf:
Fixes the Major findings raised by the automated review on #192 (the v1.7.0 release-assembly PR). Targets `main`; the commit is cherry-picked onto `release/v1.7.0` separately so the two branches stay convergent.

## Four findings, three defects

Two of the four Major findings are **the same finding filed twice**: the text describes `_append_locked` in `decision_log.py`, and the second copy was anchored on `eval_judge.py:296` — a file that contains neither that function nor any `flock` call. So three distinct defects.

Each was reproduced before being fixed, and each has a test that fails without the fix.

## Why this matters for v1.7.0

All three are pre-existing on `main`, but **not** pre-existing for anyone running v1.6.2: the decision log arrived in #79/#80, and the eval harness and semantic spec-coverage pass are new in this release too. v1.7.0 is the first time any of this reaches users.

## 1. `decision_log`: the append lock could hang a command

`_append_locked` took `fcntl.LOCK_EX` with no `LOCK_NB` and no timeout. A sibling process holding the lock — hung, or killed without releasing it — froze the caller inside instrumentation, and `record()`'s `except Exception` cannot intercept it, because a blocking call never raises.

Reproduced directly against `main` by holding the lock and calling `record()`:

```
### main:        lock held; calling record() ...   → killed at 15s (never returned)
### with fix:    lock held; calling record() ...   → record returned: True
```

The fix routes the append through `atomic_io.with_file_lock` with a bounded timeout, degrading to the unlocked append already used off POSIX.

This is not a new pattern — it is the one **already used by the read path in the same module**. `_READ_LOCK_TIMEOUT_SECONDS` was introduced when #136 hit this exact defect in `doc_index`; only the append side was missed. The new bound is named `_APPEND_LOCK_TIMEOUT_SECONDS` to pair with it, and is deliberately shorter (2s vs 5s): a read is something the user asked for and will wait a moment for, while this runs on the hot path of every command and nobody asked for it.

## 2. `eval_harness`: one bad byte aborted the whole suite

`load_run` caught only `OSError` around `read_text(encoding="utf-8")`. `UnicodeDecodeError` is a `ValueError`, so a single phase file of non-UTF-8 bytes propagated through `load_cases` into `run_suite` and aborted the entire evaluation, discarding every other scenario's result.

Now caught alongside `OSError`, matching the contract the function already documents: a declared phase file that cannot be read is simply absent from `phase_texts`.

## 3. `spec_coverage`: NaN silently disabled the gate

`--min-coverage` and its three siblings used a bare `type=float`, so NaN survived parsing. `_requested_thresholds` counts a threshold as demanded only when `value > 0`, and `nan > 0` is `False` — the demanded guarantee silently vanished, and an empty scope exited 0 instead of 2. A misconfigured CI threshold reported success for a gate that never ran.

### Where I deliberately did not follow the review

The review suggested `math.isfinite(v) and v >= 0`. **That lower bound is wrong for this command**, and I only found out because it broke a test: a negative floor is met by any scope, so it demands nothing — which is exactly what `_requested_thresholds` documents and what `test_enforcement_empty_scan::test_a_threshold_no_scope_can_miss_is_not_a_demanded_guarantee` pins with `--min-coverage -5`.

Clamping the range broke that test. Only finiteness is checked now, which fixes the reported defect without changing the accepted range.

## Verification

- Full suite against `main`: **no regressions**. The single differing failure is the known `test_migrate_from_cypilot` order-dependent flake, which reproduces on `main` in 2 runs out of 4 and is unrelated to this change.
- Each new test was confirmed to fail without its fix.
- `make spec-coverage` gate passes (exit 0).

## A marker note, and a fragile gate

`_finite_threshold` carries its own `inst-threshold-argtype` marker with a matching CDSL step in `architecture/features/spec-coverage.md`, rather than joining `inst-build-parser`. That is not cosmetic: folding 25 lines into that already-large block costs marker granularity, and on the `release/v1.7.0` branch it was enough to turn `make spec-coverage` **red** (0.4600 → 0.4599 against a 0.46 floor). Its own narrow block puts it back at 0.4601 there, and leaves `main` at 0.4605 where it started.

### Which points at something worth fixing separately

The repo's own spec-coverage gate runs `--min-granularity 0.46`, and `main` currently sits at **0.4605**. That margin is 0.0005 wide. This PR moves it to 0.4603 — still passing, but any future change that adds a few unmarked lines will turn the gate red for reasons that have nothing to do with that change. This already bit once, in this very change. Worth either raising marker coverage or lowering the threshold to something with real headroom; I have not touched it here.

## Summary by CodeRabbit

- **New Features**
  - Human-readable cascade output now includes Tier-2 cost comparisons, verdicts, and bundle locations.
  - Evaluation scoring resolves required sections by workflow and warns about unrecognized workflow names.
  - Added automated fixing guidance for common documentation and codebase issues.

- **Bug Fixes**
  - Invalid or non-finite coverage thresholds are rejected during parsing.
  - Undecodable evaluation phase files no longer abort complete runs.
  - Evaluation CLI errors return structured JSON responses.
  - Reports no longer expose absolute local filesystem paths.

- **Security**
  - Decision logs receive owner-only permissions and improved integrity protections.


tests/test_decision_log.py
```diff
@@ -1252,3 +1253,37 @@ def test_the_cap_boundary_is_exact(tmp_path: Path, monkeypatch: pytest.MonkeyPat
...
+        log_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+    """record() must return even while another process holds the append lock.
+
+    The append lock used to be a blocking ``flock(LOCK_EX)`` -- the read lock next to
+    it was already bounded. A hung sibling, or one killed without releasing, would
+    freeze the caller inside instrumentation, and record()'s ``except Exception``
+    cannot intercept a call that blocks rather than raises. The wait is now bounded and
+    a timeout degrades to an unlocked append.
+    """
+    fcntl = pytest.importorskip("fcntl")
```
ainetx, line 1271: ### Non-POSIX unlocked-fallback path of `_append_locked` has no platform-independent test

**Severity:** Minor

**Problem**
The new `_append_locked` delegates to `with_file_lock`, which on non-POSIX platforms (Windows) degrades to calling `fn()` directly when `fcntl` is unavailable. The only new test uses `pytest.importorskip('fcntl')` and is entirely skipped on Windows. No test monkeypatches `with_file_lock` to raise `ImportError` or otherwise simulates the non-POSIX path, so a regression in `with_file_lock`'s Windows fallback would go undetected on all CI platforms.

<details>

**How to reproduce**
1. Run the test suite on Windows (or any platform without fcntl). 2. `test_a_lock_held_by_someone_else_does_not_block_the_append` is skipped. 3. No other test exercises `_append_locked` with locking disabled.

**Expected behavior**
A platform-independent test verifies that `_append_locked` still writes the event when locking is unavailable (e.g., by monkeypatching `with_file_lock` to raise `ImportError` or to simulate the no-lock path).

**Actual behavior**
Only one test covers the new behavior, and it is gated on `pytest.importorskip('fcntl')`, making it a no-op on non-POSIX platforms.

```text
Windows CI run
→ test_a_lock_held_by_someone_else_does_not_block_the_append
→ pytest.importorskip('fcntl') → SKIP
→ with_file_lock ImportError path → UNTESTED
```

**Impact**
Regression in `with_file_lock`'s non-POSIX fallback would go undetected on all CI platforms; telemetry writes could silently fail on Windows.

**Suggested correction**
Add a test that monkeypatches `with_file_lock` to simulate an ImportError (or patches `_append_locked`'s import to fail), then asserts the event is still written. This test should run on all platforms.

**How to verify**
Run the new platform-independent test on Linux/macOS and verify it passes without `fcntl` available.

</details>

ainetx:
### `atomic_io.py` module docstring claims separation from `decision_log` that the new import contradicts

**Severity:** Minor

**Problem**
The `atomic_io.py` module docstring explicitly states it was 'kept separate from [decision_log] since it also bakes in log-rotation behavior these two callers don't need.' The diff introduces `from.atomic_io import with_file_lock` inside `decision_log._append_locked`, making `decision_log` a direct consumer of `atomic_io`. The stated architectural rationale is now factually wrong and no comment, docstring update, or import-boundary rule acknowledges the new coupling.

<details>

**How to reproduce**
1. Read `atomic_io.py`'s module docstring. 2. Read the new `from.atomic_io import with_file_lock` in `decision_log.py`. 3. The docstring contradicts the actual import relationship.

**Expected behavior**
`atomic_io.py`'s docstring (or a comment in `decision_log.py`) is updated to acknowledge that `decision_log` is now a consumer, and explains why the coupling is safe.

**Actual behavior**
The docstring still claims separation; future maintainers will not understand why changes to `with_file_lock` affect decision-log write behavior.

```text
atomic_io.py docstring: 'kept separate from decision_log'
↕ CONTRADICTS
decision_log.py: from.atomic_io import with_file_lock
```

**Impact**
Developer confusion; future maintainers may not realize changes to `with_file_lock`'s contract affect decision-log write behavior, especially since `record()`'s `except Exception` silently absorbs failures.

**Suggested correction**
Update `atomic_io.py`'s module docstring to remove or correct the claim about separation from `decision_log`, listing it as a third consumer alongside `okf.py` and `doc_index.py`.

**How to verify**
Confirm the docstring accurately describes all current consumers of `with_file_lock`.

</details>

---
_Original location: skills/studio/scripts/studio/utils/atomic_io.py:9 -- inline anchoring could not be resolved after 1 attempt(s)._

tests/test_decision_log.py
ainetx: **RESOLVED** -- Confirmed fixed as of 98c2fb938dc8acbc4516415a1defc0c5a98c95a4 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `tests/test_decision_log.py:test_append_without_fcntl_still_writes` monkeypatches `fcntl` import failure, calls `record()`, and verifies one persisted event; it now exercises the `with_file_lock` ImportError fallback through `_append_locked`.
- **Verifier:** The search confirms `test_append_without_fcntl_still_writes` exists at tests/test_decision_log.py:353 — a pre-existing (pre-PR) platform-independent test that monkeypatches fcntl import failure, calls record(), and verifies a persisted event. After the refactor, this test's fcntl-import patch now exercises `with_file_lock`'s ImportError fallback path rather than `_append_locked`'s old direct `if fcntl is None:` branch, achieving equivalent coverage. The claude-sdk FINDING was based solely on the new tests visible in the diff and missed this pre-existing test. Opencode's PASS was correct.

</details>

ainetx:
### CONTRIBUTING.md omits the granularity gate that `make spec-coverage` enforces

**Severity:** Minor

**Problem**
CONTRIBUTING.md describes `make spec-coverage` as enforcing '≥90% overall, ≥60% per file' in both the Makefile Targets table and the GitHub Actions CI pipeline list, but makes no mention of the `--min-granularity` threshold. This PR explicitly adds a 10-line Makefile comment framing the 0.45 granularity floor as a documented policy decision (citing issue #132), yet CONTRIBUTING.md is not updated. A contributor reading CONTRIBUTING.md to understand CI requirements will see two thresholds but encounter three enforced checks.

<details>

**How to reproduce**
1. Read CONTRIBUTING.md, locate the `make spec-coverage` description in both the Makefile Targets table and the GitHub Actions pipeline list.
2. Note that only '≥90% overall, ≥60% per file' thresholds are listed.
3. Submit a PR that passes those two thresholds but has granularity < 0.45.
4. The CI gate fails on `--min-granularity 0.45` with no prior documentation of that requirement.

**Expected behavior**
CONTRIBUTING.md should list all three enforced thresholds (≥90% overall coverage, ≥60% per-file coverage, ≥0.45 granularity score) so contributors have an accurate picture of what CI enforces.

**Actual behavior**
CONTRIBUTING.md describes only two thresholds while the Makefile enforces three, and the Makefile comment added by this PR explicitly frames the granularity floor as a permanent, documented policy — making the omission in contributor-facing docs more conspicuous.

```text
Contributor reads CONTRIBUTING.md
→ sees: ≥90% coverage, ≥60% per-file
→ submits PR passing both
→ CI fails: --min-granularity 0.45 (undocumented)
→ contributor has no CONTRIBUTING.md section to consult
```

**Impact**
Contributors whose PRs are rejected by the granularity gate will not find the threshold in the document they were directed to read. The CI failure message names the flag, but CONTRIBUTING.md — the intended reference — remains misleading.

**Suggested correction**
In CONTRIBUTING.md, update both the Makefile Targets table row for `spec-coverage` and the GitHub Actions pipeline list item to read '≥90% overall, ≥60% per file, ≥0.45 granularity score' (or add a note pointing to the Makefile comment for rationale).

**How to verify**
After the update, search CONTRIBUTING.md for 'spec-coverage' and confirm all occurrences mention the granularity threshold alongside the coverage thresholds.

</details>

---
_Original location: CONTRIBUTING.md:292 -- inline anchoring could not be resolved after 1 attempt(s)._

SanjeevSolanki:
Reviewed the changes in my areas (spec-coverage, eval, eval-harness/structural, decision-log, semantic-coverage) — all correct, no objections. Good catches on the `--min-coverage nan` vacuous-pass and the decision-log privacy/lock-hang. 👍

ainetx (approved):
Solid PR across the board — the security hardening (least-privilege CI permissions, `persist-credentials: false`, log file permission restrictions), the NaN-threshold gate fix, the bounded append lock, the `UnicodeDecodeError` containment in the eval harness, the absolute-path leak in `_relative_posix`, and the suite-wide logger isolation in `conftest.py` are all correct and well-tested. One minor documentation gap worth a follow-up:

- **`CONTRIBUTING.md` doesn't mention the `--min-granularity` threshold** — the doc describes two thresholds enforced by `make spec-coverage` but not the third (the 0.45 granularity floor added here). A contributor reading it to understand CI requirements will be surprised when the check fails. Not blocking, but worth a quick one-liner in the Makefile Targets table. ([thread](https://github.com/constructorfabric/studio/pull/200#issuecomment-5674107974))

SanjeevSolanki:
@vasylcf — worked through the `main`→#200 conflict (it's in my decision-log area, so I took a pass). Three files: `decision_log.py`, `test_decision_log.py`, `core-infra.md`. All resolutions are "keep both sides" — your fixes and the newer main ones don't actually collide in behaviour:

1. **Append lock:** kept your bounded `with_file_lock`+timeout, and re-added `newline="\n"` to the inner `_append` — main added it after you branched, and it's the file's own documented invariant, so the refactor shouldn't drop it.
2. **`record()`:** kept main's `line = _bounded_event(record_obj)` **and** your `parent_is_new` — both are used downstream.
3. **`test_decision_log.py`:** both import adds (`re` + `time`) and both test sets (your append-lock tests + main's rotation/event-bound classes).
4. **`core-infra.md`:** merged the numbered steps so every marker resolves — main's `inst-log-event-bound` / `inst-log-segment-bytes` / SHA-256 rotation plus your `inst-log-restrict-perms` (folded in as a creation-time sub-step). `validate` → 0 errors, 241/241.

One **real fix** the merge surfaced: your new `_restrict_to_owner` logs the raw exception (`logger.debug(..., exc)`), which trips the redaction policy test main added after you branched (an `OSError` carries `$HOME`). Changed it to `_describe(exc)` to match every other handler in the module — worth folding into #200.

Gates on the resolved merge: `validate` 0 errors / 241 coverage, decision-log 144 pass, full suite 6110 pass — the only reds are the pylint-4.x checker-plugin tests, which fail the same on plain `main` (env, not us).

Minor, pre-existing (not merge-introduced, your call): the `.lock` literal is 3× → S1192.

Happy to hand over the exact resolved hunks if that's easier than re-resolving — just say.

skills/studio/scripts/studio/utils/decision_log.py
```diff
@@ -390,6 +390,33 @@ def _show_notice_once(path: Path) -> None:
...
+    must never be the thing that breaks a command.
+    """
+    if os.name != "posix":
+        return
+    try:
+        os.chmod(path, mode)
+    except OSError as exc:
+        # `_redact` on the exception too, not only the path: an OSError renders as
+        # "[Errno 13] Permission denied: '/home/<user>/...'", so passing it raw puts
+        # back the $HOME this module strips everywhere else.
```
ainetx, line 414: ### chmod failure in decision-log permission hardening is debug-only and indistinguishable from success to callers

**Severity:** Minor

**Problem**
`_restrict_to_owner` catches `OSError` from `os.chmod` and logs it only at `logger.debug`, then returns normally. `record()` subsequently returns `True` whether `chmod` succeeded or failed. The module's own docstring for this feature states its purpose is to ensure 'a log this module calls private is not left world-readable by the ambient umask' (0o644 from umask 0o022 = world-readable). When `chmod` fails, the privacy guarantee is silently not established, but no signal reaches the caller or the user.

<details>

**How to reproduce**
1. On a POSIX system where `os.chmod` on the log path will fail (e.g. a filesystem mounted noexec/with ACLs overriding chmod, or the path is on a network share). 2. Call `decision_log.record(...)`. 3. Observe: `record()` returns `True`; no warning is emitted; the log file and/or its parent directory retain ambient-umask permissions (0o644/0o755); the debug line is only visible at `--log-level DEBUG`.

**Expected behavior**
A failure to enforce the module's stated privacy guarantee should surface at `WARNING` level so that users on shared systems can know the guarantee was not established, while still not breaking the command (the event write itself should proceed).

**Actual behavior**
`OSError` from `chmod` is caught, debug-logged (invisible by default), and the function returns without any further signal. `record()` returns `True`, identical to the case where `chmod` succeeded.

```text
os.chmod(path, 0o600) → OSError
└─ logger.debug(...) ← invisible to user
└─ return ← no failure signal
record() → True ← looks like success
Log file: 0o644 (world-readable, ambient umask)
```

**Impact**
Users on shared machines unknowingly retain world-readable decision-log files that record which commands ran against which paths, despite the module documenting the log as 'private'. The failure is undetectable at normal log levels and indistinguishable from a successful privacy enforcement.

**Suggested correction**
Change `logger.debug` in `_restrict_to_owner` to `logger.warning` for the OSError path, keeping the same 'telemetry must not break a command' design by still returning without raising. This preserves the non-blocking guarantee while giving users on shared systems a visible indication that the privacy constraint could not be applied.

**How to verify**
After the change, trigger a chmod failure (e.g. mock `os.chmod` to raise `OSError`) and assert `caplog.records` contains a WARNING-level entry from `_restrict_to_owner`; also assert `record()` still returns `True`.

</details>

skills/studio/scripts/studio/utils/fixing.py
```diff
@@ -764,6 +843,7 @@ def _build_fixing_prompt(issue: Dict[str, object], project_root: Optional[Path]
         _prompt_for_cross_ref_coverage,
```
ainetx, line 843: ### No _build_fixing_prompt dispatch tests for the six JIT-readiness error codes routed through the new _prompt_for_jit_readiness builder

**Severity:** Minor

**Problem**
fixing.py inserts _prompt_for_jit_readiness into the _build_fixing_prompt builders list and dispatches EC.TOC_HEADING_DUPLICATE, EC.TOC_HEADING_DEPTH_JUMP, EC.TOC_SECTION_TOO_LONG, EC.TOC_MISSING_DESCRIPTION, EC.CODEBASE_ENTRY_EMPTY, and EC.FILE_TOO_LARGE through it (the first two via explicit if-branches, the last four via a fixed-dict lookup). The diff adds no tests to tests/test_fixing.py for any of these codes at the _build_fixing_prompt level. A future edit that misspells a key in the fixed dict, omits a code, or reorders the builders list would cause _build_fixing_prompt to return None for the affected code, and no test would catch the regression.

<details>

**How to reproduce**
1. Call _build_fixing_prompt({'code': EC.TOC_SECTION_TOO_LONG, 'loc': 'docs/foo.md:1',...}, None). 2. Observe that the result is currently a non-None string. 3. Remove EC.TOC_SECTION_TOO_LONG from the fixed dict in _prompt_for_jit_readiness. 4. Re-run the test suite — no test fails despite _build_fixing_prompt now returning None for that code.

**Expected behavior**
Each of the six JIT-readiness codes should have at least one test calling _build_fixing_prompt with an issue dict carrying that code, asserting the result is not None and contains a key phrase from the expected prompt string.

**Actual behavior**
The test suite has no _build_fixing_prompt-level test for any of EC.TOC_HEADING_DUPLICATE, EC.TOC_HEADING_DEPTH_JUMP, EC.TOC_SECTION_TOO_LONG, EC.TOC_MISSING_DESCRIPTION, EC.CODEBASE_ENTRY_EMPTY, or EC.FILE_TOO_LARGE; the existing TestBuildFixingPromptDirect class only covers CDSL_STEP_UNCHECKED and REF_NO_DEFINITION.

```text
_build_fixing_prompt(issue)
└─ builders list (ordered)
├─ _prompt_for_cross_ref_coverage (tested)
├─ _prompt_for_code_traceability (tested)
├─ _prompt_for_toc_and_warnings (partially tested)
├─ _prompt_for_jit_readiness ← NEW, UNTESTED at dispatch layer
│ ├─ if TOC_HEADING_DUPLICATE → returns string
│ ├─ if TOC_HEADING_DEPTH_JUMP → returns string
│ └─ fixed dict: TOC_SECTION_TOO_LONG, TOC_MISSING_DESCRIPTION,
│ CODEBASE_ENTRY_EMPTY, FILE_TOO_LARGE → or None
└─ _prompt_for_cdsl_structure (tested)
```

**Impact**
Any of the six JIT-readiness codes producing None from _build_fixing_prompt means the fixer emits no remediation guidance when it encounters that warning. The failure is invisible: callers that treat None as 'no prompt' silently swallow it, and no test alerts the team.

**Suggested correction**
Add a parameterized test in tests/test_fixing.py calling _build_fixing_prompt with a minimal issue dict for each code and asserting the result is not None and contains a code-specific key phrase (e.g., 'repeats an earlier one' for TOC_HEADING_DUPLICATE, 'depth jumps' for TOC_HEADING_DEPTH_JUMP, 'Split it under narrower' for TOC_SECTION_TOO_LONG, 'short description' for TOC_MISSING_DESCRIPTION, 'resolves to no files' for CODEBASE_ENTRY_EMPTY, 'past the size ceiling' for FILE_TOO_LARGE).

**How to verify**
Run pytest tests/test_fixing.py -k 'jit_readiness' — all six parameterized cases should pass; remove one key from the fixed dict and confirm the corresponding case fails.

</details>

PR #203, opened 2026-09-15
fix: point default kit source at the constructorfabric org in tests

lobster40:
## Summary

Across the org, docs/config/code sometimes still link to our previous GitHub org `cyberfabric`, which is still live and serves stale content. Since these links resolve successfully, staleness goes unnoticed. Tracking: constructorfabric/governance#1.

In this repo, production already points the default SDLC kit source at the new org:
- `skills/studio/scripts/studio/commands/update.py` — `_KNOWN_KIT_SOURCES["sdlc"] = "github:constructorfabric/studio-kit-sdlc"`
- `skills/studio/scripts/studio/commands/init.py` — `_DEFAULT_KIT_SOURCE = "constructorfabric/studio-kit-sdlc"`

However, three test fixtures still hard-coded the stale `github:cyberfabric/cyber-pilot-kit-sdlc` value as the kit's `source`, which no longer matches what production emits/expects. Fixed:

- `tests/test_update.py:2669` — `TestMigrateKitSources.test_already_has_source` fixture
- `tests/test_kit.py:6550` — `TestCmdKitUpdateCli.test_update_from_github_source` fixture

**Note:** `tests/test_cli_setup_e2e.py:183` was also touched in the initial version of this PR, but that change was a bug — see "Follow-up" below. It has been reverted and now intentionally keeps the legacy `github:cyberfabric/cyber-pilot-kit-sdlc` value.

## Explicitly out of scope (left untouched, on purpose)

The `cf-migrate-*` tooling exists specifically to migrate projects **away from** `cyberfabric`, so `cyberfabric` there is the subject matter, not a stale link:

- `skills/studio/scripts/studio/commands/migrate_from_cypilot.py` (`LEGACY_KIT_SOURCE = "github:cyberfabric/cyber-pilot-kit-sdlc"` is the value the tool matches on and rewrites — renaming it would break the migration path)
- `skills/studio/agents/cf-migrate-scanner.md`, `cf-migrate-planner.md`, `cf-migrate-migrator.md`
- `tests/test_migrate_from_cypilot.py`

Also left untouched per the sweep's hard rules: the four `cyberfabric/*` repo names that 404 under the new org (blocked on constructorfabric/governance#1), SonarCloud `cyberfabric_*` project keys, `github.com/orgs/cyberfabric/projects/N` links, the archived `.prs/81/` snapshots, and historical/prose mentions of `cyberfabric` in ADRs and archived plan output (e.g. `architecture/ADR/0020-*.md`, `.bootstrap/.plans/analyze-full-audit-cypilot/out/*.md`) that document what things *were* named before the rebrand.

## Follow-up (addressing review from @ainetx)

Thanks to @ainetx for catching a real regression in the original version of this PR, plus a maintainability issue. Both addressed in a follow-up commit:

1. **Bug fix (major).** `tests/test_cli_setup_e2e.py`'s `_bootstrap_legacy_project` fixture had its `source` field rewritten to the new `constructorfabric` URL, but that field was the fixture's only remaining legacy signal (`format` and `path` were already the migrated values). With it changed, `had_legacy_kit_signal` in `_rewrite_legacy_kit_entry` (`migrate_from_cypilot.py`) went `False`, so `version = "0"` was never forced during migration and the E2E suite silently stopped exercising that sentinel path. It also modeled an impossible state: no real cypilot install would carry the new org's URL. Reverted to the legacy source, with a comment explaining why it must stay that way.
2. **Maintainability fix (minor).** `tests/test_update.py::TestMigrateKitSources::test_already_has_source` repeated `"github:constructorfabric/studio-kit-sdlc"` as a raw literal, duplicating `_KNOWN_KIT_SOURCES` in `update.py` with no import tying the two together — a future rename would need a manual grep-and-replace and a missed spot would leave the test green while asserting the wrong URL. Now imports and uses `_KNOWN_KIT_SOURCES["sdlc"]` from `studio.commands.update`.
   - `tests/test_kit.py::TestCmdKitUpdateCli::test_update_from_github_source` was also flagged for the same literal, but left as-is: `cmd_kit_update`'s github-source branch only checks `source.startswith("github:")`, so the exact org/repo in that fixture is genuinely arbitrary "some github source" data, not a value the code under test is parameterized by. Coupling it to the SDLC constant would misleadingly imply the exact value matters.

## Test plan

- [x] `pytest tests/test_update.py tests/test_kit.py tests/test_cli_setup_e2e.py -q` → 446 passed
- [x] `pytest tests/ -q` → 6024 passed, 5 skipped, 15 xfailed, 58 subtests passed, 4 failed (pre-existing, unrelated: macOS/APFS rejects illegal-byte-sequence filenames used by `test_change_summary_cmd.py`/`test_change_summary_links.py`/`test_decision_log.py` hostile-filename tests; confirmed these fail identically on the pre-follow-up commit)
- [x] Confirmed locally that with the reverted fixture, `_rewrite_legacy_kit_entry` again produces `version = "0"` in the migrated `core.toml`
- [x] Confirmed production defaults (`_KNOWN_KIT_SOURCES`, `_DEFAULT_KIT_SOURCE`) already point at `constructorfabric`; only test fixtures were stale

Part of constructorfabric/governance#1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Updated test coverage to reflect the current Studio SDLC kit repository.
  * Verified that existing known kit sources remain unchanged during migration.
  * Maintained coverage for legacy project migration and version handling.


tests/test_update.py
```diff
@@ -2666,7 +2666,7 @@ def test_already_has_source(self):
         with TemporaryDirectory() as td:
             config = Path(td)
             toml_utils.dump({
-                "kits": {"sdlc": {"source": "github:cyberfabric/cyber-pilot-kit-sdlc"}},
+                "kits": {"sdlc": {"source": "github:constructorfabric/studio-kit-sdlc"}},
```
ainetx: ### test_kit.py still hardcodes canonical kit source URL as a raw string literal

**Severity:** Minor

**Problem**
The PR partially fixes the URL-duplication problem by importing _KNOWN_KIT_SOURCES in test_update.py, but test_kit.py line 6550 was updated only by a manual string substitution (old cyberfabric URL → new constructorfabric URL), leaving it as an independent raw literal with no link to any authoritative constant. Production constants exist — CONSTRUCTOR_KIT_SOURCE in migrate_from_cypilot.py and _KNOWN_KIT_SOURCES in update.py — but test_kit.py imports neither.

<details>

**How to reproduce**
1. Rename the canonical kit source URL in _KNOWN_KIT_SOURCES['sdlc'] (update.py). 2. Search for all occurrences of the old string. 3. Observe that test_kit.py:6550 is not caught by any import-based dependency, only by manual grep.

**Expected behavior**
test_kit.py should import _KNOWN_KIT_SOURCES (or CONSTRUCTOR_KIT_SOURCE) and use the constant, so a rename propagates automatically.

**Actual behavior**
test_kit.py line 6550 contains the literal string 'github:constructorfabric/studio-kit-sdlc' with no import of any authoritative constant, identical in structure to the test_update.py pattern this PR just fixed.

```text
migrate_from_cypilot.py: CONSTRUCTOR_KIT_SOURCE = "github:constructorfabric/studio-kit-sdlc" (not imported by test_kit.py)
update.py: _KNOWN_KIT_SOURCES["sdlc"] = same string (not imported by test_kit.py)
test_update.py: imports _KNOWN_KIT_SOURCES["sdlc"] <-- fixed by this PR
test_kit.py:6550: "github:constructorfabric/studio-kit-sdlc" <-- still a raw literal
```

**Impact**
A future rename of the canonical kit source URL will silently diverge in test_kit.py with no compile-time or import-resolution signal, replicating exactly the consistency problem this PR is otherwise fixing.

**Suggested correction**
Add `from studio.commands.update import _KNOWN_KIT_SOURCES` (or `from studio.commands.migrate_from_cypilot import CONSTRUCTOR_KIT_SOURCE`) to the relevant test class/method in test_kit.py and replace the raw literal at line 6550 with `_KNOWN_KIT_SOURCES["sdlc"]`.

**How to verify**
After the fix, grep the repo for the literal string 'studio-kit-sdlc' — it should appear only in the canonical constant definitions, not in any test file.

</details>

tests/test_cli_setup_e2e.py
```diff
@@ -180,7 +180,7 @@ def _bootstrap_legacy_project(root: Path, legacy_dir: str = "cypilot", version:
         'format = "CFS"\n'
         'path = "config/kits/sdlc"\n'
         'version = "1.0.0"\n'
-        'source = "github:cyberfabric/cyber-pilot-kit-sdlc"\n',
+        'source = "github:constructorfabric/studio-kit-sdlc"\n',
```
ainetx, line 183: ### E2E fixture now models an impossible legacy-project state, masking the version='0' force-update sentinel path

**Severity:** Major

**Problem**
In migrate_from_cypilot.py, _rewrite_legacy_kit_entry sets had_legacy_kit_signal=True (and thus writes version='0') only when the kit entry has path==LEGACY_KIT_PATH ('config/kits/cypilot-sdlc'), source==LEGACY_KIT_SOURCE ('github:cyberfabric/cyber-pilot-kit-sdlc'), or format=='Cypilot'. The changed _bootstrap_legacy_project fixture writes source='github:constructorfabric/studio-kit-sdlc', path='config/kits/sdlc', format='CFS' — so all three conditions are False, had_legacy_kit_signal stays False, and version='0' is never written. The E2E tests at lines 878/939/967 patch _run_followup_update, so the follow-up update never runs, completely hiding the behavioral difference. Additionally, no real-world cypilot install would ever carry the new constructorfabric URL, making this fixture represent an impossible real-world state.

<details>

**How to reproduce**
1. Read migrate_from_cypilot.py:43-48 — documents version='0' as the sentinel forcing kit force-download. 2. Read _rewrite_legacy_kit_entry:1297-1319 — had_legacy_kit_signal triggers on LEGACY_KIT_PATH, LEGACY_KIT_SOURCE, or format=='Cypilot'. 3. Read _bootstrap_legacy_project in the diff — all three signals are now non-legacy values. 4. Confirm E2E tests patch _run_followup_update, never executing a real kit update.

**Expected behavior**
The E2E fixture writes legacy signals (old source URL, legacy path, or Cypilot format) so that _rewrite_legacy_kit_entry produces version='0' in the migrated core.toml, exercising the force-download sentinel path.

**Actual behavior**
The fixture writes only non-legacy values; had_legacy_kit_signal is never True; version='0' is never written; _run_followup_update is patched so the omission is masked. The E2E suite no longer exercises the documented force-download mechanism.

```text
_bootstrap_legacy_project writes:
source = "github:constructorfabric/studio-kit-sdlc" ← NOT LEGACY_KIT_SOURCE
path = "config/kits/sdlc" ← NOT LEGACY_KIT_PATH
format = "CFS" ← NOT "Cypilot"
↓
_rewrite_legacy_kit_entry: had_legacy_kit_signal = False
↓
version = "0" NEVER written ← sentinel missing
↓
_run_followup_update PATCHED ← behavioral gap fully hidden
```

**Impact**
A regression that prevents the version='0' sentinel from being written (or prevents kit update from treating version='0' as requiring a force-download) can pass the migration E2E suite without detection. The fixture also models an impossible real-world state (a cypilot project with the new constructorfabric source), reducing migration test fidelity.

**Suggested correction**
Restore the fixture's source to 'github:cyberfabric/cyber-pilot-kit-sdlc' (the LEGACY_KIT_SOURCE constant), or add a separate E2E test case that uses the legacy source and asserts that the migrated core.toml contains version='0' before the patched _run_followup_update is called.

**How to verify**
After the fix, confirm that the E2E test asserts the migrated core.toml kit entry contains version='0', and that a test with the new source URL separately confirms the non-sentinel path.

</details>

tests/test_update.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The string 'github:constructorfabric/studio-kit-sdlc' appears as independent raw literals in tests/test_update.py:2669, tests/test_kit.py:6550, and tests/test_cli_setup_e2e.py:183. The same value is defined as CONSTRUCTOR_KIT_SOURCE in migrate_from_cypilot.py:37 and inside _KNOWN_KIT_SOURCES in update.py:1686, but no test file imports or references those constants. This PR itself is the pattern: it is a grep-and-replace across 3 test files with no compile-time check that the literals match the production constants.

Problem (now): The PR partially fixes the URL-duplication problem by importing _KNOWN_KIT_SOURCES in test_update.py, but test_kit.py line 6550 was updated only by a manual string substitution (old cyberfabric URL → new constructorfabric URL), leaving it as an independent raw literal with no link to any authoritative constant. Production constants exist — CONSTRUCTOR_KIT_SOURCE in migrate_from_cypilot.py and _KNOWN_KIT_SOURCES in update.py — but test_kit.py imports neither.

</details>

tests/test_cli_setup_e2e.py
ainetx: **RESOLVED** -- Confirmed fixed as of 54023fae99567194311a06bb461de4808c8572d1 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior FINDING was based on a factually incorrect reading of the diff: the verifier claimed 'the changed fixture writes source=github:constructorfabric/studio-kit-sdlc' but the actual diff shows ONLY a comment was added — the source line remains `'source = "github:cyberfabric/cyber-pilot-kit-sdlc"\n'` (the legacy URL) unchanged. The comment explicitly documents why: this old URL is the fixture's only legacy signal (format and path are already migrated values), so it IS what triggers `had_legacy_kit_signal = True` and forces version = "0" in the migrated core.toml. The sentinel contract is intact; this PR adds defensive documentation of that invariant, not a regression.
- **Reviewer:** The E2E fixture retains `github:cyberfabric/cyber-pilot-kit-sdlc`, so `_rewrite_legacy_kit_entry` sets `had_legacy_kit_signal` and rewrites the version to the `"0"` force-update sentinel; the added comment documents this invariant.

</details>

lobster40: Thanks — sharp eye. This one was a genuine catch and it was the most valuable comment on the PR.

One correction for the record, since the re-verification reads as though the original finding was a misreading: it was not. Against commit `106d3a9` the fixture really did write `source = "github:constructorfabric/studio-kit-sdlc"` — the org-wide sweep this PR belongs to had rewritten that line along with everything else. That killed the fixture's only legacy signal (`format` was already `CFS`, `path` already `config/kits/sdlc`), so `had_legacy_kit_signal` went `False` and the `version = "0"` sentinel stopped being written, while the suite stayed green because `_run_followup_update` is patched.

`54023fa` restores the legacy URL and documents the invariant, which is why the current diff shows only an added comment. Verified the contract directly rather than relying on the suite passing:

```
before: {"format": "CFS", "path": "config/kits/sdlc", "version": "1.0.0", "source": "github:cyberfabric/cyber-pilot-kit-sdlc"}
after:  {"format": "CFS", "path": "config/kits/sdlc", "version": "0",     "source": "github:constructorfabric/studio-kit-sdlc"}
```

So: finding valid, now fixed, and the comment is there so the next sweep does not repeat it.

lobster40: Thanks — sharp eye. This was the most valuable comment on the PR.

Fixed in `54023fa`: the fixture keeps `github:cyberfabric/cyber-pilot-kit-sdlc`, with a comment recording why that value must stay legacy. Verified the sentinel contract directly rather than relying on the suite passing, since `_run_followup_update` is patched:

```
before: {"format": "CFS", "path": "config/kits/sdlc", "version": "1.0.0", "source": "github:cyberfabric/cyber-pilot-kit-sdlc"}
after:  {"format": "CFS", "path": "config/kits/sdlc", "version": "0",     "source": "github:constructorfabric/studio-kit-sdlc"}
```

tests/test_update.py
lobster40: Thanks — sharp eye. Done.

All three sites now reference the canonical constant instead of carrying their own copy of the string:

- `tests/test_update.py:2669` — `_KNOWN_KIT_SOURCES["sdlc"]` (`54023fa`)
- `tests/test_kit.py:6550` — `_KNOWN_KIT_SOURCES["sdlc"]` (`df354ca`)
- `tests/test_cli_setup_e2e.py:183` — intentionally the legacy URL, per the other thread

`pytest tests/test_kit.py -q` → 280 passed.

ainetx: **RESOLVED** -- Confirmed fixed as of df354ca563804ca1032207d2d4e681da1ba6547d -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff fully addresses the duplication concern. `tests/test_kit.py` adds `from studio.commands.update import _KNOWN_KIT_SOURCES` and replaces the hardcoded literal with `_KNOWN_KIT_SOURCES["sdlc"]`; `tests/test_update.py` does the same. The prior verifier's claim that test_kit.py still had a raw string literal is contradicted by the actual diff, which shows both the import and the substitution. `tests/test_cli_setup_e2e.py` retains the old `cyberfabric` legacy URL intentionally (with a new explanatory comment) — it is a different URL used as migration-detection input and cannot use `_KNOWN_KIT_SOURCES["sdlc"]`. The previously accepted this finding no longer applies to the current state of the PR.
- **Reviewer:** The current diff removes the asserted duplication: both `test_kit.py` and `test_update.py` import `_KNOWN_KIT_SOURCES['sdlc']`; the remaining E2E literal is intentionally the distinct legacy migration input, so the posted finding's premise no longer holds.

</details>

ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_cli_setup_e2e.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

ainetx (approved):
Clean, low-risk diff — all three test files now reference the shared `_KNOWN_KIT_SOURCES["sdlc"]` constant instead of hardcoding the `github:cyberfabric/cyber-pilot-kit-sdlc` source string, plus a clarifying comment explaining why that literal must remain in the legacy-fixture builder. This removes duplication and keeps the tests aligned with the single source of truth if the kit source ever changes. No production code paths are touched, test intent and assertions are unchanged, and the added comment accurately documents the fixture's role. Nothing blocking here — approved.

tests/test_cli_setup_e2e.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

PR #199, opened 2026-09-15
feat(execution-plans): let a plan answer a gate, or say why it cannot

SanjeevSolanki:
A gate that resolves without asking has to resolve against something, and nothing in Studio attempted a plan lookup. The record side was already built — `decision_key → value → provenance → status`, every field defaulting to the literal `unspecified` — so the columns existed and nothing filled them. This fills them.

## What a plan can declare

```toml
[[gate_decisions]]
key = "runtime.base-image"
value = "ubuntu-24.04"

[[gate_decisions]]
key = "review.follow-up-depth"
dimension = "register.classification"
[gate_decisions.policy]
serious = "full-review"
normal  = "spot-check"
```

The second form is the one permitted indirection: a policy table keyed by an enum **the plan names**, so a plan can say "follow-up depth depends on classification" once instead of listing every classification an author can imagine.

## Exact match, or ask

No casefold, no strip, no separator folding, no prefix or fuzzy fallback. Normalisation belongs at plan authoring, where a user sees the correction; doing it at resolution readmits similarity matching, which is the failure this design exists to prevent.

Six near misses have tests — case, leading space, trailing space, separator, truncation, plural. None resolve. A seventh asserts they are near misses *of a key that is present*, so editing the fixture cannot quietly make the other six pass for the wrong reason.

## `absent` and `ambiguous` stay apart, though both ask

| | meaning |
|---|---|
| `resolved` | exactly one declaration answers the key |
| `absent` | the plan is silent — nothing is wrong with the plan |
| `ambiguous` | the plan spoke and did not decide |

`ambiguous` covers two declarations for one key, and a policy that covers this dimension with no row for this case. Collapsing the two would file a gap in the plan under the same heading as a question the plan never undertook to answer, and only one of those is worth an author's attention.

**The dimension is declared, never inferred.** A resolver that guessed which enum a policy was keyed on would be doing the inference this refuses everywhere else — and would lose that distinction outright, unable to tell a policy that misses this case from one that was never about this dimension.

## Nothing is cached

The plan is re-read per call, so an edit takes effect on the next lookup and no `resolved` outlives the text it came from. Tested both ways: changing a value changes the answer, removing a declaration stops it resolving.

**An unreadable plan is `ambiguous`, not `absent`**, and warned about. `absent` asserts something about the plan's contents that an unparseable file cannot support, and a plan that will not parse stops every gate that consults it — a caller logging only the status would otherwise see a run of ordinary absences with nothing saying the plan is broken.

[...]

PR #207, opened 2026-09-15
feat(core-infra): assert a reversal is armed before an autonomous edit, or refuse

SanjeevSolanki:
The task this implements rules one thing out in its own words: the filter **must assert that a reversal mechanism is armed, never judge whether an action looks undoable**.

The shipped eligibility modules do the second. They classify an option's *visible action path* and decide whether it reads as reversible — a guess about an action. This asks a question about the environment instead: **if this goes wrong, is there something that puts it back?**

## Three mechanisms, weakest last, first answer wins

| | |
|---|---|
| `worktree` | a linked git worktree — discarding it costs nothing |
| `branch` | a named branch is checked out — the edit can be committed and reset |
| `snapshot` | per-file copies. **Not built** |

The order is a strength ranking, not a preference: each is strictly weaker than the one before, so the first that answers is the strongest available. A test pins it, because "try them in any order" is how the weakest silently becomes the default.

**The gap, stated rather than hidden.** `snapshot` is the mechanism for a project that is not a git repository at all — which is the case the design note names as real, since the audit session that prompted this row was not a repo. So such a project gets a refusal and no way through. That is the safe direction and a real usability cost, and it is the reason this PR is the check only: building the snapshot mechanism is its own piece of work, not a subclause of this one.

## Unknown is refusal, never permission

A failed probe fires exactly when the environment is unusual. Reading it as "git is not saying no, carry on" would arm nothing while allowing everything, in the circumstances least likely to be recoverable.

## Every refusal names all three mechanisms

"No reversal is armed" tells an operator that something stopped and nothing about what to do. Naming a worktree, a branch and a snapshot gives them a list, and the first two usually take seconds.

It reports rather than gating, like the plan lookup in #199 — the caller decides what a refusal means. A test asserts the module has no side effect beyond reading git.

## Two guards keep it from sliding back into judging

It takes **no description of the action**, and a test pins the signature — a parameter carrying what the action does is all it would take. A second sweep rejects any parameter named for an action anywhere in the module.

## What the pre-commit checklist found

Three things the tests could not see, and the worst was not in the code.

**A path could forge its own verdict.** A directory named with a newline rendered a second line in the refusal:

```
| /tmp/…/proj
| ARMED: yes — reversal confirmed is not inside a git work tree…
```

A fabricated outcome, stating the opposite of the real one, inside the check whose whole job is to refuse. The value is stripped of non-printables, whitespace-collapsed and delimited now, and the sweep covers the class rather than the newline that was found: carriage return, tab, escape sequence, vertical tab.

**The refusal carried the username.** A project root is usually under `$HOME`, so it read `/home/<username>/project is not inside a git work tree`. Routed through the ledger's own redact-then-cap transform rather than a second rule that drifts.

**The tests were testing themselves.** They iterate `MECHANISMS`, and nothing pinned it — shrinking it to one entry left all eighteen green while checking one mechanism. Pinned to literals now.

## A claim of mine that was false

The worktree comparison's `resolve()` was documented as guarding against git answering one path relative and one absolute. Asking both flags absolutely made that untrue, and I could construct no case where `resolve()` changes the verdict — including through a symlinked project root.

```
resolve() removed        -> 0 tests fail
bare --git-dir instead   -> 5 tests fail
```

The flags are the guard. The docstring says that now, and says plainly that `resolve()` is belt-and-braces with no defect behind it — a guard justified by an invented scenario reads as evidence the scenario is real.

## Not yet wired, deliberately

Nothing consults this, and no autonomous path in Studio edits files today — those paths select menu options. The lock is fitted before the door is used, which is the only order in which it is cheap. It is whitelisted for the dead-code scan with a removal trigger naming the consumer, rather than described as closing a live hole.

**One conflict this does not resolve.** The shipped `brave-new-world-*` judging path remains authoritative, so two systems now answer one question. Retiring the judgement is assigned to a different task, and replacing a safety rule before its replacement is proven is how it gets weaker mid-refactor. Recorded here so whoever picks that up inherits the finding rather than rediscovering it.

## What the second review round changed

- **`_git()` caught only `ImportError` and `AttributeError`.** Importing a module runs it, so a `SyntaxError` after an edit — or anything raised at import time — went straight through a probe whose whole contract is that unknown is refusal. The one direction this must never fail is open, so the handler is broad and says why. `_said` in the same file had already been broadened for the same reason, and an asymmetry inside one file is how the second one gets missed.
- **The `$HOME` fallback gave up without a word.** When the shared redactor is unavailable this collapses the home directory itself, so `Path.home()` failing is the one path where the username guard stops guarding — and it set `home = ""` and carried on. Every other handler here either refuses or warns.
- **The dead-code whitelist's removal trigger is now checked rather than written.** "Delete these once a consumer appears" was operator intent with no mechanism behind it. A test asserts the correlation directly: the whitelist entries and the first consumer of this module must never exist at the same time.
- **The test's git ran with the caller's environment.** `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE` and friends are honoured even when `cwd` is an isolated `tmp_path`, so a developer with one of them exported would have had these fixtures initialising and committing into whatever repository it named.

One review point is **declined, with the reason**: a failed worktree probe does not force a refusal before the branch is checked. On git 2.30 that probe fails because `--path-format=absolute` does not exist yet, while `symbolic-ref` works — and a named branch with a commit behind it genuinely is a way back. Refusing there would break the module on older git without making any outcome safer; the final verdict is still a refusal whenever nothing was confirmed.

## Verification

Fourteen mutations, each caught by exactly one test: the failed probe read as permission, worktree no longer winning over branch, a detached HEAD reported as armed, the raw-string path comparison, the refusal dropping the mechanism list, the action passed in again, `MECHANISMS` shrunk to one entry, control characters left unstripped, delimiters removed, the path left unredacted, the cap removed, the home-directory failure swallowed again, the import probe narrowed back to `ImportError`, and a consumer of the module appearing while its dead-code whitelist entries stay.

Gates run locally on this branch rebased onto `main` at `8532cec`: `cfs validate` PASS (0 errors, 0 warnings, 242/242); `make test` **6166 passed**; spec-coverage 90.9% coverage, **0.4602** granularity; pylint, vulture-ci, test-coverage, test-gates, self-check, check-versions and validate-kits clean.

`make ci` cannot pass on this repo for anyone (#197), so the evidence above is CI's own targets, run individually.

## Summary by CodeRabbit

* **New Features**
  * Added a safety check that reports whether a project has an available reversal mechanism before autonomous edits.
  * Recognizes linked worktrees, named branches, and snapshots, prioritizing them in that order.
  * Provides clear refusal messages when the project state is unknown, unavailable, or lacks a confirmed reversal option.
* **Documentation**
  * Documented the reversal-checking behavior and added it to the architecture documentation contents.


skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+    # newline forged a second line in the refusal reading `ARMED: yes — reversal
+    # confirmed`: a fabricated verdict, saying the opposite of the real one, inside the
+    # check whose whole job is to refuse. Collapsing whitespace removes the forgery and
+    # the quotes show a reader where the path ends, which a bare interpolation cannot.
+    printable = "".join(ch if ch.isprintable() else " " for ch in text)
+    return '"{}"'.format(" ".join(printable.split()))
+
+
+def _refusal(reason: str) -> ReversalCheck:
+    """Not armed, with what would have satisfied it.
```
ainetx, line 246: ### Path.home() failure silently swallowed in `_said()` fallback path

**Severity:** Minor

**Problem**
Inside `_said()`, when the primary `from.decision_log import capped_text` import fails (triggering the `except Exception as exc:` branch which correctly logs a warning), the code then tries `home = str(Path.home())` in a nested try-except. The `except (OSError, RuntimeError): home = ""` block swallows the exception with no re-raise and no visible-signal call — neither `logger.warning`, `logger.error`, nor any equivalent. This violates the project's silent_exceptions plugin contract that every except block must either re-raise, be a control-flow exempt exception, or route to a visible signal.

<details>

**How to reproduce**
1. Make `decision_log.capped_text` unavailable (e.g., block the import).
2. Configure an environment where `Path.home()` raises `OSError` (e.g., a container with no HOME and no passwd entry).
3. Call `ar._said("/some/path")`.
4. Observe: the `OSError` is swallowed silently; no log record is emitted for the `Path.home()` failure.

**Expected behavior**
The `except (OSError, RuntimeError)` block should emit a `logger.warning(...)` (or equivalent visible signal) before falling back to `home = ""`, consistent with every other except block in this module.

**Actual behavior**
The block executes `home = ""` and continues with no log entry, leaving operators with no indication that the home-collapse step of the fallback redactor also failed.

```text
call _said(value)
→ import decision_log fails → logger.warning() ✓
→ try Path.home()
→ OSError/RuntimeError → home = "" ← NO SIGNAL HERE ✗
→ continues with partial fallback, unreported
```

**Impact**
A degraded redaction path (home not collapsed) runs silently, violating the project's exception-handling contract and potentially hiding a secondary environment problem from operators.

**Suggested correction**
Add a `logger.warning(...)` call in the inner except block before the `home = ""` assignment, e.g.: `except (OSError, RuntimeError) as exc: logger.warning("armed reversal: Path.home() unavailable in redactor fallback, home prefix not collapsed: %s", type(exc).__name__); home = ""`

**How to verify**
After the fix, a test that blocks both `decision_log` import and `Path.home()` should find a warning record for the home failure; the existing `test_the_redactor_fallback_still_collapses_home` test covers only the successful-home-collapse path.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+    A project root is very often under `$HOME`, so the refusal read
+    `/home/<username>/project is not inside a git work tree` and put a username into
+    whatever logs it. That is the leak class the ledger's redactor exists for, and reusing
+    it beats a second rule that drifts -- redaction has to happen before the cap, and that
+    ordering is already settled there.
+    """
+    try:
+        from .decision_log import capped_text  # pylint: disable=import-outside-toplevel
+        text = capped_text(str(value))
+    except ImportError:
```
ainetx: ### `_said()` claims never-raises but only catches `ImportError` — runtime exceptions from `capped_text` escape and break the safety contract

**Severity:** Major

**Problem**
`_said()` (lines 157–164) uses `try: text = capped_text(str(value)) except ImportError: text = str(value)[:200]`. The code comment reads 'never raising out of a safety check'. Once `capped_text` is successfully imported, any exception it raises at call time — `RuntimeError` from `Path.home()`, `OSError` from a filesystem codec, or any exception from internal redactor logic — propagates uncaught. `_git()` in the same file correctly delegates to `_git_query` which 'already carries... the never-raises contract'; `_said()` makes no equivalent arrangement.

<details>

**How to reproduce**
1. Arrange an environment where `capped_text` can be imported but `Path.home()` raises `RuntimeError` (e.g., uid with no `/etc/passwd` entry). 2. Call `armed_reversal(some_non_git_path)`. 3. `_inside_work_tree` returns `(False, False)`. 4. `_said(project_root)` is called. 5. `capped_text` imports successfully then raises `RuntimeError`. 6. Exception propagates out of `armed_reversal()` uncaught.

**Expected behavior**
`armed_reversal()` returns a `ReversalCheck` with `armed=False` in all failure modes, never raises.

**Actual behavior**
`armed_reversal()` raises the exception from `capped_text`, breaking every caller that depends on report-never-raise semantics.

```text
armed_reversal() -> _refusal(f"{_said(root)}...") -> _said() -> capped_text() -> RuntimeError (uncaught) -> propagates to caller
```

**Impact**
Any environment where the redactor's runtime logic raises causes `armed_reversal()` to throw instead of refusing, the opposite of fail-closed.

**Suggested correction**
Wrap the `capped_text` call: `try: text = capped_text(str(value)) except Exception: logger.warning(...); text = str(value)[:200]`.

**How to verify**
Monkeypatch `capped_text` in `_said()` to raise `RuntimeError`; assert `armed_reversal(any_path).armed is False` rather than propagating the exception.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Reuses the repo's bounded query rather than running a second subprocess of its own.
+    A duplicate implementation is how a fix gets lost -- that helper already carries a
+    timeout, the never-raises contract, and the filesystem codec a ref needs, and each of
+    those was added in response to something.
+
+    It also keeps apart the two halves of "no answer", which matters here even though both
+    end in a refusal: "there is no branch" and "git did not run" are different things to
+    tell someone, and only one of them is about their project.
+    """
```
ainetx, line 122: ### Private `_git_query` consumed across modules with no existence assertion — silent permanent refusal if renamed

**Severity:** Minor

**Problem**
`armed_reversal._git()` imports `_git_query` (underscore-prefixed, no stability guarantee) inside `try/except ImportError`. A rename of `_git_query` in `change_summary` causes `ImportError: cannot import name '_git_query'` to be caught, logging a WARNING and returning `(None, True)` on every call. Every subsequent `armed_reversal()` call returns the 'git could not be consulted' refusal. No test in the diff asserts the import succeeds, so this regression would not be caught.

<details>

**How to reproduce**
1. Rename `_git_query` to `_run_git_query` in `change_summary.py`. 2. Run `armed_reversal(any_git_repo)`. 3. `_git()` catches `ImportError`, logs WARNING, returns `(None, True)`. 4. `armed_reversal()` returns refusal 'git could not be consulted'. 5. All autonomous edits refuse permanently with a misleading message.

**Expected behavior**
A test asserts `from studio.utils.change_summary import _git_query` succeeds, making a rename a detectable test failure.

**Actual behavior**
No existence assertion exists; a rename silently causes permanent refusal.

```text
change_summary refactor renames _git_query -> _git() catches ImportError -> logs WARNING -> returns (None, True) -> armed_reversal() always refuses
```

**Impact**
A routine internal refactor of `change_summary` silently disables all reversal checks with no clear root-cause signal beyond a WARNING log.

**Suggested correction**
Add a test: `from studio.utils.change_summary import _git_query; assert callable(_git_query)`. Optionally document `_git_query` in `change_summary` as a stable internal API consumed by `armed_reversal`.

**How to verify**
Run the new test; confirm it fails if `_git_query` is renamed and passes with the correct name.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _refusal(reason: str) -> ReversalCheck:
+    """Not armed, with what would have satisfied it.
+
+    Every refusal names all three mechanisms. An operator reading "no reversal is armed"
+    learns that something stopped and nothing about what to do; one that names a worktree,
+    a branch and a snapshot gives them a list to act on, and the first two usually take
+    seconds.
+    """
+    return ReversalCheck(
```
ainetx, line 257: ### Unimplemented `snapshot` mechanism listed as actionable option in detached-HEAD and git-failure refusals without 'not built' caveat

**Severity:** Minor

**Problem**
`_refusal()` always appends the static suffix 'Any of these would arm one: worktree, branch, snapshot'. Only the non-git path prepends '...snapshot mechanism that would cover that is not built'. The detached-HEAD refusal and git-failure refusal call `_refusal()` with no snapshot caveat. The module docstring explicitly marks snapshot as 'Not implemented'. An operator on a detached HEAD reading 'Any of these would arm one: worktree, branch, snapshot' has no way to know `snapshot` cannot be armed.

<details>

**How to reproduce**
1. In a git repo, detach HEAD: `git checkout --detach HEAD`. 2. Call `armed_reversal(repo)`. 3. `why` contains 'Any of these would arm one: worktree, branch, snapshot'. 4. Operator reads `snapshot` as actionable and searches for it. 5. Nothing exists to arm.

**Expected behavior**
Every refusal that lists `snapshot` clarifies it is not yet implemented.

**Actual behavior**
Detached-HEAD and git-failure refusals list `snapshot` as an equal, actionable option with no 'not built' qualifier.

```text
detached HEAD -> _refusal('not a linked worktree and no named branch') -> appends 'Any of these: worktree, branch, snapshot' -> operator tries to arm snapshot -> nothing exists
```

**Impact**
Operators on a detached HEAD or under a git failure spend time searching for a mechanism that cannot be armed, rather than being directed to checkout a branch or add a worktree.

**Suggested correction**
Move the 'not built' caveat into `_refusal()` itself, e.g. change the suffix to note snapshot is not yet available, or make `_refusal()` accept a flag indicating whether snapshot is relevant.

**How to verify**
Extend `test_every_refusal_names_all_three_mechanisms` to also assert 'not built' appears near 'snapshot' in every refusing path's `why`.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+    except ImportError:
+        # Bounded even without the redactor, and never raising out of a safety check.
+        text = str(value)[:200]
+    # Stripped and delimited, not merely redacted and capped. A directory named with a
+    # newline forged a second line in the refusal reading `ARMED: yes — reversal
+    # confirmed`: a fabricated verdict, saying the opposite of the real one, inside the
+    # check whose whole job is to refuse. Collapsing whitespace removes the forgery and
+    # the quotes show a reader where the path ends, which a bare interpolation cannot.
+    printable = "".join(ch if ch.isprintable() else " " for ch in text)
+    return '"{}"'.format(" ".join(printable.split()))
```
ainetx: ### `_said()` does not escape embedded double-quote characters, breaking the stated delimiter convention for paths containing `"`

**Severity:** Minor

**Problem**
`_said()` (line 171) formats output as `'"{}"'.format(" ".join(printable.split()))`. The `"` character is printable, passes `isprintable()` unchanged, and is not replaced before interpolation. For a path like `/tmp/my"project`, the output is `'"my"project"'` — three double-quote characters where start delimiter, embedded quote, and end delimiter are indistinguishable. The docstring guarantees 'the quotes show a reader where the path ends', which does not hold.

<details>

**How to reproduce**
1. Call `ar._said('some/my"project')`. 2. Output: `'"some/my"project"'`. 3. Count of `"` is 3, not 2; start and end boundaries are ambiguous.

**Expected behavior**
`_said('some/my"project')` produces output with exactly 2 double-quote characters (start and end delimiters only).

**Actual behavior**
`_said('some/my"project')` produces `'"some/my"project"'` with three double-quote characters and no clear boundary.

```text
path with embedded " -> isprintable() passes -> no escape/replace -> '"{}"'.format(...) -> ambiguous 3-quote output
```

**Impact**
An operator reading a refusal for a path containing `"` cannot reliably identify where the path ends, undermining the stated purpose of the delimiter convention.

**Suggested correction**
Replace `"` inside the path before formatting: add `.replace('"', "'")` (or `\"`) on `printable` before the `format()` call. Add a test: `assert ar._said('a"b').count('"') == 2`.

**How to verify**
Run `assert ar._said('a"b').count('"') == 2` and confirm passing after the fix.

</details>

vulture_whitelist.py
```diff
@@ -196,3 +196,17 @@
 # The algo itself is declared in architecture/features/traceability-validation.md (### CPT Reference Scan).
 from studio.utils.cpt_reference_scan import graph_for  # noqa: E402
 graph_for  # noqa: B018
+
+# armed_reversal — the whole module, which nothing imports yet and that is the point.
+# It asserts a reversal mechanism is armed before an autonomous edit; the caller that will
+# consult it is the PDSL side of the same task, and the autonomy paths that exist today
```
ainetx, line 202: ### Vulture whitelist removal trigger is a manual comment with no automated CI enforcement

**Severity:** Minor

**Problem**
vulture_whitelist.py lines 202–209 add whitelist entries for `armed_reversal`, `ReversalCheck.mechanism`, and `ReversalCheck.refused` with a comment 'REMOVAL TRIGGER — delete these entries once a dispatch or eligibility module consults it (grep `armed_reversal` outside its own module and tests).' This is operator-intent documentation, not a machine-checkable constraint. No CI step, Makefile target, or test asserts the correlation between 'a real consumer exists' and 'these entries are absent.' The import graph confirms zero files currently import armed_reversal.py, so the whitelist correctly suppresses warnings today — but there is no mechanism that fires when a consumer appears.

<details>

**How to reproduce**
1. Add a dispatch or eligibility module that imports `armed_reversal`. 2. Run `vulture` / `make vulture-ci`. 3. Observe: vulture passes, whitelist entries remain, no warning is raised about redundant suppressions. 4. Later remove the consumer. 5. Run vulture again: the whitelist entry still suppresses the now-genuinely-dead symbol — no dead-code report is produced.

**Expected behavior**
Either a CI step that greps for `armed_reversal` outside its own module/tests and fails if a real importer exists while the whitelist entry is present, or a test in `test_armed_reversal.py` that asserts the whitelist entry is absent once a non-test importer is detected.

**Actual behavior**
The removal trigger is a free-text comment. Vulture does not warn on redundant whitelist entries; once the entry exists it permanently suppresses dead-code detection for those symbols unless a human removes it manually.

```text
consumer added
└─► whitelist entry stays (no CI fires)
└─► consumer later removed
└─► armed_reversal is dead code
└─► vulture: SILENT (whitelist suppresses it)
```

**Impact**
The vulture dead-code inventory silently becomes inaccurate if a consumer is ever added then removed; genuinely dead symbols in this module go undetected indefinitely. This exact pattern has been confirmed as a defect three prior times on this same file (pr-127/,, ).

**Suggested correction**
Add a test (e.g., in `tests/test_armed_reversal.py` or a dedicated whitelist-hygiene test) that: (a) searches the repo for importers of `armed_reversal` outside the module itself and its tests, and (b) asserts that if any such importer exists, the whitelist entries for `armed_reversal` are absent from `vulture_whitelist.py`. This converts the manual comment into a machine-checked invariant.

**How to verify**
After adding the test: add a dummy consumer import, run the test suite, confirm the hygiene test fails. Remove the whitelist entries, rerun — confirm it passes. Remove the dummy consumer — confirm the whitelist entries' absence is now flagged by vulture.

</details>

tests/test_armed_reversal.py
```diff
@@ -0,0 +1,315 @@
...
+    subprocess.run(["git", *args], cwd=str(cwd), check=True,
+                   capture_output=True, text=True)
+
+
+@pytest.fixture
+def repo(tmp_path: Path) -> Path:
+    """A git repository with one commit, on a named branch."""
+    root = tmp_path / "project"
+    root.mkdir()
+    _git(root, "init", "--quiet", "--initial-branch=main")
```
ainetx: ### repo fixture requires git ≥ 2.28 with no version guard or skip

**Severity:** Minor

**Problem**
`git init --initial-branch=main` was introduced in git 2.28 (July 2020). The `repo` fixture in `tests/test_armed_reversal.py` calls this unconditionally at line 33 with no `subprocess.run(['git', '--version'])` check and no `pytest.skip`. Every test class that depends on the `repo` fixture fails with `subprocess.CalledProcessError` (unrecognised option) on git < 2.28 rather than a clear skip message.

<details>

**How to reproduce**
1. Install git < 2.28 (common in older Debian/Ubuntu LTS, Docker base images). 2. Run `pytest tests/test_armed_reversal.py`. 3. Every test that uses the `repo` fixture fails at setup with `CalledProcessError` from the `git init` call.

**Expected behavior**
Tests skip with a clear message (e.g., `pytest.skip('git >= 2.28 required')`) on older git installations.

**Actual behavior**
`subprocess.CalledProcessError` from `_git(root, 'init', '--quiet', '--initial-branch=main')` propagates as a setup failure for all `repo`-dependent tests.

```text
git init --initial-branch=main
→ git < 2.28: unrecognized option
→ subprocess.run(check=True) raises CalledProcessError
→ pytest reports ERROR in setup (not SKIP)
```

**Impact**
The new test module fails entirely on older developer and CI git environments with a confusing error rather than a clear version-mismatch message.

**Suggested correction**
Add a session-scoped autouse fixture or a module-level check: `v = subprocess.run(['git', '--version'], capture_output=True, text=True).stdout; if not (major, minor) >= (2, 28): pytest.skip('git >= 2.28 required')`. Alternatively, replace `--initial-branch=main` with `git init` followed by `git checkout -b main` (portable to older git).

**How to verify**
Run the test suite on git 2.27 and confirm tests skip rather than error.

</details>

tests/test_armed_reversal.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def test_a_path_cannot_forge_a_second_line_in_the_refusal(self, tmp_path: Path) -> None:
+        """A directory named with a newline wrote its own verdict into the refusal.
+
+        It rendered as a second line reading `ARMED: yes — reversal confirmed`: a
+        fabricated outcome, saying the opposite of the real one, inside the check whose
+        whole job is to refuse. Whitespace is collapsed and the value delimited now.
+        """
+        hostile = tmp_path / "proj\nARMED: yes — reversal confirmed"
+        hostile.mkdir()
```
ainetx: ### Newline-in-filename test fails at OS level on Windows with no platform guard

**Severity:** Minor

**Problem**
`test_a_path_cannot_forge_a_second_line_in_the_refusal` does `hostile = tmp_path / 'proj\nARMED: yes — reversal confirmed'` then `hostile.mkdir()` unconditionally. Windows forbids newlines (and other control characters) in filenames; `Path.mkdir()` raises `FileNotFoundError` or `OSError` before `armed_reversal` is ever called. No `sys.platform` check, `pytest.skipif`, or `pytest.mark.skipif` decorator is present.

<details>

**How to reproduce**
1. On Windows. 2. Run `pytest tests/test_armed_reversal.py::TestTheChecksThatGuardTheseTests::test_a_path_cannot_forge_a_second_line_in_the_refusal`. 3. Observe OSError from `hostile.mkdir()` rather than the intended assertion.

**Expected behavior**
Test is skipped on Windows with a clear message, or uses a platform-conditional path without the newline on Windows.

**Actual behavior**
`hostile.mkdir()` raises `OSError`/`FileNotFoundError` on Windows before the sanitisation logic is exercised.

```text
hostile.mkdir() on Windows
→ OS rejects '\n' in filename
→ OSError raised
→ test fails at setup, sanitisation logic never reached
```

**Impact**
The test suite fails on Windows CI environments before the security-critical sanitisation property is tested.

**Suggested correction**
Add `if sys.platform == 'win32': pytest.skip('Windows forbids control characters in filenames')` before `hostile.mkdir()`, or restructure the test to pass the hostile string directly to `_said()` without filesystem creation (the sanitisation happens in `_said()`, not at the filesystem level).

**How to verify**
Run on Windows and confirm test skips (or restructured test passes) rather than erroring.

</details>

tests/test_armed_reversal.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        import threading  # noqa: PLC0415
+
+        out: list = []
+        worker = threading.Thread(target=lambda: out.append(ar.armed_reversal(tmp_path)),
+                                  daemon=True)
+        worker.start()
+        worker.join(timeout=30)
+        assert not worker.is_alive(), "the reversal check did not return"
+        assert out[0].armed is False, out[0]
```
ainetx, line 351: ### Thread-hang test indexes out[0] without guarding against an empty list

**Severity:** Minor

**Problem**
The daemon thread's lambda is `lambda: out.append(ar.armed_reversal(tmp_path))`. If `armed_reversal()` raises an unhandled exception, `out.append` is never called. The thread exits via exception (daemon threads swallow it), `worker.is_alive()` returns False, `assert not worker.is_alive()` passes, and then `assert out[0].armed is False` raises `IndexError` — an uninformative failure that hides the actual root-cause exception.

<details>

**How to reproduce**
1. Introduce an unhandled exception in `armed_reversal()` (e.g., by breaking an import). 2. Run `test_the_check_returns_rather_than_hanging`. 3. Observe `IndexError: list index out of range` rather than a diagnostic message.

**Expected behavior**
Test reports the actual exception from the worker thread, or at minimum asserts `len(out) == 1` before indexing.

**Actual behavior**
`IndexError` masking the real failure; the test's stated purpose (detecting hangs) also conflates with detecting raises.

```text
armed_reversal() raises
→ out remains []
→ worker exits via exception (suppressed by daemon thread)
→ worker.is_alive() → False
→ assert not worker.is_alive() PASSES
→ out[0] → IndexError (root cause hidden)
```

**Impact**
Diagnostic failures in the no-raise path produce a confusing IndexError in CI output instead of the actual exception.

**Suggested correction**
After `worker.join(timeout=30)`: `assert len(out) == 1, f'worker did not return a result (raised or hung)'` before `assert out[0].armed is False`.

**How to verify**
Inject a raise in `armed_reversal` and confirm the test now reports a meaningful failure rather than `IndexError`.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def _on_named_branch(project_root: Path) -> Tuple[bool, bool]:
+    """Whether a named branch is checked out, so an edit can be committed and reset.
+
+    `symbolic-ref --quiet` answers nothing on a detached HEAD and exits non-zero, which is
+    a valid negative rather than a tool failure -- the distinction the shared helper keeps.
+    """
+    answer, failed = _git(project_root, ["symbolic-ref", "--quiet", "--short", "HEAD"])
+    return bool(answer) and not failed, failed
```
ainetx: ### Branch mechanism arms on an unborn branch where no prior commit exists to reset to

**Severity:** Major

**Problem**
`_on_named_branch` calls `git symbolic-ref --quiet --short HEAD`. On a freshly initialized repository (`git init`) with no commits, this command succeeds and returns the branch name (e.g., 'main'), exit code 0. `bool(answer) and not failed` is True. The function returns armed=True, mechanism=BRANCH, with the message 'a named branch is checked out, so the edit can be committed and reset.' But on an unborn branch there is no prior commit to reset to — the stated recovery path does not exist.

<details>

**How to reproduce**
1. Create a directory, run `git init`, do NOT make any commit. 2. Call `ar.armed_reversal(that_directory)`. 3. Observe `armed=True, mechanism='branch'` and `why='a named branch is checked out, so the edit can be committed and reset'`. 4. Attempt the implied recovery (git reset HEAD~1) — fails with 'fatal: ambiguous argument HEAD~1'.

**Expected behavior**
The branch mechanism only arms when HEAD resolves to an existing commit (i.e., `git rev-parse --verify HEAD` succeeds), so an autonomous edit with a real prior state to return to is confirmed.

**Actual behavior**
`_on_named_branch` treats any non-empty symbolic ref as sufficient evidence that the branch is resettable, including unborn branches with no prior commit.

```text
git init (no commits)
git symbolic-ref --quiet --short HEAD → 'main', exit 0
bool('main') and not False → True
armed_reversal → armed=True, mechanism=BRANCH
'the edit can be committed and reset'
git reset HEAD~1 → fatal: ambiguous argument HEAD~1
```

**Impact**
An autonomous edit can be authorized for a repository with no prior commit, leaving no actual recovery target despite the affirmative report.

**Suggested correction**
In `_on_named_branch` (or in `armed_reversal` after the branch check passes), add a verification step: call `git rev-parse --verify HEAD` and treat a failure as a refusal rather than a positive. This is a read-only check the existing `_git` helper can handle.

**How to verify**
Add a test: `git init` without any commit, call `armed_reversal`, assert `armed=False` with a message indicating no prior commit exists.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+def _git(project_root: Path, args: List[str]) -> Tuple[Optional[str], bool]:
+    """Ask git one read-only question: ``(answer or None, the tool failed)``.
+
+    Reuses the repo's bounded query rather than running a second subprocess of its own.
+    A duplicate implementation is how a fix gets lost -- that helper already carries a
+    timeout, the never-raises contract, and the filesystem codec a ref needs, and each of
+    those was added in response to something.
+
+    It also keeps apart the two halves of "no answer", which matters here even though both
+    end in a refusal: "there is no branch" and "git did not run" are different things to
```
ainetx, line 120: ### Narrow except clause in `_git()` breaks the fail-closed contract on non-ImportError import failures

**Severity:** Major

**Problem**
`_git()` catches only `(ImportError, AttributeError)` around its lazy `from.change_summary import _git_query`. A `SyntaxError`, `RuntimeError`, or any other exception raised during that import propagates uncaught through `_git()` → `_inside_work_tree()` → `armed_reversal()`. This violates the module's stated invariant 'Unknown is refusal, never permission.' By contrast, `_said()` uses `except Exception` with an explicit comment explaining why broad handling is required here — the asymmetry is internal and the pattern is already established in the same file.

<details>

**How to reproduce**
1. Introduce a `SyntaxError` or `RuntimeError` at module level in `change_summary.py` (simulating a broken/partially-written.pyc or a transitive import failure). 2. Call `armed_reversal(some_project_root)`. 3. Observe that instead of returning a `ReversalCheck(armed=False)`, the call raises an uncaught exception through `_git()` → `_inside_work_tree()` → `armed_reversal()`.

**Expected behavior**
`armed_reversal()` returns `ReversalCheck(armed=False,...)` with a warning logged — consistent with the module's 'Unknown is refusal, never permission' contract and matching the pattern `_said()` already establishes.

**Actual behavior**
The exception propagates uncaught as a traceback, bypassing the fail-closed refusal path entirely.

```text
from.change_summary import _git_query
→ raises SyntaxError (not ImportError/AttributeError)
→ NOT caught by except (ImportError, AttributeError)
→ propagates through _git() → _inside_work_tree() → armed_reversal()
→ caller receives traceback instead of ReversalCheck(armed=False)
```

**Impact**
In the one scenario most likely to trigger unusual import failures (broken module state, partial.pyc), the safety guard raises instead of refusing, breaking the module's core invariant and potentially crashing the caller.

**Suggested correction**
Change `except (ImportError, AttributeError) as exc:` in `_git()` to `except Exception as exc:` with a comment matching the one in `_said()` — explaining that broad handling is required because this is a safety-critical path and any exception from the import must be treated as 'tool failed', not propagated.

**How to verify**
Add a test that patches `change_summary` to raise `SyntaxError` on import (in addition to the existing `ImportError` test) and asserts that `armed_reversal()` returns `ReversalCheck(armed=False)` rather than raising.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior accepted finding was factually wrong. The actual code catches `except Exception as exc:` (broad), not just `ImportError`. The comment explicitly documents the design: 'Broad, because the docstring promises this cannot raise and `ImportError` was only the failure imagined: `capped_text` is another module's function and any exception from it would escape a safety check through its reporting path.' Runtime exceptions from `capped_text` are fully caught.
- **Reviewer:** `_said()` now catches `Exception` around both the `capped_text` import and call, logs the degradation, and uses a local bounded fallback.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff introduces `_check_borrowed_helper_exists()` which runs at module import time via `_check_borrowed_helper_exists()` at module scope, calls `hasattr(change_summary, "_git_query")`, and raises `RuntimeError` (not just a warning) if the symbol is gone — making the failure loud and immediate, not silent. `test_the_borrowed_private_helper_is_asserted_to_exist` exercises this guard by `monkeypatch.delattr(change_summary, "_git_query")` and asserts `RuntimeError`. The prior verifier's 'no test asserts the symbol exists' concern is fully addressed by this guard and its test.
- **Reviewer:** Import-time `_check_borrowed_helper_exists()` raises if `change_summary._git_query` is absent, and `test_the_borrowed_private_helper_is_asserted_to_exist` verifies that contract.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `_refusal()` already fixes this in the diff: `offered = ", ".join(m if m != SNAPSHOT else f"{m} (not built)" for m in MECHANISMS)` — ALL refusal paths call `_refusal()`, so every path (detached-HEAD, git-failure, non-git) emits `snapshot (not built)`. The inline comment confirms it was raised in review. `test_a_refusal_does_not_offer_a_mechanism_that_is_not_built` explicitly asserts `"snapshot (not built)"` in both the non-git and detached-HEAD refusals. Prior FINDING verdicts described the code before this fix was applied.
- **Reviewer:** `_refusal()` now renders `snapshot (not built)` for every refusal, and `test_a_refusal_does_not_offer_a_mechanism_that_is_not_built` covers non-git and detached-HEAD paths.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior accepted finding was factually wrong about the implementation. The actual code in the diff uses `return json.dumps(" ".join(printable.split()))` — not a hand-rolled `'"{}"'.format(...)`. `json.dumps()` properly escapes embedded `"` as `\"`, so a path like `'proj" ARMED: yes'` produces valid JSON with the quote escaped. The test `test_a_path_containing_a_quote_cannot_escape_the_delimiter` explicitly verifies round-trip correctness via `assert json.loads(rendered) == 'proj" ARMED: yes'`. No defect exists.
- **Reviewer:** `_said()` uses `json.dumps`, which escapes embedded double quotes, and the added quote test round-trips the rendered value with `json.loads`.

</details>

vulture_whitelist.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `vulture_whitelist.py` (lines 202–209) adds a 'REMOVAL TRIGGER — delete these entries once a dispatch or eligibility module consults it (grep `armed_reversal` outside its own module and tests)' comment. This is documentation of intent, not enforcement. When a consumer of `armed_reversal` is added, `vulture` does not fail — it silently ignores used symbols in a whitelist. The stale whitelist entries then suppress legitimate dead-code detection for other symbols that may be added nearby. This exact pattern has been flagged three times before on this file (pr-127/, pr-127/, pr-127/).

Problem (now): vulture_whitelist.py lines 202–209 add whitelist entries for `armed_reversal`, `ReversalCheck.mechanism`, and `ReversalCheck.refused` with a comment 'REMOVAL TRIGGER — delete these entries once a dispatch or eligibility module consults it (grep `armed_reversal` outside its own module and tests).' This is operator-intent documentation, not a machine-checkable constraint. No CI step, Makefile target, or test asserts the correlation between 'a real consumer exists' and 'these entries are absent.' The import graph confirms zero files currently import armed_reversal.py, so the whitelist correctly suppresses warnings today — but there is no mechanism that fires when a consumer appears.

</details>

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The `repo` fixture explicitly does NOT use `--initial-branch`. Its own comment reads: '`--initial-branch` arrived in git 2.28 (2020). Set the name after `init` instead.' The code calls `_git(root, "init", "--quiet")` then `_git(root, "checkout", "--quiet", "-B", "main")`. The prior verifier FINDING was based on a misread; the compatibility workaround is already in place and documented inline.
- **Reviewer:** The fixture does not use `git init --initial-branch`; it uses portable `git init --quiet` followed by `git checkout -B main`, with a comment explaining the older-Git compatibility.

</details>

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The test wraps `hostile.mkdir()` in `try/except OSError: pytest.skip("this filesystem does not allow a newline in a name")` — on Windows, where `\n` in a filename is illegal, `mkdir()` raises `OSError` and the test skips gracefully rather than failing with a confusing error. The prior verifier FINDING incorrectly stated 'no guard is present'; the `try/except OSError` with `pytest.skip` is the correct and present mechanism.
- **Reviewer:** The newline-path creation is wrapped in `try/except OSError` and calls `pytest.skip` when the filesystem disallows the name, including Windows.

</details>

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The prior accepted finding was wrong — the fix is already present in the diff. The `_run()` function wraps the call in `try: out.append(ar.armed_reversal(tmp_path)) except BaseException as exc: out.append(exc)`, guaranteeing `out` is non-empty regardless of whether the call succeeds or raises. The subsequent `assert not isinstance(out[0], BaseException)` then produces a meaningful error message. The code's own comment documents the exact history: 'Appending only on success left `out` empty when the call raised, so the assertion below failed with an IndexError and hid what actually went wrong.'
- **Reviewer:** The thread target catches `BaseException` and appends it to `out`, then the test explicitly asserts the captured result is not an exception before accessing `.armed`.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of 968b2405e133427bfdc9a79666fe5ec66ad475f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The fix is already present in this diff. `_on_named_branch` explicitly runs `git rev-parse --verify --quiet HEAD` after `symbolic-ref` succeeds, returning `(False, failed)` when HEAD resolves to nothing. The code comment says 'Raised in review as Major', and `TestWhatReviewFoundOnTheFirstPush.test_a_repository_with_no_commits_does_not_arm` covers exactly the unborn-branch case and asserts `armed is False, mechanism is None`.
- **Reviewer:** After confirming a symbolic branch, `_on_named_branch()` runs `git rev-parse --verify --quiet HEAD` and returns armed only when that pre-edit commit exists; the new no-commits test covers the unborn-branch case.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: Re-verified against the current code -- this write-up has been updated. Severity: Minor -> Major.

<details>
<summary>Why</summary>

Problem (was): Both `_git` (importing `_git_query` from `change_summary`) and `_said` (importing `capped_text` from `decision_log`) catch only `ImportError`. If either dependency module raises `SyntaxError`, `AttributeError`, `TypeError`, or any other exception during import (e.g., due to a partially-written.pyc, a broken transitive dependency, or a regression in those modules), the exception propagates uncaught through the safety guard. This defeats the module's documented 'never-raises' and fail-closed contract.

Problem (now): `_git()` catches only `(ImportError, AttributeError)` around its lazy `from.change_summary import _git_query`. A `SyntaxError`, `RuntimeError`, or any other exception raised during that import propagates uncaught through `_git()` → `_inside_work_tree()` → `armed_reversal()`. This violates the module's stated invariant 'Unknown is refusal, never permission.' By contrast, `_said()` uses `except Exception` with an explicit comment explaining why broad handling is required here — the asymmetry is internal and the pattern is already established in the same file.

</details>

tests/test_armed_reversal.py
```diff
@@ -0,0 +1,421 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "skills/studio/scripts"))
+
+from studio.utils import armed_reversal as ar  # noqa: E402
+
```
ainetx, line 21: ### New test git launcher duplicates an incomplete git-env sanitization list instead of reusing the shared helper

**Severity:** Minor

**Problem**
tests/test_armed_reversal.py defines its own `_clean_env()`/`_REDIRECTS` (8 variables: GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, GIT_COMMON_DIR, GIT_OBJECT_DIRECTORY, GIT_ALTERNATE_OBJECT_DIRECTORIES, GIT_CEILING_DIRECTORIES, GIT_NAMESPACE) for its local `_git()` subprocess helper used by every fixture, instead of importing/reusing `change_summary`'s shared sanitized-environment helper. The hardcoded list omits git config-injection variables (GIT_CONFIG_NOSYSTEM, GIT_CONFIG_PARAMETERS/COUNT/KEY_*/VALUE_*, GIT_DISCOVERY_ACROSS_FILESYSTEM) that this repo's own history required adding to the production module's redirect-variable list for the identical attack class.

<details>

**How to reproduce**
1. Export an ambient `GIT_CONFIG_PARAMETERS` or `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0` set that overrides a git config value (e.g. `core.worktree` or `user.email`).
2. Run tests/test_armed_reversal.py's `repo` fixture, which calls `_git()` for `init`/`checkout`/`config`/`commit`/`worktree add`.
3. Observe the fixture's git invocations inherit the injected config unmodified, since `_clean_env()` never strips `GIT_CONFIG_*`.

**Expected behavior**
The test helper should reuse the repo's shared `_git_env()`/equivalent sanitization helper (or its full variable list) so fixture git invocations are isolated from ambient config-injection the same way production probes are.

**Actual behavior**
`_clean_env()` only strips a fixed local list of 8 redirect variables, leaving all `GIT_CONFIG_*` injection vectors inherited into every fixture git call.

```text
ambient GIT_CONFIG_PARAMETERS set -> pytest calls repo fixture -> _git() -> subprocess.run(env=_clean_env()) -> GIT_CONFIG_PARAMETERS still present -> git init/commit behavior altered by injected config
```

**Impact**
Under unusual ambient environments (some CI systems, developer shells exporting GIT_CONFIG_*), fixture setup could be silently influenced by unintended git configuration, producing flaky or misleading test results; the duplicated list will also drift from the production list as the latter is extended.

**Suggested correction**
Have the test's `_git()` helper import and reuse `change_summary`'s shared git-env sanitization helper (or its full variable list) instead of maintaining a separate, narrower `_REDIRECTS` tuple.

**How to verify**
Set GIT_CONFIG_COUNT/GIT_CONFIG_KEY_0/GIT_CONFIG_VALUE_0 in the test environment and confirm the fixture's `_git()` calls are unaffected (i.e. the env passed to subprocess.run strips these too).

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The `except ImportError` block in `_said` (lines ~153-155 of armed_reversal.py) catches the failure to import `decision_log.capped_text` and silently falls back to `str(value)[:200]`. It emits no `logger.warning` call, violating the project's `silent_exceptions` plugin contract which requires every except block to either re-raise, be an exempt control-flow exception, or route to a visible-signal call. The `_git` function's `except ImportError` (lines ~88-93) correctly calls `logger.warning(...)` before returning — the inconsistency is within the same file. Additionally the `str(value)[:200]` fallback performs no `$HOME` redaction, meaning if `decision_log` is unavailable the refusal `why` text can contain the operator's home directory username.

Problem (now): Inside `_said()`, when the primary `from.decision_log import capped_text` import fails (triggering the `except Exception as exc:` branch which correctly logs a warning), the code then tries `home = str(Path.home())` in a nested try-except. The `except (OSError, RuntimeError): home = ""` block swallows the exception with no re-raise and no visible-signal call — neither `logger.warning`, `logger.error`, nor any equivalent. This violates the project's silent_exceptions plugin contract that every except block must either re-raise, be a control-flow exempt exception, or route to a visible signal.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
```diff
@@ -0,0 +1,229 @@
...
+        return _refusal("git could not be consulted, so no mechanism could be confirmed")
+    if not inside:
+        # Named precisely rather than folded into a generic refusal: this is the case the
+        # snapshot mechanism exists for, and the one a reader of the refusal can do least
+        # about today.
+        return _refusal(f"{_said(project_root)} is not inside a git work tree, and the "
+                        "snapshot "
+                        "mechanism that would cover that is not built")
+
+    linked, failed = _in_linked_worktree(project_root)
```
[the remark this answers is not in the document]
SanjeevSolanki: Declined, with the reason. A failed worktree probe does not force a refusal before the branch is checked.

`--path-format=absolute` arrived in git 2.31. On 2.30 the worktree probe fails because the flag does not exist, while `symbolic-ref` works — and a named branch with a commit behind it genuinely is a way back. Refusing there would break the module on older git without making any outcome safer.

The contract holds where it matters: the final verdict is a refusal whenever nothing was confirmed. Arming on `branch` is not a guess — `_on_named_branch` verifies `HEAD` resolves to a commit, so it reports a mechanism that exists.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Fixed. It now warns before falling back, like every other handler in the file.

This is the last line of the guard: when the shared redactor is unavailable the fallback collapses `$HOME` itself, so `Path.home()` failing is the one path where the username guard stops guarding — and it set `home = ""` and carried on. Test asserts the warning is emitted and the path still renders.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Already fixed at head — `_said` catches `Exception` with a comment saying why: `capped_text` is another module's function and any exception from it would escape a safety check through its own reporting path.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Already fixed at head. `_check_borrowed_helper_exists()` runs at import and raises if `change_summary._git_query` is gone, so a rename fails loudly instead of turning every probe into a permanent silent refusal.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Already fixed at head. `_refusal` renders the mechanism as `snapshot (not built)` in every refusal, not only the non-git one.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Already fixed at head — `json.dumps` does the delimiting rather than a hand-rolled pair of quotes, exactly because a `"` is printable, survives the strip, and closes a hand-rolled delimiter early.

vulture_whitelist.py
SanjeevSolanki: Fixed — the trigger is checked rather than written.

A test asserts the correlation directly: it scans the tree for any module importing `armed_reversal` (excluding the module itself, its tests and the whitelist) and requires that a consumer existing and the whitelist entries existing are never both true. Verified by mutation — adding a consumer module fails the test.

tests/test_armed_reversal.py
SanjeevSolanki: Already fixed at head. The fixture runs `git init --quiet` then `git checkout --quiet -B main`, so it does not depend on `--initial-branch` (git 2.28) and cannot fail with an error about a flag rather than about the thing under test.

tests/test_armed_reversal.py
SanjeevSolanki: Already fixed at head — `hostile.mkdir()` is wrapped and skips with a message when the filesystem rejects the name. A newline is legal on POSIX and not on Windows, and the forgery this guards is reachable wherever the name is legal.

tests/test_armed_reversal.py
SanjeevSolanki: Already fixed at head. The thread body catches `BaseException` and appends it, and the test asserts `not isinstance(out[0], BaseException)` before touching `.armed` — so a raise reports as itself rather than as `IndexError`.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Already fixed at head. `_on_named_branch` now also runs `rev-parse --verify --quiet HEAD`, so an unborn branch does not arm. You were right that this was the worst failure here — a way back that does not exist, reported as one that does.

skills/studio/scripts/studio/utils/armed_reversal.py
SanjeevSolanki: Fixed, and thank you for naming the asymmetry — that is what made it obvious.

`_git()` now catches `Exception`. Importing a module runs it, so a `SyntaxError` after an edit, or anything raised at import time, went straight through a probe whose stated contract is that unknown is refusal. A test forces the import to raise `SyntaxError` and asserts `(None, True)`.

tests/test_armed_reversal.py
SanjeevSolanki: Fixed. The fixture helper now passes an environment with git's redirect variables removed — `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_COMMON_DIR`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_CEILING_DIRECTORIES`, `GIT_NAMESPACE`.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of f8f7901f8260f2d5f0a39410d8567569d589d654 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Re-reading the actual current diff for _said(): the nested `except (OSError, RuntimeError) as home_exc:` block that handles Path.home() failure now contains `logger.warning("armed reversal: the home directory could not be read...")` before setting home = "" — this is a visible-signal call, so it no longer silently swallows the exception. All three except blocks in this file (in _check_borrowed_helper_exists, _git, and both the outer and nested blocks in _said) route to logger.warning; the previously-posted finding describing the nested block as lacking a log call does not match the code shown in this diff, which already includes the fix.
- **Reviewer:** Every new exception handler in the current diff calls `logger.warning`, including the nested `Path.home()` failure handler; the posted finding reflects an older version.

</details>

tests/test_armed_reversal.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The local `_git()` fixture helper in `tests/test_armed_reversal.py` invokes `subprocess.run(['git', *args], cwd=str(cwd),...)` without an `env` argument, inheriting the full ambient process environment. Git respects redirect variables such as `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, and `GIT_COMMON_DIR` even when `cwd` points to an isolated `tmp_path` repo, so any of those being set in the caller's environment would redirect fixture init, commit, and worktree operations to whatever repo those variables name.

Problem (now): tests/test_armed_reversal.py defines its own `_clean_env()`/`_REDIRECTS` (8 variables: GIT_DIR, GIT_WORK_TREE, GIT_INDEX_FILE, GIT_COMMON_DIR, GIT_OBJECT_DIRECTORY, GIT_ALTERNATE_OBJECT_DIRECTORIES, GIT_CEILING_DIRECTORIES, GIT_NAMESPACE) for its local `_git()` subprocess helper used by every fixture, instead of importing/reusing `change_summary`'s shared sanitized-environment helper. The hardcoded list omits git config-injection variables (GIT_CONFIG_NOSYSTEM, GIT_CONFIG_PARAMETERS/COUNT/KEY_*/VALUE_*, GIT_DISCOVERY_ACROSS_FILESYSTEM) that this repo's own history required adding to the production module's redirect-variable list for the identical attack class.

</details>

vulture_whitelist.py
ainetx: **RESOLVED** -- Confirmed fixed as of f8f7901f8260f2d5f0a39410d8567569d589d654 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `test_the_whitelist_entries_go_when_a_consumer_arrives` scans non-test Python consumers and asserts the whitelist exists if and only if no consumer exists; the import graph currently confirms zero importers.
- **Verifier:** The diff itself adds `tests/test_armed_reversal.py::TestTheRemovalTriggerIsCheckedRatherThanWritten::test_the_whitelist_entries_go_when_a_consumer_arrives`, which scans the whole repo for any non-vulture-whitelist, non-self file containing the string 'armed_reversal' and asserts `whitelisted is not bool(consumers)` — i.e. the whitelist entry must exist iff no consumer exists. This is a real, machine-checkable test (not a free-text comment) that would fail the existing pytest run (already wired into `make test`/CI via the pre-existing `.PHONY: test...` target shown in the Makefile preview, which every other file in `tests/` already relies on for discovery — no new CI wiring is needed for a standard `test_*.py` file in that directory) the moment a real consumer of `armed_reversal` is added without removing the whitelist entries, since at that instant `consumers` becomes non-empty while `whitelisted` is still True, making the assertion False. This exactly matches the check's own stated 'Expected evidence' bullet ('a test in tests/test_armed_reversal.py that asserts the whitelist entry is only present if no non-test importer exists'). The prior FINDING verdicts and posted this finding assert 'No CI step, Makefile target, or test asserts the correlation' — that claim is factually contradicted by this test's presence and logic; the pr-127 precedents cited (//) concerned a whitelist with *only* a prose comment and no such test, which is a materially different situation from this one where an executable correlation-check exists in the same PR.

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of f8f7901f8260f2d5f0a39410d8567569d589d654 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Re-checked against the current diff: `_git()` now uses `except Exception as exc` (broad), not the narrower `(ImportError, AttributeError)` the finding describes, with a comment explicitly explaining the broadening ('Broad for that reason... a SyntaxError after an edit... went straight through a probe'). `_said()` similarly uses `except Exception`. Both lazy-import boundaries already fail closed on any exception, matching the module's 'never-raises'/refusal contract, and test_the_probe_refuses_when_the_import_fails_any_way_at_all pins this for SyntaxError. The posted finding reflects an earlier version of the code; the current diff already applies the fix.
- **Reviewer:** Both lazy-import boundaries now catch `Exception`; `_git` logs and returns `(None, True)` for a fail-closed refusal, and `_said` logs then uses its local fallback. The tests simulate a `SyntaxError` import failure.

</details>

tests/test_armed_reversal.py
ainetx: Re-verified against the current code (f8f7901f8260f2d5f0a39410d8567569d589d654) and this is still an issue.

<details>
<summary>Why</summary>

- **Reviewer:** The new test helper `_git()`/`_clean_env()` in tests/test_armed_reversal.py is a brand-new git launch site that does not reuse the repo's shared sanitized-environment helper (`change_summary._git_env()`); instead it hardcodes its own local `_REDIRECTS` tuple of only 8 variables, omitting config-injection vectors (`GIT_CONFIG_NOSYSTEM`, `GIT_CONFIG_PARAMETERS`, `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*`/`GIT_CONFIG_VALUE_*`, `GIT_DISCOVERY_ACROSS_FILESYSTEM`) that this repo's own history (pr-133/pr-149/pr-154 findings against change_summary.py's `_GIT_REDIRECT_VARS`) already found necessary to close the exact same redirection/injection class. A duplicate, less-complete list is precisely the failure mode this module's own docstring warns about ('a duplicate implementation is how a fix gets lost') — an ambient `GIT_CONFIG_PARAMETERS` setting a fake `core.worktree` would still be inherited by every fixture git invocation in this test file.
- **Reviewer:** The new test git launcher sanitizes several redirect variables but retains git configuration injection variables such as GIT_CONFIG_COUNT, GIT_CONFIG_KEY_*, and GIT_CONFIG_VALUE_*.
- **Verifier:** tests/test_armed_reversal.py adds a new git launch site (`_git()`/`_clean_env()`) that spawns `subprocess.run(["git",...])` directly with its own locally-defined `_REDIRECTS` tuple of 8 variables, rather than reusing the shared sanitized-environment helper this repo already established (`change_summary._git_env()`, confirmed by historical finding pr-133/ covering the exact same redirect-variable class). The local list omits git config-injection vectors (`GIT_CONFIG_NOSYSTEM`, `GIT_CONFIG_PARAMETERS`, `GIT_CONFIG_COUNT`/`KEY_*`/`VALUE_*`, `GIT_DISCOVERY_ACROSS_FILESYSTEM`) that the check explicitly calls out and that this repo's own precedent required closing for production code — a duplicated, less-complete implementation of the exact pattern the module's own docstring warns against ('a duplicate implementation is how a fix gets lost').

</details>

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

vulture_whitelist.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/utils/armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

ainetx (approved):
Solid piece of work overall — the `armed_reversal` module does exactly what its docstring promises (assert, never judge), the git probes correctly separate "valid negative" from "tool failure," refusals stay bounded and redacted, and the test suite is unusually thorough about the failure modes that matter here (unborn-branch false-positive, log injection via hostile paths, silent import failures). No correctness, security, or test-coverage issues block this.

One small thing worth a look before it lingers:

- **Test git helper duplicates an incomplete env-sanitization list** — `tests/test_armed_reversal.py` rolls its own `_clean_env()`/`_REDIRECTS` (8 vars) for its `_git()` fixture helper instead of reusing `change_summary`'s shared sanitized-environment helper, and the hardcoded list is missing the `GIT_CONFIG_*`/`GIT_DISCOVERY_ACROSS_FILESYSTEM` variables that were added to the production module's redirect list for this exact same attack class. Worth consolidating so the two lists can't drift again. ([comment](https://github.com/constructorfabric/studio/pull/207#discussion_r4013833008))

tests/test_armed_reversal.py
SanjeevSolanki: You are right, and this one is about my fix rather than the original defect — so it needed its own change. This PR had already merged, so it is #217.

I wrote a second list where the repo already had one. Mine named eight redirect variables and omitted the whole configuration mechanism `change_summary._GIT_REDIRECT_VARS` carries: `GIT_CONFIG_PARAMETERS`, `GIT_CONFIG_COUNT` with the indexed `KEY_n`/`VALUE_n` pairs it gates, `GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `GIT_CONFIG_NOSYSTEM`, and `GIT_DISCOVERY_ACROSS_FILESYSTEM`. The comments beside them record why each is there — `GIT_CONFIG_SYSTEM` pointing at a file that set `core.excludesFile` emptied the untracked sweep until `GIT_CONFIG_NOSYSTEM=1` brought it back — which is exactly the kind of knowledge a copy loses.

The fixture helper calls `change_summary._git_env()` now. A test asserts the reuse so the duplicate cannot come back, and a second one asserts the returned environment carries none of those variables, so it would also fail if the shared helper regressed.

The rule about reusing the repo's helper rather than writing a second one is one I applied to `_git_query` a few lines above and did not apply to the environment beside it.

Issue #206, opened 2026-09-15
AC#2 blocked: Codex and Cursor have no safe dedicated skill-output path for a real ask-tool binding


tkcoding:
**Found while implementing #142 AC#2** ("adding support for a new harness requires changing the per-target binding only") on top of the AC#6 split (#191, merged). Binding Codex and Cursor to their real native question tools turned out to be blocked by an architecture gap that predates both PRs — filing separately since it needs its own scoping/plan rather than a quick follow-on.

## Problem

`_ASK_TOOL_BINDING` can name any tool per-target (post-#191), but naming one for **Codex** or **Cursor** isn't safe to ship yet, because the *output file path* those bindings would need to reach is still shared across multiple tools — PR 191 only made the in-memory template independent per tool, not the on-disk location it's written to.

Per this repo's own architecture docs (`architecture/DESIGN.md`, `architecture/specs/cli.md`, `architecture/features/project-extensibility.md`):

- **Codex/OpenAI has no dedicated skill-output path at all, by explicit upstream design.** Quoting `project-extensibility.md`: *"OpenAI uses shared `.agents/skills/` outputs only... any `[[agents]]` entry targeting `openai` must be reported as skipped rather than generating a separate OpenAI agent file contract."* `.codex/agents/` exists but is for subagent **roles** (model/reasoning-effort config via `_render_toml_agent`) — a different concept from skill/workflow content. There is currently nowhere to put a Codex-specific binding without it also being read by windsurf/cursor/copilot/OpenCode, which all consume the same shared file.
- **Cursor is only partially better.** It has one genuinely dedicated, non-shared path — `.cursor/commands/cf.md` — but that covers only the *root* `cf` skill. The other five kit-workflow skills (`cf-generate`, `cf-analyze`, `cf-plan`, `cf-explore`, `cf-workspace`) exist only in the shared `.agents/skills/` bucket for Cursor, identically to windsurf/copilot/openai.

## Failure scenario if bound anyway

`.agents/skills/cf/SKILL.md` is a single literal path, identical for windsurf/cursor/copilot/openai. The write logic (`_write_or_skip`) decides "changed vs. unchanged" purely by content comparison, not by which tool last wrote it. If Cursor were bound to `AskQuestion` while windsurf/copilot/openai stay unbound:

1. A project with both Cursor and Windsurf configured runs `cfs generate-agents`.
2. Cursor's pass writes `ask_tool_name = "AskQuestion"` into the shared file.
[...]

PR #191, opened 2026-09-14
refactor(agents): give each generation target its own skill-output list (#142 AC#6)

tkcoding:
## Summary
- Issue #142's AC#6 requires that adding native-dialog support for a new harness only need a per-target binding change. Today it doesn't hold: `_agents_skill_outputs()` and the kit-workflow-as-skill template return one Python list object reused verbatim across windsurf/cursor/copilot/openai, so none of them can carry a differing `ask_tool_name` without first breaking that shared-file structure.
- Both now take a `tool` parameter and thread `_ASK_TOOL_BINDING.get(tool)` through, giving each tool its own independent list.
- **Pure refactor, no behavior change** — `_ASK_TOOL_BINDING` still only has `"claude"`, so every other tool's `.get(tool)` still returns `None` and output is content-identical to before.
- This is PR 1 of 4 in a planned sequence closing #142's remaining gaps (AC#2 real Codex/Cursor bindings, then #185, then #186) — kept separate and reviewable rather than one large PR, per feedback on #184's review process.

## Test plan
- [x] Updated `test_shared_skill_outputs_have_no_exact_ask_tool_binding` and `test_shared_files_no_custom_content_leak` to loop over all 4 tools individually instead of one bare call.
- [x] Verified content-preservation directly: all 4 tools still render `ask_tool_name = unset`, no `AskUserQuestion` leak.
- [x] Ran a multi-angle deep code review on the diff — no findings.
- [x] Full test suite: 5976 passed, 4 pre-existing/unrelated failures (macOS filesystem-encoding edge cases, confirmed via `git stash` against the clean branch base), zero regressions.
- [x] pylint clean, no new line-length violations.

Refs #142

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected generated agent skill and workflow templates so tool-specific configurations are applied consistently across supported tools.
  - Preserved identical generated content for tools without a specific configuration.
  - Prevented unsupported interactive prompts from appearing in generated templates.
  - Improved validation to confirm each tool receives only its own configuration and that unbound tools remain consistent.

Issue #185, opened 2026-09-11
Native-dialog binding detection can misclassify a hand-edited non-Claude file as pure generated


tkcoding:
**Found during:** deep review of #142's native-dialog implementation ([PR #184](https://github.com/constructorfabric/studio/pull/184)), not part of ainetx's review — filing separately per the maintainer's request to track it rather than block the PR.

## Problem

`_pure_generated_stub_matches()` (`skills/studio/scripts/studio/commands/agents.py`) decides whether a generated file is "pure Studio output" (safe to silently delete/overwrite during cleanup/regeneration) by trying every known `ask_tool_name` candidate and accepting a match against **any** of them — because the function has no way to know which tool/bucket actually produced the file it's checking.

```python
candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v]
for binding in candidate_bindings:
    ...
    if nonblank == expected_protocol:
        return True
```

Today there are only two buckets: Claude's dedicated files (which legitimately carry `ask_tool_name = "AskUserQuestion"`) and the shared `.agents/skills/` bucket used by windsurf/cursor/copilot/codex (which should only ever carry `unset`). The PR's own test (`test_shared_skill_outputs_have_no_exact_ask_tool_binding`) asserts that invariant on the *generation* side — but nothing enforces it on the *detection* side.

## Failure scenario

A user hand-edits a non-Claude generated file (e.g. `.cursor/agents/...` or a file under `.agents/skills/`) and, via a bad merge or copy-paste from a Claude-generated file, it ends up containing the literal line `- ask_tool_name = "AskUserQuestion"`. `_pure_generated_stub_matches` now matches the Claude candidate and returns `True`, so `_is_pure_studio_generated` reports the file as a pristine, untouched Studio stub — even though it belongs to a bucket that should never contain that line. That result gates silent deletion/overwrite in `_cleanup_claude_workflow_command`, `_cleanup_legacy_skill_files`, `_cleanup_legacy_manifest_skill_outputs`, and ownership classification in `_classify_generated_output_owner` — discarding what should have been treated as user modification.

This is a low-likelihood scenario today (requires a coincidental exact-content match), but it's a structural gap that grows with every future `_ASK_TOOL_BINDING` entry, and there's currently zero regression coverage for it.

## Suggested fix

[...]

PR #208, opened 2026-09-15
feat(pdsl): bind Cursor's root cf skill to AskQuestion (#142 AC#2, scoped)

### LINKED ISSUE #142: Studio CLI — Raise Blocking Questions Through the Harness's Native Question Dialog

tkcoding:
## Summary

While implementing #142's AC#2 ("adding support for a new harness requires changing the per-target binding only" — extended to mean Codex and Cursor should present via each harness's own affordance), I hit a real architecture blocker: PR #191 (AC#6) made the *in-memory* per-tool template independent, but the *output file path* is still one literal string (`.agents/skills/cf/SKILL.md` etc.) shared by windsurf/cursor/copilot/openai, and read natively by OpenCode too. Binding any one of those tools there would make the shared file's content depend on generation order and could misroute another tool onto a tool it doesn't have.

Filed as **#206** with three scoped options (A: small/safe, B: full fix pending a Codex-capability unknown, C: document only). This PR implements **Option A** — the smallest slice that's safe to ship today.

## What this does

- Cursor's root `cf` skill has one genuinely dedicated, non-shared output — `.cursor/commands/cf.md` — unlike its other five kit-workflow skills. That's the one place a real binding can land without the collision.
- Binds it to `AskQuestion` (Cursor's native question tool — staff-corroborated across forum threads, not from an official schema dump; availability is gated by Cursor's own mode/model support, and it has no native free-text escape hatch, unlike Claude's/Codex's tools). Documented these caveats in `pdsl-execution-card.md`.
- Introduced `_ASK_TOOL_BINDING_BY_OUTPUT` (keyed by output path, not tool) rather than a one-off constant, so a second such case is a one-line table addition.
- **Deliberately did NOT** fold this into `_pure_generated_stub_matches`'s tool-agnostic candidate pool — during review this turned out to open the exact cross-tool collision issue #185 already tracks (verified empirically: a value valid for one tool's dedicated output becomes a false "pure generated" match for an unrelated tool's legacy path). `.cursor/commands/cf.md` doesn't reach that check in practice today (verified directly, not assumed — its only writer is a plain content-diff with no purity gate). If that ever changes, the fix is #185's own direction, not widening this pool.

## What this does NOT do

Codex and Cursor's other five skills remain unbound. This does not close #142's AC#2 — only its smallest safely-achievable slice. The rest is tracked in #206.

## Test plan
- [x] Verified the binding lands only in `.cursor/commands/cf.md`'s rendered template, and does not leak into `_agents_skill_outputs("cursor")`, `_kit_workflow_skill_template("cursor")`, or any other cursor config output (regression tests for both directions).
- [x] Verified the cross-tool collision scenario is closed: a Windsurf-legacy-shaped body matching the new `AskQuestion` binding is correctly NOT classified as pure-generated.
- [x] Ran a multi-agent deep review across several rounds — caught and fixed: unsafe leak into the global purity-check candidate pool, an unguarded dict-index vs `.get()`, and test duplication with existing generic coverage.
- [x] Full test suite: 6025 passed, 5 pre-existing/unrelated failures (macOS filesystem-encoding edge cases, confirmed via `git stash` against the branch base), zero regressions.
- [x] pylint clean, `cfs pdsl validate` shows no new findings beyond the already-tolerated one for this file.

Refs #142, #206

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of configured question tools that are unavailable in the current session. These bindings now behave as unset and preserve the existing menu display behavior.
  - Updated the dedicated Cursor `cf` command to correctly support interactive question prompts.
  - Prevented Cursor-specific prompt behavior from affecting other shared generated outputs or workflow templates.
  - Improved classification of Cursor launcher content so it is recognized as Studio-owned workflow content before other checks are applied.


tests/test_subagent_registration.py
```diff
@@ -1272,6 +1272,50 @@ def test_ask_tool_binding_carries_exact_tool_name_for_claude(self):
...
+        reached by separate legacy-cleanup paths for deprecated file naming),
+        so there's no equivalent "untouched stub" assertion to make here --
+        the binding's presence in the rendered template is the whole
+        contract for this file.
+        """
+        from studio.commands.agents import _default_agents_config
+
+        config = _default_agents_config()
+        outputs = config["agents"]["cursor"]["skills"]["outputs"]
+        cf_md = next(o for o in outputs if o["path"] == ".cursor/commands/cf.md")
```
ainetx, line 1292: ### Bare `next()` in test raises StopIteration instead of AssertionError on path rename

**Severity:** Minor

**Problem**
`next(o for o in outputs if o["path"] == ".cursor/commands/cf.md")` has no `default=` argument. Any rename or removal of the `.cursor/commands/cf.md` output entry raises `StopIteration` propagating out of the test method rather than a targeted `AssertionError` with a descriptive message.

<details>

**How to reproduce**
1. Rename `.cursor/commands/cf.md` output path in `_default_agents_config()`.
2. Run `test_cursor_root_skill_binds_ask_question_via_dedicated_launcher`.
3. Observe the test runner reports ERROR with an unhelpful `StopIteration` traceback rather than FAILURE identifying the missing binding.

**Expected behavior**
A legible `AssertionError` (or explicit `self.fail()`) identifying what entry was expected but not found.

**Actual behavior**
Unhandled `StopIteration` propagates, reported as a test error not a failure.

```text
path renamed -> generator exhausted -> StopIteration -> ERROR (not FAILURE)
```

**Impact**
A path rename that removes the AskQuestion binding would show as a test infrastructure error rather than a binding-contract failure, making it harder to triage.

**Suggested correction**
Replace with `cf_md = next((o for o in outputs if o["path"] == ".cursor/commands/cf.md"), None)` followed by `self.assertIsNotNone(cf_md, "Expected.cursor/commands/cf.md output entry not found in cursor config")`.

**How to verify**
Remove the `.cursor/commands/cf.md` entry from `_default_agents_config()` and confirm the test now reports FAILURE with the expected message rather than ERROR.

</details>

tests/test_subagent_registration.py
```diff
@@ -1272,6 +1272,50 @@ def test_ask_tool_binding_carries_exact_tool_name_for_claude(self):
...
+        entry may carry the AskQuestion binding, none of cursor's other
+        config outputs should. (The per-helper invariant -- that
+        _agents_skill_outputs('cursor')/_kit_workflow_skill_template
+        ('cursor') carry no exact binding -- is already covered generically
+        for every non-Claude tool by test_shared_skill_outputs_have_no_exact_
+        ask_tool_binding and test_kit_workflow_skill_templates_have_no_exact_
+        binding_for_non_claude below; no need to re-assert it here.)"""
+        from studio.commands.agents import _default_agents_config
+
+        outputs = _default_agents_config()["agents"]["cursor"]["skills"]["outputs"]
```
ainetx, line 1380: ### cursor.workflows.template (.cursor/commands/cf-*.md workflow proxies) is not covered by any AskQuestion-leak regression test

**Severity:** Minor

**Problem**
The new regression guard test_cursor_shared_bucket_outputs_unaffected_by_root_skill_binding only inspects config['agents']['cursor']['skills']['outputs']. The commit that claimed to close this gap (f7d2c34a) added an AskQuestion assertion to test_kit_workflow_skill_templates_have_no_exact_binding_for_non_claude, but that test exercises _kit_workflow_skill_template('cursor'), which is a separate production function producing.agents/skills/{skill_id}/SKILL.md files. Neither test ever reads config['agents']['cursor']['workflows']['template'], the dict (agents.py lines 1930-1943) that actually renders Cursor's.cursor/commands/cf-*.md workflow proxy files.

<details>

**How to reproduce**
1. Open tests/test_subagent_registration.py and search for every reference to `workflows` combined with `template` under cursor's config. 2. Confirm no test constructs or inspects config['agents']['cursor']['workflows']['template']. 3. Compare with _kit_workflow_skill_template('cursor') in agents.py:2793-2817, which outputs to '.agents/skills/{skill_id}/SKILL.md', not '.cursor/commands/cf-*.md'.

**Expected behavior**
A regression test should assert that config['agents']['cursor']['workflows']['template'] never contains 'AskQuestion' (or any exact ask_tool binding), mirroring the guard already in place for cursor.skills.outputs, so a future change that threads ask_tool_name into this template would be caught.

**Actual behavior**
No test inspects this specific template list; a future edit that adds a dynamic ask_tool_name parameter to the.cursor/commands/cf-*.md workflow proxy template would not be caught by any existing test.

```text
cursor.workflows.template (agents.py:1930-1943, static, no ask_tool_name today)
|
+-- rendered into.cursor/commands/cf-*.md
|
X-- NOT checked by test_cursor_shared_bucket_outputs_unaffected_by_root_skill_binding (checks skills.outputs only)
X-- NOT checked by test_kit_workflow_skill_templates_have_no_exact_binding_for_non_claude (checks _kit_workflow_skill_template ->.agents/skills/ bucket, a different path)
```

**Impact**
If a future change (e.g. extending the _ASK_TOOL_BINDING_BY_OUTPUT pattern already introduced in this PR) adds a per-output ask_tool_name to the cursor workflows template, a leak of an exact tool binding into the.cursor/commands/cf-*.md workflow proxies would go undetected by the test suite.

**Suggested correction**
Add an assertion in a test (or extend test_cursor_shared_bucket_outputs_unaffected_by_root_skill_binding) that reads config['agents']['cursor']['workflows']['template'], joins it, and asserts 'AskQuestion' is not present.

**How to verify**
Add the assertion, then temporarily inject an ask_tool_name into the workflows.template definition and confirm the new test fails, then revert and confirm it passes.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -389,6 +389,24 @@ def _pure_generated_stub_matches(stripped: str) -> bool:
...
+    # here, unlike _ASK_TOOL_BINDING's. This candidate list is tool-agnostic
+    # -- it's tried against content from *any* tool's legacy path -- so
+    # adding a value here makes it a valid "pure generated" match everywhere,
+    # not just for the one dedicated output it's actually bound to. That's
+    # already the cross-tool collision risk tracked in
+    # constructorfabric/studio#185 for _ASK_TOOL_BINDING itself; widening the
+    # pool further here would let e.g. a Windsurf legacy file matching
+    # Cursor's "AskQuestion" content get misclassified as Windsurf's own pure
+    # stub and silently deleted by `_cleanup_legacy_skill_files`, even though
+    # Windsurf never emits that binding. `.cursor/commands/cf.md` itself
```
ainetx, line 403: ### Claimed `_classify_generated_output_owner` short-circuit for Cursor launcher is asserted in comment but not locked by any test

**Severity:** Minor

**Problem**
The comment added at lines 403–407 of agents.py states `.cursor/commands/cf.md` 'never reaches `_pure_generated_stub_matches` in practice today' because `_classify_generated_output_owner` short-circuits on `_extract_studio_owned_target` first. This is the sole safety argument for not adding `AskQuestion` to `candidate_bindings`. No regression test exercises this short-circuit path — a future refactor could silently break it.

<details>

**How to reproduce**
1. Refactor `_classify_generated_output_owner` to remove or reorder the `_extract_studio_owned_target` short-circuit.
2. A `.cursor/commands/cf.md` content with `ask_tool_name = "AskQuestion"` now reaches `_pure_generated_stub_matches`.
3. No test fails to signal the regression.

**Expected behavior**
A test that passes a pure-generated `.cursor/commands/cf.md`-style content through `_classify_generated_output_owner` and asserts the short-circuit fires before `_pure_generated_stub_matches` is reached.

**Actual behavior**
No such test exists in the diff or is referenced.

```text
_classify_generated_output_owner (refactored, no short-circuit) -> _pure_generated_stub_matches -> AskQuestion matches -> wrong classification -> silent deletion risk
```

**Impact**
If the short-circuit is removed, AskQuestion-bearing Cursor files could be misclassified as pure stubs and silently deleted by `_cleanup_legacy_skill_files`.

**Suggested correction**
Add a test that passes a Cursor launcher stub content (with `ask_tool_name = "AskQuestion"`) to `_classify_generated_output_owner` with path `.cursor/commands/cf.md` and asserts it is handled by `_extract_studio_owned_target` (not reaching the purity check).

**How to verify**
Temporarily remove the `_extract_studio_owned_target` short-circuit in `_classify_generated_output_owner` and confirm the new test fails.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -894,14 +912,44 @@ def _file_has_studio_follow_link(path: Path) -> bool:
...
 _ASK_TOOL_BINDING: Dict[str, Optional[str]] = {
     "claude": "AskUserQuestion",
 }
 _CLAUDE_ASK_TOOL_NAME = _ASK_TOOL_BINDING.get("claude")
 
+# Bindings scoped to one specific dedicated output path rather than a whole
+# tool (constructorfabric/studio#206, Option A) -- for a tool's outputs that
+# are NOT all part of a shared bucket, only some of them. Keyed by output
+# path so a second such case is a one-line addition here, not a new
+# module-level constant plus a new hand-verified exclusion elsewhere; its
```
ainetx, line 936: ### `_ASK_TOOL_BINDING_BY_OUTPUT` docstring falsely claims its values are folded into `_pure_generated_stub_matches` candidate set

**Severity:** Major

**Problem**
The dict-level docstring added in this PR states: 'its values are folded into `_pure_generated_stub_matches`'s candidate set below for the same reason.' The code at `candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v]` uses only `_ASK_TOOL_BINDING`. The comment block added immediately above that assignment in this same PR explicitly says `_ASK_TOOL_BINDING_BY_OUTPUT`'s values are 'deliberately NOT folded in here.' The docstring is factually wrong and directly contradicts the safety rationale documented two dozen lines above it.

<details>

**How to reproduce**
1. Read `_ASK_TOOL_BINDING_BY_OUTPUT`'s docstring (lines ~936–938).
2. Observe it says values are folded into `candidate_bindings`.
3. Read `candidate_bindings` assignment: only `_ASK_TOOL_BINDING.values()` is iterated.
4. Read the comment above `candidate_bindings`: values are 'deliberately NOT folded in here.'

**Expected behavior**
The docstring accurately states that `_ASK_TOOL_BINDING_BY_OUTPUT`'s values are deliberately excluded from `_pure_generated_stub_matches`'s candidate pool.

**Actual behavior**
The docstring says the opposite of what the code and the explanatory comment both say.

```text
docstring: 'folded in' <-> code: uses _ASK_TOOL_BINDING only <-> comment: 'deliberately NOT folded in'
```

**Impact**
A future contributor trusting the dict docstring may add `_ASK_TOOL_BINDING_BY_OUTPUT.values()` to `candidate_bindings`, causing Windsurf/other-tool legacy files containing `AskQuestion` content to be misclassified as pure stubs and silently deleted by `_cleanup_legacy_skill_files`.

**Suggested correction**
Fix the docstring to read: 'its values are deliberately NOT folded into `_pure_generated_stub_matches`'s candidate set — see the comment above `candidate_bindings` for the rationale.'

**How to verify**
Grep for 'folded into' near `_ASK_TOOL_BINDING_BY_OUTPUT`; confirm the statement now accurately describes the exclusion.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -1895,7 +1943,10 @@ def _default_agents_config() -> dict:
                                 "# /cf",
                                 "",
                                 "{custom_content}",
-                                *_follow_protocol_lines("{target_skill_path}"),
+                                *_follow_protocol_lines(
```
ainetx, line 1946: ### Path literal `.cursor/commands/cf.md` duplicated without a shared constant — silent binding drop on partial rename

**Severity:** Minor

**Problem**
The string `".cursor/commands/cf.md"` is hard-coded independently as the `_ASK_TOOL_BINDING_BY_OUTPUT` dict key and as the `.get()` argument inside `_default_agents_config()`. No shared constant links them. A rename requires updating three independent sites (dict key, `.get()` argument, `path` field in config); any mismatch silently demotes the binding to description fallback.

<details>

**How to reproduce**
1. Rename the `path` field for Cursor's root cf skill in `_default_agents_config()` to `.cursor/commands/cf-new.md`.
2. Forget to update the `.get(".cursor/commands/cf.md")` call.
3. `ask_tool_name` is now `None` in the rendered template — silently no binding.

**Expected behavior**
A single named constant (e.g. `_CURSOR_CF_LAUNCHER_PATH = ".cursor/commands/cf.md"`) used by both the dict definition and the `.get()` call so a rename fails loudly at one site.

**Actual behavior**
Two independent string literals; a partial rename silently produces `None` from `.get()` with no error or test failure.

```text
dict key ".cursor/commands/cf.md" !=.get(".cursor/commands/cf.md") after rename -> None -> ask_tool_name=unset in template
```

**Impact**
Native Cursor dialog binding silently disappears whenever the path literal drifts between definition and lookup site.

**Suggested correction**
Introduce `_CURSOR_CF_LAUNCHER_PATH = ".cursor/commands/cf.md"` and replace both string literals with this constant.

**How to verify**
Confirm there is only one literal occurrence of `.cursor/commands/cf.md` after the change (the constant definition), and both the dict key and `.get()` call reference the constant.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -389,6 +389,24 @@ def _pure_generated_stub_matches(stripped: str) -> bool:
...
+    # Cursor's "AskQuestion" content get misclassified as Windsurf's own pure
+    # stub and silently deleted by `_cleanup_legacy_skill_files`, even though
+    # Windsurf never emits that binding. `.cursor/commands/cf.md` itself
+    # never reaches this check in practice today (`_classify_generated_
+    # output_owner` short-circuits on `_extract_studio_owned_target` first,
+    # and its only writer, `_write_or_skip`, has no purity gate at all) --
+    # verified directly, not assumed. If that ever changes, fix it by
+    # threading tool/path identity through this function (#185's own
+    # direction), not by widening this already-risky global pool further.
     candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v]
```
ainetx, line 410: ### No negative test verifies `_pure_generated_stub_matches` rejects AskQuestion-bearing content

**Severity:** Minor

**Problem**
The diff explicitly documents (in the comment above `candidate_bindings`) that `_ASK_TOOL_BINDING_BY_OUTPUT`'s values must NOT be in the candidate pool, warning of cross-tool silent-deletion risk. The docstring on `_ASK_TOOL_BINDING_BY_OUTPUT` says the opposite. No test passes AskQuestion-bearing content to `_pure_generated_stub_matches` and asserts `False`. If `candidate_bindings` is ever widened to include `_ASK_TOOL_BINDING_BY_OUTPUT.values()`, no test catches it.

<details>

**How to reproduce**
1. Add `+ [v for v in _ASK_TOOL_BINDING_BY_OUTPUT.values() if v]` to `candidate_bindings` in `_pure_generated_stub_matches`.
2. Run the full test suite.
3. Observe no test fails despite the widening introducing the cross-tool deletion risk.

**Expected behavior**
A test that calls `_pure_generated_stub_matches` (or `_is_pure_studio_generated`) on content containing `ask_tool_name = "AskQuestion"` and asserts `False`.

**Actual behavior**
No such test exists; the exclusion is enforced only by comment.

```text
candidate_bindings widened to include 'AskQuestion' -> _pure_generated_stub_matches matches Windsurf file -> _cleanup_legacy_skill_files deletes it -> no test failure
```

**Impact**
Silent deletion of a non-Cursor tool's legacy file that contains AskQuestion content, if the candidate pool is accidentally widened.

**Suggested correction**
Add a test: construct stub content with `ask_tool_name = "AskQuestion"` and call `_is_pure_studio_generated` on it, asserting the result is `False`.

**How to verify**
Add `_ASK_TOOL_BINDING_BY_OUTPUT.values()` to `candidate_bindings` and confirm the new test now fails.

</details>

tests/test_subagent_registration.py
tkcoding: Fixed in f7d2c34a: switched to next((...), None) plus self.assertIsNotNone with a descriptive message, so a path rename now reports FAILURE with a clear 'entry not found' message instead of an unhandled StopIteration ERROR.

tests/test_subagent_registration.py
tkcoding: Fixed in f7d2c34a: added self.assertNotIn('AskQuestion', template, ...) to test_kit_workflow_skill_templates_have_no_exact_binding_for_non_claude, right after the existing AskUserQuestion check, so a leak into _kit_workflow_skill_template('cursor') specifically would now be caught.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Fixed in f7d2c34a: added test_cursor_launcher_ownership_check_short_circuits_before_purity_check, which builds real AskQuestion-bound cf.md content, asserts _extract_studio_owned_target matches it (proving the short-circuit condition is met), and asserts _classify_generated_output_owner(...) returns 'workflow' -- locking the exact safety argument the comment made, rather than leaving it as an assertion nothing enforces.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Good catch -- this was a real, dangerous inconsistency. Fixed in f7d2c34a: the docstring now correctly states the values are deliberately NOT folded into candidate_bindings, matching the code and the rationale comment below it (which was correct all along -- only the docstring above the dict was stale, left over from an earlier draft before I reverted the fold-in during review). Thanks for catching this before it could mislead someone into reintroducing the #185-class bug.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Fixed in f7d2c34a: introduced _CURSOR_CF_LAUNCHER_PATH as a single named constant, now used at the dict key, the .get() call, and the config "path" field (and in the tests too) -- a rename now only needs to happen in one place.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Fixed in f7d2c34a: added test_pure_generated_stub_matches_rejects_ask_question_bearing_content, which builds AskQuestion-bound content directly and asserts _pure_generated_stub_matches returns False -- if candidate_bindings is ever widened to include _ASK_TOOL_BINDING_BY_OUTPUT's values, this test fails immediately.

tests/test_subagent_registration.py
ainetx: **RESOLVED** -- Confirmed fixed as of f7d2c34afd2cc78b81eddc29dbc96abeae28e291 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current test uses `next((...), None)` at line 1360 and immediately follows it with `self.assertIsNotNone(..., "Expected... output entry not found...")`, so a missing or renamed output produces a clear AssertionError rather than StopIteration.
- **Reviewer:** The current worktree code at tests/test_subagent_registration.py:1360 already uses `next((o for o in outputs if o["path"] == _CURSOR_CF_LAUNCHER_PATH), None)` with a default of `None`, followed by `self.assertIsNotNone(cf_md, f"Expected {_CURSOR_CF_LAUNCHER_PATH!r} output entry not found in cursor config")` at lines 1361-1364. This matches the fix described in the discussion thread (commit f7d2c34a): a path rename or removal now produces a clear AssertionError with a descriptive message rather than an unhandled StopIteration. The originally-flagged bare `next()` without a default no longer exists in this code.
- **Verifier:** Verified directly in tests/test_subagent_registration.py lines 1360-1364: the code now reads `cf_md = next((o for o in outputs if o["path"] == _CURSOR_CF_LAUNCHER_PATH), None)` followed by `self.assertIsNotNone(cf_md, f"Expected {_CURSOR_CF_LAUNCHER_PATH!r} output entry not found in cursor config")`. This provides a default of None and a clear, descriptive AssertionError on a missing/renamed entry instead of an unhandled StopIteration, matching the fix applied in commit f7d2c34a per the discussion thread. The originally-flagged bare `next()` without a default no longer exists in the current diff/code.

</details>

tests/test_subagent_registration.py
ainetx: Re-verified against the current code (f7d2c34afd2cc78b81eddc29dbc96abeae28e291) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** The named guard reads only `cursor.skills.outputs`; the newly added generic assertion checks `_kit_workflow_skill_template("cursor")`, which generates shared `.agents/skills/` kit-workflow skills, not the distinct `cursor.workflows.template` used for `.cursor/commands/` workflow proxies.
- **Verifier:** Confirmed by direct code read: test_cursor_shared_bucket_outputs_unaffected_by_root_skill_binding (tests/test_subagent_registration.py:1368-1389) only iterates config['agents']['cursor']['skills']['outputs']; no test in the file ever reads config['agents']['cursor']['workflows']['template'] (agents.py:1930-1943, the dict rendering.cursor/commands/cf-*.md). The 'fix' commit f7d2c34a instead added an AskQuestion assertion to test_kit_workflow_skill_templates_have_no_exact_binding_for_non_claude, which calls _kit_workflow_skill_template('cursor') (agents.py:2793-2817) -- a distinct function that renders.agents/skills/{skill_id}/SKILL.md, not.cursor/commands/cf-*.md. So the specific config path this check asks about remains genuinely unchecked by any test, despite the discussion thread's claim of resolution. Risk is currently low since that template hardcodes _follow_protocol_lines('{target_workflow_path}') with no ask_tool_name argument, so it cannot dynamically leak the binding today -- this is a real but narrow test-coverage gap, not an active leak.

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of f7d2c34afd2cc78b81eddc29dbc96abeae28e291 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `test_cursor_launcher_ownership_check_short_circuits_before_purity_check` builds an AskQuestion-bound `.cursor/commands/cf.md` payload, confirms its Studio target is extractable, and asserts `_classify_generated_output_owner(...) == "workflow"`; the implementation evaluates `_extract_studio_owned_target(content) or _is_pure_studio_generated(content)`, so the tested content is correctly recognized through the claimed ownership path.
- **Reviewer:** test_cursor_launcher_ownership_check_short_circuits_before_purity_check (tests/test_subagent_registration.py:1247-1289) builds real AskQuestion-bound.cursor/commands/cf.md-style content, asserts _extract_studio_owned_target(content) is not None (proving the short-circuit condition is met), and asserts _classify_generated_output_owner(_CURSOR_CF_LAUNCHER_PATH, content) == 'workflow'. Cross-checked against the actual implementation at agents.py:5168-5171 (`if not (_extract_studio_owned_target(content) or _is_pure_studio_generated(content))`), which confirms the `or` short-circuit the comment claims -- Python's `or` never evaluates `_is_pure_studio_generated` (and therefore never reaches `_pure_generated_stub_matches`) once `_extract_studio_owned_target` returns truthy. This test now locks that exact invariant with concrete evidence rather than leaving it as an unverified comment.
- **Verifier:** Verified directly in tests/test_subagent_registration.py: test_cursor_launcher_ownership_check_short_circuits_before_purity_check (line 1247) builds AskQuestion-bound.cursor/commands/cf.md-style content, asserts _extract_studio_owned_target(content) is not None (proving the short-circuit's left operand is truthy), and asserts _classify_generated_output_owner(_CURSOR_CF_LAUNCHER_PATH, content) == 'workflow'. Cross-checked against the actual implementation at agents.py:5165-5179, which reads `if not (_extract_studio_owned_target(content) or _is_pure_studio_generated(content)): return None` -- confirming Python's `or` short-circuits and never evaluates `_is_pure_studio_generated` (hence never reaches `_pure_generated_stub_matches`) once `_extract_studio_owned_target` returns truthy. This concretely locks the exact safety invariant the comment at lines 403-407 claims.

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of f7d2c34afd2cc78b81eddc29dbc96abeae28e291 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current dict-level comment explicitly says `_ASK_TOOL_BINDING_BY_OUTPUT` values are deliberately NOT folded in (lines 936–940), matching the explanatory comment and `candidate_bindings`, which iterates only `_ASK_TOOL_BINDING.values()` (line 410).
- **Reviewer:** Verified current code directly: line 410's candidate_bindings assignment (`[None] + [v for v in _ASK_TOOL_BINDING.values() if v]`) uses only _ASK_TOOL_BINDING, and the _ASK_TOOL_BINDING_BY_OUTPUT docstring at lines 936-940 now correctly reads 'Its values are deliberately NOT folded into `_pure_generated_stub_matches`'s candidate set below', matching both the code and the rationale comment at lines 393-409. The prior contradiction was fixed in commit f7d2c34a per the discussion thread, and no discrepancy remains.
- **Verifier:** Verified directly in current code: the _ASK_TOOL_BINDING_BY_OUTPUT docstring (agents.py lines 932-940) reads 'Its values are deliberately NOT folded into `_pure_generated_stub_matches`'s candidate set below', which matches both the explanatory comment above line 410 and the actual code `candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v]` (line 410), which uses only _ASK_TOOL_BINDING. The contradiction described in the check (docstring claiming values ARE folded in) has been fixed per the discussion thread (commit f7d2c34a); no discrepancy remains in the current diff.

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of f7d2c34afd2cc78b81eddc29dbc96abeae28e291 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `_CURSOR_CF_LAUNCHER_PATH` is the single source of `.cursor/commands/cf.md` (line 956) and is used as the binding-map key (line 958), the generated output path (line 1948), and the `.get()` lookup key (line 1956), so a path rename cannot silently desynchronize these sites.
- **Reviewer:** Verified current code: `_CURSOR_CF_LAUNCHER_PATH = ".cursor/commands/cf.md"` (agents.py line 956) is a single named constant referenced at the `_ASK_TOOL_BINDING_BY_OUTPUT` dict key (line 958), the config's `path` field (line 1948), and the `.get()` call inside `_default_agents_config()` (line 1956) -- no independently duplicated string literal remains at any of the three sites, so a path rename only needs to happen in one place, matching the fix already merged in commit f7d2c34a.
- **Verifier:** Verified directly in the current file: _CURSOR_CF_LAUNCHER_PATH = ".cursor/commands/cf.md" (agents.py:956) is a single named constant, referenced at the _ASK_TOOL_BINDING_BY_OUTPUT dict key (line 958), the config output's `path` field (line 1948), and the `.get()` lookup inside _default_agents_config() (line 1956); tests also import and use the same constant (lines 1257, 1287, 1356-1382). No independently duplicated string literal remains at any of the sites the check names, so a path rename only needs to happen in one place -- confirming the fix already applied (commit f7d2c34a per the discussion thread).

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of f7d2c34afd2cc78b81eddc29dbc96abeae28e291 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `test_pure_generated_stub_matches_rejects_ask_question_bearing_content` constructs protocol lines with `ask_tool_name="AskQuestion"`, passes the resulting content directly to `_pure_generated_stub_matches`, and asserts `False`; it will fail if output-scoped bindings are added to `candidate_bindings`.
- **Reviewer:** tests/test_subagent_registration.py:1223-1245 defines test_pure_generated_stub_matches_rejects_ask_question_bearing_content, which builds protocol content via _follow_protocol_lines(..., ask_tool_name="AskQuestion") and asserts _pure_generated_stub_matches(body) is False; cross-checked against agents.py:410 where candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v] deliberately excludes _ASK_TOOL_BINDING_BY_OUTPUT's values, so the test matches current behavior and would fail if candidate_bindings were ever widened to include those values.
- **Verifier:** tests/test_subagent_registration.py:1223-1245 defines test_pure_generated_stub_matches_rejects_ask_question_bearing_content, which builds content via _follow_protocol_lines(..., ask_tool_name="AskQuestion") and asserts _pure_generated_stub_matches(body) is False; verified against agents.py:410 where candidate_bindings = [None] + [v for v in _ASK_TOOL_BINDING.values() if v] deliberately excludes _ASK_TOOL_BINDING_BY_OUTPUT's values, so this test both matches current behavior and would catch any future widening of candidate_bindings.

</details>

ainetx (approved):
Approved — the review is otherwise clean, and these are the only open non-blocking items worth a follow-up glance:

- **Cursor workflow proxy coverage** — Add regression coverage for the actual `config['agents']['cursor']['workflows']['template']` path that renders `.cursor/commands/cf-*.md`, rather than only adjacent shared-skill helpers. [Discussion](https://github.com/constructorfabric/studio/pull/208#discussion_r4013158903)
- **Absent-tool fallback coverage** — Exercise the new rule that an unavailable bound tool falls back exactly like `unset`, specifically in combination with Cursor's `AskQuestion` binding.
- **Concrete launcher contract coverage** — Render `.cursor/commands/cf.md` and assert both its `AskQuestion` binding and its required bootstrap handoff together, matching the runtime contract.

PR #211, opened 2026-09-15
ci: run the pipeline on release branches, not only on main

vasylcf:
## The gap

`ci.yml` filtered both triggers to `main`:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

A PR targeting `release/v1.7.0` matches neither, so the whole pipeline is skipped. What a release PR actually shows today is six checks, all from apps that trigger regardless of this filter — DCO, CodeRabbit, SonarCloud, code-ranker — and **nothing** from this workflow:

| Not running on release branches |
|---|
| Test (Python 3.11 / 3.12 / 3.13 / 3.14) |
| Coverage (≥90%) |
| Pylint |
| Dead Code Scan |
| Enforcement Gates (seeded violations must fail) |
| Spec Coverage |
| Validate Artifacts (× 4 Python versions) |
| Validate Kits (× 4 Python versions) |
| Version Consistency |

Found while preparing v1.7.0 (#192). The `CI` workflow had never run on `release/v1.7.0-integration` — not once across the branch's history.

## Why it matters

This is backwards for the release process the branches exist to serve. A release branch is cut from the previous tag, reviewed and tested there, and only then tagged — so the one branch whose entire purpose is *being tested before a tag* was the only one the pipeline did not cover.

For v1.7.0 that meant the evidence the release passes (5,723 tests, `cfs validate`, the spec-coverage gate) was a local run on one contributor's machine, with nothing in GitHub to corroborate it.

## The change

```yaml
on:
  push:
    branches: [main, 'release/**']
  pull_request:
    branches: [main, 'release/**']
```

`release/**` rather than a single branch name, so 1.7.1 and everything after it is covered without another edit.

## Note on where this has to live

The same change is already on the v1.7.0 branch in #192, and that is not duplication. A `pull_request` trigger is evaluated against the PR's own workflow file, so only the copy on the release branch makes the pipeline run on the release PR **before** it merges. This PR is what gives every future release branch the same behaviour from the moment it is cut.

## Cost

More runner time: pushes to release branches now run the full matrix. That is the point — it is the same work the pipeline already does for `main`, applied to the branch a tag is actually created from.

## Summary by CodeRabbit

* **Chores**
  * Continuous integration checks now run for release-candidate branches in addition to the main branch.
  * Release branches receive the same automated testing, linting, coverage, and artifact validation before tagging.


.github/workflows/ci.yml
```diff
@@ -1,11 +1,16 @@
...
 
+# `release/**` alongside `main`: releases are cut on a branch taken from the previous
+# tag, reviewed and tested there, and only then tagged. Filtering CI to `main` meant the
+# one branch whose whole purpose is being tested before a tag was the one branch this
+# pipeline never ran on -- a release PR showed only the checks that come from apps
+# (DCO, SonarCloud, code-ranker), and no tests, lint, coverage or artifact validation.
 on:
   push:
-    branches: [main]
+    branches: [main, 'release/**']
```
ainetx, line 11: ### MAJOR: `release/**` trigger never matches the documented `vX.Y.Z-beta`/`vX.Y.Z` release branch naming convention

**SEVERITY: MAJOR**

**Problem**
CONTRIBUTING.md line 146 instructs `git checkout -b vX.Y.Z-beta`, and the Branch and Release Workflow section (lines 188-191) shows release branches named `vX.Y.Z-beta` and `vX.Y.Z`. The new glob `release/**` in ci.yml matches only branch names that literally begin with `release/`. Branches named per the documented convention (e.g. `v1.2.0-beta`, `v1.2.0`) do not start with `release/` and never match this pattern.

<details>

**How to reproduce**
1. Follow CONTRIBUTING.md's documented release process: `git checkout -b v1.2.0-beta`. 2. Push the branch. 3. Observe CI does not trigger on push/PR for this branch despite the new `release/**` entry, because the branch name doesn't match the glob.

**Expected behavior**
The CI trigger pattern should match whatever branch naming convention is actually used for releases, so that the full test/lint/coverage/artifact-validation pipeline runs before a release is tagged.

**Actual behavior**
The `release/**` glob is disjoint from the documented `vX.Y.Z-beta`/`vX.Y.Z` naming, so release branches created per CONTRIBUTING.md's own instructions never trigger the expanded CI checks the PR intended to add.

```text
CONTRIBUTING.md says: create branch `vX.Y.Z-beta` --> ci.yml trigger: `branches: [main, 'release/**']` --> glob requires prefix `release/` --> `vX.Y.Z-beta` does not start with `release/` --> no match --> CI never runs full suite on release branch
```

**Impact**
Release branches can be reviewed and tagged without the tests, lint, coverage, and artifact-validation checks this PR was specifically meant to enable on them, defeating the PR's stated purpose.

**Suggested correction**
Either update the trigger glob to match the actual branch naming convention (e.g. `v[0-9]*` for `vX.Y.Z-beta`/`vX.Y.Z`), or update CONTRIBUTING.md to document `release/*` branches and change the actual release process to use that naming — ideally both, matching what real branches are cut as.

**How to verify**
Push a branch named per CONTRIBUTING.md's documented convention (e.g. `v1.2.0-beta`) and confirm CI triggers; also confirm the fix doesn't regress matching for whatever branch naming is actually used in practice for cutting releases.

</details>

.github/workflows/ci.yml
```diff
@@ -1,11 +1,16 @@
...
   pull_request:
-    branches: [main]
+    branches: [main, 'release/**']
```
ainetx, line 13: ### pull_request.branches filter targets an undocumented hotfix scenario, not the stated release-PR use case

**Severity:** Minor

**Problem**
GitHub Actions evaluates `pull_request.branches` against the PR's base (target) branch, not the head (source) branch. A PR from a release branch into `main` has base=`main`, which already matched before this change. Adding `release/**` to `pull_request.branches` enables CI only for PRs that target a `release/**` branch (i.e., hotfix PRs into a release branch). The PR comment describes the missing CI as being on the release branch itself (push side), which is correctly addressed by the `push.branches` change — the `pull_request.branches` addition addresses a separate, undocumented scenario.

<details>

**How to reproduce**
1. Create a release branch `release/1.2.0`. 2. Open a PR from `release/1.2.0` into `main`. 3. Observe that base=`main`, which matched `pull_request.branches: [main]` before this PR — CI already ran on that PR.

**Expected behavior**
The `pull_request.branches` addition should implement the rationale stated in the comment, or the comment should describe the hotfix-targeting-release-branch scenario it actually enables.

**Actual behavior**
The `pull_request.branches: release/**` addition enables CI for PRs whose base is a `release/**` branch (hotfix PRs), while the stated motivation (release branch lacking test/lint/coverage) is fully addressed by the `push.branches` change alone.

```text
Before: push→release/** ✗CI | PR(head=release,base=main) ✓CI(already via main)
After: push→release/** ✓CI | PR(head=any,base=release/**) ✓CI(NEW, undocumented)
Stated gap: push→release/** ✗CI (fixed by push change, not PR change)
```

**Impact**
The comment misleads future maintainers about what the `pull_request.branches` line does. Additionally, if the actual branch naming is `vX.Y.Z-beta` (not `release/*`), both additions may be dead on arrival.

**Suggested correction**
Either: (a) document that `pull_request.branches: release/**` is intentional for hotfix PRs targeting release branches, or (b) remove it if only the push-side gap was the intent. Also align the PR comment to describe the actual semantics.

**How to verify**
Create a test PR with base=`release/x.y.z` and confirm CI triggers; create a PR with head=`release/x.y.z` and base=`main` and confirm CI was already running before this change.

</details>

.github/workflows/ci.yml
ainetx, line 11: ### Test job retains checkout credentials while executing repository-controlled Make targets

**Severity:** Minor

**Problem**
The `gates` job sets `persist-credentials: false` with the explicit comment 'Keep the job token out of.git/config so the test corpus this job executes cannot read it back'. The `test` job runs the same pattern — `make install` and `make test` against repository-controlled Makefiles — but has no `persist-credentials: false`. The GitHub token therefore remains in `.git/config` during the full test run. This pre-existing asymmetry now also applies to the newly admitted `release/**` branches.

<details>

**How to reproduce**
1. Push to any `release/**` branch (or open a PR targeting one). 2. CI triggers the `test` job. 3. `actions/checkout@v4` with no `persist-credentials: false` writes the workflow token into `.git/config`. 4. Any code path reached by `make install` or `make test` can read that token.

**Expected behavior**
The `test` job should apply `persist-credentials: false` for the same reason the `gates` job does — both execute repository-controlled code with the checkout token in scope.

**Actual behavior**
The `test` job retains the checkout token throughout `make install` and `make test`; the `gates` job correctly strips it. The asymmetry is unaddressed and now extends to release branches.

```text
test job: checkout(token persisted) → make install → make test [token readable]
gates job: checkout(persist-credentials:false) → make install → make test-gates [token cleared]
```

**Impact**
A compromised dependency pulled by `make install`, or a malicious path exercised during `make test`, can exfiltrate the workflow token. For release branches the token may have broader write permissions than for a fork-originated PR.

**Suggested correction**
Add `persist-credentials: false` to the `test` job's `actions/checkout` step, matching the `gates` job's existing pattern.

**How to verify**
After adding `persist-credentials: false` to the test job's checkout step, inspect the runner's `.git/config` mid-run and confirm no `extraheader` credential entry is present.

</details>

ainetx, line 11: ### SonarQube action uses mutable `@v6` tag while the repository's own SHA-pin policy is demonstrated by code-ranker

**Severity:** Minor

**Problem**
`.github/workflows/code-ranker.yml` line 7 explicitly documents a 'supply-chain hardening' rationale and pins to `42c2fba1acd7c99a71875c8eefa171224d8e683c`. `.github/workflows/ci.yml` line 120 uses `SonarSource/sonarqube-scan-action@v6`, a mutable major-version tag that resolves to whatever commit SonarSource last pushed to that tag. No policy document grants an exemption. The diff expands the sonarqube job to execute on `release/**` branches — precisely the branches the PR describes as the last gate before a version tag.

<details>

**How to reproduce**
1. Push to a `release/**` branch. 2. CI triggers the sonarqube job. 3. The job runs `SonarSource/sonarqube-scan-action@v6`, resolving the tag at runtime. 4. If SonarSource updates `@v6` between CI runs, different code executes with access to `SONAR_TOKEN`.

**Expected behavior**
The SonarQube action is pinned to an immutable commit SHA, consistent with the SHA-pin policy demonstrated by code-ranker.yml, especially for CI that now runs on release branches.

**Actual behavior**
`SonarSource/sonarqube-scan-action@v6` is a mutable label; a tag update or compromise at SonarSource silently changes the code that runs with `SONAR_TOKEN` in the environment.

```text
code-ranker.yml: uses:...@42c2fba1acd7c99a71875c8eefa171224d8e683c ← immutable
ci.yml sonarqube: uses: SonarSource/sonarqube-scan-action@v6 ← mutable
```

**Impact**
A changed or compromised `@v6` tag executes arbitrary code with `SONAR_TOKEN` in scope during release CI, a higher-value target than ordinary feature-branch CI.

**Suggested correction**
Pin the action to its current commit SHA: look up the SHA for the `v6` tag on the SonarSource GitHub repo and replace `@v6` with `@<sha> # v6`.

**How to verify**
After pinning, confirm `ci.yml` line 120 references a full 40-character SHA. Run `git log --oneline` on the SonarSource repo to cross-check the SHA matches the intended `v6` release.

</details>

ainetx, line 11: ### MINOR: SonarQube job silently reports success without scanning when SONAR_TOKEN is absent

**SEVERITY: MINOR**

**Problem**
The `sonarqube` job (ci.yml:101-121) sets `SONAR_TOKEN` at the job level from secrets but has no job-level `if:` guard. The checkout step (108-110) runs unconditionally regardless of whether the token is set. Only the 'Download coverage report' (112-116) and 'SonarQube Scan' (118-120) steps are individually guarded with `if: env.SONAR_TOKEN != ''`. When the secret is unavailable (e.g. GitHub Environment secrets scoped to `main` only, now that release branches also trigger this workflow), the job runs checkout, skips both scan steps, and exits with overall status `success`.

<details>

**How to reproduce**
1. Push to or open a PR against a `release/**` branch where the SONAR_TOKEN secret/environment is not available. 2. Observe the `sonarqube` job executes checkout, both guarded steps are skipped due to the empty token condition. 3. The job completes with a green checkmark despite no scan having run.

**Expected behavior**
A meaningful quality gate should either skip the whole job explicitly (reporting 'skipped', not 'passed') or fail loudly when the token is required but missing, rather than reporting success for a no-op run.

**Actual behavior**
The job reports success ('SonarQube: passed') on the CI dashboard even though zero scanning was performed, misleading anyone checking release readiness.

```text
release branch push --> sonarqube job starts --> checkout succeeds (unconditional) --> SONAR_TOKEN empty --> download-artifact skipped --> scan skipped --> job exits 0 --> dashboard shows green 'SonarQube' with no scan run
```

**Impact**
A release branch's SonarQube check can misleadingly show as passed even though no static analysis was ever performed, undermining confidence in the pre-tag quality gate this PR is intended to strengthen.

**Suggested correction**
Add a job-level `if: ${{ env.SONAR_TOKEN != '' }}` (or equivalent condition sourced from secrets context) so the job is honestly reported as skipped when the token is unavailable, rather than green with no scan performed.

**How to verify**
Run the workflow on a branch/context where SONAR_TOKEN is absent and confirm the job now shows as 'skipped' rather than 'success'.

</details>

ainetx, line 11: ### MAJOR: Version consistency check compares release branches against a moving `origin/main`, producing spurious 'version not bumped' failures

**SEVERITY: MAJOR**

**Problem**
`scripts/check_versions.py` line 184 defaults `--base` to `origin/main`. `check_kit_version_bump` (line 115) diffs kit files against that base and, at line 168, fails if the release branch's kit version is `<=` the version found in `origin/main`'s conf.toml. The `versions` CI job (ci.yml lines 148-161) passes no `--base` override. When a release branch is cut from `main` at an older point and `main` later bumps a kit's version (which necessarily accompanies a content change, since content_changes filters out conf.toml-only diffs at lines 134-141), the release branch — which never touched that kit — will fail this check listing files it never changed.

<details>

**How to reproduce**
1. Cut release branch `release/vX.Y.Z` from `main` at commit A. 2. `main` continues, and a later commit B bumps kit `foo`'s version and content. 3. CI runs `make check-versions` on the release branch (now triggered per this PR's changed `on:` block). 4. `git diff origin/main -- kits/foo/` on the release branch shows files changed (relative to main's newer HEAD), and `cur_int <= base_int` evaluates true since the release branch's version predates main's bump. 5. CI reports a false 'version not bumped' error for files the release branch never touched.

**Expected behavior**
The version-bump check on a release branch should compare against the branch's own fork point (merge-base) or the tag it was cut from, not against the continuously advancing `origin/main` HEAD.

**Actual behavior**
The unmodified `origin/main` default causes the check to run on release branches now that this PR added them to CI, producing spurious failures whenever main advances kit versions after the branch point.

```text
main tag N --> release/vX.Y.Z branch cut --> main advances, bumps kit foo to N+1 (content+version) --> CI now runs on release branch (this PR) --> check_versions.py diffs against origin/main HEAD (N+1) --> release branch kit is still N --> cur_int(N) <= base_int(N+1) --> false failure naming untouched files
```

**Impact**
Release branches can fail CI with confusing, spurious 'version not bumped' errors purely because unrelated changes landed on main after the branch was cut, blocking or confusing release validation for a check whose failure mode was previously unreachable (the job never ran on release branches before this PR).

**Suggested correction**
Change the `--base` used for release-branch runs to the merge-base with the tag/commit the branch was cut from (or pass an explicit `--base` computed via `git merge-base` in the `versions` job) rather than the moving `origin/main` tip.

**How to verify**
Simulate a release branch cut from an older main commit, have main bump a kit's version afterward, then run `make check-versions` on the release branch and confirm it no longer fails for kits it didn't touch.

</details>

ainetx (approved):
The core change is correct — adding `release/**` to `push.branches` closes the genuine gap where release branches ran no pipeline tests before tagging. Approving. Four minor observations are threaded inline; none block merging, but they're worth a follow-up pass:

- **`pull_request.branches` targets hotfix PRs, not release PRs** — `pull_request.branches` matches the PR's *base*, so `release/**` here covers PRs *into* a release branch, not the release-branch push path the commit comment describes. The push side is the right fix; this addition is incidental. [thread](https://github.com/constructorfabric/studio/pull/211#discussion_r4014982834)

- **`test` job keeps credentials in `.git/config` during Make execution** — the `gates` job explicitly drops them with `persist-credentials: false`; `test` doesn't, leaving the token readable by any Makefile target during the run. Now applies to `release/**` branches as well. [thread](https://github.com/constructorfabric/studio/pull/211#discussion_r4014983743)

- **SonarQube action pinned to mutable `@v6` tag** — `code-ranker.yml` SHA-pins its actions with an explicit supply-chain rationale; `ci.yml` line 120 uses a floating major-version tag for the SonarQube scan, which now runs on the pre-tag release branches. [thread](https://github.com/constructorfabric/studio/pull/211#discussion_r4014984334)

- **SonarQube job exits green with no scan when `SONAR_TOKEN` is absent** — the two scan steps are individually gated but the job itself is not, so a missing secret produces a ✓ on the dashboard with nothing actually scanned. [thread](https://github.com/constructorfabric/studio/pull/211#discussion_r4014984932)

ainetx (changes requested):
Changes requested — this reintroduces the exact gap CI was trying to close, just moved: the release branch trigger and the version check it now runs alongside don't match how this project actually cuts releases.

- **`release/**` glob never matches real release branches** — CONTRIBUTING.md and the branch/release diagram both name release branches `vX.Y.Z-beta` / `vX.Y.Z` (e.g. `v1.2.0-beta`), which doesn't start with `release/`. As written, CI still never runs on a release branch — the trigger looks fixed but isn't. ([comment](https://github.com/constructorfabric/studio/pull/211#discussion_r4014982034))
- **Version check will false-fail on every release branch once `main` moves on** — `check_versions.py` diffs against `origin/main` by default, with no `--base` override in the `versions` CI job. A release branch cut from `main` keeps its kit at version N; if `main` bumps that kit to N+1 afterward, the check reads the diff as "version not bumped" and fails even though the release branch never touched that kit. ([comment](https://github.com/constructorfabric/studio/pull/211#discussion_r4014985788))

Both need fixing before this can merge: the branch pattern has to actually match the documented naming, and the version-check base needs to compare against the branch's own fork point (or an equivalent baseline), not a moving `main`.

.github/workflows/ci.yml
ainetx: Let's update CONTRIBUTING guide

vasylcf: Valid, and this is the one finding in this review that changed the code.

`CONTRIBUTING.md:146` does instruct `git checkout -b vX.Y.Z-beta`, and `release/**` cannot match that name.

Fixed in #254 (4b96424), which matches both conventions: `branches: [main, 'release/**', 'v[0-9]*']`. Both rather than one, because the documented name and the names actually cut for the last release (`release/v1.7.0`, `release/v1.7.0-integration`) are different habits, and matching only one leaves the other untested -- which is the state the change exists to end.

One correction to the framing: "the glob matches zero branches ever created" and "the entire change is a no-op" hold for the documented naming, but not for what was actually cut. `release/v1.7.0` exists and `release/**` does match it. The pattern was under-inclusive, not inert. The severity stands; the reproduction is the accurate part.

This PR is superseded by #254, which carries the same change plus this fix.

.github/workflows/ci.yml
```diff
@@ -1,11 +1,16 @@
 # @cpt-algo:cpt-studio-spec-init-structure-change-infrastructure:p1
 name: CI
 
+# `release/**` alongside `main`: releases are cut on a branch taken from the previous
+# tag, reviewed and tested there, and only then tagged. Filtering CI to `main` meant the
```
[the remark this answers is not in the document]
vasylcf: Confirmed, and the two sources really do disagree -- but the workflow comment is the one that matches practice.

Checked what was actually done rather than what is written: `git merge-base origin/release/v1.7.0 origin/main` is `e3a283fd`, and `git describe --tags --exact-match e3a283fd` resolves it to `v1.6.2`. The last release branch was cut from the previous tag. `CONTRIBUTING.md:143` ("Create a release branch from `main`") and `CONTRIBUTING.md:191` ("Tag `main` after merge") describe a procedure that was not followed: v1.7.0 was tagged at the head of the release branch, which is not an ancestor of `main`.

So the fix is not to reword the comment to agree with the document. It is to decide which one is right and make reality follow it. That decision is recorded in #257, which takes the CONTRIBUTING side: the next release branch is cut from `main`, and the branch is merged back after tagging.

Not changing it in this PR -- it is superseded by #254.

.github/workflows/ci.yml
vasylcf: Correct on the mechanism: `pull_request.branches` is evaluated against the PR's base branch, so a release-branch-into-`main` PR already matched `[main]` before the change, and adding `release/**` there only admits PRs whose base is a release branch.

The scenario is not undocumented in intent -- a PR into a live release branch is where fixes land while the branch is being stabilised before a tag, and that is exactly the window this pipeline was blind to. But you are right that the comment argued the push case and then changed both filters, so the `pull_request` half was never justified in writing.

Related and better handled together: on #254 the same area drew a Major -- with both `push` and `pull_request` admitting release branches, a release-branch PR now runs the whole pipeline twice, because the concurrency group is `${{ github.workflow }}-${{ github.ref }}` and a push ref never equals the PR's `refs/pull/N/merge` ref, so the two runs do not coalesce.

Both points are about whether the `pull_request` half earns its place. I am answering them on #254, which supersedes this PR.

.github/workflows/ci.yml
vasylcf: Confirmed against `main`, not just against the diff: `gates` sets `persist-credentials: false` at `ci.yml:55` with the stated reason that the corpus it executes must not read the job token back; `test` checks out at `ci.yml:23` with no such setting and then runs `make install` and `make test` against the same repository-controlled Makefiles.

The asymmetry is real and it is an oversight, not a decision -- one job states the reason and the other ignores it while doing the same kind of work.

It is pre-existing, though, and a trigger change is the wrong PR to fix it in: nothing here made the `test` job keep the token; admitting more branches only means it does so on more branches. It belongs in a credential-hygiene pass over the checkout steps, together with the two SonarQube findings in this review.

Not fixing it here -- this PR is superseded by #254, which is also only a trigger change. I have the follow-up written up and will file it as an issue so it does not close with this PR.

vasylcf: Confirmed, and the contrast is exactly as stated: `code-ranker.yml:8` pins `@42c2fba1acd7c99a71875c8eefa171224d8e683c` and documents supply-chain hardening as the reason, while `ci.yml:115` uses `SonarSource/sonarqube-scan-action@v6`, a tag its owner can move. The action that is not pinned is the one that receives `SONAR_TOKEN`.

No exemption exists; it is simply inconsistent.

Pre-existing and not caused by the trigger change, so same disposition as the `persist-credentials` finding above: it goes in a follow-up issue with the other hardening items rather than into a PR that only edits `on:`. This PR is superseded by #254.

vasylcf: Confirmed, line by line: `SONAR_TOKEN` is set at job level (`ci.yml:101`), the two substantive steps each carry `if: env.SONAR_TOKEN != ''` (`ci.yml:108`, `ci.yml:114`), and the job itself has no guard. With the secret absent the job checks out, skips both steps and exits green.

Agreed that this is worse than a skipped job: a green check that scanned nothing is a check that actively misinforms. Either a job-level `if:` (which reports the job as skipped, honestly) or an explicit failure would be accurate.

Pre-existing; grouped with the two findings above for a follow-up issue rather than fixed in a trigger change. This PR is superseded by #254.

vasylcf: Confirmed, and this is the one of the six that the trigger change actually provokes rather than merely widens. `scripts/check_versions.py:184` defaults `--base` to `origin/main` and the `versions` job passes no override, so before this change the job simply never ran on a release branch and the failure had nowhere to appear.

Verified the mechanism in `check_kit_version_bump` rather than taking it on trust, because there is a filter that could have saved it and does not. `content_changes` (lines 134-141) drops `kits/<slug>/conf.toml` itself, so a `main` commit that touched *only* the version line is filtered out and `continue`s harmlessly. The false failure needs `main` to have changed kit content as well.

That caveat does not rescue it: the version is bumped precisely because content changed, so the two travel together, and the pair is what `git diff origin/main -- kits/<slug>/` reports. `base_int` is then `main`'s newer version, `cur_int` the release branch's older one, and `cur_int <= base_int` fails at line 167 naming files the release branch never touched.

Not fixed here. This PR is superseded by #254, which lands the same trigger, so the exposure moves there -- and the fix (compare against the branch point, not the moving tip) is a change to `check_versions.py` and not to `on:`. Carrying it into the follow-up issue with the other hardening findings, flagged as the one to decide against #254 before that PR merges.

.github/workflows/ci.yml
ainetx: @vasylcf This concern still stands.

<details>
<summary>Why</summary>

The human explicitly concedes the finding's core claims: the mechanism (base-branch matching) is correct, and "the pull_request half was never justified in writing" by the comment — which is exactly the finding's complaint. Arguing the scenario is legitimate in intent doesn't change that the PR comment as written only argues the push case, so the misleading-comment problem stands as described. Deferring full resolution to a separate PR (#254) doesn't dismiss the finding on this PR; it's a scope/sequencing choice, not a rebuttal of the technical claim.

</details>

.github/workflows/ci.yml
ainetx: @vasylcf **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

Confirmed in the actual file: `test` checks out at line 28 with no `persist-credentials: false`, while `gates` sets it at line 60 with exactly the stated comment. So the factual claim is accurate.

But the diff here only widens `on.push`/`on.pull_request` branch filters (lines 9-13) — it doesn't touch the checkout steps or introduce the credential asymmetry, which predates this PR. The pushback's scoping point holds: this is a pre-existing gap that deserves its own credential-hygiene fix across all checkout steps, not a one-line patch bolted onto a trigger-only PR, and the human has committed to tracking it as a follow-up issue.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a security or hygiene gap (e.g. missing persist-credentials: false, an unset flag, an absent check), diff the PR to confirm it introduces or touches the affected lines; if the gap predates the diff and the PR only changes unrelated scope (e.g. trigger filters), treat it as a pre-existing issue to track separately rather than a defect in this PR."

</details>

ainetx: @vasylcf **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The full diff (confirmed via get_file_diff) shows this PR only changes the `on:` trigger block (branches [main] -> [main, 'release/**']); line 120's `SonarSource/sonarqube-scan-action@v6` is untouched pre-existing code, not something introduced or modified here. The human's factual claims check out: code-ranker.yml does pin to a SHA with a stated hardening rationale while ci.yml:120 uses the mutable `@v6` tag -- a real inconsistency, but pre-existing and orthogonal to what this PR actually changes (adding release/** as a trigger). Scoping this to a follow-up hardening issue rather than blocking a PR that only edits `on:` is a reasonable disposition, consistent with how the sibling persist-credentials finding was handled.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging a pre-existing security inconsistency (e.g., an unpinned action version) as a blocking issue, verify via the actual diff whether the PR touches that code at all -- if it only modifies unrelated parts of the same file (e.g., trigger conditions), scope the finding to a follow-up issue rather than blocking the PR."

</details>

ainetx: @vasylcf Re-verified against the current code (c7924927f398297afc71870d04b3431b1971026e) and this is still an issue. Severity has also been reassessed from Critical to Major given the current code. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** CONTRIBUTING.md documents release branches as vX.Y.Z-beta and vX.Y.Z, but release/** only matches branch names beginning with release/, so the added trigger does not cover the documented release workflow.
- **Verifier:** CONTRIBUTING.md:146 documents release branches as `vX.Y.Z-beta`/`vX.Y.Z` (e.g. `git checkout -b vX.Y.Z-beta`), which never begins with the literal `release/` prefix, so the added `release/**` glob never matches the documented branch naming convention. The discussion thread confirms this was accepted as a real, human-verified finding and already fixed in a follow-up PR (#254) that added `v[0-9]*` alongside `release/**` — though the human also clarified the actual branches cut for the last release were `release/v1.7.0`-style, making the pattern under-inclusive rather than fully inert. Either way, a real gap existed relative to the documented convention.

</details>

ainetx: @vasylcf Re-verified against the current code (c7924927f398297afc71870d04b3431b1971026e) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** The newly included release-branch runs expose a SonarQube job that unconditionally checks out the code but skips both scan steps when SONAR_TOKEN is empty, leaving a green job that performed no scan.
- **Verifier:** Confirmed from ci.yml lines 101-121: the `sonarqube` job has no job-level `if:` guard; the `actions/checkout` step (108-110) runs unconditionally; only 'Download coverage report' (112-116) and 'SonarQube Scan' (118-120) are gated on `env.SONAR_TOKEN != ''`. When SONAR_TOKEN is absent (e.g. environment secrets scoped only to `main`), the job still runs checkout, skips both substantive steps, and completes with status success, showing a misleading green check on release branches. Both reviewers agree and the human thread confirms this as a real, pre-existing gap now exposed by the trigger expansion; posted as Minor severity by the finding author, which is reasonable given it's a reporting/observability issue rather than a functional break.

</details>

ainetx: @vasylcf Re-verified against the current code (c7924927f398297afc71870d04b3431b1971026e) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** Adding release-branch execution activates versions CI with its existing origin/main default, so a release branch cut from an older tag can compare against later main-only kit changes and fail spuriously.
- **Verifier:** Confirmed from scripts/check_versions.py: `--base` defaults to `origin/main` (line 184), `check_kit_version_bump` (line 115) diffs kit files against that base (line 129) and compares conf.toml versions (lines 144-168), failing at line 168 if `cur_int <= base_int`. ci.yml's `versions` job (lines 148-161) invokes `make check-versions` with no `--base` override. The human-confirmed thread traced the exact mechanism: the `content_changes` filter (lines 134-141) only excludes commits touching conf.toml alone, but a real version bump on main always accompanies a content change, so the filter doesn't save it — a release branch cut before main bumps a kit's version+content will spuriously fail. This is a genuine new failure mode introduced by running this same job (unmodified) on release branches, confirmed correct in both mechanism and severity (Major) by the human review thread.

</details>

ainetx:
### MAJOR: Version consistency check compares release branches against a moving `origin/main`, producing spurious 'version not bumped' failures

**SEVERITY: MAJOR**

**Problem**
`scripts/check_versions.py` line 184 defaults `--base` to `origin/main`. `check_kit_version_bump` (line 115) diffs kit files against that base and, at line 168, fails if the release branch's kit version is `<=` the version found in `origin/main`'s conf.toml. The `versions` CI job (ci.yml lines 148-161) passes no `--base` override. When a release branch is cut from `main` at an older point and `main` later bumps a kit's version (which necessarily accompanies a content change, since content_changes filters out conf.toml-only diffs at lines 134-141), the release branch — which never touched that kit — will fail this check listing files it never changed.

<details>

**How to reproduce**
1. Cut release branch `release/vX.Y.Z` from `main` at commit A. 2. `main` continues, and a later commit B bumps kit `foo`'s version and content. 3. CI runs `make check-versions` on the release branch (now triggered per this PR's changed `on:` block). 4. `git diff origin/main -- kits/foo/` on the release branch shows files changed (relative to main's newer HEAD), and `cur_int <= base_int` evaluates true since the release branch's version predates main's bump. 5. CI reports a false 'version not bumped' error for files the release branch never touched.

**Expected behavior**
The version-bump check on a release branch should compare against the branch's own fork point (merge-base) or the tag it was cut from, not against the continuously advancing `origin/main` HEAD.

**Actual behavior**
The unmodified `origin/main` default causes the check to run on release branches now that this PR added them to CI, producing spurious failures whenever main advances kit versions after the branch point.

```text
main tag N --> release/vX.Y.Z branch cut --> main advances, bumps kit foo to N+1 (content+version) --> CI now runs on release branch (this PR) --> check_versions.py diffs against origin/main HEAD (N+1) --> release branch kit is still N --> cur_int(N) <= base_int(N+1) --> false failure naming untouched files
```

**Impact**
Release branches can fail CI with confusing, spurious 'version not bumped' errors purely because unrelated changes landed on main after the branch was cut, blocking or confusing release validation for a check whose failure mode was previously unreachable (the job never ran on release branches before this PR).

**Suggested correction**
Change the `--base` used for release-branch runs to the merge-base with the tag/commit the branch was cut from (or pass an explicit `--base` computed via `git merge-base` in the `versions` job) rather than the moving `origin/main` tip.

**How to verify**
Simulate a release branch cut from an older main commit, have main bump a kit's version afterward, then run `make check-versions` on the release branch and confirm it no longer fails for kits it didn't touch.

</details>

---
_Original location: scripts/check_versions.py:184 -- inline anchoring could not be resolved after 1 attempt(s)._

vasylcf:
Closing as superseded by #254, and recording where each of the seven findings went, because this PR's review found real things and none of them should close with it.

**Why superseded rather than merged.** #254 carries this same change plus the fix the Critical finding here asked for: `branches: [main, 'release/**', 'v[0-9]*']`. `release/**` alone cannot match `vX.Y.Z-beta`, which is what `CONTRIBUTING.md:146` instructs, so this PR's pattern was under-inclusive. #254's is a strict superset.

**Why this PR went stale.** The change it makes reached `release/v1.7.0` directly and never reached `main`, so this PR stayed open against a branch that already had the code, while `main` did not. #254 exists to port it, and #257 records the process fix that prevents the next one.

**Where the findings went:**

| Finding | Disposition |
|---|---|
| `release/**` never matches `vX.Y.Z-beta` (Critical) | Fixed in #254 (4b96424) |
| `pull_request.branches` matches the base, not the head | Answered on #254, where the same area drew a Major |
| CONTRIBUTING.md and the workflow comment disagree | #254 (13b72dc8) stopped the comment restating the procedure; the disagreement itself is settled in #257 |
| `test` job keeps the checkout token | **Fixed** in #229 — all ten checkouts now set `persist-credentials: false`, and `permissions: contents: read` is declared at workflow level |
| `sonarqube-scan-action@v6` is a mutable tag | #261 |
| `sonarqube` reports success without scanning | #261 |
| `check_versions.py` vs an advancing `origin/main` (Major) | #261, flagged as the one to decide against #254 |

Every one was verified against `main` before being filed rather than taken on trust — including the CONTRIBUTING disagreement, where `git merge-base origin/release/v1.7.0 origin/main` resolves to the `v1.6.2` tag, meaning the last release branch really was cut from the previous tag and the document is the side that does not match practice.

Thanks to @ainetx and CodeRabbit for the review; the findings outlive the PR.

PR #212, opened 2026-09-15
fix(eval): name the failing checks on screen, and fail --check on an empty suite

SanjeevSolanki:
## What

Two QA-found defects in the `eval` structural scorer — reporting + exit code only; the detector is unchanged and verified sound.

**1. The human report never named the finding.** A plain run printed only `structural compliance: N%`, naming neither the failing scenario nor the check that failed; that detail existed only in the global `--json` payload. `_report_scenarios` now prints one line per scenario:

```
compliant-run      PASS  100.0
non-compliant-run  FAIL  0.0   findings=['structural: manifest-matches-files: declared=[1, 2], found=[1]; …', 'structural: phase-total-matches-count: total=[2], count=1']
```

Each failing **check** is named, prefixed by its scorer; only deterministic checks are named (advisory notes stay on their own line); every scenario is shown so a passing control is visible.

**2. `eval --check` passed on a suite that scored nothing.** An empty (or entirely unscoreable) scenarios dir exited `0`, so a lost or misconfigured path in CI reported success. `gate_exit_code` now fails a `--check` that demanded a positive floor but scored nothing — the "cannot assess is not PASS" rule `spec-coverage` already applies. A non-positive floor still clears an empty suite.

## Verdict honesty

A scenario with nothing deterministically assessed (an unloadable run, or advisory-only) now reads `UNKNOWN`, not `PASS` — consistent with the gate. A `FAIL` with no finding string still names its scorer, and a malformed payload never crashes the report.

## Tests

- `gate_exit_code`: the full truth table incl. `None + --check + positive floor → 2`, `--min 0 → 0`.
- CLI: empty `--check → 2`, plain run `→ 0`, `--min 0 → 0`; the per-scenario rows are asserted through `_human_report` (wiring), not only the leaf helper.
- Human report: names the specific scenario + the specific failing check; `UNKNOWN`/malformed/detail-less-FAIL cases pinned.

Two pre-existing tests that enshrined the old "empty is honest zero" behaviour are updated with the rationale.

## Summary by CodeRabbit

- **New Features**
  - Human-readable evaluation reports now show each scenario’s **FAIL**, **PASS**, or **UNKNOWN** verdict, along with failing checks, scorer names, and finding details.
  - Failures appear first, with long reports indicating omitted scenarios.

- **Bug Fixes**
  - `--check` now fails empty or entirely unscored suites when a positive minimum compliance threshold is required.
  - Reports remain resilient when scenario data or findings are malformed.

- **Documentation**
  - Updated `--check` help text and evaluation guidance to describe behavior for unscored suites.


architecture/features/eval-harness.md
```diff
@@ -114,7 +115,7 @@ under the gate contract (only deterministic verdicts affect the exit code).
 3. [x] - `p1` - Load one scenario's run and apply every scorer to it, isolating a raising scorer as UNKNOWN - `inst-run-scenario`
 4. [x] - `p1` - Run every scenario under the root through the scorers and aggregate into a report - `inst-run-suite`
 5. [x] - `p1` - Compute structural compliance (deterministic pass ratio) per scenario and in aggregate, `None` when nothing was scored - `inst-compliance`
```
ainetx, line 117: ### Feature doc's user-facing flow omits new nothing-scored gate-exit-2 case

**Severity:** Minor

**Problem**
gate_exit_code and the CLI --check help text were updated so a positive --min floor also exits 2 when nothing was scored (empty/all-unscoreable suite), and the algo-level inst-gate doc bullet was updated to match. But the user-facing 'Run Eval Harness' flow's Success Scenarios and Error Scenarios sections still list only the two pre-existing triggers (below-floor compliance, baseline regression), per search_document_content confirming the exact unchanged text.

<details>

**How to reproduce**
1. Read architecture/features/eval-harness.md's 'Run Eval Harness' flow (Success/Error Scenarios).
2. Compare to gate_exit_code's new behavior and the updated --check help text.
3. Observe the flow section does not mention the new nothing-scored-under-positive-floor exit-2 case.

**Expected behavior**
The Success/Error Scenarios bullets should mention that --check with a positive --min also exits 2 when nothing was scored (empty or all-unscoreable suite).

**Actual behavior**
Those bullets still describe only 'below --min' and 'baseline regression' as --check exit-2 triggers.

```text
code: gate_exit_code() -> exit 2 on (below-floor OR positive-floor-and-nothing-scored)
doc inst-gate step: updated to match
doc 'Run Eval Harness' flow (user-facing): still only (below-floor OR baseline-regression)
```

**Impact**
A user reading only the flow-level Success/Error Scenarios section (the primary user-facing description) would be misled about when --check fails, missing the new empty-suite failure mode.

**Suggested correction**
Update the Success Scenarios and Error Scenarios bullets in the 'Run Eval Harness' flow to mention the new nothing-scored-under-a-positive-floor exit-2 case.

**How to verify**
Re-read the flow section after the doc update and confirm it lists all three --check exit-2 triggers, matching gate_exit_code's actual logic.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +167,53 @@ def _human_report(data: Dict[str, object]) -> None:
...
+        else:
+            verdict = "UNKNOWN"                # nothing deterministic assessed — not a pass
+        findings: List[str] = []
+        for r in deterministic:
+            raw = r.get("findings")
+            named = [f"{r.get('scorer')}: {finding}"
+                     for finding in (raw if isinstance(raw, list) else [])]
+            # A FAIL that carried no finding string still names its scorer, so the failing
+            # check is never anonymous; a non-list `findings` is ignored, never iterated.
+            findings += named or ([f"{r.get('scorer')}: (no detail)"]
```
ainetx: ### Inconsistent compliance percentage formatting between aggregate and per-scenario lines

**Severity:** Minor

**Problem**
_human_report renders aggregate structural compliance as `f"{compliance * 100:.0f}%"` right before calling _report_scenarios, which renders per-scenario compliance as `f"{comp * 100:.1f}"` — different decimal precision and a missing % unit, undocumented.

<details>

**How to reproduce**
1. Run `cfs eval` on any suite with mixed compliance. 2. Observe the printed 'structural compliance: 80%' line followed by per-scenario rows like ' my-scenario FAIL 80.0' with no % sign and an extra decimal digit.

**Expected behavior**
Both the aggregate and per-scenario compliance values use the same format (e.g. both `:.0f%` or both `:.1f%`).

**Actual behavior**
The two adjacent lines of the same report present the same metric with different precision and unit, risking reader confusion.

```text
structural compliance: 80% (aggregate,.0f%) \n my-scenario FAIL 80.0 (per-scenario,.1f, no %) -- same metric, different rendering
```

**Impact**
A human reader could mistake the per-scenario number for a different metric or unit than the aggregate compliance shown just above it.

**Suggested correction**
Format per-scenario compliance identically to the aggregate line, e.g. `f"{comp * 100:.0f}%"`.

**How to verify**
Run the human report and confirm both aggregate and per-scenario compliance lines share the same decimal precision and unit suffix.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +167,53 @@ def _human_report(data: Dict[str, object]) -> None:
...
+    failed, and UNKNOWN when nothing deterministic was actually assessed (an unloadable run,
+    or advisory-only) -- the same "cannot assess is not PASS" rule the gate applies, so the
+    console never labels an unscored scenario PASS. Every scenario is shown so a passing
+    control is visible too. Tolerates any malformed shape without raising."""
+    for row in per_scenario if isinstance(per_scenario, list) else []:
+        if not isinstance(row, dict):
+            continue
+        results = row.get("results")
+        results = results if isinstance(results, list) else []
+        deterministic = [r for r in results if isinstance(r, dict)
```
ainetx: ### No test exercises the FAIL-overrides-PASS verdict precedence in _scenario_verdict

**Severity:** Minor

**Problem**
_scenario_verdict checks `if VERDICT_FAIL in verdicts` before `elif VERDICT_PASS in verdicts` so a scenario with a mix of passing and failing deterministic scorers correctly renders FAIL. Both new tests (test_human_report_names_each_scenario_and_its_failing_check and test_report_scenarios_honest_verdict_and_malformed_tolerance) supply only a single deterministic result per scenario row.

<details>

**How to reproduce**
1. Inspect _scenario_verdict in eval.py. 2. Inspect the two new tests in test_eval_cli.py — each scenario dict's 'results' list contains exactly one deterministic entry.

**Expected behavior**
A test with a scenario containing two deterministic results (one PASS, one FAIL) asserting the row renders FAIL, protecting the precedence order.

**Actual behavior**
No such test exists; a future refactor swapping the if/elif order would go undetected by CI.

```text
scenario.results = [PASS, FAIL] -> _scenario_verdict -> (untested) -> should be FAIL
```

**Impact**
A regression that inverts precedence would silently show PASS for a scenario that actually has a real failing check, undermining the human report's reliability.

**Suggested correction**
Add a test case with two deterministic results of differing verdicts within one scenario row and assert the rendered verdict is FAIL.

**How to verify**
Add the test, then verify it fails if the if/elif order in _scenario_verdict is swapped.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +167,53 @@ def _human_report(data: Dict[str, object]) -> None:
...
+                         and r.get("kind") == eval_harness.ScorerKind.DETERMINISTIC.value]
+        verdicts = [r.get("verdict") for r in deterministic]
+        if eval_harness.VERDICT_FAIL in verdicts:
+            verdict = "FAIL"
+        elif eval_harness.VERDICT_PASS in verdicts:
+            verdict = "PASS"
+        else:
+            verdict = "UNKNOWN"                # nothing deterministic assessed — not a pass
+        findings: List[str] = []
```
ainetx: ### _scenario_findings renders findings from non-FAIL deterministic results

**Severity:** Minor

**Problem**
In _scenario_findings, `named = [f"{r.get('scorer')}: {finding}" for finding in (raw if isinstance(raw, list) else [])]` is computed from a result's `findings` list unconditionally, before any check of `r.get('verdict')`. The FAIL-gated fallback string only fires when `named` is empty. Thus a deterministic result with verdict PASS or UNKNOWN but a non-empty `findings` list (e.g. informational notes) is rendered identically to an actual failure.

<details>

**How to reproduce**
1. Construct a per_scenario row with one deterministic result: verdict='PASS', findings=['note: informational'].
2. Call _report_scenarios([row]).
3. Observe the printed line includes `findings=['scorer: note: informational']` even though the scenario passed.

**Expected behavior**
Only results whose verdict is VERDICT_FAIL should contribute to the row's findings= output.

**Actual behavior**
Any deterministic result with a non-empty findings list is included in findings=, regardless of its own verdict.

```text
for r in deterministic:
named = [scorer: f for f in r.findings] # <-- no verdict check here
out += named or (FAIL-gated fallback) # fallback only used when named empty
```

**Impact**
A passing or unscoreable scenario row can display findings=[...] that look like failure diagnostics, misleading a reader into thinking a passing/unknown scenario actually failed a check.

**Suggested correction**
Gate the findings collection on `r.get('verdict') == eval_harness.VERDICT_FAIL` before building `named`, so only failed results contribute (with the no-detail fallback for a FAIL lacking a findings list).

**How to verify**
Add a test where a deterministic result has verdict=PASS (or UNKNOWN) and a non-empty findings list, and assert the rendered row's findings= is empty / that string does not appear.

</details>

tests/test_eval_cli.py
```diff
@@ -367,6 +427,10 @@ def test_human_report_shows_compliance(capsys, tmp_path: Path) -> None:
     out = capsys.readouterr().out
     assert rc == 0
     assert "compliance" in out
+    # Wiring: _human_report must call _report_scenarios so the per-scenario rows reach the human
```
ainetx, line 498: ### Wiring test for _human_report -> _report_scenarios doesn't guard the PASS-row-shown invariant

**Severity:** Minor

**Problem**
test_human_report_shows_compliance was extended with `assert 'non-compliant-run' in out` and `assert 'FAIL' in out` to guard the wiring between _human_report and _report_scenarios, but never asserts the fixture's passing scenario ('compliant-run') or 'PASS' appears in the output, despite _report_scenarios's docstring explicitly promising 'every scenario is shown so a passing control is visible too'.

<details>

**How to reproduce**
1. Introduce a regression in _human_report (or a wrapper) that filters per_scenario to only FAIL rows before calling _report_scenarios.
2. Run test_human_report_shows_compliance.
3. Observe it still passes because it never checks for 'compliant-run' or 'PASS'.

**Expected behavior**
The wiring-level integration test should assert both the FAIL scenario and the PASS scenario (and their verdict strings) appear in the human output.

**Actual behavior**
Only the FAIL scenario/verdict is asserted at the integration/wiring level; the PASS scenario is only covered by a unit test that calls _report_scenarios directly, bypassing the real wiring.

```text
_human_report(data) --calls--> _report_scenarios(per_scenario)
test asserts: 'non-compliant-run' + 'FAIL' present
test does NOT assert: 'compliant-run' + 'PASS' present
=> a PASS-row-dropping regression in the wiring goes undetected
```

**Impact**
A future regression in _human_report's call to _report_scenarios (e.g. an accidental filter dropping PASS rows) would not be caught by any integration test, only by the unit test that bypasses the real wiring.

**Suggested correction**
Add `assert 'compliant-run' in out` and `assert 'PASS' in out` to test_human_report_shows_compliance alongside the existing FAIL assertions.

**How to verify**
Mutate _report_scenarios (or its call site) to skip PASS-verdict rows and confirm the updated test now fails.

</details>

architecture/features/eval-harness.md
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): gate_exit_code and the CLI help now document a third exit-2 trigger (positive --min with compliance=None), but architecture/features/eval-harness.md's 'Run Eval Harness' flow Success/Error Scenarios list only 'compliance below --min' and 'baseline regression' as --check failure causes.

Problem (now): gate_exit_code and the CLI --check help text were updated so a positive --min floor also exits 2 when nothing was scored (empty/all-unscoreable suite), and the algo-level inst-gate doc bullet was updated to match. But the user-facing 'Run Eval Harness' flow's Success Scenarios and Error Scenarios sections still list only the two pre-existing triggers (below-floor compliance, baseline regression), per search_document_content confirming the exact unchanged text.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: Re-verified against the current code -- this write-up has been updated. Severity: Major -> Minor.

<details>
<summary>Why</summary>

Problem (was): The findings-collection loop in _report_scenarios computes `named` from `r.get('findings')` for every deterministic result unconditionally. It only checks `r.get('verdict') == VERDICT_FAIL` in the fallback branch used when `named` is empty. If a PASS- or UNKNOWN-verdict deterministic result carries a non-empty `findings` list (e.g. an informational note), those strings are appended to the row's findings= output even though the scenario's overall verdict may be PASS.

Problem (now): In _scenario_findings, `named = [f"{r.get('scorer')}: {finding}" for finding in (raw if isinstance(raw, list) else [])]` is computed from a result's `findings` list unconditionally, before any check of `r.get('verdict')`. The FAIL-gated fallback string only fires when `named` is empty. Thus a deterministic result with verdict PASS or UNKNOWN but a non-empty `findings` list (e.g. informational notes) is rendered identically to an actual failure.

</details>

tests/test_eval_cli.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): test_human_report_shows_compliance is the only integration-level test that exercises _human_report calling _report_scenarios with a real report payload (fixtures contain both a FAIL scenario 'non-compliant-run' and a PASS scenario 'compliant-run'). The diff adds assertions only for the FAIL scenario ('non-compliant-run' in out, 'FAIL' in out), never asserting the PASS scenario or 'PASS' string appear in the output.

Problem (now): test_human_report_shows_compliance was extended with `assert 'non-compliant-run' in out` and `assert 'FAIL' in out` to guard the wiring between _human_report and _report_scenarios, but never asserts the fixture's passing scenario ('compliant-run') or 'PASS' appears in the output, despite _report_scenarios's docstring explicitly promising 'every scenario is shown so a passing control is visible too'.

</details>

ainetx (approved):
Solid change — the gate/CLI semantics for a nothing-scored suite under a positive `--min` are correctly threaded through `gate_exit_code`, the `--check` help text, and the algo-level doc bullet, and the new tests (empty-suite gating, mixed --min=0, per-scenario rendering, malformed-row tolerance) back it up well. Nothing here blocks merge — just a few small polish items worth a glance next time you're in this area:

- **Feature doc's user flow is now slightly behind the algo doc** — the new nothing-scored gate-exit-2 case was added to the CLI help and the algo-level `inst-gate` bullet, but the user-facing "Run Eval Harness" flow's Success/Error Scenarios still only mention the two original triggers (below-floor, baseline regression). ([comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4015298009))
- **Compliance formatting is inconsistent between aggregate and per-scenario lines** — the aggregate line renders `NN%` while the new per-scenario line renders `NN.N` with no `%`, and nothing indicates this is intentional. ([comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4015298253))
- **FAIL-overrides-PASS precedence isn't exercised by a mixed-scorer test** — both new tests only ever give a scenario a single deterministic result, so the `FAIL in verdicts` before `PASS in verdicts` branch is unverified against a scenario with both. ([comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4015298507))
- **`_scenario_findings` can surface non-FAIL findings as if they were failures** — `findings` is read unconditionally before checking `verdict`, so a PASS/UNKNOWN result with informational notes renders identically to an actual failure. ([comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4015298748))
- **The `_human_report` → `_report_scenarios` wiring test doesn't check the PASS row** — it asserts the failing scenario and "FAIL" appear, but never checks the passing scenario or "PASS", even though the docstring promises the passing control stays visible. ([comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4015299025))

skills/studio/scripts/studio/commands/eval.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): _human_report renders aggregate compliance as f"{compliance * 100:.0f}%" (integer percent with unit) immediately before calling _report_scenarios, which renders per-scenario compliance as f"{comp * 100:.1f}" (one decimal, no unit). No comment or spec text in inst-human-scenarios documents this as intentional.

Problem (now): _human_report renders aggregate structural compliance as `f"{compliance * 100:.0f}%"` right before calling _report_scenarios, which renders per-scenario compliance as `f"{comp * 100:.1f}"` — different decimal precision and a missing % unit, undocumented.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `_report_scenarios`'s verdict logic checks `if VERDICT_FAIL in verdicts` before `elif VERDICT_PASS in verdicts`, so a scenario with a mix of passing and failing deterministic scorers must show FAIL. Both new tests only ever supply a single deterministic result per scenario (either PASS or FAIL alone), so this precedence branch is never exercised with a mixed set.

Problem (now): _scenario_verdict checks `if VERDICT_FAIL in verdicts` before `elif VERDICT_PASS in verdicts` so a scenario with a mix of passing and failing deterministic scorers correctly renders FAIL. Both new tests (test_human_report_names_each_scenario_and_its_failing_check and test_report_scenarios_honest_verdict_and_malformed_tolerance) supply only a single deterministic result per scenario row.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +167,59 @@ def _human_report(data: Dict[str, object]) -> None:
...
+    The per-scenario rows are already in the payload; this only renders them, so a reader can
+    act on the result without re-running under the global ``--json`` flag. A verdict without
+    its findings is a bare boolean -- the most common report defect. Only deterministic checks
+    are named (an advisory scorer's notes, e.g. an unwired judge, are on their own line, and
+    repeating them per row would clutter every pass); every scenario is shown so a passing
+    control is visible too. Tolerates any malformed shape without raising."""
+    for row in per_scenario if isinstance(per_scenario, list) else []:
+        if not isinstance(row, dict):
+            continue
+        results = row.get("results")
```
ainetx: ### _report_scenarios has no output cap despite _REGRESSION_CAP existing for the same verbosity concern

**Severity:** Minor

**Problem**
_report_regression caps its output at _REGRESSION_CAP=10 rows specifically to avoid flooding the terminal. _report_scenarios, introduced in this diff, emits one ui.info line per scenario unconditionally, with no comparable cap, placed above the (capped) regression section in the human report.

<details>

**How to reproduce**
1. Run `cfs eval` against a suite with hundreds of scenarios. 2. Observe the human report: the compliance summary is followed by hundreds of per-scenario lines. 3. The (already-capped) regression summary, meant to be the actionable signal, scrolls far off-screen.

**Expected behavior**
Either an explicit cap/truncation (e.g. `_SCENARIOS_CAP`) analogous to _REGRESSION_CAP, or the regression/summary sections printed before the (uncapped) per-scenario ledger.

**Actual behavior**
Per-scenario rows are unbounded, unlike the regression rows, which the author explicitly capped for the same terminal-verbosity reason.

```text
human report: [compliance] -> [N scenario rows, uncapped] -> [regression rows, capped at 10] -- large N buries the capped, actionable section
```

**Impact**
For large eval suites, the most actionable part of the human report (regressions) becomes practically invisible without scrolling, reducing the report's usefulness at scale.

**Suggested correction**
Add a cap (with a '+N more' style continuation) to _report_scenarios consistent with _REGRESSION_CAP, or reorder sections so capped/summary sections print first.

**How to verify**
Run the CLI against a suite with >10 scenarios and confirm the regression section remains visible without excessive scrolling, or that scenario output is truncated with a count of remaining rows.

</details>

architecture/features/eval-harness.md
SanjeevSolanki: Fixed in b1971ed2 — the *Run Eval Harness* flow's Success and Error Scenarios bullets now list all three `--check` exit-2 triggers, including a positive `--min` that scored nothing (an empty or all-unscoreable suite), so the user-facing flow matches `gate_exit_code` and the `inst-gate` step.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Fixed in b1971ed2 — per-scenario compliance now formats as `{:.0f}%` (e.g. `100%`), matching the aggregate `structural compliance: N%` line exactly, so the two adjacent lines present the metric identically.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Fixed in b1971ed2 — added `test_report_scenarios_verdict_precedence_and_findings_are_fail_only`: a scenario with one PASS and one FAIL deterministic result asserts the row renders FAIL, pinning the FAIL-over-PASS precedence.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Fixed in b1971ed2 — `_scenario_findings` now skips any result whose verdict isn't FAIL before collecting, so a passing/unknown scorer's informational findings are never rendered as failures. The same new test asserts a PASS result's findings string does not appear.

tests/test_eval_cli.py
SanjeevSolanki: Fixed in b1971ed2 — the wiring test now also asserts the passing control (`compliant-run` / PASS) appears, not just the FAIL row, guarding the 'every scenario shown' invariant end-to-end.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Fixed in b1971ed2 — added `_SCENARIOS_CAP` (20) with a `+N more — see --json` continuation, mirroring `_REGRESSION_CAP`, so a large suite no longer buries the capped regression section below it.

architecture/features/eval-harness.md
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Resolved in b1971ed2; eval-harness.md Success/Error Scenarios match CLI --check help and gate_exit_code for positive-floor empty-suite gating.
- **Reviewer:** Verified against the current checked-out file: architecture/features/eval-harness.md lines 77 and 84 (Success Scenarios and Error Scenarios bullets under 'Run Eval Harness') now both list all three --check exit-2 triggers, including the new 'positive --min scored nothing (empty or all-unscoreable — failure to assess is not a pass)' case, matching gate_exit_code's logic and the CLI --check help text — confirming the discussion thread's note that this was fixed in commit b1971ed2.
- **Verifier:** Verified via search_document_content and the diff itself: architecture/features/eval-harness.md's 'Run Eval Harness' Success Scenarios and Error Scenarios bullets both now list the new 'positive --min scored nothing (empty or all-unscoreable — failure to assess is not a pass)' trigger, matching gate_exit_code's logic and the CLI --check help text exactly; the doc-code contract is consistent (fixed in b1971ed2, confirmed present in the current diff/worktree).

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Per-scenario compliance in `_report_scenarios` (line 221) formats as `f"{comp * 100:.0f}%"`, matching the zero-decimal precision and `%` suffix of aggregate compliance in `_human_report` (line 162).
- **Reviewer:** Direct read of current code confirms both lines use identical formatting: _human_report's aggregate line (eval.py:162) is f"{compliance * 100:.0f}%" and _report_scenarios' per-scenario line (eval.py:221) is now `pct = f"{comp * 100:.0f}%"` -- same precision (0 decimals) and same % unit. The previously-flagged.1f/no-% discrepancy has been fixed.
- **Verifier:** Verified directly against both the current worktree file and get_file_diff's ground-truth diff for eval.py: _report_scenarios' per-scenario line is `pct = f"{comp * 100:.0f}%"` (eval.py:221), which exactly matches _human_report's aggregate line `f"{compliance * 100:.0f}%"` (eval.py:162) — same zero-decimal precision and same % unit. The originally-flagged `.1f`/no-% discrepancy (visible only in the stale diff text pasted into this check's prompt) has been fixed in the actual PR content, consistent with the discussion thread's claim that this was fixed in b1971ed2.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_report_scenarios_verdict_precedence_and_findings_are_fail_only in tests/test_eval_cli.py exercises a scenario with both PASS and FAIL deterministic scorers and asserts that 'mixed FAIL' is rendered.
- **Reviewer:** tests/test_eval_cli.py now includes test_report_scenarios_verdict_precedence_and_findings_are_fail_only, which supplies one PASS ('a') and one FAIL ('b') deterministic result for the same scenario ('mixed') and asserts the row renders 'mixed FAIL' while the passing scorer's 'fyi note' is not rendered — this directly exercises the FAIL-over-PASS precedence branch in _scenario_verdict (checks VERDICT_FAIL before VERDICT_PASS), matching the confirmed fix from commit b1971ed2 referenced in the discussion thread.
- **Verifier:** tests/test_eval_cli.py now contains test_report_scenarios_verdict_precedence_and_findings_are_fail_only, which supplies a scenario with two deterministic results (one PASS from scorer 'a', one FAIL from scorer 'b') and asserts the row renders 'mixed FAIL' (FAIL wins over PASS) and only the FAIL's finding 'b: broke' appears while the PASS's 'fyi note' does not -- this directly exercises and pins the FAIL-over-PASS precedence branch in _scenario_verdict that was previously unprotected.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** _scenario_findings gates collection with an explicit check for VERDICT_FAIL, preventing non-failing results from contributing findings to the row as verified by test_report_scenarios_verdict_precedence_and_findings_are_fail_only.
- **Reviewer:** Current code (eval.py lines 186-199) shows _scenario_findings has `if r.get("verdict") != eval_harness.VERDICT_FAIL: continue` before collecting findings, so only deterministic results whose verdict is FAIL contribute to the row's findings= output; a FAIL with no findings list still falls back to '{scorer}: (no detail)'. This is confirmed by test_report_scenarios_verdict_precedence_and_findings_are_fail_only in tests/test_eval_cli.py (lines 183-201), which asserts a PASS result's 'fyi note' finding is never rendered while a co-occurring FAIL result's 'broke' finding is shown, and by test_report_scenarios_honest_verdict_and_malformed_tolerance covering UNKNOWN/malformed cases. This matches the fix described in the PR discussion thread (commit b1971ed2) and the prior attempt-3 verdict.
- **Verifier:** Verified directly against the actual diff and current file: _scenario_findings (eval.py lines 186-199) begins its loop with `if r.get("verdict") != eval_harness.VERDICT_FAIL: continue` before ever reading `r.get('findings')`, so only deterministic results whose own verdict is FAIL contribute to the row's findings= output; a FAIL lacking a findings list falls back to '{scorer}: (no detail)'. Tests confirm this: test_report_scenarios_verdict_precedence_and_findings_are_fail_only shows a PASS result's 'fyi note' is never rendered while a co-occurring FAIL's 'broke' is shown ('fyi note' not in out), and test_report_scenarios_honest_verdict_and_malformed_tolerance shows an UNKNOWN-verdict result with non-empty findings ('advisory-only' with ['no judge']) produces no PASS/finding leakage. The earlier attempt-1/attempt-2 finding described an unguarded version of this code that has since been fixed (per commit b1971ed2 referenced in the discussion thread), and the current diff shown in this round already contains the guard.

</details>

tests/test_eval_cli.py
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_human_report_shows_compliance in tests/test_eval_cli.py lines 453-456 explicitly asserts that 'compliant-run' and 'PASS' appear alongside the failing scenario.
- **Reviewer:** Verified directly against the checked-out worktree file (tests/test_eval_cli.py:453-456): test_human_report_shows_compliance now contains both `assert "compliant-run" in out` and `assert "PASS" in out` alongside the existing `non-compliant-run`/`FAIL` assertions, so the wiring test guards the 'every scenario shown, including passing ones' invariant end-to-end — matching the discussion thread's note that this was fixed in commit b1971ed2.
- **Verifier:** Verified directly against the current worktree file (tests/test_eval_cli.py:453-456): test_human_report_shows_compliance now contains `assert "non-compliant-run" in out`, `assert "FAIL" in out`, `assert "compliant-run" in out`, and `assert "PASS" in out`, all guarding real fixture scenario IDs confirmed in tests/fixtures/eval/compliant/scenario.toml (id="compliant-run") and tests/fixtures/eval/non_compliant/scenario.toml (id="non-compliant-run"). _human_report calls `_report_scenarios(data.get("per_scenario"))` unconditionally with the full unfiltered payload (eval.py:164), and `_report_scenarios` iterates every row with no PASS-filtering logic (eval.py:213-226), so this test would genuinely fail under the hypothesized mutation. The gap identified in earlier rounds and the posted this finding has been fixed in commit b1971ed2, as the discussion thread notes.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Confirmed fixed as of ca918ad8c77077acb5d86d62e8cab0ea3f1b6cfb -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** _SCENARIOS_CAP = 20 caps scenario rows in _report_scenarios with a '+N more — see --json' summary line when exceeded, mirroring _REGRESSION_CAP and preventing terminal flooding.
- **Reviewer:** Verified directly in the current file (skills/studio/scripts/studio/commands/eval.py lines 39, 215, 227-228): _SCENARIOS_CAP = 20 is defined alongside _REGRESSION_CAP, and _report_scenarios now iterates only rows[:_SCENARIOS_CAP] and emits a '(+N more — see --json for the full list)' line when the suite exceeds the cap, mirroring _report_regression's truncation pattern and resolving the terminal-verbosity concern raised in this check.
- **Verifier:** Verified directly in skills/studio/scripts/studio/commands/eval.py: `_SCENARIOS_CAP = 20` is defined at line 39 alongside `_REGRESSION_CAP = 10` (line 36), and `_report_scenarios` (lines 202-228) iterates only `rows[:_SCENARIOS_CAP]` and, when `len(rows) > _SCENARIOS_CAP`, emits `' (+{N} more — see --json for the full list)'` — this exactly mirrors `_report_regression`'s `regressed[:_REGRESSION_CAP]` truncation pattern (lines 307+), so the terminal-verbosity concern that motivated the original finding has been resolved in the current code.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +170,65 @@ def _human_report(data: Dict[str, object]) -> None:
...
+                         if isinstance(r, dict)
+                         and r.get("kind") == eval_harness.ScorerKind.DETERMINISTIC.value]
+        comp = row.get("compliance")
+        pct = f"{comp * 100:.0f}%" if isinstance(comp, (int, float)) else "n/a"
+        line = f"  {row.get('scenario', '?')}  {_scenario_verdict(deterministic)}  {pct}"
+        findings = _scenario_findings(deterministic)
+        if findings:
+            line += f"  findings={findings}"
+        ui.info(line)
+    if len(rows) > _SCENARIOS_CAP:
```
ainetx: ### Exact _SCENARIOS_CAP boundary (no-overflow case) is untested

**Severity:** Minor

**Problem**
The overflow condition in _report_scenarios is `len(rendered) > _SCENARIOS_CAP`. The new regression test (test_report_scenarios_shows_failures_first_and_counts_omitted_failures) only constructs 21, 23, and 27-row lists to check the overflow path, never exactly 20 rows to check the non-overflow path.

<details>

**How to reproduce**
1. Change `len(rendered) > _SCENARIOS_CAP` to `len(rendered) >= _SCENARIOS_CAP` in eval.py. 2. Run the full test_eval_cli.py suite. 3. All tests still pass because none supplies exactly 20 rows and asserts the overflow line is absent.

**Expected behavior**
A test should build a per_scenario list of exactly 20 rows and assert the output contains all 20 lines and does NOT contain a '+N more' line.

**Actual behavior**
No such test exists; only >20-row overflow cases are asserted.

```text
20 rows -> render all 20, no overflow line (never asserted) -> mutate `>` to `>=` -> false '+0 more' line emitted -> undetected by CI
```

**Impact**
A boundary regression in the truncation condition could cause a complete, non-truncated human report to falsely claim scenarios were omitted, without any test catching it.

**Suggested correction**
Add a test that calls _report_scenarios with exactly _SCENARIOS_CAP rows and asserts no '+N more' overflow line is present in the output.

**How to verify**
Add the exact-20-row test, then confirm it fails against a `>=` mutant and passes against the current `>` implementation.

</details>

skills/studio/scripts/studio/commands/eval.py
```diff
@@ -164,6 +170,65 @@ def _human_report(data: Dict[str, object]) -> None:
...
+    act on the result without re-running under the global ``--json`` flag. A verdict without
+    its findings is a bare boolean -- the most common report defect. Only deterministic checks
+    are named (an advisory scorer's notes, e.g. an unwired judge, are on their own line, and
+    repeating them per row would clutter every pass); every scenario is shown so a passing
+    control is visible too. The compliance percentage is formatted like the aggregate line
+    above it, and the ledger is capped so a large suite does not bury the regression section
+    below. Tolerates any malformed shape without raising."""
+    candidates = per_scenario if isinstance(per_scenario, list) else []
+    rows = [r for r in candidates if isinstance(r, dict)]
+    for row in rows[:_SCENARIOS_CAP]:
```
ainetx: ### Scenario-summary cap truncates by discovery order, so a failing scenario past row 20 can be invisible in the human report

**Severity:** Major

**Problem**
_report_scenarios slices `rows[:_SCENARIOS_CAP]` in the incoming per_scenario order (alphabetical by scenario directory, per load_scenarios' `sorted(root.glob(...))`), with no prioritization of FAIL verdicts and no FAIL count surfaced in the overflow message ('+N more — see --json for the full list').

<details>

**How to reproduce**
1. Create an eval suite with 21 scenarios where the first 20 alphabetically all PASS and the 21st FAILs. 2. Run `cfs eval --check` in human mode. 3. Observe the console output.

**Expected behavior**
The human summary should surface the failing scenario (e.g. by prioritizing FAIL rows, or at minimum noting a FAIL count in the remainder) so a reader debugging a failed exit code doesn't have to guess and fall back to --json.

**Actual behavior**
The console prints 20 PASS rows and '(+1 more — see --json for the full list)', giving no indication that the single omitted scenario is the one that failed and caused the exit-2.

```text
20 PASS scenarios (alphabetically first) -> printed
scenario_21 (FAIL) -> not printed, hidden in '+1 more' with no verdict hint
exit code 2 -> reader sees failure but not which scenario, contra stated goal
```

**Impact**
For exactly the reader this feature was built for (someone debugging a failing gate without --json), a real regression can be completely invisible in the default human-mode report for suites over 20 scenarios.

**Suggested correction**
Sort or partition rows so FAIL (and UNKNOWN) verdicts are shown first / always within the cap, or at minimum include a FAIL count in the overflow message (e.g. '+N more, M of which FAILED — see --json').

**How to verify**
Add a test with >20 scenarios where a FAIL sits past position 20 in list order, and assert the FAIL scenario's name/verdict appears in the printed output (or that the overflow message discloses it).

</details>

ainetx (changes requested):
Changes requested due to a regression from the earlier approval: the new capped human-readable scenario summary can hide failing scenarios, undermining the report’s usefulness for diagnosing evaluation failures.

- **Scenario-summary cap can hide failures** — Rows are truncated in discovery order, so a failed scenario after the first 20 may not appear in the human report, and the overflow notice does not indicate that failures were omitted. Prioritize failing scenarios or otherwise surface omitted failure counts. [Review comment](https://github.com/constructorfabric/studio/pull/212#discussion_r4025324808)

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Addressed in d37788d6. `_report_scenarios` now sorts FAIL → UNKNOWN → PASS (via `_VERDICT_RANK`) before applying `_SCENARIOS_CAP`, so a failing scenario is never the row the cap drops — it is shown while passes are trimmed first. When failures themselves exceed the cap, the overflow line names how many omitted rows are failures: `(+N more; M failing — see --json for the full list)`. Both halves covered — failures prioritised and omitted-failure count surfaced.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Addressed in d37788d6. `test_report_scenarios_shows_failures_first_and_counts_omitted_failures` exercises the cap boundary directly — 21, 23, and a 27-row mix — asserting the `+N more` line, FAIL-first ordering, and the omitted-failure count, including a mixed 22-fail/5-pass tail (`+7 more; 2 failing`) so the failing count cannot be confused with the total.

architecture/features/eval-harness.md
SanjeevSolanki: The user-facing "Run Eval Harness" flow already covers this: both its Success Scenarios and Error Scenarios name the empty / all-unscoreable → exit 2 case (`cfs eval --check` "… or when a positive `--min` scored nothing (an empty or all-unscoreable suite — a failure to assess is not a pass)"). So the user flow and the `inst-gate` algo bullet are consistent — no omission remains.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: These match now — the aggregate line in `_human_report` and the per-scenario line in `_rendered_scenario_row` both render `{x * 100:.0f}%` (same precision, and the `%` unit on both). The `.1f`/missing-unit divergence is gone.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: Covered by `test_report_scenarios_verdict_precedence_and_findings_are_fail_only`: it feeds one scenario mixing a PASS and a FAIL deterministic result and asserts it renders `FAIL` (FAIL wins), plus that the passing scorer's note is not shown.

skills/studio/scripts/studio/commands/eval.py
SanjeevSolanki: `_scenario_findings` guards this — the loop skips non-FAIL results first (`if r.get('verdict') != VERDICT_FAIL: continue`) before building the findings list, so only a FAIL contributes. The precedence test above also asserts a passing scorer's `fyi note` is never rendered.

architecture/features/eval-harness.md
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

skills/studio/scripts/studio/commands/eval.py
ainetx: Re-verified against the current code (d37788d6e768d88e509dbbf28bfa52d91c8055b5) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** The new test covers 21+ rows and validates overflow counts, but it never invokes `_report_scenarios` with exactly `_SCENARIOS_CAP` rows to prove that the equality boundary emits no overflow line.
- **Verifier:** Verified by reading tests/test_eval_cli.py directly: the new test_report_scenarios_shows_failures_first_and_counts_omitted_failures covers 21, 23, and 27-row overflow cases correctly, but no test anywhere in the file constructs exactly _SCENARIOS_CAP (20) rows to assert that NO '+N more' line is printed; the equality/no-truncation boundary of `len(rendered) > _SCENARIOS_CAP` (eval.py) is therefore still unexercised, so a `>` to `>=` mutation would go undetected.

</details>

skills/studio/scripts/studio/commands/eval.py
ainetx: **RESOLVED** -- Confirmed fixed as of d37788d6e768d88e509dbbf28bfa52d91c8055b5 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `_report_scenarios` stably sorts rendered rows by `_VERDICT_RANK` (FAIL=0, UNKNOWN=1, PASS=2) before applying `_SCENARIOS_CAP`, and its overflow line reports the omitted FAIL count; the new 21-, 23-, and mixed 27-row tests explicitly cover failure prioritization and omitted-failure accounting.
- **Reviewer:** Verified in eval.py:240 (`rendered.sort(key=lambda lv: _VERDICT_RANK.get(lv[1], 1))`) that rows are sorted FAIL < UNKNOWN < PASS before the `rendered[:_SCENARIOS_CAP]` slice at line 241, so a FAIL is never the row dropped by the cap even though scenarios are discovered alphabetically. The overflow line at 243-247 also names the omitted-failure count (`+N more; M failing`). This is exercised by test_report_scenarios_shows_failures_first_and_counts_omitted_failures (tests/test_eval_cli.py:204-239), which covers a late-failure-after-20-passes case, an all-failure overflow case, and a mixed 22-fail/5-pass tail that distinguishes total-omitted (7) from failing-omitted (2) — resolving the previously accepted finding, as also noted in the discussion thread (fixed in d37788d6).
- **Verifier:** Confirmed in eval.py:238-247 that _report_scenarios builds (line, verdict) tuples via _rendered_scenario_row, then sorts them with rendered.sort(key=lambda lv: _VERDICT_RANK.get(lv[1], 1)) using _VERDICT_RANK={FAIL:0, UNKNOWN:1, PASS:2} BEFORE slicing to rendered[:_SCENARIOS_CAP] — so FAIL rows always sort to the front and are never among the dropped rows unless FAILs themselves exceed the cap. The overflow branch computes omitted_fail = count of FAIL in the omitted tail and includes it in the message ('+N more; M failing'). This directly refutes the original finding (which was based on a pre-fix version lacking this sort). Verified test_report_scenarios_shows_failures_first_and_counts_omitted_failures (tests/test_eval_cli.py:204-241) exercises exactly the failure scenario described in the check: 20 passes + 1 late failure still shows the failure ('late-failure FAIL' in out, 'failing' not in the omitted-PASS overflow line), an all-FAIL 23-row overflow correctly reports '+3 more; 3 failing', and a mixed 22-FAIL/5-PASS case correctly distinguishes total omitted (7) from failing-omitted (2). The fix is real, present in the current diff, and adequately tested.

</details>

tests/test_eval_cli.py
```diff
@@ -108,16 +108,137 @@ def test_check_help_documents_baseline_regression() -> None:
...
+
+    def _pass(i): return {"scenario": f"pass-{i:02d}", "compliance": 1.0,
+                          "results": [{"scorer": "s", "kind": det, "verdict": "PASS", "findings": []}]}
+
+    def _fail(name): return {"scenario": name, "compliance": 0.0,
+                             "results": [{"scorer": "s", "kind": det,
+                                          "verdict": eval_harness.VERDICT_FAIL,
+                                          "findings": ["broke"]}]}
+
+    # exactly _SCENARIOS_CAP passes, then one failure that load-order would push past the cap
```
ainetx, line 223: ### Exact _SCENARIOS_CAP boundary (no-overflow case) is untested

**Severity:** Minor

**Problem**
The overflow condition in _report_scenarios is `len(rendered) > _SCENARIOS_CAP`. The new regression test (test_report_scenarios_shows_failures_first_and_counts_omitted_failures) only constructs 21, 23, and 27-row lists to check the overflow path, never exactly 20 rows to check the non-overflow path.

<details>

**How to reproduce**
1. Change `len(rendered) > _SCENARIOS_CAP` to `len(rendered) >= _SCENARIOS_CAP` in eval.py. 2. Run the full test_eval_cli.py suite. 3. All tests still pass because none supplies exactly 20 rows and asserts the overflow line is absent.

**Expected behavior**
A test should build a per_scenario list of exactly 20 rows and assert the output contains all 20 lines and does NOT contain a '+N more' line.

**Actual behavior**
No such test exists; only >20-row overflow cases are asserted.

```text
20 rows -> render all 20, no overflow line (never asserted) -> mutate `>` to `>=` -> false '+0 more' line emitted -> undetected by CI
```

**Impact**
A boundary regression in the truncation condition could cause a complete, non-truncated human report to falsely claim scenarios were omitted, without any test catching it.

**Suggested correction**
Add a test that calls _report_scenarios with exactly _SCENARIOS_CAP rows and asserts no '+N more' overflow line is present in the output.

**How to verify**
Add the exact-20-row test, then confirm it fails against a `>=` mutant and passes against the current `>` implementation.

</details>

architecture/features/eval-harness.md
```diff
@@ -94,6 +94,7 @@ advisory verdict can never gate a build.
 - [x] - `p1` - Load an optional baseline report JSON, degrading to no-diff on error - `inst-load-baseline`
 - [x] - `p1` - Save this run's report JSON to serve as a later baseline - `inst-save-report`
 - [x] - `p1` - Render a short human-readable summary when not in JSON mode - `inst-human-report`
+- [x] - `p1` - In the human summary, name each scenario's verdict and the checks it failed (findings prefixed by their scorer), rendering the per-scenario rows already in the payload so a reader need not fall back to the global `--json` flag - `inst-human-scenarios`
```
ainetx: ### Feature doc claims the human summary names every scenario, but output is capped and falls back to --json

**Severity:** Minor

**Problem**
The new inst-human-scenarios bullet in architecture/features/eval-harness.md states the human summary will 'name each scenario's verdict and the checks it failed... so a reader need not fall back to the global --json flag'. The implementation it documents (_report_scenarios in eval.py) caps rendered rows at _SCENARIOS_CAP=20 and for any suite larger than that prints '(+N more... see --json for the full list)', i.e. it explicitly directs the reader back to --json for the omitted scenarios.

<details>

**How to reproduce**
1. Read architecture/features/eval-harness.md line 97 (inst-human-scenarios bullet). 2. Read skills/studio/scripts/studio/commands/eval.py lines 227-247 (_report_scenarios). 3. Run a suite with more than 20 scenarios in human (non-JSON) mode and observe the '+N more — see --json for the full list' line.

**Expected behavior**
The doc should state the cap and that a reader may need to consult --json for suites beyond the cap, matching the actual _SCENARIOS_CAP=20 behavior.

**Actual behavior**
The doc asserts an unconditional 'each scenario... reader need not fall back to --json', which is false whenever a suite exceeds 20 scenarios.

```text
doc: 'each scenario named, no --json needed' --> code: _SCENARIOS_CAP=20 truncation --> overflow line says 'see --json for the full list' (contradiction)
```

**Impact**
A reader of the feature doc could reasonably expect the human console output to be a complete scenario ledger for any suite size and be surprised when large suites truncate and redirect them to --json.

**Suggested correction**
Add a clause to the inst-human-scenarios bullet noting the cap (e.g. 'up to _SCENARIOS_CAP rows, failures prioritised; the rest are summarised with a --json pointer').

**How to verify**
Re-read the updated bullet alongside _report_scenarios and confirm the cap/overflow behavior is now stated.

</details>

ainetx (approved):
Approved — the review is otherwise clean, and the remaining items are non-blocking notes worth a glance:

- **Exact `_SCENARIOS_CAP` boundary is untested** — The overflow behavior is covered, but an exactly-at-cap, non-overflow case is not. [Thread](https://github.com/constructorfabric/studio/pull/212#discussion_r4025314732)
- **Exact `_SCENARIOS_CAP` boundary is untested** — A second review thread notes the same missing boundary coverage for 20 rendered rows. [Thread](https://github.com/constructorfabric/studio/pull/212#discussion_r4034351513)
- **Human-summary documentation overstates full scenario coverage** — The docs say readers need not use `--json`, while capped output explicitly directs readers there for omitted scenarios. [Thread](https://github.com/constructorfabric/studio/pull/212#discussion_r4034352766)

tests/test_eval_cli.py
SanjeevSolanki: Addressed in 8b363f9a. `test_report_scenarios_shows_failures_first_and_counts_omitted_failures` now includes the exact-boundary case — exactly `_SCENARIOS_CAP` (20) rows — asserting all 20 render and that no `+N more` line appears, so flipping the cap condition from `>` to `>=` would fail it.

architecture/features/eval-harness.md
SanjeevSolanki: Addressed in 8b363f9a. The `inst-human-scenarios` bullet now states the cap and the fallback explicitly: rows are "capped at `_SCENARIOS_CAP` with any remainder deferred to `--json` (its count, and how many are failing, named on the overflow line)" — so it no longer claims a reader never needs `--json`.

PR #217, opened 2026-09-15
test(core-infra): reuse the shared sanitised git environment in the reversal fixtures

SanjeevSolanki:
A follow-up to #207, closing a review finding that arrived after it merged.

## One list, not two

The reversal fixtures run git through a helper of their own. It inherited the caller's environment, which review flagged: `GIT_DIR`, `GIT_WORK_TREE` and friends are honoured even when `cwd` points at an isolated `tmp_path`, so a developer with one of them exported would have had the fixtures initialising, committing to and creating worktrees in whatever repository it named.

**The first fix wrote a second list, and that is what this corrects.** It named eight redirect variables — the ones that came to mind — and omitted every *configuration* channel the shipped helper already carries:

- `GIT_CONFIG_PARAMETERS`, and `GIT_CONFIG_COUNT` with the indexed `GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n` pairs it gates
- `GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `GIT_CONFIG_NOSYSTEM`
- `GIT_DISCOVERY_ACROSS_FILESYSTEM`

Those are not a longer list of the same kind. They are a second mechanism — git takes *configuration* from the environment as well as location — and each was added to `change_summary._GIT_REDIRECT_VARS` by measurement, in response to a real failure. The comments there record one: `GIT_CONFIG_SYSTEM` pointing at a file that set `core.excludesFile` emptied the untracked sweep, and `GIT_CONFIG_NOSYSTEM=1` brought it back.

A copy of a list like that is the one that stops being updated. The fixture helper now calls `change_summary._git_env()`, so there is a single place this is maintained and the next variable added there reaches the tests without anyone remembering to copy it.

## What holds it

A test asserts the reuse directly — that the helper calls `_git_env()` rather than building an environment of its own — and, separately, that the returned environment carries none of the redirect variables. The first assertion stops the duplicate coming back; the second would fail if the shared helper itself regressed.

## Verification

Gates run locally on this branch: `cfs validate` PASS (0 errors, 0 warnings, 242/242); `make test` **6167 passed**; spec-coverage 90.9% coverage, **0.4602** granularity; pylint, vulture-ci, test-coverage, test-gates, self-check, check-versions and validate-kits clean.

`make ci` cannot pass on this repo for anyone (#197), so the evidence above is CI's own targets, run individually.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protection against Git environment settings redirecting repository operations or affecting repository discovery.
  * Git commands now consistently use a sanitized environment, including namespace-related settings.

* **Tests**
  * Expanded coverage for environment sanitization, preservation of unrelated settings, and prevention of repository redirection.
  * Consolidated validation around shared environment handling to improve reliability and reduce duplicated test logic.


tests/test_armed_reversal.py
```diff
@@ -492,6 +489,27 @@ def _broken(name, *args, **kwargs):
...
+        source = inspect.getsource(_git)
+        assert "change_summary._git_env()" in source, source
+        # And the helper still removes the config channels a local list forgot.
+        for channel in ("GIT_CONFIG_PARAMETERS", "GIT_CONFIG_COUNT", "GIT_CONFIG_GLOBAL",
+                        "GIT_CONFIG_SYSTEM", "GIT_CONFIG_NOSYSTEM",
+                        "GIT_DISCOVERY_ACROSS_FILESYSTEM"):
+            assert channel in change_summary._GIT_REDIRECT_VARS, channel
+        # Asserted on behaviour too, not only on the call: a sanitised environment is one
+        # that does not carry these out to the subprocess.
+        assert not set(change_summary._git_env()) & set(change_summary._GIT_REDIRECT_VARS)
```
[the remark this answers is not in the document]
SanjeevSolanki: Correct, and confirmed by mutation before fixing: with the sanitisation removed from `_git_env` entirely, this test still passed. No redirect variable is set in a normal shell or under pytest, so the intersection was empty either way and the assertion proved nothing.

It now seeds **every** name in `_GIT_REDIRECT_VARS` through `monkeypatch` before calling the helper, and asserts none of them survive — naming the ones that do, so a partial failure says which. It also checks the result is a sanitised copy rather than an empty dict, since an empty environment would pass the first assertion while leaving the fixtures without a `PATH`.

Three mutations now fail it where the previous version caught none: sanitisation removed entirely, only the first variable removed, and the environment emptied instead of sanitised. That middle one matters most here — it is the shape this PR exists to prevent, a list that covers some channels and not others.

One note on my own process, since it nearly let this through. My first mutation check reported "caught" for a test node ID that did not exist: pytest exits 4 on an unmatched node and my harness read any non-zero exit as a caught mutation. I now assert the baseline is green and that a test actually ran before trusting the result.

tests/test_armed_reversal.py
```diff
@@ -492,6 +490,39 @@ def _broken(name, *args, **kwargs):
 class TestTheRemovalTriggerIsCheckedRatherThanWritten:
     """A comment saying "delete these when a consumer appears" is not a constraint."""
 
+    def test_the_fixture_git_uses_the_shared_sanitised_environment(
```
ainetx, line 498: ### New git-env-dedup test filed under an unrelated 'removal trigger' test class

**Severity:** Minor

**Problem**
test_the_fixture_git_uses_the_shared_sanitised_environment (line 493) tests that _git delegates to change_summary._git_env()/_GIT_REDIRECT_VARS instead of duplicating the list -- unrelated to TestTheRemovalTriggerIsCheckedRatherThanWritten's stated theme (a vulture-whitelist deletion trigger being checked rather than merely commented). The class docstring was not updated and its only other member is genuinely about the whitelist trigger.

<details>

**How to reproduce**
1. Open tests/test_armed_reversal.py. 2. Read class docstring at line 491. 3. Read the new test's docstring/body at lines 493-524. 4. Observe no shared subject between the class docstring and the new test.

**Expected behavior**
The git-env sanitisation test should live in its own class (or alongside other _git/env-sanitisation coverage, e.g. near TestWhatReviewFoundOnTheFirstPush), with a docstring describing that theme.

**Actual behavior**
The test is nested in a class whose docstring and sole sibling test are about the vulture whitelist removal trigger, making the class docstring inaccurate about its full contents.

```text
new import/env-dedup concern --> dropped into nearest existing class --> class docstring (whitelist removal trigger) no longer matches all members
```

**Impact**
Reduces discoverability of the git-env-sanitisation guard for future maintainers relying on class docstrings, raising the odds the shared-list invariant is duplicated again without a maintainer noticing a test already exists.

**Suggested correction**
Move test_the_fixture_git_uses_the_shared_sanitised_environment into its own class (e.g. TestTheFixtureGitEnvironmentIsShared) or into TestWhatReviewFoundOnTheFirstPush with an updated docstring.

**How to verify**
Re-read the class docstring against all its members and confirm each member's subject matches.

</details>

ainetx, line 498: ### No end-to-end test that a leaked redirect variable is actually kept out of the fixtures' real git subprocess

**Severity:** Minor

**Problem**
The new test verifies change_summary._git_env()'s own output and _git's source text, but never exercises _git()/the repo fixture with an ambient redirect variable set and observes the resulting git invocation still targets tmp_path.

<details>

**How to reproduce**
1. Set GIT_DIR in the ambient environment via monkeypatch.setenv before any test. 2. Call _git(root, 'init') via the repo fixture. 3. Observe: no test in this file asserts the resulting repository is still rooted at tmp_path rather than redirected.

**Expected behavior**
A test that seeds one or more _GIT_REDIRECT_VARS in the ambient environment, then runs the repo fixture / _git() directly, and asserts the operation still targets tmp_path (e.g. via git rev-parse --show-toplevel).

**Actual behavior**
Only structural/unit assertions exist (source-text substring check, direct _git_env() call), which a subtle env-merging regression in _git could satisfy while still leaking ambient vars into the real subprocess.

```text
_git changed to env={**os.environ, **change_summary._git_env()} --> substring check passes, _git_env() output still clean --> but merged env leaks ambient GIT_DIR into subprocess.run --> undetected
```

**Impact**
A regression reintroducing exactly the ambient-redirect bug this file exists to catch could ship with this test suite green.

**Suggested correction**
Add a behavioural test that monkeypatches a redirect variable, invokes the repo fixture/_git(), and asserts the git operation still resolves to tmp_path.

**How to verify**
Introduce the env={**os.environ, **change_summary._git_env()} mutation locally and confirm the new behavioural test fails while the existing structural test still passes.

</details>

tests/test_armed_reversal.py
```diff
@@ -18,26 +18,24 @@
...
+    `GIT_CONFIG_COUNT` and the indexed `KEY_n`/`VALUE_n` pairs it gates,
+    `GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `GIT_CONFIG_NOSYSTEM` and
+    `GIT_DISCOVERY_ACROSS_FILESYSTEM`. Those were added to the production list by
+    measurement, in response to real failures, and a second copy is exactly how that work
+    gets lost: the copy is the one that stops being updated. Raised in review, twice —
+    the first fix wrote the duplicate, the second removed it.
+    """
     subprocess.run(["git", *args], cwd=str(cwd), check=True,
-                   capture_output=True, text=True, env=_clean_env())
+                   capture_output=True, text=True, env=change_summary._git_env())
```
ainetx, line 38: ### Switching the fixtures to the shared redirect list silently drops GIT_NAMESPACE sanitisation

**Severity:** Minor

**Problem**
The removed local _REDIRECTS tuple cleared GIT_NAMESPACE; change_summary._GIT_REDIRECT_VARS, now the sole source for the fixtures' git env, does not include it anywhere, and a full-repo search confirms no code clears it. GIT_NAMESPACE redirects which refs/objects a git invocation resolves, the same class of vector as GIT_DIR/GIT_WORK_TREE this list exists to neutralise.

<details>

**How to reproduce**
1. grep the repo for GIT_NAMESPACE -- zero results. 2. Compare to the diff's removed _REDIRECTS tuple, which listed it explicitly. 3. Confirm change_summary._GIT_REDIRECT_VARS (lines 110-157) omits it.

**Expected behavior**
Either GIT_NAMESPACE is added to change_summary._GIT_REDIRECT_VARS (restoring parity with the old local list) or a comment documents why it is deliberately excluded (as core.worktree is documented as a non-vector).

**Actual behavior**
GIT_NAMESPACE is cleared nowhere in the codebase after this diff, and the new test's docstring implies the migration only adds coverage, not that it also drops a variable.

```text
old local _REDIRECTS (clears GIT_NAMESPACE) --removed by diff--> fixtures now use change_summary._GIT_REDIRECT_VARS (no GIT_NAMESPACE) --> ambient GIT_NAMESPACE no longer sanitised for these fixtures
```

**Impact**
A developer with GIT_NAMESPACE exported now has these test fixtures (and all production change_summary git calls) unprotected against namespace-based ref redirection.

**Suggested correction**
Add GIT_NAMESPACE to change_summary._GIT_REDIRECT_VARS, and extend the new test's channel list to assert it stays covered.

**How to verify**
Set GIT_NAMESPACE in the environment, call change_summary._git_env(), and confirm it is stripped.

</details>

ainetx (approved):
Approved — the shared git-environment sanitisation change is otherwise clean. The following non-blocking notes are the only remaining items worth a follow-up:

- **Test placement** — The new shared-environment test sits under a removal-trigger test class, which makes the test's purpose harder to discover. [Comment thread](https://github.com/constructorfabric/studio/pull/217#discussion_r4025236532)
- **End-to-end fixture coverage** — Consider exercising the fixture’s real git subprocess with an ambient redirect variable to verify it still operates in `tmp_path`. [Comment thread](https://github.com/constructorfabric/studio/pull/217#discussion_r4025236784)
- **`GIT_NAMESPACE` sanitisation** — Reusing the shared list drops the fixture’s previous clearing of `GIT_NAMESPACE`, leaving a git redirection channel unsanitised. [Comment thread](https://github.com/constructorfabric/studio/pull/217#discussion_r4025237081)

tests/test_armed_reversal.py
SanjeevSolanki: Fixed here rather than deferred, because this one is a hole **this PR opens** rather than a pre-existing gap.

Confirmed: the shared tuple had 13 variables and `GIT_NAMESPACE` was not among them, while the local list it replaced did carry it. So the change arguing "one maintained list beats a stale copy" dropped a channel the copy had closed. Both lists were incomplete, in different places.

`GIT_NAMESPACE` is now in `_GIT_REDIRECT_VARS` rather than back in a local list — the production queries have the same exposure, and it is a third mechanism distinct from the two the tuple already covers: with it set, every ref lookup is rewritten under `refs/namespaces/<name>/`, so `rev-parse --abbrev-ref HEAD` and `symbolic-ref` answer about a namespace the caller never asked about. Location and configuration both correct, refs redirected.

The reversal test now names it among the channels it asserts are removed; deleting it from the tuple fails that test.

This does mean the incompleteness argument cuts both ways, and the comment in the tuple says so: one list is still right, but only once it holds the union rather than the larger of the two.

tests/test_armed_reversal.py
SanjeevSolanki: Agreed, and taking this as the follow-up you suggested rather than growing an approved PR with test-shape changes.

You are right that the class is wrong: the test asserts a property of the shared environment helper and sits under a removal-trigger class whose subject is the dead-code whitelist. Someone looking for environment sanitisation will not find it there.

Noted for the follow-up alongside the end-to-end coverage point, which belongs in the same change — a fixture-level test exercising the real git subprocess with an ambient redirect set is the thing that would have caught the `GIT_NAMESPACE` drop directly, rather than by list comparison.

SanjeevSolanki: Agreed, and this is the one I would most like in the follow-up.

Everything currently asserted about sanitisation is a comparison of lists — that the helper's tuple contains certain names, and that its output does not. None of it proves the fixture's **real** git subprocess still operates in `tmp_path` when a redirect variable is set in the ambient environment.

That gap is not theoretical: the `GIT_NAMESPACE` drop in the sibling thread was found by reading the two lists against each other, and an end-to-end test would have caught it by behaviour instead — set `GIT_NAMESPACE`, run the fixture, observe the refs it actually resolves.

Will pair it with the test-placement move so the follow-up lands as one coherent change rather than three small ones.

tests/test_armed_reversal.py
SanjeevSolanki: **Correction to my reply above: the mechanism I gave for `GIT_NAMESPACE` is wrong, and I asserted it without testing it.**

I claimed that with it set, `rev-parse --abbrev-ref HEAD` and `symbolic-ref` answer about a namespace. They do not. Measured against every command this helper guards — `rev-parse --is-inside-work-tree`, `--absolute-git-dir`, `--verify --quiet HEAD`, `symbolic-ref --quiet --short HEAD`, `rev-parse main`, `show-ref`, `for-each-ref`, `branch --list` — the output is identical with and without it set. `GIT_NAMESPACE` is a ref-*advertisement* mechanism: it bites `fetch`, `ls-remote` and the pack protocols, not local resolution.

So the entry stays, but for your reason rather than mine: the fixtures' local list carried it, this tuple did not, and adopting the tuple dropped it — one list should hold the union of two partial ones. Not because it changes what these queries answer. The comment in the tuple now records the measurement, including that an earlier version of it claimed the opposite.

One thing the measurement turned up, outside this PR. `git_utils._run_git` passes no `env=` at all, so it inherits the ambient environment wholesale, and `git_utils.py:320` runs `fetch --quiet origin` — a remote-facing command, which is exactly where `GIT_NAMESPACE` does bite. That looks like a real exposure to me, but it is a different module and predates this change; happy to raise it as its own issue if you agree it is worth one.

tests/test_armed_reversal.py
SanjeevSolanki: **Fixed here after all, rather than deferred** — you had already approved, but three deferrals on one PR is worse than one more push.

Moved to its own class, `TestTheFixtureGitCannotBeRedirected`, whose docstring says why it moved: the removal-trigger class is about the dead-code whitelist, and someone looking for environment sanitisation would not have found it there.

SanjeevSolanki: **Fixed here after all, rather than deferred.**

`test_an_exported_git_dir_does_not_move_the_fixture_repository` exports `GIT_DIR` pointing at a decoy repository, builds the fixture, and asserts git wrote into `tmp_path` and not into the decoy. Behaviour, running the real subprocess, rather than a comparison of lists.

Worth recording that I first wrote it with `GIT_NAMESPACE` — exactly as I said in this thread I would — **and it was vacuous: it passed with the entry removed from the tuple.** That is what sent me to measure, and the answer is in the sibling thread: `GIT_NAMESPACE` changes none of the commands this helper runs. `GIT_DIR` is the right probe because it does.

Mutation over the whole tuple now, rather than over the one entry I changed: removing any of the fourteen entries fails a test somewhere in the suite, and removing `GIT_DIR` fails this one specifically.

ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The finding is resolved in the current diff: `test_the_fixture_git_uses_the_shared_sanitised_environment` now belongs to the separate `TestTheFixtureGitCannotBeRedirected` class, immediately before `TestTheRemovalTriggerIsCheckedRatherThanWritten`. The new class docstring explicitly identifies environment sanitisation as its subject, so the removal-trigger class no longer has inaccurate contents.

Noted -- recording this as a repo-wide rule for future reviews: "Always verify the current test-class boundaries and docstrings before concluding that a test is grouped under an unrelated class."

</details>

ainetx: **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The finding is addressed by `test_an_exported_git_dir_does_not_move_the_fixture_repository`: it sets ambient `GIT_DIR`, invokes `_git(root, "init", "--quiet")`, and asserts `.git` is created under `tmp_path` while the decoy receives no refs. This directly exercises the fixture subprocess rather than only checking `_git_env()` output or `_git` source.

Noted -- recording this as a repo-wide rule for future reviews: "Always verify whether an integration test already exercises the real subprocess with the relevant ambient environment before concluding that environment sanitization lacks end-to-end coverage."

</details>

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Confirmed fixed as of a8d7dd6e9fe9e06b68bcd4ba0acab645bc478c60 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff adds `GIT_NAMESPACE` to `change_summary._GIT_REDIRECT_VARS` (line 172), documents why the shared list must retain the former fixture coverage, and the new test explicitly asserts its presence (tests/test_armed_reversal.py:515-521).
- **Reviewer:** The current worktree already contains the fix: change_summary._GIT_REDIRECT_VARS (skills/studio/scripts/studio/utils/change_summary.py:157-172) now includes "GIT_NAMESPACE", with a comment documenting that it was re-added because the removed local _REDIRECTS list in tests/test_armed_reversal.py had cleared it while the shared tuple originally did not, so adopting the shared tuple alone would have dropped it. The new test TestTheFixtureGitCannotBeRedirected.test_the_fixture_git_uses_the_shared_sanitised_environment (tests/test_armed_reversal.py:512-521) explicitly asserts "GIT_NAMESPACE" is a member of change_summary._GIT_REDIRECT_VARS, and the behavioural half (lines 522-533) sets every name in that tuple -- including GIT_NAMESPACE -- via monkeypatch and asserts none leak through _git_env(). So the fixtures do still sanitise GIT_NAMESPACE post-migration; the coverage regression this check describes has been closed in the current diff, matching the discussion thread's confirmed correction and fix.
- **Verifier:** Verified directly in the current worktree: change_summary._GIT_REDIRECT_VARS (skills/studio/scripts/studio/utils/change_summary.py:110-173) now includes "GIT_NAMESPACE" at line 172, with a comment (lines 157-171) explaining it was re-added because the removed local _REDIRECTS list in tests/test_armed_reversal.py had cleared it while the original shared tuple did not, and documenting (via the discussion thread's measurement) that it doesn't affect the local git queries this module runs but is kept so the shared list holds the union of both partial lists. tests/test_armed_reversal.py's TestTheFixtureGitCannotBeRedirected.test_the_fixture_git_uses_the_shared_sanitised_environment explicitly asserts "GIT_NAMESPACE" is a member of _GIT_REDIRECT_VARS (line 520-521) and the behavioural loop (lines 528-532) sets every name in that tuple, including GIT_NAMESPACE, via monkeypatch and asserts none leak through _git_env(). The coverage regression this check describes has been closed in the current diff; it is not silently dropped.

</details>

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (dismissed-by-agreement), and this doesn't change that.

ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (dismissed-by-agreement), and this doesn't change that.

tests/test_armed_reversal.py
ainetx: **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

PR #218, opened 2026-09-16
test(migrate): stop the migration tests calling the real GitHub API

### LINKED ISSUE #214: Tests: the migrate-from-cypilot family calls the real GitHub API, so it fails at random in CI
`tests/test_migrate_from_cypilot.py` fails intermittently with `AssertionError: assert 'WARN' == 'PASS'`, on a different subset of tests each run. It is not order-dependent noise: these tests make **real network calls to the GitHub API**.

## Evidence

CI, PR #211 — a change touching only `ci.yml` triggers, which cannot affect test outcomes:

```
Test (Python 3.13):  14 failed, 6118 passed
Test (Python 3.11):  pass
Test (Python 3.12):  pass
Test (Python 3.14):  pass
```

Re-running the same job with no code change: **25/25 pass**.

Locally on clean `main`, `tests/test_migrate_from_cypilot.py tests/test_cli_setup_e2e.py` — one failure per run, three runs in a row, a different test each time:

```
run 1: 1 failed, 174 passed
run 2: 1 failed, 174 passed
run 3: 1 failed, 174 passed
```

Observed failing members so far: `test_supported_legacy_versions_migrate_directly[3.9.0]`, `[3.10.0]`, `test_interactive_init_prompts_and_accepting_migrates`, `test_unsupported_legacy_version_prompts_to_update_then_migrates`, `test_non_interactive_unsupported_legacy_updates_with_explicit_approval`, `test_implicit_migration_uses_install_dir_as_target`, `test_implicit_migration_allows_install_dir_matching_legacy_dir`, `test_update_migrate_yes_on_legacy_project_migrates_without_prompt`, `test_update_unsupported_legacy_version_can_update_baseline_then_migrate`, `test_init_update_legacy_studio_yes_updates_baseline_then_migrates` — i.e. effectively the whole `followup_update_ok` family.

## Root cause

The migration result is `WARN` rather than `PASS` when any warning mentions a failure:

```python
# commands/migrate_from_cypilot.py:379
has_failure_warning = any("failed" in warning.lower() for warning in state.warnings)
result = {"status": "PASS" if update_rc in (None, 0) and not has_failure_warning else "WARN", ...}
```

There are **two** follow-up calls, and the shared fixture stubs only one of them:

```python
# tests/test_migrate_from_cypilot.py:137
@pytest.fixture
def followup_update_ok(monkeypatch):
    monkeypatch.setattr(migration, "_run_followup_update", lambda ...: (0, {"status": "PASS", ...}))
```

`_run_followup_kit_update` is left live, and it fails into exactly that warning:

```python
kit_update_rc = _run_followup_kit_update(project_root=state.project_root, yes=state.yes)
if kit_update_rc:
    state.warnings.append("follow-up kit update failed")
```

And that call goes to the network. Its own docstring says so — *"queries each kit's (now renamed) source for the latest tag, downloads…"* — and the path bottoms out in:

```python
# commands/kit.py:400
url = apply_override(f"https://api.github.com/repos/{owner}/{repo}/tags?per_page=100")
req = urllib.request.Request(url, headers=_github_headers())
```

So every test using `followup_update_ok` performs an unauthenticated GitHub API request. `make test` runs with `-n 6`, so six workers issue them concurrently. Unauthenticated GitHub API allows 60 requests/hour per IP; on a shared CI runner that budget is often already spent. A throttled or transient request makes `cmd_kit_update` return non-zero → the warning → `WARN` → the assertion fails.

This explains every symptom: non-deterministic, a different victim each run, worse under `-n 6`, different results across the four Python jobs of a single CI run, and green on a plain re-run.

## Why it matters now

Until #211 the pipeline did not run on release branches, so this was invisible there. It now runs on `release/**`, which means a red `Test (…)` job will appear at random during release acceptance — precisely when a red check is most expensive to interpret, and precisely where "just re-run it" is the habit worth not teaching.

It is also a correctness point beyond flakiness: a unit test suite should not depend on api.github.com being reachable, unthrottled, and returning the tags it expects. The suite currently cannot run offline.

## Suggested fix

Stub the second call in the same fixture, the way the first one already is:

```python
@pytest.fixture
def followup_update_ok(monkeypatch):
    monkeypatch.setattr(migration, "_run_followup_update",
                        lambda project_root, *, yes: (0, {"status": "PASS", "project_root": project_root.as_posix()}))
    monkeypatch.setattr(migration, "_run_followup_kit_update",
                        lambda *, project_root, yes: 0)
```

`test_run_followup_kit_update_passes_project_root` (line 2853) already covers that function's own contract directly, so stubbing it here loses no coverage — it only stops every *other* test from exercising it by accident.

Worth considering alongside:

- [ ] A guard that fails loudly if the suite reaches the network at all (e.g. a `socket` autouse fixture refusing outbound connections), so the next such dependency is caught at authoring time rather than as a flake months later.
- [ ] An audit for other tests reaching `api.github.com` — this fixture is unlikely to be the only one.

## Scope

- [ ] Stub `_run_followup_kit_update` in `followup_update_ok`
- [ ] Confirm the family is stable over repeated runs, including with `-n 6`
- [ ] Decide on the network guard

vasylcf:
Closes #214.

## The cause

`tests/test_migrate_from_cypilot.py` failed intermittently with `assert 'WARN' == 'PASS'`, a different subset each run. Not test ordering — **these tests were making real network calls**.

The migration makes two follow-up calls. The shared fixture stubbed one:

```python
@pytest.fixture
def followup_update_ok(monkeypatch):
    monkeypatch.setattr(migration, "_run_followup_update", lambda ...: (0, {"status": "PASS", ...}))
    # _run_followup_kit_update left live
```

`_run_followup_kit_update` is not local. `cmd_kit_update` queries `api.github.com` for each kit's latest tag and downloads from `codeload.github.com`. Its non-zero return appends `"follow-up kit update failed"`, and:

```python
has_failure_warning = any("failed" in warning.lower() for warning in state.warnings)
result = {"status": "PASS" if update_rc in (None, 0) and not has_failure_warning else "WARN", ...}
```

Unauthenticated GitHub allows 60 requests/hour per IP, `make test` runs `-n 6`, and a CI runner's IP is shared. A throttled request became a failing assertion in tests that have nothing to do with kits.

## Measured, not inferred

Wrapping `socket.connect` and counting, for this module alone:

| | connections | to |
|---|---|---|
| before | **48** | 36 × `api.github.com` (20.205.243.168), 12 × `codeload.github.com` (20.205.243.165) |
| after | **0** | — |

Runtime of `test_migrate_from_cypilot.py` + `test_cli_setup_e2e.py`: **16–18s → ~5s**. That difference was time spent waiting on GitHub.

Stability, same two modules:

```
before:  run 1: 1 failed, 174 passed     after:  6 consecutive runs, 175 passed
         run 2: 1 failed, 174 passed
         run 3: 1 failed, 174 passed     (a different test each time)
```

Full suite: **6166 passed, 0 failed.**

## The change

Test-side only — no production code is touched.

- `followup_update_ok` stubs `_run_followup_kit_update` too. Its own contract is already covered directly by `test_run_followup_kit_update_passes_project_root`, so this costs no coverage; it only stops two dozen unrelated tests exercising it, and the network, by accident.
- `test_internal_migration_preserves_partial_success_when_followup_update_fails` stubs it inline for the same reason: it patched the first follow-up and not the second, putting four GitHub requests behind an assertion about the *update* path.

## What this PR does not fix

This module is not the only offender. Measured the same way:

| module | connections |
|---|---|
| `tests/test_kit.py` | 22 |
| `tests/test_cli_setup_e2e.py` | 12 |
| `tests/test_whatsnew.py` | did not finish within 180s |

A full-suite audit run with the recorder attached was still going at **51 minutes**, against the 8–17 minutes the suite normally takes — so the network share across the whole suite is substantial.

Left out of this PR on purpose: fixing those is a larger change, and a suite-wide guard that fails loudly on any outbound connection needs to know which tests legitimately need one before it can be turned on. Both are recorded on #214 so the remaining work is visible rather than assumed done.

## Summary by CodeRabbit

* **Tests**
  * Improved migration test reliability by stubbing follow-up update operations that could trigger external API and download activity.
  * Preserved coverage for partial-success behavior when a follow-up update fails.


ainetx:
### Sibling e2e tests in test_cli_setup_e2e.py still exercise the live GitHub kit-update path this PR set out to eliminate

**Severity:** Major

**Problem**
test_init_from_dir_migrates_explicit_legacy_directory, test_init_migrate_yes_migrates_legacy_project_without_prompt, and test_init_update_legacy_studio_yes_updates_baseline_then_migrates each call `init --migrate-from-cypilot yes` without --skip-update (init.py hardcodes skip_update=False at line 464), and each only patches `studio.commands.migrate_from_cypilot._run_followup_update` (lines 887, 948, 989) -- none patches `_run_followup_kit_update`. `_run_followup_steps` (migrate_from_cypilot.py:352-365) unconditionally calls both when skip_update/dry_run are False. `_bootstrap_legacy_project` writes a real `source = "github:cyberfabric/cyber-pilot-kit-sdlc"` kit entry into core.toml, which is exactly the shape that drives cmd_kit_update to hit api.github.com/codeload.github.com for real.

<details>

**How to reproduce**
1. Run tests/test_cli_setup_e2e.py::test_init_from_dir_migrates_explicit_legacy_directory (or either sibling) in CI. 2. `_run_followup_kit_update` runs live, unstubbed, and calls cmd_kit_update against the real GitHub API for the 'sdlc' kit.

**Expected behavior**
All test paths that exercise migrate_from_cypilot with skip_update=False should have _run_followup_kit_update stubbed, exactly as this PR now does in test_migrate_from_cypilot.py.

**Actual behavior**
Three tests in test_cli_setup_e2e.py still reach the live network path, reproducing the same GitHub-throttling flakiness (issue #214) this PR fixes elsewhere.

```text
init(skip_update=False) -> migrate_from_cypilot -> _run_followup_steps -> _run_followup_update (patched) + _run_followup_kit_update (NOT patched) -> cmd_kit_update -> api.github.com/codeload.github.com
```

**Impact**
CI can still be throttled/made flaky by GitHub rate limiting through this second test file, so the PR's stated goal of eliminating live network calls from the migration test suite is only partially achieved.

**Suggested correction**
Add `patch("studio.commands.migrate_from_cypilot._run_followup_kit_update", return_value=0)` (or equivalent) alongside the existing _run_followup_update patch in all three e2e tests.

**How to verify**
Run the three e2e tests with network access disabled/mocked and confirm they still pass without any outbound call to api.github.com or codeload.github.com.

</details>

---
_Original location: tests/test_cli_setup_e2e.py:887 -- inline anchoring could not be resolved after 1 attempt(s)._

tests/test_migrate_from_cypilot.py
```diff
@@ -136,13 +136,29 @@ def _snapshot_tree(root: Path) -> dict[str, bytes]:
...
+    unrelated tests from exercising it, and the network, by accident.
+    """
     import studio.commands.migrate_from_cypilot as migration
 
     monkeypatch.setattr(
         migration,
         "_run_followup_update",
         lambda project_root, *, yes: (0, {"status": "PASS", "project_root": project_root.as_posix()}),
     )
+    monkeypatch.setattr(migration, "_run_followup_kit_update", lambda *, project_root, yes: 0)
```
ainetx, line 161: ### No systemic network-blocking fixture guards against recurrence of the exact bug class this PR fixes

**Severity:** Minor

**Problem**
tests/conftest.py centralizes cross-cutting test isolation via three autouse fixtures (_enable_json_mode, _restore_studio_logger, _isolate_decision_log) but none of them blocks real outbound sockets/HTTP. The only mitigation for the '_run_followup_kit_update left live' bug class (issue #214) is per-test/per-file manual monkeypatching, which this very PR shows is easy to miss (it had to be applied twice in this one file, and a sibling call site in test_cli_setup_e2e.py was also missed, per ).

<details>

**How to reproduce**
1. A new test is added that exercises migrate_from_cypilot or a similar command with skip_update=False. 2. The author forgets to monkeypatch _run_followup_kit_update (as almost happened twice in this very diff). 3. The test suite silently makes real calls to api.github.com/codeload.github.com instead of failing fast with a clear 'network blocked' error.

**Expected behavior**
A shared, autouse fixture in tests/conftest.py (e.g. blocking socket.socket or monkeypatching urlopen/requests) should make any accidental real network call fail immediately and loudly, rather than allowing flaky, rate-limited, or slow test runs.

**Actual behavior**
No such guard exists; protection relies entirely on remembering to stub each individual follow-up call in each individual test file.

```text
New test -> calls migrate_from_cypilot(skip_update=False) -> forgets to stub _run_followup_kit_update -> cmd_kit_update -> real HTTP to api.github.com/codeload.github.com -> flaky/slow/rate-limited CI, no fast failure
```

**Impact**
Future tests can become slow, flaky, rate-limited (60 req/hr unauthenticated), or dependent on external GitHub availability, with failures misreported as migration WARN/ERROR status rather than as unintended network access.

**Suggested correction**
Add an autouse fixture in tests/conftest.py that blocks real socket connections (e.g. patches socket.socket to raise, or monkeypatches urllib/requests) for the whole test session, with an explicit opt-in for the few tests that intentionally need network access.

**How to verify**
Add the guard fixture, then temporarily remove one of the _run_followup_kit_update monkeypatches added in this diff and confirm the test fails immediately with a clear network-blocked error rather than making a real HTTP call.

</details>

ainetx (changes requested):
Requesting changes: this PR set out to remove the live GitHub kit-update path from tests, but the fix is incomplete -- one sibling test file still exercises it.

- **Sibling e2e tests still hit the live GitHub kit-update path** -- `test_init_from_dir_migrates_explicit_legacy_directory`, `test_init_migrate_yes_migrates_legacy_project_without_prompt`, and `test_init_update_legacy_studio_yes_updates_baseline_then_migrates` in `test_cli_setup_e2e.py` call `init --migrate-from-cypilot yes` without `--skip-update`, and each only patches `_run_followup_update`, leaving `_run_followup_kit_update` live. Since `_bootstrap_legacy_project` writes a real `github:cyberfabric/cyber-pilot-kit-sdlc` kit source, these tests will still make real calls to `api.github.com`/`codeload.github.com`, reintroducing exactly the flakiness this PR is fixing. ([comment](https://github.com/constructorfabric/studio/pull/218#issuecomment-5695797536))

tests/test_migrate_from_cypilot.py
vasylcf: Valid, and still unimplemented -- I checked rather than assuming it had been picked up elsewhere.

`tests/conftest.py` on `main` today has autouse fixtures for JSON mode, the studio logger and the decision log, and `grep -n "socket\|urlopen" tests/conftest.py` returns nothing. There is no systemic guard. #221 ("two more modules call the real GitHub API") was closed as completed on 2026-09-16, but it closed by fixing the two call sites it named, not by adding the fixture that would stop the third.

That is the argument for the finding, not against it: the bug class has now recurred twice and been fixed twice by hand, which is the pattern a guard exists to break.

The measurement from this PR is the case for it being cheap, too -- stubbing the follow-up took the migrate tests from 48 outbound TCP connections and 20.98s to 0 and 3.73s. A socket guard would have surfaced all 48 as a loud failure on the first run instead of as CI flakiness under load.

Not adding it in this PR: it is superseded by #254, which cherry-picks this exact commit onto `main` (`git patch-id` on `1a3b7ce7` and `412cb1fb` is identical -- `dfabf09c`). The guard is a change to `conftest.py` that touches every test in the suite and deserves its own PR and its own CI run, not a rider on a port. I will file it so it does not close with this one.

ainetx: @vasylcf Re-verified against the current code (1a3b7ce78c86e6b6b0c2a8736218b1d88e5b6321) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** tests/conftest.py has no socket or HTTP-blocking autouse fixture, while this diff only adds per-test monkeypatches, so a future test can still accidentally reach GitHub through the migration path.
- **Verifier:** Confirmed by direct read of tests/conftest.py: only autouse fixtures are _enable_json_mode, _restore_studio_logger, and _isolate_decision_log -- none blocks sockets/urlopen/HTTP. A repo-wide grep for socket/urlopen/HTTP-mocking libs in tests/ turns up nothing relevant to a systemic blocker. This diff itself demonstrates the recurring bug class it's meant to guard against: it has to manually stub _run_followup_kit_update in two separate places within the same file (the followup_update_ok fixture at line 161, and again standalone at line 657 in test_internal_migration_preserves_partial_success_when_followup_update_fails, per its own comment noting 'this is the same omission as the fixture above'). Combined with the sibling gap already found in test_cli_setup_e2e.py (per ) and issue #214 being a repeat occurrence, this is a genuine, non-hypothetical systemic gap: nothing stops a future test from silently reintroducing live GitHub calls. The posted this finding and its discussion thread (human confirmed it's valid and unimplemented, filing a separate follow-up issue rather than dismissing it) corroborate this.

</details>

vasylcf:
Closing as superseded by #254, which cherry-picks this exact commit onto `main`.

Not "an equivalent change" -- the same patch. `git patch-id --stable` on `1a3b7ce7` (this branch) and on `412cb1fb` (#254) both give `dfabf09c`.

**Why this PR went stale.** The commit reached `release/v1.7.0` directly and never reached `main`, so this PR stayed open against a branch that already had it. #254 ports it; #257 records the process fix so the next release branch is cut from `main` and merged back after tagging.

**The one review finding stays open, and is unimplemented.** @ainetx asked for a systemic network-blocking fixture rather than per-call-site stubbing, on the grounds that the bug class had already recurred. I checked rather than assumed: `tests/conftest.py` on `main` today has no socket guard, and #221 -- which looked like it might have covered this -- closed by fixing the two call sites it named, not by adding the fixture that would stop the third.

So the argument is stronger now than when it was written: the class has been fixed by hand twice. The measurement from this PR is also the case for it being cheap -- stubbing the follow-up took these tests from 48 outbound TCP connections and 20.98s to 0 and 3.73s, and a socket guard would have surfaced all 48 as one loud failure on the first run instead of as CI flakiness under load.

It is not being folded into #254, which is a port and should stay one: a `conftest.py` fixture touches every test in the suite and deserves its own PR and its own CI run. Tracking it separately rather than letting it close here.

Thanks @ainetx -- the finding outlives the PR.

vasylcf:
Filed as #262, so the finding has somewhere to live now that this PR is closed.

PR #220, opened 2026-09-16
feat(spec-coverage): name the flagged requirements on screen, not just their count

### LINKED ISSUE #195: spec-coverage --semantic: human output reports files, not which requirements are covered/uncovered
## Context

`spec-coverage --semantic` prints per-file coverage percentages, where the number is *lines of code that carry markers*. The human output never names a **requirement**.

As a result, a reader of the console cannot see:

- which requirement maps to which code, or
- whether some requirement has **no code at all**.

The pairing does exist — `block_id`, `path` and `line` are all in the JSON report. It is just not present in what a person reads on screen.

## Ask

Either surface requirement-level coverage in the human-readable output — in particular the **uncovered** requirements — or document explicitly that the JSON report is the requirement-level surface and the console is a bulk health number.

## Why

The console is currently the first (and for most readers, only) thing seen. "Which requirement is uncovered?" is the question a reviewer actually has, and today it is answerable only by reading the JSON. Surfaced during review of the advisory semantic-coverage pass.

SanjeevSolanki:
## What

`spec-coverage --semantic` prints a one-line advisory summary that *counts* how many marked
blocks came back weak/wrong, but to learn *which* ones a reader has to open `--json`. This names
them: beneath the summary line, each finding whose verdict is `wrong` or `partial` is listed by
`block_id` and `path:line`, so the result is actionable on screen.

Resolves #195.

## Why these are (and aren't) listed

- **Weak/wrong** (`wrong`/`partial`) — listed. These are the actionable findings, the exact set
  the summary line already counts as "weak/wrong".
- **Presumed-covered** — not listed. The passing majority; their count stays on the summary line.
- **Unjudgeable** — deliberately not listed. With no judge wired every block is unjudgeable, so
  per-block rows would be pure noise. Their count stays on the summary line.

The list is capped at 20 rows with a `+N more — see --json` continuation, and a malformed or
absent shape yields no lines rather than raising.

## Scope

Advisory and rendering-only — the new output never touches the coverage status or exit code, exactly
like the summary line above it. This extends the existing one-line summary; it does not change what
`--semantic` assesses or how the structural gate is computed.

## Testing

- `spec-coverage --system studio --min-coverage 90 --min-file-coverage 60 --min-granularity 0.45` —
  all thresholds met (granularity 0.4629; the two changed source files at 98.95% / 88.25%).
- `validate` — all checks pass.
- Full test suite green; new tests cover: weak/wrong named while covered + unjudgeable are omitted;
  the cap and its `+N more` row; and malformed/empty/non-dict inputs returning no lines. A CLI-level
  test asserts the flagged location reaches human output (not just the summary count).
- pylint and vulture clean on the changed files.

## Summary by CodeRabbit

* **New Features**
  * Human-readable semantic coverage reports now list weak or incorrect requirements with their IDs and source locations.
  * Incorrect findings appear before partial findings for easier review.
  * Reports show up to 20 flagged requirements, followed by a continuation notice when additional findings exist.
  * Presumed-covered and unjudgeable findings remain excluded from the advisory list.


skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -155,3 +155,34 @@ def summary_line(semantic: Dict[str, object]) -> str:
...
+        return []
+    findings = semantic.get("findings")
+    findings = findings if isinstance(findings, list) else []
+    out = [f"  {f['block_id']}  {f['verdict']}  {f['path']}:{f['start_line']}"
+           for f in findings if isinstance(f, dict)
+           and f.get("verdict") in (eval_semantic.SEM_WRONG, eval_semantic.SEM_PARTIAL)
+           and isinstance(f.get("block_id"), str) and f["block_id"]
+           and isinstance(f.get("path"), str) and f["path"]
+           and isinstance(f.get("start_line"), int)]
+    if len(out) > _SEMANTIC_CAP:
```
ainetx, line 191: ### flagged_lines() truncates the human report by pre-filter rank order, not verdict severity, so a 'wrong' finding can be hidden behind a shown 'partial'

**Severity:** Minor

**Problem**
flagged_lines() (semantic_coverage.py) builds its output list by filtering semantic['findings'] and truncating at _SEMANTIC_CAP without sorting by verdict. The 'findings' order originates from eval_semantic.rank_pairings(), which orders weak links only by ascending token-overlap score (a judge-budget heuristic), not by the eventual judge verdict. Overlap score and verdict severity are independent (judge_fn is a pluggable seam), so this ordering is not a proxy for severity.

<details>

**How to reproduce**
1. Produce >20 weak-link findings via eval_semantic.assess() with a judge_fn. 2. Arrange for a 'wrong' verdict to occur on a pairing with higher overlap score than a 'partial' verdict on another pairing. 3. Call flagged_lines(semantic) on the resulting report. 4. Observe the 'partial' appears in the first 20 shown rows while the 'wrong' is folded into the '+N more — see --json' line.

**Expected behavior**
Since the human report's stated purpose (spec-coverage.md step 5) is to let a reader act without opening --json, the more severe verdict class ('wrong') should never be de-prioritized behind a less severe one ('partial') by an arbitrary pre-filter ordering.

**Actual behavior**
The cap is applied directly to the incoming findings order (ranked weakest-overlap-first), so severity and display priority are decoupled — a 'wrong' can be truncated while a 'partial' is shown.

```text
rank_pairings (sort by ascending overlap, NOT verdict) -> findings list -> flagged_lines() cap at 20 (preserves order) -> wrong finding pushed past cap, partial shown first
```

**Impact**
On a report with more than 20 weak/wrong findings, a human relying on the console report to prioritize fixes could address a less urgent 'partial' shown above the fold while missing a more severe 'wrong' silently folded into '+N more'.

**Suggested correction**
Sort by verdict severity (wrong before partial) before applying _SEMANTIC_CAP in flagged_lines(), or at minimum document that ordering is overlap-based and intentional, then add a test with a mixed wrong/partial set at the cap boundary.

**How to verify**
Add a test with >20 findings mixing wrong/partial in overlap-ascending order such that a wrong finding would sort after the cap under naive order-preservation, and assert it is never dropped ahead of a shown partial.

</details>

tests/test_semantic_coverage.py
```diff
@@ -364,3 +364,49 @@ def test_relative_posix_still_relativises_against_a_root(tmp_path: Path):
...
+def test_flagged_lines_tolerates_malformed_and_caps():
+    assert sc.flagged_lines(None) == []                              # absent/malformed -> no lines
+    assert sc.flagged_lines({"error": "boom"}) == []                 # errored pass -> no lines
+    assert sc.flagged_lines({"findings": "nope"}) == []              # non-list findings -> no raise
+    assert sc.flagged_lines({"findings": []}) == []                  # empty findings -> no lines
+    assert sc.flagged_lines({"findings": [None, 7]}) == []           # non-dict rows skipped, no raise
+    # a wrong/partial record missing (or null-ing) a required field is skipped, not rendered as
+    # "None:None" — so the docstring's "malformed shape yields no lines" holds per-record too.
+    assert sc.flagged_lines({"findings": [{"verdict": sem.SEM_WRONG}]}) == []            # all missing
+    assert sc.flagged_lines({"findings": [{"verdict": sem.SEM_WRONG, "path": "a.py",
```
ainetx, line 402: ### No test pins flagged_lines() behavior for a finding dict missing the 'verdict' key entirely

**Severity:** Minor

**Problem**
flagged_lines()'s filter comprehension uses f.get('verdict') in (eval_semantic.SEM_WRONG, eval_semantic.SEM_PARTIAL) to select rows. When a dict has block_id/path/start_line all present but no 'verdict' key at all, this silently excludes the row (None not in the tuple) rather than raising — but this exact shape is absent from test_flagged_lines_tolerates_malformed_and_caps, which only covers None, {'error':...}, non-list findings, non-dict rows ([None, 7]), and dicts with verdict present but another field missing.

<details>

**How to reproduce**
1. Call sc.flagged_lines({'findings': [{'block_id': 'b:1', 'path': 'a.py', 'start_line': 10}]}) (no 'verdict' key at all). 2. Observe it returns [] today via the.get() None-not-in-tuple path — but no assertion in the test suite currently exercises this exact input.

**Expected behavior**
The docstring's 'malformed/absent shape yields no lines, never raises' invariant should be pinned by a test for a dict missing the 'verdict' key entirely, since it is a distinct branch from the already-tested missing-block_id/path/start_line cases.

**Actual behavior**
The test suite has no case for this specific shape, so a future refactor (e.g. switching '.get("verdict")' to bracket-indexing 'f["verdict"]' under an assumption the field is 'always present') could silently start raising KeyError on this shape without any existing test catching the regression.

```text
dict missing 'verdict' key -> f.get('verdict') -> None -> None in (WRONG, PARTIAL) -> False -> row excluded (untested branch)
```

**Impact**
A future refactor tightening verdict access could silently break the 'never raises' guarantee on this shape, undetected by CI until it manifests in production on real judge output missing the field.

**Suggested correction**
Add an assertion in test_flagged_lines_tolerates_malformed_and_caps for sc.flagged_lines({'findings': [{'block_id': 'b:1', 'path': 'a.py', 'start_line': 10}]}) == [] (verdict key absent entirely, all other fields well-formed).

**How to verify**
Add the test case and confirm it passes against the current implementation; it should also fail if a future refactor replaces.get('verdict') with direct indexing.

</details>

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -155,3 +155,34 @@ def summary_line(semantic: Dict[str, object]) -> str:
...
+    if not isinstance(semantic, dict) or semantic.get("error"):
+        return []
+    findings = semantic.get("findings")
+    findings = findings if isinstance(findings, list) else []
+    out = [f"  {f['block_id']}  {f['verdict']}  {f['path']}:{f['start_line']}"
+           for f in findings if isinstance(f, dict)
+           and f.get("verdict") in (eval_semantic.SEM_WRONG, eval_semantic.SEM_PARTIAL)
+           and isinstance(f.get("block_id"), str) and f["block_id"]
+           and isinstance(f.get("path"), str) and f["path"]
+           and isinstance(f.get("start_line"), int)]
```
ainetx: ### flagged_lines' start_line type guard silently admits booleans due to Python's bool/int subtyping

**Severity:** Minor

**Problem**
The row-validity filter in flagged_lines uses `isinstance(f.get('start_line'), int)` to decide whether a finding row is well-formed enough to render. Because `bool` is a subclass of `int` in Python, `isinstance(True, int)` and `isinstance(False, int)` both evaluate to True, so a finding dict with `start_line=True` or `start_line=False` passes the guard alongside valid int values.

<details>

**How to reproduce**
1. Call `flagged_lines({'findings': [{'block_id': 'b:1', 'path': 'a.py', 'start_line': True, 'verdict': eval_semantic.SEM_WRONG}]})`.
2. The list comprehension's `isinstance(f.get('start_line'), int)` check passes since `isinstance(True, int) == True`.
3. The row is formatted via `f" {f['block_id']} {f['verdict']} {f['path']}:{f['start_line']}"`.

**Expected behavior**
Per the function's docstring, a malformed record (including a wrong-typed field) should be 'skipped rather than rendered' -- consistent with how the code already skips records with null/missing block_id, path, or start_line.

**Actual behavior**
The record is rendered verbatim as `b:1 wrong a.py:True`, corrupting the printed location instead of being dropped like other malformed-field cases the same guard style is meant to catch.

```text
finding.start_line = True --> isinstance(True, int) == True (bool subclasses int) --> guard passes --> f"{path}:{start_line}" --> "a.py:True" printed to human report
```

**Impact**
Low today because the only current producer (run_semantic_pass) always emits genuine int line numbers, so no real payload triggers this. But the guard does not deliver the robustness its docstring and sibling tests explicitly claim for every other malformed-field case, and would silently render a nonsensical location (`path:True`/`path:False`) if any future or manually-constructed semantic payload supplied a boolean start_line.

**Suggested correction**
Exclude bool explicitly, e.g. change the guard to `isinstance(f.get('start_line'), int) and not isinstance(f.get('start_line'), bool)`.

**How to verify**
Add a test asserting `sc.flagged_lines({'findings': [{'block_id': 'b:1', 'path': 'a.py', 'start_line': True, 'verdict': sem.SEM_WRONG}]}) == []`, confirming the row is skipped rather than rendered as `a.py:True`.

</details>

skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -155,3 +155,34 @@ def summary_line(semantic: Dict[str, object]) -> str:
...
+#: Cap on named requirement rows in the human report; the rest are summarised as "+N more".
+_SEMANTIC_CAP = 20
+
+
+# @cpt-begin:cpt-studio-algo-semantic-coverage-pass:p1:inst-scov-flagged
+def flagged_lines(semantic: Dict[str, object]) -> List[str]:
+    """The weak/wrong requirements, named one per line for the human report, so a reader can act on
+    the result without opening ``--json``: a finding whose ``verdict`` is ``wrong`` or ``partial``
+    (the set ``summary_line`` counts as "weak/wrong") named by ``block_id`` + ``path:line``.
+    Presumed-covered blocks are the passing majority and are not listed; *unjudgeable* blocks are
```
ainetx, line 171: ### flagged_lines()/spec-coverage.md rationale for omitting all 'unjudgeable' rows only explains judge-wiring noise, not the permanent pre-filter gaps also included in that list

**Severity:** Minor

**Problem**
eval_semantic.assess() populates SemanticReport.unjudgeable from two independent sources: pre-filter gaps in _rank_one/rank_pairings (no retrievable requirement for the block's id, or too little text on either side to compare — unrelated to whether judge_fn is supplied) and judge-stage gaps in _judge_one (no judge_fn wired, or judge_fn raised/returned a malformed verdict). The new flagged_lines() docstring and architecture/features/spec-coverage.md:187 justify blanket-omitting unjudgeable rows from the human report with the single claim 'with no judge wired every block is unjudgeable' — which is only true for the judge-stage source.

<details>

**How to reproduce**
1. Wire a real judge_fn. 2. Include a pairing whose block_id cannot be resolved to a requirement (e.g. a stale/typo'd algo id). 3. Run assess(); this pairing's Ranked.score is None (never even reaches the judge, since weak_link stays False), so it lands in unjudgeable with reason 'no retrievable requirement for this id'. 4. Run flagged_lines() on the resulting semantic dict / view the human report: this entry is never named, indistinguishable from ordinary judge-wiring noise, even though a real judge is wired and this is a genuinely actionable data problem.

**Expected behavior**
The design rationale documented in code and in spec-coverage.md should distinguish 'no judge wired yet' (transient, until wiring happens) from 'this block's requirement id could not be resolved' (a persistent, actionable data gap unrelated to judge wiring), so future maintainers don't wrongly assume all unjudgeable rows become moot once a judge is wired.

**Actual behavior**
Both sources are conflated under a single justification that only accurately describes the judge-wiring case, so a maintainer reading the docstring/spec line would incorrectly conclude that wiring a real judge makes all unjudgeable rows non-issues, when pre-filter-stage gaps remain hidden from the human report indefinitely regardless of judge wiring.

```text
rank_pairings: score=None (no requirement / below floor) --> unjudgeable (source A, judge-wiring-independent)
judge_weak_links: no judge_fn / crash / malformed --> unjudgeable (source B, judge-wiring-dependent)
flagged_lines(): omits ALL unjudgeable rows, docstring cites only source B's cause
```

**Impact**
A maintainer relying on the docstring to reason about future behavior (e.g. once a real judge is wired) could wrongly assume unjudgeable rows become noise-free/actionable automatically and never build a way to surface persistent pre-filter-stage gaps (unresolved requirement ids) in the human-readable report.

**Suggested correction**
Correct the docstring/spec wording to name both sources of 'unjudgeable' explicitly, and/or acknowledge that pre-filter gaps remain permanently omitted from the human report by design (not just 'noise until a judge is wired'); consider a follow-up to surface 'no retrievable requirement' entries distinctly once actionable.

**How to verify**
Re-read the corrected docstring/spec text and confirm it no longer implies wiring a judge eliminates all unjudgeable-row noise; optionally add a test asserting a score-is-None unjudgeable entry stays unjudgeable even when a stub judge_fn is supplied.

</details>

ainetx (approved):
Approved — the review is otherwise clean, and these are the only open non-blocking notes worth a follow-up glance:

- **Prioritize more severe findings in capped output.** The human report currently caps findings in pre-filter rank order, so a `wrong` result may be omitted while a `partial` result is shown. [Discussion](https://github.com/constructorfabric/studio/pull/220#discussion_r4025410674)
- **Clarify why unjudgeable rows are omitted.** The documentation attributes all unjudgeable rows to missing judge wiring, but that category also includes permanent pre-filter gaps. [Discussion](https://github.com/constructorfabric/studio/pull/220#discussion_r4025599263)
- **Add coverage for a missing verdict key.** The malformed-input tests do not cover an otherwise complete finding dictionary that lacks `verdict`. [Discussion](https://github.com/constructorfabric/studio/pull/220#discussion_r4025411053)
- **Reject boolean line numbers.** The current integer guard accepts `True` and `False` as `start_line` values because booleans subclass `int` in Python. [Discussion](https://github.com/constructorfabric/studio/pull/220#discussion_r4025411238)

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0ab11a9fd37df98b473fcad114d95b890a1eda17 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** `flagged_lines` now stably sorts valid findings with `wrong` before `partial` before applying `_SEMANTIC_CAP` (lines 188-192), and `test_flagged_lines_prioritises_wrong_over_partial_when_capped` explicitly verifies wrong findings survive after 20 input-first partials.
- **Reviewer:** Current code already fixes this: flagged_lines() sorts `valid` by verdict (wrong before partial, stable) at semantic_coverage.py:189 BEFORE truncating at _SEMANTIC_CAP, so severity — not the upstream overlap-ranked findings order — determines what survives the cap; test_flagged_lines_prioritises_wrong_over_partial_when_capped (tests/test_semantic_coverage.py:428-441) directly exercises a mixed set with 20 partials-then-3-wrongs straddling the cap and asserts all 3 wrongs survive and appear before any shown partial, exactly the scenario this check worried about.

</details>

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0ab11a9fd37df98b473fcad114d95b890a1eda17 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The updated implementation and feature acceptance criterion explicitly distinguish no-judge noise from permanent pre-filter gaps and intentionally state that both unjudgeable categories are omitted as non-actionable while their aggregate count remains in the summary.
- **Reviewer:** This diff's docstring and spec-coverage.md line now explicitly distinguish the two unjudgeable sources: 'unjudgeable... is dominated by no-judge-wired noise (with no judge wired every block is unjudgeable) but also holds permanent pre-filter gaps; neither is an actionable requirement here' (semantic_coverage.py:171-174; spec-coverage.md step 5) — this is precisely the acknowledgement the check's 'Expected evidence' asked for, correcting the prior round's single-cause framing by naming both the judge-stage cause and the permanent pre-filter cause and stating the design choice (omit both from the human report) is intentional for both.

</details>

tests/test_semantic_coverage.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0ab11a9fd37df98b473fcad114d95b890a1eda17 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current malformed-input test explicitly supplies an otherwise complete finding dict without `verdict` and asserts `flagged_lines()` returns no rows (tests/test_semantic_coverage.py:408-410), pinning the `.get("verdict") -> None` exclusion path.
- **Reviewer:** Read tests/test_semantic_coverage.py lines 408-410 directly: `assert sc.flagged_lines({"findings": [{"block_id": "b:1", "path": "a.py", "start_line": 10}]}) == [] # verdict absent` supplies exactly a well-formed dict (block_id/path/start_line all present and valid) with the 'verdict' key entirely absent, with a comment explicitly calling out 'an otherwise-complete record with NO verdict key at all is skipped, not raised on'. This pins the `f.get('verdict') in (...)` -> None branch that prior review rounds claimed was untested; that claim does not hold against the current file content.

</details>

skills/studio/scripts/studio/utils/semantic_coverage.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0ab11a9fd37df98b473fcad114d95b890a1eda17 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current predicate explicitly excludes booleans after its integer check (`and not isinstance(f.get("start_line"), bool)`) and the new test supplies `start_line=True` and asserts it is skipped (tests/test_semantic_coverage.py:411-413).
- **Reviewer:** Read the current guard at semantic_coverage.py line 187: `isinstance(f.get('start_line'), int) and not isinstance(f.get('start_line'), bool)` — the bool-exclusion clause is already present in this exact diff (it appears in the added lines shown), directly contradicting the check's premise that the guard is 'isinstance(...,int) alone (without excluding bool)'. test_flagged_lines_tolerates_malformed_and_caps lines 411-413 explicitly asserts a `start_line: True` record is skipped, confirming the fix is both present and tested; the described defect does not exist in the current code.

</details>

PR #222, opened 2026-09-16
test(kit,setup): stop the last two modules calling the real GitHub API

### LINKED ISSUE #221: Tests: two more modules call the real GitHub API, so the suite cannot run offline
Follow-up to #214, which fixed one module. The same dependency is present in two others, and this issue exists so the rest is tracked rather than assumed done.

## What was measured

`socket.connect` wrapped for the whole suite, recording every destination. Run on `main` @ `916dee73`, before #218 lands:

```
total: 85 connections
  69 → 20.205.243.168   api.github.com
  15 → 20.205.243.165   codeload.github.com
   1 → /tmp/.../overwork.sock   (local unix socket, not a concern)
```

Per module:

| module | connections | fixed by | remaining |
|---|---|---|---|
| `tests/test_migrate_from_cypilot.py` | 28+ | **#218** | — |
| `tests/test_cli_setup_e2e.py` | 12 | — | **yes** |
| `tests/test_kit.py` | 8–22 (varies per run) | — | **yes** |

No other module in the 157 reaches the network.

## `tests/test_cli_setup_e2e.py` — same cause as #214, same fix

Three tests, all in `TestValidateKitsAndInitE2E`:

- `test_init_from_dir_migrates_explicit_legacy_directory`
- `test_init_migrate_yes_migrates_legacy_project_without_prompt`
- `test_init_update_legacy_studio_yes_updates_baseline_then_migrates`

Each patches the first follow-up call and leaves the second live:

```python
with patch("studio.commands.migrate_from_cypilot._run_followup_update", return_value=(0, {"status": "PASS"})):
    ...  # _run_followup_kit_update not patched → cmd_kit_update → api.github.com
```

Identical to the fixture bug #214 describes. The fix is the same shape: patch `_run_followup_kit_update` alongside it, in all three places.

These are also the tests that were failing in the same flake family — `test_init_update_legacy_studio_yes_updates_baseline_then_migrates` appeared in the original CI failure list, so #218 alone does not make that family deterministic.

## `tests/test_kit.py` — different cause, different fix

Four tests:

- `TestPartialGithubSourceFailures::test_all_kits_fail_returns_structured_errors`
- `TestPartialGithubSourceFailures::test_one_good_one_bad_github_kit`
- `TestCmdKitUpdateCli::test_update_github_download_failure`
- `TestCmdKitInstallGithubPath::test_download_failure`

These are *failure-path* tests, and they obtain the failure from the real GitHub rather than from a stub — a request to a nonexistent repository returning 404 is what makes the assertion pass.

That works until it doesn't, and it is fragile in both directions:

- **Rate limiting**: 60 requests/hour unauthenticated, per IP, shared on a CI runner. A 403 is not a 404; whether the assertions still hold depends on how the error is classified.
- **A false green**: if GitHub returned something unexpected, these tests could pass for the wrong reason — they assert a failure occurred, not *which* failure.

The fix here is to stub the HTTP layer (`urllib.request.urlopen` in `commands/kit.py`, or the helper above it) and assert against a controlled error, so the test states which failure it is exercising.

## Why it is worth doing now

The suite currently cannot run offline — on a plane, in a locked-down container, or on a runner without egress, these fifteen tests fail for reasons unrelated to the code.

More practically: since #211 the pipeline runs on `release/**`, so the remaining flake surface now lands during release acceptance, where a red check is most expensive to interpret and "just re-run it" is the worst habit to teach.

## Scope

- [ ] `tests/test_cli_setup_e2e.py`: patch `_run_followup_kit_update` in the three migration tests
- [ ] `tests/test_kit.py`: stub the HTTP layer in the four failure-path tests, asserting a controlled error
- [ ] Re-measure: the suite should report **0** remote connections
- [ ] Then, and only then, add a suite-wide guard that fails loudly on any outbound connection — turning it on before the above is fixed converts a rare flake into a guaranteed failure

## Method, for whoever picks this up

```python
# tests/conftest.py — temporary, not for commit
import socket as _sock
_NET_LOG = []

@pytest.fixture(autouse=True)
def _record_network(monkeypatch, request):
    real = _sock.socket.connect
    def _spy(self, address, *a, **k):
        _NET_LOG.append((request.node.nodeid, address))
        return real(self, address, *a, **k)
    monkeypatch.setattr(_sock.socket, "connect", _spy)
    yield

def pytest_sessionfinish(session, exitstatus):
    import collections
    print("\ntotal:", len(_NET_LOG), collections.Counter(n for n, _ in _NET_LOG).most_common(20))
```

Run it plainly (`pytest tests -q > out.txt`). Piping through `sed`/`head` closes the pipe early and makes the run appear to hang — that artefact produced a bogus "51 minutes" measurement in an earlier comment on #214; instrumented, the suite takes its usual ~8 minutes.

## Related

- #214 — the original flake, and the diagnosis
- #218 — the fix for `test_migrate_from_cypilot.py`
- #211 — why this now surfaces on release branches

vasylcf:
Closes #221. Completes what #218 started: with both merged, the suite makes **zero remote connections** and can run offline.

## Measured

`socket.connect` wrapped for the whole suite, before and after:

| module | before | after |
|---|---|---|
| `tests/test_cli_setup_e2e.py` | 12 | **0** |
| `tests/test_kit.py` | 8–22 (varies) | **0** |

### `test_cli_setup_e2e.py`

Three migration tests patched `_run_followup_update` and left `_run_followup_kit_update` live — exactly the omission #214 describes in the shared fixture. `cmd_kit_update` then queried `api.github.com`, and a throttled request turned the migration result into `WARN`.

One of them, `test_init_update_legacy_studio_yes_updates_baseline_then_migrates`, appeared in the original CI failure list on #214. **#218 alone did not make that family deterministic** — this PR is the other half.

### `test_kit.py`

Four tests patched `_download_kit_from_github`. But `cmd_kit_update` and `cmd_kit_install` both call `_download_kit_from_github_with_authority`, which resolves the ref over the network *before* delegating to the patched function:

```
cmd_kit_update
└─ _download_kit_from_github_with_authority
   ├─ _resolve_github_ref          ← network, raises first
   └─ _download_kit_from_github    ← the patched one, never reached
```

The failure path then calls `_resolve_github_ref` again to enrich the error record — caught and degraded, but a second live request. Both are stubbed now.

### A correction to the issue

#221 said these tests "obtain the failure from the real GitHub rather than from a stub". That was wrong. They *do* stub the failure; the network call is an unstubbed **prerequisite** step. It is the same defect as the migration one, not a different kind — which also means the fix is the same shape rather than a rewrite.

## A test that was not testing its premise

`test_one_good_one_bad_github_kit` — *"One kit downloads OK, one fails → partial failure in results"* — asserted `rc == 2`, `status == FAIL`, both slugs present, and that badkit failed. **It never asserted that goodkit succeeded.**

It could not have. The unpatched ref resolve raised for *both* kits, so both failed and the test passed regardless. Observed directly: the captured output carried `GitHub API error (403): rate limit exceeded` for both kits while the test was green. The suite was already rate-limited and nothing said so.

With the stub at the right seam the scenario is exercised for the first time, and the missing half is pinned:

```python
good_r = next(r for r in out["results"] if r["kit"] == "goodkit")
self.assertNotEqual(good_r["action"], "failed", good_r)
```

Verified to have teeth: forcing the mock to fail both kits makes the test fail.

## Verification

- Both modules: 0 remote connections, measured.
- `tests/test_kit.py` 280 passed, `tests/test_cli_setup_e2e.py` 24 passed.
- Full suite: **6166 passed, 0 failed**.
- No production code touched — two test files only.

## What remains on #221

The suite-wide guard that fails loudly on any outbound connection. It is now unblocked: with this and #218 merged there is nothing left for it to break, and it is the only thing that stops the next such dependency arriving unnoticed.

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for legacy project migrations, including scenarios involving follow-up updates.
  * Strengthened validation of GitHub-based kit installation and updates when downloads or source resolution fail.
  * Added checks to ensure successful kits remain successful when another kit in the same operation encounters an error.
  * Reduced test flakiness caused by unintended external network requests.


ainetx (approved):
Reviewed the test-isolation updates around legacy migration and GitHub kit installation/update flows, including the corrected mocking seams and strengthened partial-failure assertion. The changes make the affected tests deterministic without changing production behavior. No blocking correctness or test-coverage issues found; approving.

PR #230, opened 2026-09-17
fix(agents): thread known tool identity through the pure-stub check (#185)

### LINKED ISSUE #185: Native-dialog binding detection can misclassify a hand-edited non-Claude file as pure generated

tkcoding:
## Summary

Closes #185.

`_pure_generated_stub_matches()` decided whether a generated file was a pure,
untouched Studio stub (safe to silently delete/overwrite during cleanup or
regeneration) by trying **every** known `_ASK_TOOL_BINDING` value and
accepting a match against any of them — because no caller told it which
tool/bucket actually produced the file being checked.

**Failure scenario (from the issue):** a user hand-edits a non-Claude
generated file (e.g. under `.cursor/agents/...` or the shared
`.agents/skills/` bucket) and, via a bad merge or copy-paste, it ends up
containing the literal line `- ask_tool_name = "AskUserQuestion"`. The
brute-force check would match Claude's candidate and report the file as a
pristine stub, gating silent deletion/overwrite in
`_cleanup_claude_workflow_command`, `_cleanup_legacy_skill_files`,
`_cleanup_legacy_manifest_skill_outputs`, and ownership classification in
`_classify_generated_output_owner`.

## Change

- `_pure_generated_stub_matches()` and `_is_pure_studio_generated()` gain an
  optional `tool: Optional[str] = None` parameter.
- When `tool` is known, the candidate set narrows to `[None, _ASK_TOOL_BINDING.get(tool)]`
  — that tool's own current binding, plus the universal unset/no-binding
  shape (which carries no tool-specific signal, so it's safe to accept for
  any tool — this is what keeps a tool's own pre-binding legacy files
  recognized once that tool later gains a real binding). A *different*
  tool's actual bound value is no longer accepted.
- The 6 real callers that already know their own tool identity one frame up
  now pass it through: `_cleanup_studio_legacy_subagents` (agent param),
  `_cleanup_legacy_skill_dirs` (agent param), `_scan_owned_generated_outputs`
  (provider loop var), `_cleanup_claude_workflow_command` (hardcoded
  `"claude"` — only ever touches `.claude/commands/`),
  `_cleanup_legacy_skill_files` (agent param), and
  `_cleanup_legacy_manifest_skill_outputs` (target param).

## Known, explicitly out-of-scope limitation

The shared `.agents/skills/` scan bucket inside `_scan_owned_generated_outputs`
is intentionally passed `tool=None` and still falls back to the full
brute-force loop: a pre-existing on-disk file there could legitimately belong
to any of windsurf/cursor/copilot/openai/OpenCode, since that path isn't
split per-tool (only Cursor's dedicated `.cursor/commands/cf.md` was split out
so far, in #208). Closing this fully requires further per-tool output-path
separation, which is a larger, separate effort — not a regression introduced
by this PR, and the brute-force fallback here is no worse than the
pre-#185 behavior for that one bucket.

## Testing

- `tests/test_subagent_registration.py`: leaf-level regression test for the
  rejected cross-tool collision (`test_pure_generated_stub_matches_rejects_a_foreign_tools_binding`)
  and the future-binding edge case
  (`test_pure_generated_stub_matches_keeps_recognizing_pre_binding_files_after_a_new_binding_lands`).
- `tests/test_agents_coverage.py`: an end-to-end regression test
  (`test_a_foreign_tools_bound_stub_is_not_deleted_as_claudes_own`) proving a
  real call site (`_cleanup_studio_legacy_subagents`) actually forwards its
  own tool identity through to the check, not just the unit-tested helper —
  the existing `cypilot-`/`cf-constructor-` prefixed fixtures in that file
  route through a different code path (`_is_legacy_generator_stub`) that
  never looks at `tool` at all, so they couldn't have caught a regression
  here.
- Full suite: 6213 passed, 6 failed (all pre-existing/environmental —
  verified via `git stash` that each also fails on `main` without this
  branch's changes; 5 are known macOS filesystem-encoding edge cases, 1 is
  an unrelated subprocess/pytest-availability quirk in
  `test_tomllib_compat.py`), 5 skipped, 15 xfailed.
- `pylint` clean on all 3 changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of generated legacy skill and subagent files by ensuring tool-specific bindings are matched only to the appropriate tool.
  - Prevented files associated with another tool from being incorrectly classified as generated or marked for deletion.
  - Preserved compatibility for files created before tool-specific bindings were introduced.
  - Improved ownership detection for generated outputs, including shared and tool-specific locations.

- **Tests**
  - Added regression coverage for tool-specific cleanup, generated-file ownership, and legacy classification.


skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5080,7 +5118,14 @@ def _scan_owned_generated_outputs(project_root: Path) -> List[ManagedOutput]:
...
                 # @cpt-begin:cpt-studio-algo-project-extensibility-generate-agents:p1:inst-track-agent-results
                 rel = _safe_relpath(path, project_root)
-                owner_kind = _classify_generated_output_owner(rel, content)
+                # "studio" is the shared .agents/skills/ scan root (issue
+                # #185) -- a file found there could legitimately belong to
+                # any of several tools, so tool identity genuinely isn't
+                # resolvable here; every other provider key names exactly
+                # one tool.
+                owner_kind = _classify_generated_output_owner(
+                    rel, content, tool=None if provider == "studio" else provider,
```
ainetx, line 5187: ### Shared.agents/skills/ scan root stays on the pre-fix brute-force tool matching, reopening the exact cross-tool collision once a second sharer gets a binding

**Severity:** Minor

**Problem**
_scan_owned_generated_outputs passes tool=None for every file under the shared 'studio' provider root (.agents/skills/, agents.py:5121-5128). _agents_skill_outputs (agents.py:1798-2079) confirms windsurf, cursor, copilot, and openai all write identical content to this same directory. tool=None routes to the pre-#185 brute-force candidate loop in _pure_generated_stub_matches, which tries every known binding rather than scoping to one tool. This is safe only because none of those four tools currently has a non-null _ASK_TOOL_BINDING entry -- a state the _ASK_TOOL_BINDING docstring (agents.py:918-929) explicitly anticipates changing ('adding an entry for a currently-unbound tool needs no structural change').

<details>

**How to reproduce**
1. Today: none of windsurf/cursor/copilot/openai have a real _ASK_TOOL_BINDING entry, so all their.agents/skills/ files render identically as 'unset'.
2. In a future change, one of these four tools (e.g. cursor) gains a real _ASK_TOOL_BINDING entry.
3. A cursor-authored stub in.agents/skills/ now carries cursor's binding string; a copilot-authored stub (still unbound) in the same directory would, under the pre-#185 brute-force loop with tool=None, still match cursor's binding as a valid candidate.
4. The scanner can misclassify the copilot file's ownership/purity exactly as issue #185 originally described.

**Expected behavior**
The shared-root scan path should either carry a documented, accepted-risk note that this residual multi-tool collision is deferred until a second tool needs a binding, or use a disambiguation mechanism (e.g. per-tool subpaths or content markers) so tool=None isn't required for this directory.

**Actual behavior**
No comment, test, or ticket reference marks this specific residual gap; the fix silently reverts to its pre-fix vulnerable behavior for this one directory the moment the precondition (no second tool bound) stops holding.

```text
windsurf/cursor/copilot/openai --all write-->.agents/skills/ (provider='studio') --scanner--> tool=None --> brute-force loop over ALL _ASK_TOOL_BINDING values --> [today: harmless, all unbound] --> [future: 2nd tool bound] --> cross-tool misclassification re-emerges
```

**Impact**
A regression path with no test or comment guarding it: the #185 fix's coverage silently degrades back to its pre-fix vulnerable state for the one shared directory in the system as soon as a second sharer of that directory gains a binding.

**Suggested correction**
Add a code comment at the tool=None fallback explicitly flagging this residual risk and/or a regression test that fails once a second shared-root tool gains a real _ASK_TOOL_BINDING entry, so the gap is caught rather than silently reopened.

**How to verify**
Patch _ASK_TOOL_BINDING with a second shared-root tool (e.g. {'cursor': 'X'}) bound alongside claude, construct a copilot-authored (still-unbound).agents/skills/ stub, and confirm the scanner no longer misclassifies it as owned/pure under cursor's binding.

</details>

tests/test_subagent_registration.py
```diff
@@ -1220,6 +1220,88 @@ def test_pre_142_generated_files_still_recognized_as_pure_after_upgrade(self):
...
+
+        target = "{cf-studio-path}/.core/workflows/analyze.md"
+        pre_binding_windsurf_body = "\n".join(
+            line
+            for line in _follow_protocol_lines(target, required_bootstrap_path=_REQUIRED_BOOTSTRAP_PATH)
+            if line.strip()
+        )
+
+        with mock_patch.dict(_ASK_TOOL_BINDING, {"windsurf": "SomeFutureWindsurfTool"}):
+            self.assertTrue(_pure_generated_stub_matches(pre_binding_windsurf_body, tool="windsurf"))
```
ainetx, line 1452: ### No regression test combines a genuine pre-#142 stub body (context lines fully absent) with an explicit tool argument

**Severity:** Minor

**Problem**
The legacy_protocol branch (agents.py:384-400) runs unconditionally before the new tool-scoped candidate loop and never consults `tool` -- verified by reading the code. test_pre_142_generated_files_still_recognized_as_pure_after_upgrade (tests/test_subagent_registration.py:1198-1221) calls _pure_generated_stub_matches(pre_142_body) with no tool kwarg (defaults to None). The new tool-scoped tests (tests/test_subagent_registration.py:1223-1345) cover only the later 'ask_tool_name = unset' shape combined with tool=; none combines the earlier context-lines-fully-absent shape with a concrete non-None tool.

<details>

**How to reproduce**
1. Construct a body with both ask_tool_name/ask_tool_description lines fully absent (the genuine pre-#142 shape).
2. Call _pure_generated_stub_matches(body, tool='claude') or any other concrete tool.
3. No existing test asserts this returns True, even though reading the code shows it currently does (legacy_protocol branch runs before tool is consulted).

**Expected behavior**
A test should assert _pure_generated_stub_matches(pre_142_body, tool=<concrete tool>) remains True (and ideally stays True after that tool gains a binding), proving the legacy_protocol branch's tool-independence is an intentional, protected contract.

**Actual behavior**
The behavior is currently correct by code inspection but unverified by any test; a future refactor that scopes or moves the legacy_protocol branch inside the tool-aware candidate loop could silently orphan pre-#142 stubs for a specific tool without any test failure.

```text
pre-#142 body (no context lines) --> _pure_generated_stub_matches(body, tool=X) --> legacy_protocol check (tool-independent, unconditioned) --> True [correct today, but untested for tool != None]
```

**Impact**
A later refactor could regress historical-file recognition for a specific tool's legacy stubs, going undetected by the current test suite.

**Suggested correction**
Add a test asserting _pure_generated_stub_matches(pre_142_body, tool=<concrete tool, ideally after patching that tool's binding non-null>) still returns True, alongside the existing extra-content-rejection assertion.

**How to verify**
Run the new test before and after a hypothetical refactor that moves/scopes the legacy_protocol branch, confirming it catches the regression.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5080,7 +5118,14 @@ def _scan_owned_generated_outputs(project_root: Path) -> List[ManagedOutput]:
                     continue
                 # @cpt-begin:cpt-studio-algo-project-extensibility-generate-agents:p1:inst-track-agent-results
                 rel = _safe_relpath(path, project_root)
-                owner_kind = _classify_generated_output_owner(rel, content)
+                # "studio" is the shared .agents/skills/ scan root (issue
+                # #185) -- a file found there could legitimately belong to
+                # any of several tools, so tool identity genuinely isn't
+                # resolvable here; every other provider key names exactly
+                # one tool.
+                owner_kind = _classify_generated_output_owner(
```
ainetx, line 5186: ### No test exercises _scan_owned_generated_outputs / _classify_generated_output_owner / ManagedOutput at all

**Severity:** Major

**Problem**
Grepped the entire tests/ directory for `_scan_owned_generated_outputs`, `_classify_generated_output_owner`, `ManagedOutput`, and `_collect_scanned_managed_outputs` -- zero matches in tests/test_agents_coverage.py and tests/test_subagent_registration.py. This is the one production call site the PR's own comment (agents.py:5121-5128) explicitly justifies needing tool-scoping for, and it feeds `collect_managed_outputs`, used by downstream cleanup/status/uninstall reporting -- yet it ships with no direct unit test in either the resolvable-tool branch (non-studio provider root) or the unresolvable-tool branch (studio/.agents/skills root).

<details>

**How to reproduce**
1. Search tests/test_agents_coverage.py and tests/test_subagent_registration.py for any of: _scan_owned_generated_outputs, _classify_generated_output_owner, ManagedOutput, _collect_scanned_managed_outputs.
2. No results are found.

**Expected behavior**
A test constructing files under a dedicated provider root (e.g..cursor/commands/) and under the shared studio root (.agents/skills/) that verifies _scan_owned_generated_outputs/_classify_generated_output_owner classify ownership correctly, including the tool-scoping behavior this diff adds.

**Actual behavior**
Zero test coverage of this mechanism, including the pre-existing scan/classify logic that this diff modifies.

```text
PR fix at _scan_owned_generated_outputs (agents.py:5126-5128) --> feeds _classify_generated_output_owner --> feeds collect_managed_outputs --> no test anywhere exercises this chain
```

**Impact**
Regressions in the new provider-to-tool threading -- including the already-ineffective short-circuit bug identified separately -- ship without any test detection, and any future refactor here is unguarded.

**Suggested correction**
Add unit tests for _scan_owned_generated_outputs/_classify_generated_output_owner covering both a dedicated single-tool provider root (tool known) and the shared studio root (tool=None), including a foreign-tool-bound fixture.

**How to verify**
Confirm new tests fail on the pre-fix behavior (tool=None threading everywhere) and pass on the current diff; also confirm they would catch the short-circuit defect once fixed.

</details>

skills/studio/scripts/studio/commands/agents.py
```diff
@@ -5103,12 +5148,14 @@ def _read_generated_output_text(path: Path) -> Optional[str]:
...
-def _classify_generated_output_owner(rel: str, content: str) -> Optional[str]:
+def _classify_generated_output_owner(
+    rel: str, content: str, *, tool: Optional[str] = None,
+) -> Optional[str]:
     if rel.endswith(".toml"):
         return "agent" if _is_studio_managed_toml_output(content) else None
     if not (
         _extract_studio_owned_target(content)
-        or _is_pure_studio_generated(content)
+        or _is_pure_studio_generated(content, tool=tool)
```
ainetx, line 5252: ### _extract_studio_owned_target short-circuits before the new tool-scoped purity check ever runs in _classify_generated_output_owner

**Severity:** Major

**Problem**
_classify_generated_output_owner (agents.py:5151-5167) evaluates `_extract_studio_owned_target(content) or _is_pure_studio_generated(content, tool=tool)`. _extract_studio_owned_target/_extract_studio_control_target/_extract_studio_follow_target (agents.py:244-295) match purely on the 'ALWAYS open and follow' / 'LOAD... as controlling protocol' line via regex -- entirely independent of any ask_tool_name/context lines. Meanwhile _pure_generated_stub_matches (agents.py:381-383) itself requires that same follow/control target be extractable from the stripped content, or it returns False. So whenever the tool-scoped second disjunct could be True (multi-line case), the first disjunct is already True too, since both depend on the same follow-link being present -- Python short-circuits and the newly tool-scoped call never executes. The `tool=None if provider == 'studio' else provider` threading added at the _scan_owned_generated_outputs call site (agents.py:5126-5128), despite its detailed justifying comment, therefore has no observable effect on this function's outcome for the common case (an intact follow-link, which is the normal shape even for a foreign-tool-bound stub, since only the ask_tool_name lines would differ).

<details>

**How to reproduce**
1. Create a stub file under a dedicated single-tool scan root (e.g..cursor/commands/) with intact ALWAYS-open-and-follow line but Claude's ask_tool_name binding (AskUserQuestion) instead of cursor's own (unset) shape.
2. Run _scan_owned_generated_outputs / _classify_generated_output_owner('...', content, tool='cursor').
3. _extract_studio_owned_target(content) returns the follow target (non-None/truthy) because the follow-link line is untouched by the foreign binding.
4. The `or` short-circuits; _is_pure_studio_generated(content, tool='cursor') is never evaluated; the file is classified as owned regardless of the foreign binding.

**Expected behavior**
The scanner should classify a foreign-tool-bound stub as NOT owned by the wrong tool, consistent with the fix applied at every other call site in this diff.

**Actual behavior**
The file is classified as owned via the first disjunct, so the tool-scoping added by this diff at this call site is dead code for the scenario it targets.

```text
content w/ intact follow-link + foreign ask_tool_name --> _extract_studio_owned_target(content) = target (truthy) --> `or` short-circuits --> _is_pure_studio_generated(..., tool=tool) never runs --> owner_kind computed from path regardless of tool
```

**Impact**
The scanner-specific part of the #185 fix is ineffective for the common intact generated-stub shape; a copied or hand-edited generated-looking file with a foreign tool's binding can still enter the managed-output inventory (feeding gitignore-pattern generation and other downstream consumers of collect_managed_outputs) despite the new tool-scoping logic.

**Suggested correction**
Either drop `_extract_studio_owned_target(content)` from this disjunction when `tool` is known (rely solely on the tool-scoped `_is_pure_studio_generated`), or make `_extract_studio_owned_target`/its underlying regex checks tool-aware too, so the ordering can't bypass the new scoping.

**How to verify**
Add a test that constructs a foreign-tool-bound-but-follow-link-intact fixture under a dedicated provider root and asserts _classify_generated_output_owner (or _scan_owned_generated_outputs) does NOT classify it as owned by that provider's tool.

</details>

tkcoding: Fixed in 0f297f65 — added a detailed docstring to `_classify_generated_output_owner` explaining exactly this: for the common control/follow-target shape, `_extract_studio_owned_target` (tool-agnostic, no purity requirement) is always true whenever `_is_pure_studio_generated` could be, so the tool-scoped disjunct never decides the outcome there — traced the full consumer chain (`collect_managed_outputs` -> `list_managed_agent_output_paths` -> only .gitignore pattern generation in `init.py`, nothing deletion-related) to confirm the real severity.

I did not apply your suggested fix (dropping the first disjunct when `tool` is known), because it would regress a real case: it would stop reporting a Studio-originated-but-user-customized file as managed the moment it's edited, and it would break an existing, deliberate test from PR #208 (`test_cursor_launcher_ownership_check_short_circuits_before_purity_check`) that locks in this exact short-circuit as intended behavior.

One correction from my own deep review before committing: my first attempt at this docstring overclaimed universality — it's NOT redundant for a legacy single-line "endpoint-only" proxy stub using a pre-canonical path variable (e.g. `{cypilot_path}/`), since `_extract_studio_endpoint_target` only accepts the canonical `{cf-studio-path}/` prefix unlike `_extract_studio_follow_target`'s legacy-prefix allowlist — for that shape, the purity disjunct is the *only* thing that classifies the file as owned. Verified directly (`_extract_studio_owned_target` -> None, `_is_pure_studio_generated(tool=X)` -> True regardless of X) and added `test_classify_generated_output_owner_needs_the_purity_disjunct_for_legacy_endpoint_only_stubs` pinning the corrected, two-part claim.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Fixed in 0f297f65 — added `TestScanOwnedGeneratedOutputs` (tests/test_agents_coverage.py) with 3 tests giving `_scan_owned_generated_outputs`/`_classify_generated_output_owner`/`ManagedOutput` their first direct coverage: a dedicated single-tool provider root (`.claude/agents/`), the shared studio root (`.agents/skills/`), and a negative case (a file with no owned-target/purity signal at all, confirming it's excluded). Also added two more targeted tests on `_classify_generated_output_owner` directly while addressing your other findings — see the reply on the short-circuit comment.

tests/test_subagent_registration.py
tkcoding: Fixed in 0f297f65 — added `test_pure_generated_stub_matches_recognizes_a_pre_142_stub_for_any_known_tool`, combining a genuine pre-#142 body (context lines fully absent) with concrete `tool=` values including `claude` (which already has a real, non-null binding today, not just a hypothetical future one) plus `windsurf`/`cursor`/`None`, confirming the `legacy_protocol` branch's tool-independence is pinned rather than incidental.

skills/studio/scripts/studio/commands/agents.py
tkcoding: Fixed in 0f297f65 — turns out the fuller picture (see the reply on the short-circuit finding) makes this moot rather than something to guard against separately: `_classify_generated_output_owner`'s `tool` value is inert for the common shape regardless of provider, not only the shared `.agents/skills/` root, so a second sharer gaining a binding wouldn't reopen the collision through this function either way. Pinned that directly with `test_classify_generated_output_owner_ignores_tool_for_the_common_shape`, using exactly the scenario your own verification steps described (patching a second tool into `_ASK_TOOL_BINDING` and confirming classification of a still-unbound tool's file is unaffected).

ainetx: **RESOLVED** -- Confirmed fixed as of 0f297f655be38ae957b4edbb11aa500d1c58f9a6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Verified directly against current code: _classify_generated_output_owner (agents.py:5211-5296), which _scan_owned_generated_outputs routes the shared '.agents/skills/' root through with tool=None, is provably tool-blind for both content shapes it recognizes. For the common control/follow-target shape, the multi-line branch of _pure_generated_stub_matches can only reach the tool-scoped binding loop when control_target is extractable from the stripped content, but that same extraction (_extract_studio_control_target/_extract_studio_follow_target) is a strict subset of what _extract_studio_owned_target already checks on the full content — so the first disjunct in _classify_generated_output_owner is already True before the tool-scoped second disjunct's outcome could matter, and the follow-protocol's 'LOAD and RUN target_path' line (agents.py:213) is independent of ask_tool_name's value, so a future binding change can't affect this extraction either. For the legacy single-line 'endpoint-only' shape, that single-line branch in _pure_generated_stub_matches never reads tool at all. This is pinned by two new tests already in this diff (test_classify_generated_output_owner_ignores_tool_for_the_common_shape, which explicitly patches a second tool into _ASK_TOOL_BINDING and confirms classification is unaffected, and test_classify_generated_output_owner_needs_the_purity_disjunct_for_legacy_endpoint_only_stubs). Since tool's value never changes this function's outcome for either shape, tool=None on the shared.agents/skills/ scan root cannot reopen the cross-tool collision even once a second sharer of that directory gains its own _ASK_TOOL_BINDING entry.
- **Verifier:** Traced the actual code: _classify_generated_output_owner gates on `_extract_studio_owned_target(content) or _is_pure_studio_generated(content, tool=tool)`. For the common multi-line shape, _pure_generated_stub_matches's tool-scoped binding loop is only reached after `_extract_studio_control_target(stripped) or _extract_studio_follow_target(stripped)` succeeds — the same regex functions _extract_studio_owned_target already runs over the full (superset) content, so the first disjunct is provably already true whenever the second disjunct's tool-scoped outcome could matter; tool value is therefore inert here. For the legacy single-line endpoint-only shape, the check (`line.startswith('Constructor Studio endpoint only. Prompt source: ')`) is evaluated before any tool-scoped logic and never reads `tool` at all. So `tool=None` on the shared `.agents/skills/` root cannot reopen the cross-tool collision through this function even once a second sharer gains its own `_ASK_TOOL_BINDING` entry — confirmed directly in code, and pinned by the diff's own new tests (test_classify_generated_output_owner_ignores_tool_for_the_common_shape, which mocks exactly that scenario, and test_classify_generated_output_owner_needs_the_purity_disjunct_for_legacy_endpoint_only_stubs). The real #185 fix (tool-scoped exclusivity) matters only at call sites that decide purity alone without this redundant owned-target disjunct (e.g. _cleanup_legacy_skill_files, _cleanup_studio_legacy_subagents), and those already thread the concrete tool through correctly.

</details>

tests/test_subagent_registration.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0f297f655be38ae957b4edbb11aa500d1c58f9a6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff adds test_pure_generated_stub_matches_recognizes_a_pre_142_stub_for_any_known_tool (tests/test_subagent_registration.py), which builds pre_142_body by stripping both ask_tool_name and ask_tool_description lines (a genuine pre-#142 shape, not the later 'unset' shape), then asserts _pure_generated_stub_matches(pre_142_body, tool=tool) is True for tool in ('claude','windsurf','cursor',None), and explicitly confirms via assertIsNotNone(_ASK_TOOL_BINDING.get('claude')) that claude already has a real non-null binding today rather than being a hypothetical. This directly closes the gap the check describes: a concrete non-None tool combined with a fully-context-absent legacy body.
- **Verifier:** Verified directly in the current code/tests: the legacy_protocol branch (agents.py lines 384-400) runs unconditionally before the tool-scoped candidate loop and never reads `tool` at all. The diff adds test_pure_generated_stub_matches_recognizes_a_pre_142_stub_for_any_known_tool (test_subagent_registration.py:1454), which builds a genuine pre-#142 body using the exact same filter (both `- ask_tool_name` and `- ask_tool_description` lines fully absent, not merely 'unset'), then asserts _pure_generated_stub_matches(pre_142_body, tool=tool) is True for tool in ('claude','windsurf','cursor',None), and confirms via assertIsNotNone(_ASK_TOOL_BINDING.get('claude')) that claude already carries a real, non-null binding today (not hypothetical). The pre-existing test_pre_142_generated_files_still_recognized_as_pure_after_upgrade (line 1198) already covers rejection of extra user content appended to the same body shape. Together these close exactly the gap the check and posted finding described.

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0f297f655be38ae957b4edbb11aa500d1c58f9a6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff adds TestScanOwnedGeneratedOutputs in tests/test_agents_coverage.py with three tests directly calling _scan_owned_generated_outputs and asserting on the resulting ManagedOutput objects: one for a dedicated single-tool provider root (.claude/agents/, tool resolvable to 'claude'), one for the shared studio root (.agents/skills/, tool=None branch), and a negative case confirming a file with no owned/pure signal is excluded. Additionally, test_subagent_registration.py adds direct tests of _classify_generated_output_owner covering both the common shape (tool value provably inert) and the legacy endpoint-only shape (purity disjunct required). A repo-wide grep confirms these symbols now appear in both test files, resolving the previously-identified zero-coverage gap.
- **Verifier:** Confirmed directly against the code: the diff adds TestScanOwnedGeneratedOutputs (tests/test_agents_coverage.py, lines ~5453-5515) with three tests that call the real _scan_owned_generated_outputs (which internally calls _classify_generated_output_owner and builds ManagedOutput objects) — one for a dedicated single-tool provider root (.claude/agents/, provider='claude', owner_kind='agent'), one for the shared studio root (.agents/skills/, provider='studio', tool=None branch, owner_kind='skill'), and a negative case where a hand-written file with no owned/pure signal is correctly excluded. I traced this against the actual production code at agents.py:5155-5261 (_scan_owned_generated_outputs, _classify_generated_output_owner, _provider_for_output_path) and confirmed the expected provider/owner_kind values match real behavior (e.g. '/skills/' in rel -> 'skill'). tests/test_subagent_registration.py additionally adds two more direct tests of _classify_generated_output_owner covering the common control/follow-target shape and the legacy endpoint-only shape. A repo-wide grep confirms _classify_generated_output_owner, _scan_owned_generated_outputs, and ManagedOutput now appear in both test files, resolving the originally-identified zero-coverage gap for both the resolvable-tool and unresolvable-tool ('studio') branches.

</details>

skills/studio/scripts/studio/commands/agents.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0f297f655be38ae957b4edbb11aa500d1c58f9a6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Confirmed directly: _extract_studio_owned_target's three helpers (agents.py:244-303) match purely on the follow-link/control-protocol regex text and never inspect ask_tool_name lines, so for the common intact-follow-link shape this first disjunct is always true whenever the tool-scoped _is_pure_studio_generated(..., tool=tool) disjunct could be — the short-circuit claim is real. However, the current diff's _classify_generated_output_owner docstring (agents.py:5211-5247) explicitly documents this exact precedence for both shapes (common control/follow-target shape where tool is provably inert, and the legacy '{cypilot_path}/' endpoint-only shape where the purity disjunct — and hence tool — is the only thing that decides ownership), and two new tests (test_classify_generated_output_owner_ignores_tool_for_the_common_shape and test_classify_generated_output_owner_needs_the_purity_disjunct_for_legacy_endpoint_only_stubs, both verified present in tests/test_subagent_registration.py) pin both halves of this claim with passing assertions across multiple tool values. The maintainer's discussion reply also traces the consumer chain (collect_managed_outputs -> list_managed_agent_output_paths ->.gitignore pattern generation only, not deletion) confirming low blast radius. This is now fully documented, intentional, and tested behavior rather than an undisclosed dead-code regression, matching the prior round's confirmed-pass verdict which I independently re-verified against the current code.
- **Verifier:** Directly verified: _extract_studio_owned_target and the multi-line branch of _pure_generated_stub_matches both require the same follow/control-target regex match (agents.py:258-264, 381-383), so for the common shape the tool-agnostic first disjunct in _classify_generated_output_owner is always true whenever the tool-scoped purity disjunct could be — the short-circuit claim is technically real. However this predates this PR (locked in by test_cursor_launcher_ownership_check_short_circuits_before_purity_check from PR #208, agents.py's own docstring at 5211-5247 now explicitly documents it), and this diff correctly identifies and fixes the one shape where tool-scoping IS load-bearing — the legacy '{cypilot_path}/' endpoint-only proxy stub, where _extract_studio_owned_target returns None and only the tool-scoped purity check decides ownership — pinned by test_classify_generated_output_owner_needs_the_purity_disjunct_for_legacy_endpoint_only_stubs (verified present at tests/test_subagent_registration.py:1335). Both halves are now correctly documented and tested rather than silently broken, and the function's output only feeds.gitignore pattern generation, not deletion, so the redundancy for the common case is non-destructive.

</details>

tkcoding:
@ainetx please check all issue is resolve or not.

ainetx (approved):
Solid, well-targeted fix for issue #185: scoping the ask-tool-binding candidate check to the resolving tool (with a documented, deliberate fallback for the genuinely-unresolvable `studio` shared root) closes the cross-tool collision without disturbing the tool-agnostic paths that legitimately need to stay tool-blind. The inline reasoning about why `_ASK_TOOL_BINDING_BY_OUTPUT` stays out of both branches, and why `_classify_generated_output_owner`'s `tool` value doesn't change outcomes for either the common follow-target shape or the legacy endpoint-only shape, is thorough and checks out against the code. Test coverage is strong: new end-to-end and unit tests pin the foreign-binding rejection, the pre-binding/legacy-unset compatibility case, and previously untested surface (`_scan_owned_generated_outputs`, `_classify_generated_output_owner`, `ManagedOutput`). No correctness, security, or test-coverage issues found — approving as is.

PR #231, opened 2026-09-17
feat(pdsl): declare menu shape instead of inferring it from prose (#186)

### LINKED ISSUE #186: Native-dialog shape-compatibility test is ambiguous against menus with free-form multi-select replies
**Found during:** deep review of #142's native-dialog implementation ([PR #184](https://github.com/constructorfabric/studio/pull/184)), not part of ainetx's review — filing separately per the maintainer's request to track it rather than block the PR.

## Problem

`pdsl-execution-card.md`'s new "native-dialog shape-compatible" test only checks:
- at most 4 top-level `OPTIONS` entries, and
- no entry "documented as accepting free-text/arbitrary input (a path, a name, 'or describe your own', etc.)"

This wording is ambiguous against real existing menus whose *overall* reply is free-form even though each individual `OPTIONS` entry looks like an ordinary fixed choice:

- `RepoSelectionMenu` (`skills/studio/modules/workspace-discover.md`) has only 3 options, but its `TITLE` is "Reply with numbers/names or `all`" — the menu is designed to accept an arbitrary multi-select list, not a single pick among 3 buttons.
- `ZeroResultsMenu` (same file) embeds a free-text collection *inside* one option's action clause ("2 manual -> WAIT a manual source (name + path or URL)...") rather than in the option's label, so the letter of the rule doesn't clearly flag it either.

Neither menu explicitly says "describe your own" in a way the current wording is guaranteed to catch.

## Failure scenario

On a target with a real `ask_tool_name` binding (today: Claude), a model literally applying the shape-compatibility rule could judge either menu "shape-compatible" (≤4 options, no option textually says "free text") and route it through a fixed 4-choice native dialog — silently collapsing the menu's actual multi-select/free-form reply into a single button-press, losing functionality the menu's own design depends on.

There's also no stated **minimum** option count, so a degenerate 1-option `EMIT_MENU` is technically "shape-compatible" today — likely harmless, but worth deciding deliberately rather than by omission.

## Suggested fix

Tighten the shape-compatibility rule to also exclude a menu whose `TITLE` or any `OPTIONS` action clause describes accepting more than one selection, an open-ended list, or a value embedded in prose (not just a literal "free text" flag on the option label) — or require menu authors to declare shape/arity explicitly (mirroring the existing `MENU ... TYPE:` declaration pattern from #153) so the routing rule reads a fact instead of inferring one from prose.

## Scope

Tracked as a follow-up rather than blocking #142/#184, per discussion in the PR.

tkcoding:
## Summary

Closes #186.

The native-dialog shape-compatibility rule in `pdsl-execution-card.md`
inferred "free-text-accepting" from prose (option wording, entry count),
which is ambiguous against real menus whose *overall* reply is free-form
even though each `OPTIONS` entry looks like an ordinary fixed choice:

- `RepoSelectionMenu` (`skills/studio/modules/workspace-discover.md`) has
  only 3 options, but its `TITLE` is "Reply with numbers/names or `all`" --
  designed to accept an arbitrary multi-select list, not a single pick among
  3 buttons.
- `ZeroResultsMenu` (same file) embeds a free-text collection *inside* one
  option's action clause rather than in the option's label, so the letter of
  the old rule didn't clearly flag it either.

A model applying the old rule literally could judge either menu
"shape-compatible" and route it through a fixed 4-choice native dialog --
silently collapsing the menu's actual multi-select/free-form reply into a
single button-press.

## Change

- Adds an optional `SHAPE: fixed-choice | free-form` MENU header to
  `skills/studio/scripts/studio/utils/pdsl.py`, mirroring the existing
  `TYPE: confirmation | decision | blocking` gate-risk declaration from #153
  exactly: same declaration-region tracking (`state.gate_scope`), the same
  one-declaration-per-menu dedup (new `state.menu_shape_line` field), and the
  same near-miss/typo detection -- now via generalized helpers
  (`_is_declared_header_typo`, `_looks_like_declaration_attempt`) shared with
  `TYPE`'s own checks. `TYPE` and `SHAPE`'s near-miss detectors can never
  cross-fire against each other: their edit distance is 3, above the
  distance-1 threshold either one uses.
- New rule IDs: `PDSL710` (bad value), `PDSL711` (duplicate), `PDSL712`
  (declared outside scope), `PDSL713` (near-miss unrecognized).
- `pdsl-execution-card.md`'s shape-compatibility rule now reads a declared
  `SHAPE` when present -- `fixed-choice` is shape-compatible, `free-form`
  never is, regardless of option count or wording -- and falls back to the
  existing heuristic (now also excluding `TITLE`-embedded free text,
  open-ended lists, and multi-select) only when `SHAPE` is undeclared or
  holds an unvalidated value (unvalidated PDSL is still executable, so the
  runtime rule can't assume validation has run).
- The ~113 existing undeclared menus are grandfathered under
  `UNSHAPED_MENU_BASELINE` + a shrink-only ceiling in `tests/test_pdsl_keywords.py`,
  mirroring `UNTYPED_MENU_BASELINE` exactly: the surface migrates menu by
  menu and must not grow.

## Testing

- `tests/test_pdsl_validate_cli.py`: a focused directed test suite covering
  declared/undeclared classification (both shape tokens, both indentation
  shapes), bad value (`PDSL710`), duplicate (`PDSL711`), declared outside
  scope (`PDSL712`), near-miss (`PDSL713`), `TYPE`/`SHAPE` independence and
  ordering (either keyword can be declared first), and the deep-indent
  continuation-swallowing edge case mirroring `TYPE`'s own coverage.
  Deliberately does not re-clone every one of `TYPE`'s dozens of edge-case
  variants (confusable Unicode letters, exact elision boundary, etc.), since
  those exercise shared, already-tested low-level helpers
  (`_gate_header_candidate`, `_edit_distance`, `_normalize_header_name`)
  rather than `SHAPE`-specific logic. The property-based fuzzer
  (`test_property_validation_never_raises_on_arbitrary_text`) gained a few
  `SHAPE` fragments so it exercises the new dispatch path too.
- `tests/test_pdsl_keywords.py`: `test_the_unshaped_menu_surface_does_not_grow`
  mirrors `test_the_untyped_menu_surface_does_not_grow`, backed by a new
  `_menu_shape_declarations()` discovery function mirroring
  `_menu_type_declarations()`.
- `tests/test_workflow_subagents_dispatch.py`: updated
  `test_pdsl_execution_card_routes_emit_menu_through_native_ask_tool` to pin
  the new rule wording.
- Full suite: 6225 passed, 6 failed (all pre-existing/environmental --
  verified via `git stash` that each also fails without this branch's
  changes; 5 are known macOS filesystem-encoding edge cases, 1 is an
  unrelated subprocess/pytest-availability quirk in
  `test_tomllib_compat.py`), 5 skipped, 15 xfailed.
- `cfs pdsl validate` over the whole authored corpus: no new/unexpected
  findings (`test_prompt_pdsl_blocks_pass_cfs_pdsl_validate` passes
  unchanged).
- `pylint` clean on all 5 changed files.
- Independently verified via a deep multi-angle review (near-miss
  cross-fire fuzzing across ~550 edit-distance-1 candidates of both
  keywords, declaration-order testing, baseline-completeness check) --
  findings from that pass (an underspecified execution-card fallback for a
  declared-but-invalid `SHAPE` value, and two test-coverage gaps for
  declaration ordering and deep-indent continuation-swallowing) are already
  fixed in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for declaring menu input shapes as `fixed-choice` or `free-form`.
  * Added validation for invalid, duplicate, misplaced, and misspelled shape declarations.
  * Menus use the native dialog only when compatible, with no more than four options and an exact `fixed-choice` declaration or compatible wording.
  * Incompatible menus, including `free-form` menus, use the text-menu experience.

* **Tests**
  * Expanded coverage for shape declarations, compatibility rules, and validation behavior.


skills/studio/modules/runtime/pdsl-execution-card.md
```diff
@@ -56,9 +56,17 @@ RULES:
     applies to that pairing, not to text output that merely lists choices
     without waiting on a reply.
   ALWAYS, when executing `EMIT_MENU`, first check the menu is native-dialog
-    shape-compatible: at most 4 top-level `OPTIONS` entries, and no entry
-    documented as accepting free-text/arbitrary input (a path, a name, "or
-    describe your own", etc.) rather than choosing among the listed entries.
+    shape-compatible. When the owning `MENU` declares exactly `SHAPE:
+    fixed-choice` or `SHAPE: free-form`, read that fact instead of inferring
+    one: `fixed-choice` is shape-compatible, `free-form` never is, regardless
+    of option count or wording (issue #186). When `SHAPE` is undeclared, OR
```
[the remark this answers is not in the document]
tkcoding: Fixed in 38767bff: the shape-compatibility rule now always enforces the <=4-option cap regardless of a declared SHAPE — `fixed-choice` no longer bypasses it. SHAPE governs only the free-text/wording axis; the option-count check is separate and unconditional (Claude's AskUserQuestion hard-rejects >4 options, and SHAPE never overrides a tool's own mechanical limit). Also updated the later shape-incompatible restatement further down in the same rule, which my first pass left stale — it now mentions the declared-SHAPE override too, not just the old option-count/free-text heuristic.

skills/studio/modules/runtime/pdsl-execution-card.md
```diff
@@ -56,9 +56,17 @@ RULES:
...
+    declared with any other value (unvalidated PDSL is still executable;
+    PDSL validation flags this at authoring time but a runtime reader must
```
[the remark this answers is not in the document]
tkcoding: Fixed in 38767bff: the rule now states that only the first in-scope SHAPE declaration in a MENU's declaration region is read — a later or out-of-scope line is not a second declaration. Also fixed a related latent bug your investigation didn't touch but this same fix round surfaced: the pre-existing TYPE-declaration test guard (tests/test_pdsl_keywords.py) had a declaration-region-end check that only exempted TITLE, not the newly-added SHAPE header, so a menu declaring SHAPE before TYPE would have hidden its TYPE declaration from the untyped-surface growth guard. Fixed by sharing one DECLARED_HEADER_NON_TERMINATORS constant between the validator (pdsl.py) and the test-side scanner instead of two independently-maintained literals, plus a regression test covering both declaration orderings.

ainetx (approved):
Reviewed the SHAPE declaration addition for MENU blocks (issue #186) — the shared validation logic between TYPE and SHAPE headers, the near-miss/typo detection, the declaration-region handling for ordering (SHAPE before/after TYPE), and the test coverage across the validator, the CLI, and the execution-card runtime rules. The refactor into a shared `_handle_declared_menu_header` helper is clean and the test baseline/ceiling migration pattern is consistent with the existing TYPE precedent. Tests are thorough, covering edge cases like cross-firing near-misses, indentation boundaries, and declaration ordering. No blocking issues found — approving.

ainetx:
### New SHAPE declaration and PDSL710-713 rule band are undocumented in the authoritative PDSL spec/feature docs

**Severity:** Major

**Problem**
pdsl.py adds a full new MENU declaration mechanism (`SHAPE: fixed-choice|free-form`) with four new rule IDs (PDSL710-713) and pdsl-execution-card.md gives it detailed runtime semantics, but architecture/specs/PDSL.md (which documents the parallel `TYPE`/gate-risk feature in full, including a table and edge-case rules) has no mention of SHAPE at all, and architecture/features/pdsl-validation-cli.md's rule-band table still only lists 'PDSL700 for declared gate risk' with no entry for the new 710-713 band.

<details>

**How to reproduce**
1. Open architecture/specs/PDSL.md and search for 'SHAPE' -- only unrelated hits ('Core Shape', 'output shape'). 2. Open architecture/features/pdsl-validation-cli.md and search for 'SHAPE' or '710'/'711' -- no hits; the rule-band acceptance criterion only names PDSL100-700. 3. Compare to pdsl.py's MENU_SHAPE_HEADER/PDSL710-713 and pdsl-execution-card.md's SHAPE routing rules, which are fully implemented and tested.

**Expected behavior**
The authoritative spec (PDSL.md) and feature doc (pdsl-validation-cli.md) document the new SHAPE header, its valid tokens, declaration-region rules, and the PDSL710-713 rule band the same way they document TYPE/PDSL700, or explicitly note the deferral.

**Actual behavior**
Only the runtime execution-card module and test files reference SHAPE/issue #186; the two docs that are supposed to be the source of truth for PDSL syntax and the CLI's rule catalog are silent on it.

```text
pdsl.py (SHAPE, PDSL710-713) --implements--> new behavior
|
v
pdsl-execution-card.md --documents runtime routing-->
|
X (no update)
v
architecture/specs/PDSL.md / pdsl-validation-cli.md --stays silent on SHAPE-->
```

**Impact**
Authors/reviewers consulting the language spec or CLI feature doc will not learn that SHAPE exists, how it is validated, or that PDSL710-713 findings can occur, causing confusion and drift between documented and actual CLI behavior.

**Suggested correction**
Add a 'Declared menu shape' section to architecture/specs/PDSL.md mirroring the existing 'Declared gate risk' section, and extend the PDSL700-band description / acceptance criteria in architecture/features/pdsl-validation-cli.md to include the new PDSL710-713 band.

**How to verify**
Grep architecture/specs/PDSL.md and architecture/features/pdsl-validation-cli.md for 'SHAPE' and 'PDSL71' after the fix and confirm they describe the new tokens/rules.

</details>

---
_Original location: architecture/specs/PDSL.md:112 -- inline anchoring could not be resolved after 1 attempt(s)._

tkcoding:
Fixed in 08d89d6e — confirmed via `grep -n "SHAPE" architecture/specs/PDSL.md architecture/features/pdsl-validation-cli.md` returning nothing before the fix.

- Added a "Declared menu shape" section to `architecture/specs/PDSL.md` mirroring the existing "Declared gate risk" section exactly: the meaning table, a worked example (now showing `TYPE` and `SHAPE` declared together on one menu, since they share a declaration region), and every rule (region/continuation/near-miss mechanics), verified against the actual code rather than assumed — e.g. the ordinary-English near-misses of `SHAPE` listed (`SHARE`/`SHAKE`/`SHADE`/`SHAME`/`SHALE`/`SHAPED`/`SHAPES`) were confirmed by directly calling `_is_menu_shape_header_typo()`. Also called out the one real asymmetry from `TYPE`: `SHAPE`'s undeclared fallback (the execution-card prose heuristic) is live, executable behavior today, unlike `TYPE`'s undeclared-defaults-to-`blocking`, which nothing currently enforces.
- Updated `architecture/features/pdsl-validation-cli.md`'s step descriptions and acceptance criteria to name `SHAPE`/`PDSL710-713` alongside `TYPE`/`PDSL700`. The `SHAPE` code reuses the exact same `@cpt` traceability markers as `TYPE` (no new ones were added), so this was a prose-accuracy fix rather than a new spec-coverage obligation — verified with `spec-coverage --min-coverage 90 --min-file-coverage 60 --min-granularity 0.45` locally, matching CI's invocation ("All thresholds met").

One side effect worth flagging: the new worked example sits inside a ` ```pdsl ` fence, so it's scanned as a real `MENU` by the untyped/unshaped baseline guards in `tests/test_pdsl_keywords.py`. It declares both `TYPE` and `SHAPE` so it doesn't newly appear in either baseline, but inserting it before the file's existing `SubAgentApprovalMenu` example shifted that example's block index (`#13` -> `#14`); both `UNTYPED_MENU_BASELINE` and `UNSHAPED_MENU_BASELINE` were updated to match, per those tests' own documented rename-handling contract. Full suite re-verified green (6223 passed, only the 6 known pre-existing/environmental failures) and `spec-coverage` reconfirmed passing after that change.

tests/test_pdsl_validate_cli.py
```diff
@@ -697,6 +697,135 @@ def test_prose_and_control_flow_headers_are_not_gate_declarations() -> None:
...
+        "single-select",
+        "{MENU_SHAPE}",
+        "fixed-choice WHEN SIMPLE_MODE == normal",
+        "Fixed-Choice",
+        "<fixed-choice | free-form>",
+    ):
+        assert _rule_ids(_gate_menu(extra=f"SHAPE: {value}")) == ["PDSL710"], value
+
+    # A bare `SHAPE:` with no value declares nothing but is still a declaration.
+    assert _rule_ids(_gate_menu(extra="SHAPE:")) == ["PDSL710"]
```
ainetx, line 730: ### SHAPE's casing-dependent empty-value contract lacks a dedicated pinned test

**Severity:** Minor

**Problem**
PDSL.md documents that empty-valued `shape:`/`Shape:` must be reported as PDSL713 near-misses while empty-valued `SHAPE:` must be PDSL710 -- the same contract TYPE has a dedicated test for (`test_an_abandoned_declaration_is_reported_whatever_its_casing`). No equivalent test exists for SHAPE: `test_menu_shape_must_be_one_literal_enum_token` only tests bare `SHAPE:` (correct case) at line 730, and `test_a_malformed_menu_shape_header_is_reported` only tests `Shape:`/`shape:` with a non-empty value (`fixed-choice`) at lines 763-764.

<details>

**How to reproduce**
1. Inspect tests/test_pdsl_validate_cli.py for any assertion combining an empty value with `shape:` or `Shape:`. 2. None exists distinguishing PDSL710 vs PDSL713 for empty-valued SHAPE by casing.

**Expected behavior**
A test asserting `_rule_ids(_gate_menu(extra="shape:")) == ["PDSL713"]`, `_rule_ids(_gate_menu(extra="Shape:")) == ["PDSL713"]`, and `_rule_ids(_gate_menu(extra="SHAPE:")) == ["PDSL710"]`.

**Actual behavior**
Only the correctly-cased empty case and non-empty miscased cases are pinned; the miscased+empty combination is untested.

```text
shared _looks_like_declaration_attempt(name, has_separator, value, valid_tokens) -> [empty value + miscased -> PDSL713] vs [empty value + correct case -> handled directly as PDSL710] -- only the latter branch is pinned by a test for SHAPE
```

**Impact**
A future refactor of the shared miscasing-suppression guard could silently start treating empty-valued, miscased SHAPE headers as prose, leaving a menu's shape declaration attempt undetected, without any existing test failing.

**Suggested correction**
Add a SHAPE-specific test mirroring `test_an_abandoned_declaration_is_reported_whatever_its_casing`, asserting `shape:`/`Shape:` -> PDSL713 and `SHAPE:` -> PDSL710.

**How to verify**
Run the new test against the current implementation (should pass) and against a mutated guard that drops the empty-value exemption (should fail).

</details>

architecture/specs/PDSL.md
```diff
@@ -472,6 +472,83 @@ Rules:
...
+routing reads that fact from source rather than inferring it from prose
+(issue #186).
+
+```pdsl
+MENU OpenEndedReplyMenu:
+  TITLE: Reply with numbers/names or `all`
+  TYPE: decision
+  SHAPE: free-form
+  OPTIONS:
+    1 repo-a -> SET SELECTED_REPOS += repo-a
```
ainetx: ### Undocumented `+=` compound-assignment operator introduced only in a PDSL.md example

**Severity:** Minor

**Problem**
The new OpenEndedReplyMenu example uses `SET SELECTED_REPOS += repo-a` / `+= repo-b` for accumulating into a list-valued state variable, but `+=` is not defined anywhere else in the spec: the Core Shape grammar only shows `SET <name>: <allowed values>` (declaration) and `=` (assignment), the Keywords table's `SET` entry says only 'Assign state', and every other SET example in the file uses `=`. pdsl.py's validator has zero handling of `+=` as an operator.

<details>

**How to reproduce**
1. grep architecture/specs/PDSL.md for `+=` -> only lines 488-489. 2. grep pdsl.py for `+=` -> only unrelated Python state counters, no PDSL-operator handling.

**Expected behavior**
Either the spec defines `+=` semantics (accumulation into list/set-valued STATE, interaction with `SET <name>: <allowed values>`) with validator/test coverage, or the example uses only the documented `=` assignment.

**Actual behavior**
`+=` appears solely inside an illustrative example with no grammar definition, no keyword-table entry, and no validator recognition.

```text
PDSL.md example uses SET X += Y -> no grammar rule defines it -> pdsl.py validator has no operator check -> authors/tools may treat += as valid PDSL when it is unspecified
```

**Impact**
Authors or automated tooling relying on the spec as the source of truth for valid PDSL syntax could adopt `+=` believing it's supported, while the validator silently accepts it without understanding accumulation semantics, and no runtime/validator component implements list-accumulation SET behavior.

**Suggested correction**
Either add a grammar/keyword-table definition and validator support for `+=` accumulation semantics, or change the example to use documented `=` assignment (e.g. building the list via existing constructs).

**How to verify**
Confirm the spec's grammar section and SET keyword-table entry mention `+=` with defined semantics, and that pdsl.py/tests recognize and validate it -- or confirm the example no longer uses `+=`.

</details>

tests/test_pdsl_validate_cli.py
tkcoding: Fixed in e801c788 — added `test_an_abandoned_shape_declaration_is_reported_whatever_its_casing`, mirroring `test_an_abandoned_declaration_is_reported_whatever_its_casing` (TYPE's dedicated test for this exact contract). Confirmed the actual behavior directly before writing the test: `SHAPE:` -> PDSL710, `Shape:`/`shape:` -> PDSL713, and `shape: skill`/`Shape: CLI`/`shape: workflow` -> no findings (stays prose).

ainetx: **RESOLVED** -- Confirmed fixed as of e801c78852b584649df2f41aba733329292f1bf2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current code now contains `test_an_abandoned_shape_declaration_is_reported_whatever_its_casing` (tests/test_pdsl_validate_cli.py:779-798), which explicitly asserts `_rule_ids(_gate_menu(extra="SHAPE:")) == ["PDSL710"]` and `_rule_ids(_gate_menu(extra=variant)) == ["PDSL713"]` for `shape:`/`Shape:`, mirroring TYPE's dedicated casing test exactly as the discussion thread's fix (e801c788) describes -- the gap identified in prior rounds has been closed.
- **Verifier:** The current diff/code contains `test_an_abandoned_shape_declaration_is_reported_whatever_its_casing` (tests/test_pdsl_validate_cli.py:779-792), which explicitly asserts `_rule_ids(_gate_menu(extra="SHAPE:")) == ["PDSL710"]` and `_rule_ids(_gate_menu(extra=variant)) == ["PDSL713"]` for `shape:`/`Shape:`, plus that non-empty other-token values (`shape: skill`, `Shape: CLI`, `shape: workflow`) stay prose (`== []`). This precisely mirrors PDSL.md:549-550's documented contract and is even more precise than TYPE's own analogous test (line 1551-1565, which only asserts `!= []` rather than pinning exact rule IDs). The gap identified in the prior round (fixed in commit e801c788 per the discussion thread) is confirmed closed by direct inspection of the test file.

</details>

architecture/specs/PDSL.md
tkcoding: Fixed in f51b3633 — replaced the invented `+=` operator with the same prose-action-clause style the real corpus menu this example is modeled on already uses (`RepoSelectionMenu` in `workspace-discover.md`: "parse the selection into the included-sources list, then ..."). No new PDSL syntax needed; the example still demonstrates a `SHAPE: free-form` menu just as well without it.

tkcoding:
Re: the "outside diff range" finding in the review at https://github.com/constructorfabric/studio/pull/231#pullrequestreview-5244009316 (no inline comment ID, since it couldn't be posted on the diff) — **Align the earlier TYPE declaration-region rules.**

Fixed in f51b3633 — the pre-existing "Declared gate risk" section's declaration-region rules ("from the menu header up to the first section that is not TITLE or TYPE" / "any recognized section other than TITLE or TYPE ends the region") predated SHAPE and gave conflicting guidance against the new "Declared menu shape" section, which correctly states the two headers share one region. Updated both to name SHAPE alongside TYPE, matching the actual shared implementation (`DECLARED_HEADER_NON_TERMINATORS` in pdsl.py).

architecture/specs/PDSL.md
ainetx: **RESOLVED** -- Confirmed fixed as of f51b36331623d2357d610f10b4d4e9f7e2b6c8e2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Verified current PDSL.md (lines 482-495): the OpenEndedReplyMenu example no longer uses `+=` anywhere in the file (confirmed via grep, zero matches). It now uses prose action clauses ('parse the selection into the selected-repos list, then CONTINUE CurrentWorkflow') for options 1/2 and plain `=` for option 3 ('SET SELECTED_REPOS = all'), matching the discussion thread's stated fix in commit f51b3633 and consistent with the rest of the spec's SET usage.
- **Verifier:** Verified current architecture/specs/PDSL.md directly: grep for '+=' returns zero matches in the file, and the OpenEndedReplyMenu example (lines 484-497) now uses prose action clauses ('parse the selection into the selected-repos list, then CONTINUE CurrentWorkflow') for options 1/2 and plain '=' ('SET SELECTED_REPOS = all') for option 3, matching the discussion thread's stated fix in commit f51b3633. The originally flagged undocumented operator is no longer present.

</details>

tkcoding:
@ainetx please check all issues is resolved or not.

architecture/specs/PDSL.md
```diff
@@ -472,6 +474,83 @@ Rules:
...
+routing reads that fact from source rather than inferring it from prose
+(issue #186).
+
+```pdsl
+MENU OpenEndedReplyMenu:
+  TITLE: Reply with numbers/names or `all`
+  TYPE: decision
+  SHAPE: free-form
+  OPTIONS:
+    1 repo-a -> parse the selection into the selected-repos list, then CONTINUE CurrentWorkflow
```
ainetx: ### OpenEndedReplyMenu example mixes unstructured prose and structured DO-syntax action clauses with no reconciling rule

**Severity:** Minor

**Problem**
The new flagship example for SHAPE (architecture/specs/PDSL.md:485-497) has options 1 and 2 use free natural-language prose as the action clause while option 3 in the same OPTIONS block uses PDSL's structured DO-syntax. No text anywhere in the document (Menu rules, Keywords table, State rules, or the new 'Declared menu shape' section) states that an OPTIONS action clause may be unstructured prose, restricts that to SHAPE: free-form menus, or requires consistency of action-clause style within one menu. 'SELECTED_REPOS' is never declared via STATE anywhere in the doc, and options 1-2 never explicitly perform a stated action, contradicting the spec's own design goal (lines 41-42) and its explicit State rule (lines 572-573) that every referenced state variable be declared.

<details>

**How to reproduce**
1. Open architecture/specs/PDSL.md and read the new 'Declared menu shape' section (lines 477-552). 2. Compare the OpenEndedReplyMenu example's options 1-2 action clauses against option 3's and against every other MENU example in the file (lines 65-72, 331-338, 360-365, 670-676). 3. Search the document for a STATE declaration of SELECTED_REPOS or any rule permitting free-text action clauses -- none exists.

**Expected behavior**
Either the spec explicitly defines when a free-text/unstructured OPTIONS action clause is valid PDSL grammar (e.g. only under SHAPE: free-form) and requires uniform action-clause style within a menu, or the example is corrected to use consistent structured DO-syntax (with SELECTED_REPOS declared in STATE) across all three options.

**Actual behavior**
The example ships with two undefined prose action clauses alongside one structured action, with no spec text anywhere reconciling the two styles or declaring the referenced state variable.

```text
New SHAPE feature -> flagship example added -> options 1-2 use prose actions, option 3 uses SET syntax -> no spec rule distinguishes/permits this -> SELECTED_REPOS never declared in STATE -> reader/validator has no way to know if prose actions are valid PDSL or a mistake
```

**Impact**
Authors copying this example may believe free natural-language action clauses are valid PDSL syntax generally; a future validator implementer has no spec text to decide whether to accept or reject prose action clauses, undermining the fail-closed clarity SHAPE was introduced to support.

**Suggested correction**
Either add spec text formally permitting free-text action clauses (scoped to SHAPE: free-form) with a stated resolution mechanism for the referenced state, and declare SELECTED_REPOS in a STATE block; or rewrite options 1-2 to use structured DO/SET syntax consistent with option 3 and the rest of the document.

**How to verify**
Re-read the 'Declared menu shape' section and confirm it states a rule for action-clause style/consistency and that SELECTED_REPOS (or its replacement) is declared in a STATE block.

</details>

tkcoding: Fixed in 59bec196 — removed the state variable entirely and switched all three options to plain `CONTINUE CurrentWorkflow`, matching the very next example (`PlanApprovalGate`) in the same file, which sidesteps this exact class of issue the same way: no `SET`, no `STATE` block needed, no mixed action-clause styles. The `TITLE`/`INVALID` text alone already conveys the free-form/multi-select intent `SHAPE: free-form` is meant to illustrate — the example doesn't need to also model selection-accumulation logic to make that point.

ainetx: **RESOLVED** -- Confirmed fixed as of 59bec1969deba5fa9b1fcc6ccbf7dc1fb30e0f93 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Verified current PDSL.md (lines 485-497): the OpenEndedReplyMenu example was fixed per the discussion thread (commit 59bec196) — all three options now use identical structured 'CONTINUE CurrentWorkflow' action clauses with no prose action clause, and 'SELECTED_REPOS' no longer appears anywhere in the file. The mixed structured/unstructured action-clause inconsistency this check flagged is resolved in the code currently under review.
- **Verifier:** Verified current architecture/specs/PDSL.md lines 485-497: OpenEndedReplyMenu's three options all use identical structured 'CONTINUE CurrentWorkflow' action clauses (no prose action clause remains, no SET, no mixed styles), and a repo-wide search confirms 'SELECTED_REPOS' no longer appears anywhere in the file. This matches the discussion thread's claim that commit 59bec196 fixed the issue by removing the state variable and unifying all three options to plain CONTINUE, mirroring the adjacent PlanApprovalGate example. The originally flagged inconsistency is resolved in the code under review.

</details>

PR #234, opened 2026-09-17
fix(eval): stop the harness trusting its input more than it should

### LINKED ISSUE #226: eval harness: NaN baselines, an unbounded judge call, and two over-permissive structural checks
Group 2 of the review findings left open on #192 after it merged. All four are in the eval harness, which ships for the first time in v1.7.0 — but `eval` is advisory and gates nothing, so none of these block the tag.

Verified against `release/v1.7.0` @ `f9dbace9`.

## 1. A NaN baseline silently suppresses regression detection — Major

`diff_reports()` in `utils/eval_harness.py:461`:

```python
# missing / non-numeric baseline → no comparison; exclude bool (a subclass of int).
if not isinstance(before, (int, float)) or isinstance(before, bool):
    before = None
```

The guard rejects booleans and non-numerics. `float('nan')` is a float, so it passes. Then:

```python
elif now < before:   regressed.append(...)     # NaN → False
elif now > before:   improved.append(...)      # NaN → False
```

Both comparisons are `False` for NaN, so the scenario lands in no bucket at all, and `has_regression` stays `False` — even if compliance fell to 0. A corrupted or hand-edited baseline turns the regression gate off without saying anything.

This is the same defect class as the `--min-coverage nan` one fixed in #200, in a different place: a value that survives a type check and then disappears through comparison.

**Fix**: `math.isfinite()` in the guard, alongside the existing checks.

## 2. The injected judge has no execution deadline — Minor

`AdvisoryJudge.score` calls `self._judge_fn(request)` synchronously. The surrounding `except Exception` converts *raised* exceptions into an UNKNOWN verdict, which is the right degradation — but an adapter that hangs rather than raises is not an exception. The suite stops, with no verdict and no timeout.

Same shape as the decision-log append lock fixed in #200: a guard that only catches raising, protecting a call that can block.

**Fix**: bound the call. The reviewer filed this twice (`decision_log.py:260` and `eval_judge.py:296`); it is one finding — the `eval_judge` anchor is the correct one.

## 3. Any truthy scalar counts as a declared output list — Minor

`_check_every_phase_declares_output()` in `utils/eval_structural.py:339` passes a phase whenever `front.get('outputs')` or `front.get('output_files')` is truthy. `outputs = true`, `outputs = 1`, and `outputs = "step.out"` all satisfy it, so malformed frontmatter gets the same structural credit as a correctly declared list.

**Fix**: require a list of strings, and report the malformed case rather than passing it.

## 4. Tilde-fenced code blocks are not stripped — Minor

`_prose_headings()` in `utils/eval_structural.py:359` toggles `in_fence` on triple backticks only. CommonMark also allows `~~~` fences, so a `## Rules` heading inside a tilde-fenced code sample is treated as a real section and can satisfy `_check_required_sections()` — structural credit for content that is an example, not a section.

**Fix**: recognise `~~~` as a fence delimiter alongside ` ``` `.

## Why these belong together

All four are the scorer trusting its input more than it should: a number that is not a number, a call that never returns, a scalar standing in for a list, a code sample standing in for prose. Fixing them in one pass keeps the reasoning — and the tests — in one place.

## Scope

- [ ] `math.isfinite()` guard in `diff_reports`, with a NaN-baseline regression test
- [ ] Bound the injected `judge_fn` call, with a hanging-adapter test
- [ ] Require a list for phase outputs, with tests for `true` / `1` / bare string
- [ ] Strip `~~~` fences, with a test for a heading inside one

## Context

- #192 — the release PR these were raised on
- #200 — where the same two defect classes (NaN through a type check, a guard that only catches raising) were fixed elsewhere

vasylcf:
Closes #226.

Group 2 of the findings left open on #192. Four findings in the eval harness, all the same shape: **a value that passes a check and then means something other than what the check assumed.** None of them gates, so none blocks the v1.7.0 tag — but all four ship in it.

## 1. A NaN baseline switched off regression detection — Major

```python
# missing / non-numeric baseline → no comparison; exclude bool (a subclass of int).
if not isinstance(before, (int, float)) or isinstance(before, bool):
    before = None
```

NaN is a float, so it survived the guard — and then disappeared a second time:

```python
elif now < before:   regressed.append(...)   # False for NaN
elif now > before:   improved.append(...)    # False for NaN
```

The scenario joined **neither** bucket, so `has_regression` stayed `False` even for a drop to zero. A corrupt or hand-edited baseline turned the gate off and said nothing about it.

`math.isfinite` now guards both places the baseline is read — the comparison and the departed-scenario tail. Non-finite means "no usable baseline", which this function already has a behaviour for.

Same defect class as `--min-coverage nan` fixed in #200: a value that clears a type check and vanishes through a comparison.

## 2. The injected judge had no deadline — Minor

`AdvisoryJudge.score` wraps the call in `except Exception`, which turns a judge that *raises* into UNKNOWN. A judge that never returns is not an exception, and was caught by nothing — one unresponsive out-of-tree adapter stopped the whole suite, with no verdict and no message.

### The obvious fix is wrong, and it is worth saying why

A `ThreadPoolExecutor` returns promptly, but its threads are non-daemon and the interpreter joins them on the way out. The hang moves from the run to process shutdown — harder to attribute, no better. Measured on the way through:

```
_call_judge -> TimeoutError за 1.0s
дошли до конца скрипта за 1.0 s
exit=124   ← process never exited
```

A daemon `threading.Thread` has neither problem:

```
TimeoutError за 1.0s
конец скрипта за 1.0 s
exit=0
```

The honest limitation: Python cannot kill the worker, so a hung judge keeps its thread until the process ends. What this buys is that the *suite* is no longer hostage to it — a degraded report instead of no report. There is a test asserting the worker is a daemon, so the executor version cannot come back by accident.

Same shape as the decision-log append lock fixed in #200: a guard that only catches raising, protecting a call that can block.

## 3. Any truthy scalar counted as a declared output list — Minor

`_check_every_phase_declares_output` was a bare truthiness test, so `outputs = true`, `outputs = 1` and `outputs = "step.out"` all earned the same structural credit as a correct list.

The bare string is the one that matters: it looks right, it is the likeliest mistake, and accepting it hides exactly the authoring error the check exists to surface. Now requires a list or tuple carrying at least one non-empty string.

## 4. Tilde-fenced code blocks were not stripped — Minor

`_prose_headings` toggled on triple backticks only. CommonMark also fences with `~~~`, so a `## Rules` heading inside a tilde-fenced sample read as a real section and could satisfy `_check_required_sections` — deterministic structural credit for a code sample.

Both characters are recognised now, and a fence closes on the character that opened it — so a tilde line inside a backtick block stays content, rather than becoming a new way for a sample to be read as prose.

## Verification

**Fifteen new tests.** Each confirmed to fail with its fix reverted:

| | reverted |
|---|---|
| NaN / ±inf baseline | 4 failed |
| judge deadline | failed (bounded in the test, so it cannot wedge the suite) |
| truthy scalar outputs | 3 failed |
| tilde fence | 1 failed |

And each paired with a test pinning what must **not** change — a real regression is still flagged beside a NaN one, a judge that answers in time is untouched, a judge that raises is still reported as raising, a real output list still passes, and real sections after a tilde fence are still found.

Full suite: **6236 passed, 0 failed.** `cfs validate` 0 errors, spec-coverage gate passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid or non-finite baseline and aggregate compliance values are now excluded from evaluation comparisons.
  - Evaluation runs no longer remain blocked indefinitely when an injected judge fails to respond; timed-out judgments are reported as unknown and excluded from calibration metrics.
  - Structural checks now correctly ignore headings inside fenced code blocks, including tilde fences and invalid closing fences.
  - Phase output declarations must now be non-empty lists containing only valid, non-blank strings.


skills/studio/scripts/studio/utils/eval_judge.py
```diff
@@ -332,7 +377,16 @@ def score(self, run: Optional[RunArtifacts], scenario: Scenario) -> ScorerResult
...
+            # degradation as a raise, said differently so the cause is legible.
+            logger.warning("eval: judge_fn did not return within %.0fs on scenario %s",
+                           _JUDGE_TIMEOUT_SECONDS, scenario.id)
+            return self._result(
+                VERDICT_UNKNOWN,
+                [f"judge_fn did not return within {_JUDGE_TIMEOUT_SECONDS:.0f}s"],
+                "judge timeout", scenario)
         # A misbehaving injected judge must degrade to UNKNOWN, never sink the run.
         except Exception as exc:  # pylint: disable=broad-except
             logger.warning("eval: judge_fn raised on scenario %s: %s", scenario.id, exc)
```
ainetx, line 394: ### Timed-out judge_fn is scored as a real mismatch in calibrate(), unlike a crashing judge_fn

**Severity:** Major

**Problem**
AdvisoryJudge.score() reports a raised exception with coverage="judge error" and a timeout with coverage="judge timeout" (two different literal strings). _score_case() only excludes a case from calibration when "judge error" appears in coverage, so the newly-added timeout path is never recognized as an operational failure.

<details>

**How to reproduce**
1. Call calibrate(cases, judge_fn=a_judge_that_always_sleeps_past__JUDGE_TIMEOUT_SECONDS, runs=3). 2. Each score() call raises TimeoutError internally, caught and converted to coverage="judge timeout", verdict=UNKNOWN. 3. _score_case() computes errored = any("judge error" in coverage) -> False, since coverage is "judge timeout". 4. The case is scored: majority=UNKNOWN compared against gold's expected verdict (rarely UNKNOWN) -> matched=False, consistency computed normally, and it is counted in `scored`/`accuracy` instead of `excluded`.

**Expected behavior**
A judge_fn that times out should be excluded from accuracy/consistency and appear in `excluded`, exactly like a crashing judge_fn, per calibrate()'s own documented contract.

**Actual behavior**
The timed-out case is silently counted as a real UNKNOWN verdict mismatch, deflating reported accuracy/consistency with no error surfaced to the caller.

```text
judge_fn hangs -> _call_judge times out -> score() coverage="judge timeout" -> _score_case(): "judge error" not in "judge timeout" -> errored=False -> counted in `scored` -> compared to gold -> accuracy deflated silently
```

**Impact**
cfs eval --calibrate reports understated judge accuracy/consistency whenever the injected model is slow, not just when it errors -- undermining the trust signal the calibration numbers exist to provide, exactly the failure mode the crash-handling code was written to prevent.

**Suggested correction**
Match both operational-failure coverage strings in _score_case (e.g. `any(s in (result.coverage or "") for s in ("judge error", "judge timeout") for result in results)`), or better, use a shared marker/enum for both failure modes instead of two independently-typed literal strings.

**How to verify**
Add a test that calls calibrate() with a judge_fn that always exceeds the (monkeypatched, shortened) timeout and assert the scenario lands in `excluded` with accuracy computed only over remaining scoreable cases (or None if none remain), mirroring test_calibration_excludes_a_crashing_judge.

</details>

skills/studio/scripts/studio/utils/eval_harness.py
```diff
@@ -472,7 +479,7 @@ def diff_reports(report: EvalReport, baseline: Dict[str, object]) -> Dict[str, o
             improved.append({"scenario": scenario_id, "from": before, "to": now})
     for scenario_id, before in prev.items():
         if (scenario_id not in seen and isinstance(before, (int, float))
-                and not isinstance(before, bool)):
+                and not isinstance(before, bool) and math.isfinite(before)):
             # gone from the suite entirely — surfaced, but not a gate-worthy regression.
             no_longer_scoreable.append({"scenario": scenario_id, "from": before})
     baseline_summary = baseline.get("summary", {})
```
ainetx, line 529: ### aggregate_before is not hardened against the same corrupt-baseline values just fixed for per-scenario compliance

**Severity:** Minor

**Problem**
diff_reports() adds a finite-numeric + bool-exclusion guard for per-scenario `before` values explicitly to prevent a corrupt/hand-edited baseline from silently corrupting the regression report, but the very same function's `aggregate_before` assignment a few lines later only checks `isinstance(baseline_summary, dict)`, never validating the type or finiteness of `structural_compliance` itself.

<details>

**How to reproduce**
1. Construct a baseline dict with `summary.structural_compliance = float('nan')` (or `true`, or `"n/a"`). 2. Call diff_reports(report, baseline). 3. The returned dict's `aggregate_before` field is the raw NaN/bool/string value, unguarded, unlike the per-scenario `before` values a few lines above which would have been coerced to None.

**Expected behavior**
aggregate_before should apply the same finite-numeric/bool-exclusion guard as per-scenario `before`, treating a non-finite or non-numeric summary.structural_compliance as 'no prior aggregate value' (None), consistent with the fix's own stated purpose.

**Actual behavior**
A corrupt or hand-edited baseline's summary.structural_compliance value (NaN, inf, bool, or arbitrary type) flows unmodified into the returned `aggregate_before` field of the report dict that CLI consumers read/serialize.

```text
corrupt baseline summary.structural_compliance=NaN -> isinstance(baseline_summary, dict) is True -> aggregate_before = NaN (unguarded) -> emitted in JSON report (json.dumps NaN literal, non-strict-JSON) or displayed as-is
```

**Impact**
A malformed aggregate_before (NaN, bool, or string masquerading as a compliance percentage) can appear in the stable regression-report shape returned to CLI consumers, producing non-strict-JSON output or a nonsensical percentage -- inconsistent with the hardening this same diff just applied one function-scope away, and no test guards against it.

**Suggested correction**
Apply the same guard used for per-scenario `before` (isinstance check + bool exclusion + math.isfinite) to the `structural_compliance` value before assigning it to aggregate_before, e.g. via a small shared helper function.

**How to verify**
Add a test with baseline summary.structural_compliance set to NaN/inf/bool/non-numeric string and assert aggregate_before is None in the returned diff dict, mirroring the new per-scenario NaN/inf tests.

</details>

vasylcf:
Pushed `e56b5467`. All four findings taken — none needed arguing, and each is the same shape as the bugs this PR set out to fix: a check that accepts more than it means to, or a new path not wired into the place that already knew how to handle its kind.

| finding | |
|---|---|
| **Major** — timed-out judge scored as a mismatch in `calibrate()` | fixed |
| **Major** (CodeRabbit) — `any(...)` accepts a malformed output list | fixed |
| **Major** (CodeRabbit) — a fence closes on the wrong delimiter | fixed |
| **Minor** — `aggregate_before` unguarded | fixed |

## The timeout one is the sharpest, and it is mine

`_score_case` already knew how to tell an operational failure from a disagreement — that is what `errored` is for, and its docstring says so. I added a *second* operational failure, `"judge timeout"`, and never registered it. So a judge that hung produced a forced UNKNOWN, which was compared against the gold verdict and counted as the model being wrong.

The exclusion mechanism existed; I just didn't connect the new path to it. Both literals now live in `_JUDGE_OPERATIONAL_FAILURES`, and there is a test asserting that every coverage literal `score()` writes appears in that tuple — the two drifting apart is the whole finding, so it is worth a test rather than care.

## The fence one reintroduced the bug it was fixing

`fence = stripped[0] * 3` stores three characters regardless of the opener, so:

- a `~~~~` block closes on a later `~~~`, and
- ```` ```python ```` counts as a closer, because I never checked the remainder was whitespace.

Either way the `## Rules` after it becomes visible again — which is exactly the defect the tilde change in this PR was meant to remove, one case over. Now: same character, at least as long as the opener, nothing but whitespace after. The chained `startswith` becomes a tuple argument, which is also what SonarCloud flagged.

## `any` → `all`

`["step.out", 1]` and `["step.out", ""]` passed. I chose `any` to be permissive and got it wrong: a list of output names is a list of output names, and one usable entry beside a `1` gives structural credit to precisely the malformed frontmatter the check exists to surface.

## The aggregate

Guarded the per-scenario baselines against NaN, left the aggregate reading `summary["structural_compliance"]` raw. It is the number a reader actually *sees* — NaN beside a real `aggregate_after` reads as a measurement rather than as missing data, and `True` renders as `True`. Both now share one `_usable_baseline` predicate so they cannot disagree about what a number is again.

## Verification

Thirteen new tests, each confirmed to fail with its fix reverted:

```
any → all, fences        3 failed
calibrate exclusion      1 failed
aggregate guard          4 failed
```

Each is paired with one pinning what must **not** change: a judge that answers in time is still scored, a real output list still passes, a `~~~~~` closer for a `~~~` opener still closes, and a usable aggregate still comes through.

Full suite **6251 passed, 0 failed**; `cfs validate` 0 errors; spec-coverage gate passes.

skills/studio/scripts/studio/utils/eval_judge.py
ainetx: **RESOLVED** -- Confirmed fixed as of e56b546707dafd8f1112f5e6dd1af88a4c2079c0 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Current diff already defines _JUDGE_OPERATIONAL_FAILURES = ("judge error", "judge timeout") and _score_case's errored-detection iterates over both marks (`any(mark in... for mark in _JUDGE_OPERATIONAL_FAILURES)`), so a timed-out judge_fn is excluded from accuracy/consistency exactly like a crashing one; test_calibrate_treats_a_timed_out_judge_like_a_crashing_one and test_every_operational_coverage_literal_is_registered directly cover this. The described mismatch was present in an earlier version of this change but is fixed in the diff under review.

</details>

skills/studio/scripts/studio/utils/eval_harness.py
ainetx: **RESOLVED** -- Confirmed fixed as of e56b546707dafd8f1112f5e6dd1af88a4c2079c0 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Confirmed via get_file_diff: this diff adds `_aggregate_baseline()`, which applies the same `_usable_baseline` (isfinite + bool-exclusion) guard to `baseline_summary.get('structural_compliance')` before it becomes `aggregate_before`, and new tests (test_a_corrupt_aggregate_baseline_is_reported_as_missing, test_a_usable_aggregate_baseline_still_comes_through) cover NaN/inf/bool/string/None cases. The gap described in the check is already closed by the diff under review.

</details>

skills/studio/scripts/studio/utils/eval_structural.py
```diff
@@ -385,18 +405,36 @@ def _prose_headings(raw_body: str) -> str:
...
+    structural credit to a code sample.
+
+    A fence closes on its own character, per CommonMark: tildes inside a backtick block are
+    content, not a delimiter.
     """
     front, normalised = _match_frontmatter(raw_body)
     body = normalised[front.end():] if front else normalised
     visible: List[str] = []
-    in_fence = False
+    fence: Optional[Tuple[str, int]] = None
```
ainetx, line 420: ### Duplicated CommonMark fence-matching logic between eval_structural.py and eval_judge.py risks silent re-drift

**Severity:** Minor

**Problem**
eval_judge.py already contains a correct `_fence_delim()`/`_split_sections()` implementation (lines 105-164) that tracks fenced-code state as `(char, run_length)` and applies the CommonMark closer rule (same character, run length >= opener length, nothing but whitespace after). This diff adds a second, textually independent implementation of the exact same rule in eval_structural.py's new `_prose_headings()` (lines 401-438), tracking `(char, opener_length)` and doing the identical run-length comparison, but with its own separate code, own comments, and no import of or reference to eval_judge.py's helper.

<details>

**How to reproduce**
1. Diff both files: eval_judge.py's `_fence_delim` (untouched, pre-existing) and eval_structural.py's new fence-tracking block in `_prose_headings`. 2. Note both handle backtick/tilde fences, closer-length matching, and info-string rejection with separate code. 3. Grep tests/ and eval_structural.py's imports for `_fence_delim`, `_split_sections`, or `eval_judge` — none found.

**Expected behavior**
A single shared fence-matching helper (or at minimum a shared test fixture / equivalence test) used by both modules, so a future CommonMark edge case fix (e.g. tab-indented fences, fence chars beyond backtick/tilde) is applied once and both call sites benefit.

**Actual behavior**
Two independent implementations of the same parsing rule exist in eval_structural.py and eval_judge.py, both needed the same set of fixes in this very PR (proving they already drifted apart once), and there is no shared code or cross-referencing test to prevent the next fix from being applied to only one of them.

```text
eval_judge.py: _fence_delim/_split_sections (fixed independently)
|
X (no shared code, no import)
|
eval_structural.py: _prose_headings (fixed independently, same diff)

=> next CommonMark edge case fixed in one, silently missed in the other
```

**Impact**
Increases maintenance burden and risk of behavioral drift: the next fence-parsing edge case (e.g. tab-indented fence markers, fence characters mixed with leading whitespace beyond current handling) is likely to be patched in only one of the two files, silently reintroducing in the other module the exact class of bug (headings inside fenced code miscounted as real sections/content) that this diff just closed.

**Suggested correction**
Extract a single shared `fence_delim`/fence-state helper into a common module (or eval_harness.py, which both already import) and have both eval_structural.py's `_prose_headings` and eval_judge.py's `_split_sections` use it; alternatively, add a parametrized test that runs the same set of fence fixtures through both implementations and asserts identical visible-line output.

**How to verify**
After extraction, confirm both eval_structural.py and eval_judge.py call the same fence-matching function, and that a shared/parametrized test fails if either implementation's fence-closing behavior diverges from the other on the same input set.

</details>

vasylcf:
Pushed `46140de8`. Two findings resolved on re-verification; the new one is worth naming plainly.

## This PR was the drift

`eval_judge._split_sections` already had a correct CommonMark fence matcher — same `(char, run_length)` state, same closer rule. My fix to `eval_structural._prose_headings` reimplemented it, independently, without noticing.

The defect is not either implementation: it is that there are two. They drift, and the second one existing at all means the next reader cannot tell which is authoritative. This PR created that situation while fixing a bug caused by the same rule being got wrong once.

`fence_delim` and `fence_closes` now live in `eval_harness` — the module both scorers already import — and both delegate. Moved there rather than shared out of `eval_judge` because a structural scorer importing from the judge is the wrong direction: the judge is the specialised one.

The CDSL declaration moved with the code (`inst-judge-fence` → `inst-fence-delim`, under the harness's own section), so the marker and its declaration stay together. `cfs validate` caught the orphan while they were briefly apart — the second time that gate has caught something on this PR.

Full suite **6251 passed, 0 failed**; `cfs validate` 0 errors; spec-coverage gate passes.

skills/studio/scripts/studio/utils/eval_structural.py
ainetx: **RESOLVED** -- Confirmed fixed as of 8746b4a22ade03751ef704ae5732c8a1a08ae9f6 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current diff extracts fence_delim/fence_closes into eval_harness.py and both eval_judge.py and eval_structural.py import and reuse them (confirmed via grep at eval_judge.py:39,144,148 and eval_structural.py:30,423,430), eliminating the duplicate independent implementations the earlier review round flagged.

</details>

ainetx (approved):
Solid follow-up to the #234 review — deduplicates the fence-delimiter logic into a single shared implementation used by both the structural scorer and the judge, hardens baseline-diffing against NaN/inf/bool sneaking in as "numbers," tightens the output-declaration check so malformed frontmatter (bools, ints, bare strings, mixed-type lists) no longer passes as a valid declaration, and adds a timeout around the injected judge call so a hung judge degrades to UNKNOWN instead of stalling the whole suite. Each fix is paired with tests that pin down the actual failure mode (NaN silently disabling the regression gate, a corrupt aggregate rendering as a fake measurement, tilde fences being invisible to one of the two scorers, a non-daemon worker blocking process exit), and the daemon-thread choice for the timeout is well-reasoned and covered. No correctness, security, or test-coverage concerns — approving as is.

Issue #155, opened 2026-09-08
SonarCloud reports 0.0% coverage: coverage.xml paths are relative to the --cov root, not the repo root


SanjeevSolanki:
SonarCloud reports **0.0% Coverage on New Code** on every pull request, including ones whose new lines are fully covered locally. The cause looks like a path-resolution mismatch rather than missing tests, so the coverage signal is currently inert for the whole project.

## What happens

`make test-coverage` generates the report with the package as the coverage root:

```
--cov=skills/studio/scripts/studio --cov-report=xml:coverage.xml
```

so `coverage.xml` records paths **relative to that package**:

```xml
<source>/…/skills/studio/scripts/studio</source>
...
<class filename="utils/pdsl.py" …>
<class filename="commands/pdsl.py" …>
```

`sonar-project.properties` sets `sonar.sources=src,skills,scripts` and `sonar.python.coverage.reportPaths=coverage.xml`. Sonar resolves component paths from the repository root, so for that file it looks for:

```
skills/studio/scripts/studio/utils/pdsl.py
```

The report offers `utils/pdsl.py`. The two never match, so no coverage is attributed to any file.

## Why it has not been noticed

The quality gate has no coverage condition on new code, so `0.0%` renders with a green tick and the gate passes. On a recent PR the reported figure was `0.0%` while the changed file measured 100% of new lines and branches under `make test-coverage` locally.

The practical consequence is that the gate would stay green over a genuinely untested diff — the metric is not measuring anything.

## Possibly related

`.coveragerc` still pins a pre-rebrand source root:

```
[run]
source = skills/cypilot/scripts/cypilot
```

That directory does not exist in the repository. The Makefile passes `--cov` explicitly so local runs still work, but the stale config may matter for any invocation that relies on `.coveragerc` alone.

## To reproduce

1. `make test-coverage`
2. Inspect `coverage.xml` — `<source>` is the package directory, and `class/@filename` values are relative to it
3. Compare against a component path Sonar would use, e.g. `skills/studio/scripts/studio/utils/pdsl.py`

## Note

Not proposing a fix — the coverage invocation and `sonar.python.coverage.reportPaths` are owned by whoever set the CI up, and there is more than one reasonable way to reconcile them. Raising it because a green coverage badge that measures nothing is worth knowing about.

PR #235, opened 2026-09-17
fix(core-infra): make the local quality gates runnable and report the right tree

### LINKED ISSUE #197: make ci: a privacy test false-positives when the container user is root, so local CI cannot pass

### LINKED ISSUE #233: make test: 21 pylint-plugin tests fail locally for a --user pipx install, because the autouse fixture moves $HOME
`make test` fails 21 tests on a clean checkout of `main` for any contributor whose `pipx` is a `pip install --user` install. CI is unaffected, so the suite looks green in the one place it is watched and is red in the place a new contributor first runs it.

## Symptom

```
FAILED tests/test_pylint_output_channels.py::TestOutputChannelsChecker::test_flags_plain_print_to_stdout
FAILED tests/test_pylint_silent_exceptions.py::TestSilentExceptionSwallowedChecker::test_flags_except_pass
... 21 total
E   AssertionError: 'silent-exception-swallowed' not found in ''
```

All 21 are in the four files that shell out to pylint: `test_pylint_error_surfaces.py`, `test_pylint_import_boundaries.py`, `test_pylint_output_channels.py`, `test_pylint_silent_exceptions.py`. The underlying error is:

```
ModuleNotFoundError: No module named 'pipx'
```

## Cause

Those tests invoke the checker through `pipx run --spec pylint pylint` (`_run_pylint`). `tests/conftest.py:90` has an autouse fixture that moves `$HOME`:

```python
monkeypatch.setenv("HOME", str(tmp_path_factory.mktemp("cfs_home")))
```

Python derives the *user* site-packages directory from `$HOME`. With `$HOME` moved, `~/.local/lib/python3.x/site-packages` no longer resolves, and a `--user`-installed `pipx` becomes unimportable — so the `pipx` console script fails before pylint ever runs, the subprocess produces empty output, and each assertion fails on the empty string rather than on anything about the checker.

One line reproduces it without pytest:

```
$ HOME=$(mktemp -d) pipx --version
ModuleNotFoundError: No module named 'pipx'
```

The fixture's own comment explains why it moves `$HOME` rather than patching `_brand_dir` — tests that make `Path.home()` raise need the real seam intact — and that reasoning is sound. The interaction with subprocesses that resolve their own imports from `$HOME` looks unintended.

## Why it is worth fixing rather than documenting

The failure does not name its cause. It surfaces as 21 assertions about missing pylint message IDs, which reads as "the custom checkers are broken" — a plausible enough conclusion that the natural next step is to start debugging the checkers. Nothing in the output mentions `$HOME`, and `make pylint` passes, which makes the checkers look fine and the tests look wrong.

## Suggested fixes

In rough order of preference:

1. Preserve the real `$HOME` for the subprocess in `_run_pylint`, which is the only place that needs it — the isolation the fixture provides is about the brand directory, not about the tool's own imports.
2. Have the four pylint test modules opt out of the `$HOME`-moving fixture.
3. Fail loudly instead of silently: `_assert_pylint_ran` already exists, and could treat empty stdout with a non-zero return as a harness failure naming `$HOME`, so the diagnosis is in the output.

Happy to put up whichever you prefer. Found while separating pre-existing failures from a branch's own during a review pass.

SanjeevSolanki:
Closes #233. Closes #197.

Three things that made the local gates lie about their own results. None changes shipped behaviour; all three change whether a contributor can trust what they see.

## `make test` failed 21 tests on a clean checkout (#233)

Not for everyone — for anyone whose tooling is a `pip install --user` install. The four pylint-plugin families shell out through `pipx`, and the autouse fixture in `conftest.py` moves `$HOME`, which is where Python derives the *user* site-packages directory from. So `pipx` itself became unimportable.

The failure had no fingerprints. The subprocess produced empty output and 21 assertions failed claiming a checker never emitted its message, with `$HOME` mentioned nowhere — a plausible enough story that the natural next step is to start debugging the checkers, while `make pylint` passes and makes them look fine.

One line reproduces it without pytest:

```
$ HOME=$(mktemp -d) pipx --version
ModuleNotFoundError: No module named 'pipx'
```

The fixture is right to move `$HOME` — its own comment explains why, and that reasoning holds. So the fix is a shared `subprocess_env()` that restores the real one for child processes that resolve their own imports, and the four copies of `dict(os.environ)` now go through it.

## `make ci` could not pass at all (#197)

It runs the workflows through `act`, and a privacy test asserts the OS username appears in no `REASON_*` constant — while `REASON_NO_PROJECT_ROOT` legitimately contains the word "root".

Measured rather than assumed, because the reason is narrower than the issue says. A bare container leaves `USER` empty, which would skip the check entirely; it is act's default runner image that sets it:

```
$ docker run --rm catthehacker/ubuntu:act-latest sh -c 'echo $USER'
root
$ docker run --rm ubuntu:22.04          sh -c 'echo $USER'
              (empty)
```

A shared service account is not a personal identifier, and a substring test cannot tell one from ordinary English, so those names are exempt. **Widening a privacy check to make it go green is how a guard quietly retires**, so a second test re-runs the same check under a username no exemption covers and asserts it still fails. Disabling the username check fails that test.

## `.coveragerc` named a tree that no longer exists

`source = skills/cypilot/scripts/cypilot`, gone since the rename — so a bare `coverage run` measured nothing. `make test-coverage` only looked correct because it passes `--cov=` on the command line, which overrides the file. Paths are now recorded relative to the repository root as well, since a report carrying the runner's absolute paths cannot be matched against the repository by anything reading it later.

## Verification

Ten local targets green: `validate`, `self-check`, `check-versions`, `validate-kits`, `test-gates`, `test`, `test-coverage`, `pylint`, `vulture-ci`, `spec-coverage`. **6223 passed** — the first fully green `make test` on this machine, previously 21 red.

Every fix is mutation-checked:

| mutation | result |
|---|---|
| remove the `$HOME` restore | 11 tests fail |
| read `$HOME` at call time instead of import time | the new test fails, and 10 real pylint tests fail |
| drop the generic-account exemption, run under `USER=root` | reproduces #197 exactly |
| disable the username check entirely | the guard test fails |

One note on the second row. The first version of the new test compared against *this test's* `$HOME` and was **vacuous** — the autouse fixture calls `mktemp` per test, so every test gets a different temp home, and pointing `_REAL_HOME` at an unrelated directory left it green. It is anchored on pytest's basetemp now, and the docstring records what the mutation does and does not pin.

## Not in scope

**#155 is not fixed here, and its stated cause is wrong.** The `SonarQube Scan` step is `skipped` on every run — pull requests *and* pushes to `main` — because `SONAR_TOKEN` is unset, so the job has never executed and no coverage report has ever been uploaded. The relative-path problem the issue describes is real, and fixed above, but it is not why coverage reads 0%. That needs a repository secret plus disabling Automatic Analysis (SonarCloud refuses CI-based analysis while it is on), and fork PRs never receive secrets in any case. I will add the evidence to #155.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test subprocess behavior so user-installed tooling remains discoverable in isolated environments.
  - Corrected coverage reporting paths and exclusions for Studio components.
  - Refined privacy checks to avoid false positives for standard system accounts while continuing to detect genuine username leaks.

- **Tests**
  - Added coverage for subprocess environment handling and privacy-leak detection.
  - Updated Pylint integration tests to use consistent subprocess environment setup.


tests/pylint_plugin_fakes.py
```diff
@@ -1,12 +1,39 @@
...
+    """The environment for a subprocess that resolves its own imports.
+
+    Restores the real ``$HOME`` so user-installed tooling stays importable, and applies
+    ``extra`` on top. Use this rather than ``dict(os.environ)`` for any child process that
+    has to find an installed program.
+    """
+    env = dict(os.environ)
+    if _REAL_HOME is not None:
+        env["HOME"] = _REAL_HOME
+    env.update(extra)
```
ainetx, line 34: ### subprocess_env lets caller-supplied extras silently override the restored HOME

**Severity:** Minor

**Problem**
subprocess_env restores `_REAL_HOME` into `env` and then unconditionally applies `env.update(extra)`, so any `extra` key named `HOME` clobbers the restoration with no error or warning.

<details>

**How to reproduce**
1. Call `subprocess_env(HOME='/some/other/dir')`.
2. Observe the returned dict's `HOME` is `/some/other/dir`, not the restored real home.
3. No test or assertion prevents or flags this.

**Expected behavior**
The helper's contract ('restores the real $HOME') should hold regardless of caller-supplied extras, or the override should be explicitly documented/guarded as intentional.

**Actual behavior**
A caller passing HOME as an extra silently defeats the restoration, reintroducing the exact pipx/user-site import failure this helper exists to prevent.

```text
dict(os.environ) -> set HOME=_REAL_HOME -> update(extra) [HOME clobbered here] -> returned env
```

**Impact**
A future caller (even accidentally) reintroduces the opaque pipx ModuleNotFoundError failure this helper was built to eliminate, with no test catching it.

**Suggested correction**
Apply extras first, then force HOME to _REAL_HOME last (or raise if 'HOME' is in extra).

**How to verify**
Add a test asserting subprocess_env(HOME='/tmp/x')['HOME'] equals the restored real HOME, not '/tmp/x'.

</details>

tests/pylint_plugin_fakes.py
```diff
@@ -1,12 +1,39 @@
...
+
+def subprocess_env(**extra: str) -> dict:
+    """The environment for a subprocess that resolves its own imports.
+
+    Restores the real ``$HOME`` so user-installed tooling stays importable, and applies
+    ``extra`` on top. Use this rather than ``dict(os.environ)`` for any child process that
+    has to find an installed program.
+    """
+    env = dict(os.environ)
+    if _REAL_HOME is not None:
```
ainetx, line 32: ### subprocess_env silently no-ops HOME restoration when $HOME was unset at import time

**Severity:** Minor

**Problem**
`_REAL_HOME = os.environ.get('HOME')` can be None; `subprocess_env` only assigns `env['HOME'] = _REAL_HOME` when `_REAL_HOME is not None`, contradicting the docstring's unconditional claim.

<details>

**How to reproduce**
1. Run pytest in an environment where $HOME is unset at collection time.
2. conftest.py's autouse fixture later sets a temp HOME via monkeypatch.
3. Call subprocess_env(); HOME in the result remains the temp directory, not the real one.

**Expected behavior**
The docstring's guarantee should hold, or the function should surface an explicit error/warning when it cannot restore a real HOME.

**Actual behavior**
The function is a silent no-op for HOME in this case, reproducing the original opaque pipx ModuleNotFoundError/empty-output failure with no signal explaining why.

```text
HOME unset at import -> _REAL_HOME=None -> conftest sets temp HOME -> subprocess_env() skips restoration -> child inherits temp HOME -> pipx import fails silently
```

**Impact**
The portability fix does not apply on HOME-less environments, reproducing the exact opaque failure mode this PR set out to eliminate, on a subset of machines.

**Suggested correction**
Raise or warn loudly when _REAL_HOME is None instead of silently skipping the HOME key, or document/test this as an accepted limitation.

**How to verify**
Add a test that monkeypatches _REAL_HOME to None and asserts subprocess_env raises or otherwise flags the missing restoration.

</details>

tests/test_pylint_plugin_fakes.py
```diff
@@ -0,0 +1,104 @@
...
+
+from tests import pylint_plugin_fakes
+from tests.pylint_plugin_fakes import subprocess_env
+
+
+def _user_site(env: dict) -> str:
+    """Ask a real child process where it would look for user-installed packages."""
+    result = subprocess.run(
+        [sys.executable, "-c", "import site; print(site.getusersitepackages())"],
+        env=env, capture_output=True, text=True, check=True,
```
ainetx, line 19: ### _user_site discards captured stdout/stderr on subprocess failure

**Severity:** Minor

**Problem**
_user_site captures stdout/stderr via capture_output=True but relies on check=True for failure signaling; CalledProcessError's default __str__ does not include the captured streams.

<details>

**How to reproduce**
1. Cause the child `python -c 'import site;...'` invocation to fail (e.g. env missing/corrupted).
2. check=True raises CalledProcessError.
3. pytest reports only 'Command... returned non-zero exit status N', with no stdout/stderr shown.

**Expected behavior**
A failure in this diagnostic helper should surface the captured stdout/stderr so environment-sensitive failures are distinguishable.

**Actual behavior**
Only the exit code and command line are shown; the actual interpreter error (often on stderr) is discarded.

```text
child fails -> CalledProcessError(returncode) -> str() omits stdout/stderr -> pytest shows bare 'non-zero exit status'
```

**Impact**
Diagnosing a failure in exactly the test family meant to diagnose environment sensitivity becomes harder than necessary.

**Suggested correction**
Catch CalledProcessError (or use check=False) and re-raise/assert with the captured stdout and stderr included in the message.

**How to verify**
Force a failing child command and confirm the resulting test failure message includes the child's stdout/stderr text.

</details>

tests/pylint_plugin_fakes.py
```diff
@@ -1,12 +1,39 @@
...
+
+
+def subprocess_env(**extra: str) -> dict:
+    """The environment for a subprocess that resolves its own imports.
+
+    Restores the real ``$HOME`` so user-installed tooling stays importable, and applies
+    ``extra`` on top. Use this rather than ``dict(os.environ)`` for any child process that
+    has to find an installed program.
+    """
+    env = dict(os.environ)
```
ainetx, line 31: ### subprocess_env() fixes HOME but not PYTHONNOUSERSITE/PYTHONUSERBASE, which independently control user-site resolution

**Severity:** Minor

**Problem**
subprocess_env() (tests/pylint_plugin_fakes.py:24-35) restores the real $HOME into the child environment specifically so pip-install --user tooling (pipx) stays importable, per its own docstring's rationale that 'Python also derives the user site-packages directory from $HOME'. It passes through the rest of os.environ unchanged, including PYTHONNOUSERSITE (which disables user-site imports entirely, independent of HOME) and PYTHONUSERBASE (which redirects the user-site base directory to an arbitrary path, independent of HOME). A repo-wide search finds zero references to either name anywhere in the codebase, so nothing documents or rules out their presence in an ambient dev/CI environment. The committed regression test in tests/test_pylint_plugin_fakes.py (_user_site, lines 15-25) only calls `site.getusersitepackages()` and asserts on the returned path string -- that function computes the path unconditionally regardless of whether user-site is actually enabled, so the test would still pass (reporting the 'correct' path) even under PYTHONNOUSERSITE=1, while a real pipx subprocess started the same way would still fail to import from user-site.

<details>

**How to reproduce**
1. Set PYTHONNOUSERSITE=1 (or PYTHONUSERBASE=/some/other/path) in the ambient environment before running pytest. 2. Run any of the pylint-family tests (e.g. tests/test_pylint_error_surfaces.py) that call `_run_pylint`, which builds its subprocess environment via `subprocess_env(PYTHONPATH=PYTHONPATH)`. 3. The pipx-invoked pylint subprocess inherits PYTHONNOUSERSITE/PYTHONUSERBASE unchanged (only HOME was fixed), so it cannot resolve pipx's pip-install --user location and dies the same way described in the module's own comment (ModuleNotFoundError, empty output). 4. Meanwhile tests/test_pylint_plugin_fakes.py's own regression test (`test_the_shared_env_restores_the_real_home`) still passes, because `_user_site` only inspects the computed path, not whether user-site is actually enabled for imports.

**Expected behavior**
subprocess_env() should clear or reconcile PYTHONNOUSERSITE and PYTHONUSERBASE (or the accompanying regression test should assert that a real user-installed import succeeds, e.g. via `site.ENABLE_USER_SITE` or an actual `import pipx`-style check) so the fix is robust to the same class of ambient environment variables that govern the exact resolution mechanism it targets.

**Actual behavior**
subprocess_env() only overrides HOME and caller-supplied extras, leaving PYTHONNOUSERSITE/PYTHONUSERBASE unreconciled, and the committed test only checks the resolved path string rather than actual import-time enablement, so the fix's own verification cannot detect this failure mode even though it exercises the identical mechanism (user-site resolution) that HOME was fixed for.

```text
ambient env has PYTHONNOUSERSITE=1
|
v
subprocess_env() copies os.environ, fixes HOME only
|
v
child pipx subprocess: site.ENABLE_USER_SITE=False -> pipx unimportable
|
v
pylint test fails with empty output (same symptom PR fixes for HOME)

meanwhile: _user_site() test calls site.getusersitepackages() -> path is still
computed correctly regardless of ENABLE_USER_SITE -> regression test PASSES anyway
```

**Impact**
On any developer machine or CI runner that sets PYTHONNOUSERSITE or PYTHONUSERBASE ambiently, the pylint-plugin test family can still silently fail with the exact misleading symptom (empty subprocess output misread as 'checker never emitted its message') that this PR was written to fix, and the newly-added regression test would not catch the regression because it only measures the resolved path, not whether user-site imports actually work.

**Suggested correction**
In subprocess_env(), pop/clear PYTHONNOUSERSITE and reset PYTHONUSERBASE to the real user's value (or unset it) alongside the existing HOME restoration; strengthen the _user_site helper/test to assert `site.ENABLE_USER_SITE` is true (or perform a real import of a user-installed package) rather than only comparing the path string.

**How to verify**
Set PYTHONNOUSERSITE=1 in the shell, rerun tests/test_pylint_plugin_fakes.py and the pylint-family tests; after the fix they should still pass (pipx importable), and the strengthened _user_site check should fail before the fix and pass after.

</details>

ainetx (approved):
Nice work tracking down the elusive pipx/$HOME interaction and backing it with a real regression test rather than a dict-comparison stand-in — the coveragerc fix and the `root`-in-CI username exemption are similarly well-reasoned. Review covered correctness of the subprocess-env restoration, the coverage config path/relative-files fix, and the privacy test's username-leak logic, including its own test-for-the-test. Nothing blocking; a few small hardening notes worth a glance on your next pass:

- **`subprocess_env` lets caller-supplied extras silently override the restored HOME** — `env.update(extra)` runs after the HOME restoration, so an `extra` containing a `HOME` key clobbers it with no error. ([comment](https://github.com/constructorfabric/studio/pull/235#discussion_r4037207638))
- **HOME restoration silently no-ops if `$HOME` was unset at import time** — the docstring promises an unconditional restore, but `_REAL_HOME` can be `None`, in which case the assignment is skipped. ([comment](https://github.com/constructorfabric/studio/pull/235#discussion_r4037208374))
- **`subprocess_env` doesn't address `PYTHONNOUSERSITE`/`PYTHONUSERBASE`** — these independently control user-site resolution alongside `HOME`, and the new regression test only checks the computed path string, so it would still pass even if user-site were actually disabled. ([comment](https://github.com/constructorfabric/studio/pull/235#discussion_r4045244465))
- **`_user_site` drops captured stdout/stderr on subprocess failure** — it captures output but relies on `check=True`, whose default exception message doesn't include the captured streams, making failures harder to diagnose. ([comment](https://github.com/constructorfabric/studio/pull/235#discussion_r4037208678))

tests/pylint_plugin_fakes.py
```diff
@@ -1,12 +1,39 @@
...
+    if _REAL_HOME is not None:
+        env["HOME"] = _REAL_HOME
+    env.update(extra)
+    return env
```
ainetx, line 35: ### Pylint-subprocess launch scaffolding remains copy-pasted across four test files after the env fix

**Severity:** Minor

**Problem**
tests/pylint_plugin_fakes.py now exports subprocess_env() specifically to fix a $HOME-related bug shared by all four pylint-family test files, proving that this launcher logic is fragile to inconsistent application. But the surrounding _run_pylint() scaffolding (PYTHONPATH construction, TemporaryDirectory/target-file handling, and the full pipx invocation arg list) is still separately hand-maintained in each of the four files, identical apart from the --enable= value and target path.

<details>

**How to reproduce**
1. Open tests/test_pylint_error_surfaces.py, tests/test_pylint_import_boundaries.py, tests/test_pylint_output_channels.py, tests/test_pylint_silent_exceptions.py. 2. Compare their _run_pylint() bodies: PYTHONPATH build, TemporaryDirectory usage, and pipx run --spec pylint pylint --score=n --disable=all... invocation are byte-for-byte identical except for --enable= and the target filename.

**Expected behavior**
A shared launcher (e.g. a run_pylint(code, enable=..., relative_path=...) helper in pylint_plugin_fakes.py) that all four suites call, so any future environment/invocation fix (e.g. PYTHONNOUSERSITE handling) is applied once.

**Actual behavior**
Each file keeps its own copy of the launcher, so a future fix could again land in only some of the four copies, silently leaving others broken -- the same failure mode subprocess_env() was created to eliminate for the env-building step alone.

```text
subprocess_env() fix -> applied via import in all 4 files (good)
_run_pylint() body -> still hand-copied x4 -> future fix in file A only -> files B/C/D silently regress
```

**Impact**
Increases maintenance cost and reintroduces the exact drift risk category this PR's own extraction was meant to close, for the remainder of the pylint subprocess-launch logic.

**Suggested correction**
Move the common parts of _run_pylint (PYTHONPATH build, TemporaryDirectory/target-write, pipx argument list) into pylint_plugin_fakes.py as a parametrized helper, with only the --enable= value and relative path passed in by each caller.

**How to verify**
After refactor, all four _run_pylint definitions should be thin wrappers (or removed) calling the shared helper, and a change to the shared subprocess invocation should require editing only one place.

</details>

tests/pylint_plugin_fakes.py
SanjeevSolanki: Correct, and fixed — it raises now rather than letting the extra win.

```python
if "HOME" in extra:
    raise ValueError("subprocess_env restores HOME; passing it as an extra would silently undo that. ...")
```

No caller needs it, and the failure a silent override produces is precisely the fingerprint-free one this helper exists to prevent: `pipx` dies on `ModuleNotFoundError`, the subprocess yields empty output, and the assertions fail claiming a checker never emitted its message, with `$HOME` mentioned nowhere. Refusing it where it is visible is cheaper than debugging it there. Mutation-checked: removing the guard fails the new test.

tests/pylint_plugin_fakes.py
SanjeevSolanki: Correct — the docstring promised something the code did not do.

Fixed by making the behaviour match the promise rather than softening the promise: when `$HOME` was unset at import there is nothing to restore, so it is now **removed** from the child's environment instead of leaving the moved temp value in place. Skipping the assignment passed the fixture's `$HOME` straight through, which is the opposite of the intent.

tests/test_pylint_plugin_fakes.py
SanjeevSolanki: Correct. `CalledProcessError.__str__` carries the exit code and not the streams, so a failing probe reported that it failed and nothing about why — in a helper whose whole subject is a subprocess that dies during startup.

`check=False` now, with the raise here:

```python
if result.returncode != 0:
    raise AssertionError(
        f"the probe interpreter exited {result.returncode}\n"
        f"stdout: {result.stdout.strip()!r}\nstderr: {result.stderr.strip()!r}")
```

which is the shape that would have named `ModuleNotFoundError: No module named 'pipx'` on the original failure instead of leaving it to be rediscovered.

tests/pylint_plugin_fakes.py
SanjeevSolanki: This is the best finding on the PR — right about the variables, and right that the test was weak for exactly the reason given. Both confirmed by running it:

```
plain                          -> /home/sanjeev/.local/lib/python3.13/site-packages  ENABLE_USER_SITE=False
PYTHONNOUSERSITE=1             -> /home/sanjeev/.local/lib/python3.13/site-packages  ENABLE_USER_SITE=False
PYTHONUSERBASE=/tmp/elsewhere  -> /tmp/elsewhere/lib/python3.13/site-packages        ENABLE_USER_SITE=False
```

`getusersitepackages()` computes its answer whether or not user-site is enabled, so the assertion held in all three — while a real `pipx` under either would still fail to import. Exactly as you wrote it.

Both names are dropped now, alongside the `$HOME` restore, on the same reasoning: they are redirection variables for the same mechanism, and one list should hold all three rather than the largest of several partial ones.

The test is behavioural and, more to the point, **not vacuous this time** — my first attempt was. It built a local dict, set `PYTHONUSERBASE` in that, and asserted the name was absent from `subprocess_env()`, where it had never been; removing the drop left it green. It now exports both into the ambient environment, asserts the control (the redirect really does move where a child looks), then asserts the helper's environment resolves back under the real home. Removing the drop fails it.

Worth recording that this is the second vacuous assertion in this file found by mutation rather than by reading, and the file now says so in its own docstrings.

tests/pylint_plugin_fakes.py
SanjeevSolanki: Agreed, and done in this push rather than deferred — your framing is the reason. The `$HOME` bug being shared by all four is evidence about the launcher, not just about `$HOME`.

`run_pylint(code, *, enable, relative_path="sample.py")` lives in `pylint_plugin_fakes` and owns the `pipx` argument list, the `PYTHONPATH`, the temp-file handling and the environment. Each of the four modules keeps a two-line `_run_pylint` that supplies only its own `--enable=`, which is the one thing that actually differed.

It paid for itself immediately, which I would not have predicted: the §3b pass over the consolidated diff showed that **none of the four copies passed a timeout**. A hung `pipx` — a cold spec fetch, say — stopped the whole suite rather than failing one test, which reads as CI being broken rather than as a check failing. One launcher took one bound; four copies would have taken four, and on this evidence would have got two.

tests/test_pylint_plugin_fakes.py
```diff
@@ -0,0 +1,160 @@
...
+    def test_the_captured_home_is_the_real_one_not_a_temp_directory(self) -> None:
+        """`_REAL_HOME` is read at import time, which is collection — before any fixture.
+
+        If the module were ever first imported from inside a test body instead, it would
+        capture the moved `$HOME` and the restore below would be a no-op that still looked
+        correct. Anchored to the shared root so *any* test's temp home fails this, not just
+        the one this test happens to have been given.
+        """
+        assert pylint_plugin_fakes._REAL_HOME, "no $HOME was captured at import"
+        assert not pylint_plugin_fakes._REAL_HOME.startswith(str(_temp_home_root())), (
```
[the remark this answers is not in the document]
SanjeevSolanki: Correct — the assertion contradicted behaviour added in this same PR.

`subprocess_env()` removes `HOME` from the child when `_REAL_HOME` is `None`, which is the documented handling for `$HOME` unset at import. The test then asserted `_REAL_HOME` was truthy, so it would have failed on exactly that supported case.

It branches now: with nothing captured it asserts `HOME` is absent from `subprocess_env()` and returns. The two later assertions that would have taken a prefix of `None` are guarded as well — those would have raised a `TypeError` rather than said what went wrong.

Checked that the new branch actually asserts something, because the obvious mutation does not show it: forcing `_REAL_HOME = None` alone leaves all six tests green. Breaking the `HOME` removal *as well* fails two of them, which is what proves the branch has teeth.

tests/pylint_plugin_fakes.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 457f355aab6620f4599ce65c32da6982e69488d9 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current subprocess_env (tests/pylint_plugin_fakes.py:37-67) explicitly raises ValueError if "HOME" in extra before doing anything else, so a caller-supplied HOME extra can no longer silently override the restored real HOME; test_home_cannot_be_passed_as_an_extra in tests/test_pylint_plugin_fakes.py exercises this via pytest.raises(ValueError, match='restores HOME'). The prior rounds' finding was against an earlier version of the diff; the version under review here already contains the fix.
- **Verifier:** tests/pylint_plugin_fakes.py:54-58 now raises ValueError before any assignment if "HOME" in extra, so env.update(extra) can never silently override the restored real HOME; test_home_cannot_be_passed_as_an_extra in tests/test_pylint_plugin_fakes.py exercises this with pytest.raises(ValueError, match='restores HOME'). The prior FINDING targeted an earlier version of this code; the guard is present in the diff under review.

</details>

tests/pylint_plugin_fakes.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 457f355aab6620f4599ce65c32da6982e69488d9 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** subprocess_env now branches explicitly: `if _REAL_HOME is None: env.pop("HOME", None) else: env["HOME"] = _REAL_HOME`, so the None case removes HOME rather than leaving the moved temp value, matching the docstring's updated claim ('it is *removed* from the child's environment'). test_the_captured_home_is_the_real_one_not_a_temp_directory in tests/test_pylint_plugin_fakes.py exercises exactly this branch with `assert "HOME" not in subprocess_env()` when captured is None. The prior finding targeted an earlier version; this diff already contains the fix.
- **Verifier:** tests/pylint_plugin_fakes.py:60-63 branches explicitly: `if _REAL_HOME is None: env.pop("HOME", None) else: env["HOME"] = _REAL_HOME`, matching the updated docstring's claim that HOME is removed (not left as the moved temp value) when unset at import. test_the_captured_home_is_the_real_one_not_a_temp_directory in tests/test_pylint_plugin_fakes.py asserts `"HOME" not in subprocess_env()` for exactly this branch when captured is None. The docstring's claim is now honored by the code.

</details>

tests/pylint_plugin_fakes.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 457f355aab6620f4599ce65c32da6982e69488d9 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** subprocess_env now defines _USER_SITE_REDIRECTS = ("PYTHONNOUSERSITE", "PYTHONUSERBASE") and pops both from the environment before returning; test_a_redirected_user_site_base_does_not_survive in tests/test_pylint_plugin_fakes.py sets both ambient env vars, confirms the control (a real child's user-site path moves under the redirect), then confirms subprocess_env()'s output resolves back to the real home behaviorally via _user_site (a real subprocess), not just a dict-key check. This matches the exact fix and behavioral test described in the discussion thread and is present in the diff under review.
- **Verifier:** tests/pylint_plugin_fakes.py:26-34,64-65 defines _USER_SITE_REDIRECTS = ("PYTHONNOUSERSITE", "PYTHONUSERBASE") and pops both from env before returning; test_a_redirected_user_site_base_does_not_survive in tests/test_pylint_plugin_fakes.py sets both via monkeypatch.setenv in the ambient environment, confirms the control (a real subprocess's user-site path moves under the redirect via _user_site), then confirms subprocess_env()'s output both lacks the keys and resolves the real child's user-site path back to the real home -- a behavioral, non-vacuous check per the discussion thread.

</details>

tests/test_pylint_plugin_fakes.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 457f355aab6620f4599ce65c32da6982e69488d9 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** _user_site in tests/test_pylint_plugin_fakes.py now uses check=False and, on nonzero returncode, raises AssertionError including the return code plus repr'd stdout/stderr, so a failing probe surfaces exactly the diagnostic context (e.g. ModuleNotFoundError text) that a bare CalledProcessError would have discarded. This is the exact fix described in the discussion and is present in the current diff.
- **Verifier:** _user_site in tests/test_pylint_plugin_fakes.py:19-30 now uses check=False and explicitly raises AssertionError with the return code plus repr'd stdout/stderr when returncode != 0, so a failing probe surfaces exactly the diagnostic context (e.g. ModuleNotFoundError text) that a bare CalledProcessError would have discarded.

</details>

tests/pylint_plugin_fakes.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 457f355aab6620f4599ce65c32da6982e69488d9 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** run_pylint(code, *, enable, relative_path="sample.py") is now defined once in tests/pylint_plugin_fakes.py, owning PYTHONPATH construction, the pipx argument list, temp-file handling, and a timeout; all four family files (test_pylint_error_surfaces.py, test_pylint_import_boundaries.py, test_pylint_output_channels.py, test_pylint_silent_exceptions.py) import it and reduce their own _run_pylint to a one-line wrapper supplying only --enable=. Verified directly via grep across tests/ that all four files reference run_pylint. This matches the consolidation described in the discussion thread and is present in the diff under review.
- **Verifier:** run_pylint(code, *, enable, relative_path="sample.py") is now defined once in tests/pylint_plugin_fakes.py owning PYTHONPATH construction, the pipx argument list, temp-file handling, and a 600s timeout; all four family files (test_pylint_error_surfaces.py, test_pylint_import_boundaries.py, test_pylint_output_channels.py, test_pylint_silent_exceptions.py) import run_pylint and reduce their own _run_pylint to a one-line wrapper supplying only --enable=, confirmed by reading each file's diff/current content.

</details>

tests/test_pylint_plugin_fakes.py
```diff
@@ -0,0 +1,173 @@
+"""Tests for the shared harness the pylint-plugin test families run through."""
```
ainetx, line 1: ### New harness test suite never directly unit-tests run_pylint(), only subprocess_env()

**Severity:** Minor

**Problem**
tests/test_pylint_plugin_fakes.py's module docstring claims to test 'the shared harness the pylint-plugin test families run through,' but every test in TestASubprocessCanStillFindUserInstalledTools imports and calls only `subprocess_env` (and `pylint_plugin_fakes._REAL_HOME`). `run_pylint` -- the function that builds the temp file via `textwrap.dedent`, constructs the `pipx run --spec pylint pylint... --enable=` argv, sets `cwd=REPO_ROOT`, and applies `timeout=600` -- is never imported or called in this file.

<details>

**How to reproduce**
1. Open tests/test_pylint_plugin_fakes.py.
2. Note the only import from pylint_plugin_fakes is `subprocess_env` (plus the module itself for `_REAL_HOME`).
3. grep for `run_pylint` in the file: no matches.
4. `run_pylint`'s own argument-construction and temp-file logic is thus only reached transitively, when the four consumer test files (test_pylint_error_surfaces.py etc.) invoke it with a real pipx/pylint subprocess.

**Expected behavior**
A dedicated, fast unit test that calls run_pylint(...) directly and asserts on its constructed argv (e.g. via monkeypatching subprocess.run), its timeout value, or its temp-file/relative_path/textwrap.dedent handling, so a bug specific to the launcher (e.g. broken relative_path subdirectory handling, an argv ordering mistake, or a lost timeout under a future refactor) fails fast and locally rather than surfacing as unexplained failures scattered across four consumer files.

**Actual behavior**
run_pylint's internal logic is validated only indirectly, by the ~20+ integration-style tests in the four consumer files that happen to invoke real pylint checkers through it -- which is exactly the diagnostic-difficulty pattern this PR's own commentary complains about for the pre-consolidation four-copies situation, just one level up the stack.

```text
run_pylint() [argv/temp-file/timeout logic]
|-- exercised only via --> 4 consumer test files (real pipx+pylint subprocess)
`-- NOT exercised by --> test_pylint_plugin_fakes.py (only calls subprocess_env)
```

**Impact**
Low: a defect in run_pylint's own construction logic would still likely be caught (as failures in the four consumer files), just with less direct, less immediately diagnostic error output -- not a silent regression, but a maintainability/debuggability gap in the new consolidated harness.

**Suggested correction**
Add a test in test_pylint_plugin_fakes.py that calls run_pylint directly (optionally monkeypatching subprocess.run to capture argv) and asserts on the constructed command list, timeout, and relative_path/textwrap.dedent behavior for at least the subdirectory relative_path case used by test_pylint_import_boundaries.py.

**How to verify**
After adding such a test, intentionally break run_pylint (e.g. swap argument order or drop the timeout) and confirm the new unit test fails immediately and specifically, rather than only failing downstream in one of the four consumer files.

</details>

tests/test_change_summary_core.py
```diff
@@ -573,8 +588,24 @@ def test_no_reason_string_leaks_a_path_home_or_username(self):
             value = getattr(cs, name)
             assert os.sep not in value, f"{name} contains a path separator"
             assert home not in value
-            if user:
-                assert user not in value
+            if user and user.lower() not in self._GENERIC_ACCOUNTS:
```
ainetx, line 597: ### Generic-account allowlist can mask a real developer's own username leak

**Severity:** Minor

**Problem**
_GENERIC_ACCOUNTS = frozenset({'root','user','admin','runner','build','nobody'}) is checked via plain case-insensitive string membership against $USER/$USERNAME with no signal distinguishing a CI/container default account from a real person's actual account name. The new companion test only exercises a username outside the exemption set, so it cannot detect that the exemption itself is unbounded.

<details>

**How to reproduce**
1. A developer's real machine or CI service account is literally named 'admin', 'build', 'user', or similar (plausible: default macOS admin account, common corporate service accounts). 2. A future change accidentally introduces a REASON_ constant containing that literal username (a real personal-identifier leak). 3. test_no_reason_string_leaks_a_path_home_or_username runs under that account; user.lower() is in _GENERIC_ACCOUNTS, so the `if user and user.lower() not in self._GENERIC_ACCOUNTS` branch is skipped and the assertion never runs. 4. The test passes despite the real leak.

**Expected behavior**
The exemption should only suppress the check for genuinely non-personal, environment-detected contexts (e.g., detect act/CI via $CI, $ACT, or similar), not for any developer whose real account name happens to coincide with one of six common words.

**Actual behavior**
The exemption is a flat string allowlist applied regardless of whether the environment is actually CI/container or a real developer machine, silently disabling the privacy guard for that population.

```text
USER=admin (real dev account) -> user.lower() in _GENERIC_ACCOUNTS -> assertion skipped -> leaked REASON_ string containing 'admin' -> test still PASSES
```

**Impact**
Weakens (does not eliminate) the privacy regression guard for developers/CI accounts whose name coincides with a common generic word; failure mode is silent (test passes) rather than loud, so a real leak in that population would go undetected by this specific test.

**Suggested correction**
Restrict the exemption to a detected CI/container context (e.g., os.environ.get('CI') or a marker specific to `act`) rather than to a static list of common account names, or additionally require the leaked value to also appear as a full word rather than any generic-account substring only when a CI marker is present.

**How to verify**
Add a test that sets USER to an exempted name (e.g., 'admin') outside of any CI marker and injects a leaking REASON_ string; the check should still fail if the exemption is properly scoped to CI-only.

</details>

tests/test_pylint_plugin_fakes.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1692a1c241006df90c9132d0494d830c3ef33dcd -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Re-checked the current diff directly: tests/test_pylint_plugin_fakes.py now imports run_pylint (line 15) and defines TestTheSharedLauncher, whose two tests call run_pylint(...) directly with subprocess.run mocked, asserting on the constructed argv (pipx/--spec/pylint/--enable=/target path), timeout=600, cwd=REPO_ROOT, check=False, the env's PYTHONPATH/absence of PYTHONUSERBASE, and the textwrap.dedent handling of the written source file. This is exactly the gap the prior finding identified, and it is already fixed in the code under review.

</details>

tests/test_change_summary_core.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1692a1c241006df90c9132d0494d830c3ef33dcd -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Re-checked the current diff directly: _GENERIC_ACCOUNTS is now frozenset({"root"}) only, not the six-name set the check's premise describes, with an added test (test_the_exemption_covers_only_what_was_measured) that pins the set to exactly {"root"} and fails if it grows. The comment explicitly reasons through and rejects the six-name version for the exact false-negative risk this check raises, narrowing to the one measured, reproduced case (act's root-default container). The residual risk -- a real developer whose actual account is literally 'root' -- is an edge case explicitly acknowledged and deliberately traded off in the diff's own commentary, not an unbounded or undocumented gap.

</details>

SanjeevSolanki: Correct, and this was me breaking the repository's own rule. Only `root` was ever measured — `user`, `admin`, `runner`, `build` and `nobody` were speculation, added because they sounded like service accounts.

Your point that membership cannot distinguish a container default from a real person is the one that decides it: a developer whose username happens to be `build` would have had this guard switched off for them, invisibly and permanently. The rule here is to scope a fix to observed behaviour, and five of the six were not observed anywhere.

Narrowed to `frozenset({"root"})`, with the command that observed it kept in the comment.

Two assertions rather than one, because the set alone is not the claim — you were right that the companion test could not detect the exemption being unbounded:

- the exemption is exactly `{"root"}`, so re-widening it fails;
- and a generic-sounding name *outside* it is still caught, demonstrated on a constant that genuinely leaks it.

Re-adding `build` to the set now fails the first; the second is what proves the guard is still live for everyone else.

tests/test_pylint_plugin_fakes.py
SanjeevSolanki: Correct — the file's own docstring claimed more than it delivered. Every test called `subprocess_env`; the launcher that builds the argv, the working directory and the timeout was never imported.

Two tests now, with the subprocess intercepted rather than run: what is asserted is how the call is *composed*, and that it works end to end is what the four pylint families already prove forty-two times a run. Covered: the `pipx` argv and `--enable=`, the target path, `cwd`, `check=False`, the environment coming from the helper rather than a raw `os.environ`, that `PYTHONUSERBASE` is gone, that the source is dedented, and the `timeout=600` — which is the assertion that matters most, since none of the four copies this replaced carried one.

One of those assertions was initially worthless and mutation caught it: `cwd == REPO_ROOT` passes either way when pytest runs from the repository root, because `Path.cwd()` is the same value. The test changes directory first now, and swapping `REPO_ROOT` for `Path.cwd()` fails it.

ainetx: @SanjeevSolanki **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

tests/test_change_summary_core.py
ainetx: @SanjeevSolanki **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

ainetx (approved):
Solid cleanup. The `.coveragerc` fix restores real coverage measurement (the old `source` path predated the rename, so a bare `coverage run` was silently measuring nothing) and adding `relative_files` makes reports portable. The new `subprocess_env`/`run_pylint` helper in `pylint_plugin_fakes.py` consolidates four duplicated subprocess launchers into one, fixes the `$HOME`/`PYTHONUSERBASE`/`PYTHONNOUSERSITE` issue that broke `pipx`-based subprocesses under the temp-home test fixture, and adds a sensible timeout so a hung `pipx` fails one test instead of the whole suite. The privacy test's generic-account exemption is narrowed to only the one name actually observed (`root`), with tests pinning that it stays narrow and still catches real leaks. New tests thoroughly exercise the shared harness itself. Nothing blocking here — approved.

PR #237, opened 2026-09-17
fix(semantic): keep paths out of log records; settle the CPT kind grammar

### LINKED ISSUE #228: Re-verify two rewritten review findings: the CPT "feature" kind, and a path leak that moved to a debug log
Group 4 of the review findings left open on #192. Two comments the reviewer re-verified and rewrote rather than resolved, so what they claim now is **not** what they claimed when first raised. Both need reading fresh before anything is changed.

## 1. `architecture/DECOMPOSITION.md:918` — the `feature` kind segment

**Originally**: `cpt-studio-feature-eval-harness` and `cpt-studio-feature-artifact-quality` use `feature` in the kind position, which the grammar restricts to `flow|algo|dod|state|adr`.

I assessed that as a false positive during the #200 work, on this evidence:

```
56 unique  cpt-*-feature-*  identifiers across the repository
18 of them declared in DECOMPOSITION.md
```

`cpt-*-feature-*` is the feature-level identifier — a different position from the `kind` in a code marker (`@cpt-algo:`, `@cpt-flow:`), and the marker regex `(?P<id>cpt-[a-z0-9][a-z0-9-]+)` accepts it. Renaming two of fifty-six would break the convention, not restore it. `cfs validate` passes with 0 errors.

**Now**: the reviewer restates the grammar as `cpt-<project>-<kind>-<domain>-<slug>` — a five-segment form, where my reading assumed four. If that is the documented grammar, my count does not refute the finding; it might mean fifty-six identifiers are wrong rather than two.

**What to do**: find where the grammar is actually specified — not inferred from the marker regex, which is what I did — and settle it. Either the finding is right and this is a large rename, or the convention is right and the grammar document needs to say so. Both outcomes are worth having written down; right now two readings coexist.

## 2. `semantic_coverage.py:73` — the leak moved

**Originally**: `_relative_posix` returned `path.as_posix()` when `project_root is None`, putting an absolute path into the report and the judge prompt. Fixed in #200 — it now returns the bare filename.

**Now**: a *different* line in the same function. When a file resolves outside the project root, the function correctly returns a relative path but logs the raw absolute one:

```python
logger.debug("semantic: %s is outside project_root; using a relative path", path)
```

The docstring promises "never absolute ... so no local path leaks into the report or the out-of-tree judge prompt". A debug log is neither of those, so this is narrower than the original — but it is the same username and directory layout, and the module already has `_redact` for exactly this.

Worth noting the pattern: this is the third instance of an exception or a path reaching a log unredacted in this module family. #200 fixed one in `decision_log._restrict_to_owner` after a test caught it. A rule — never interpolate a raw path or exception into a log record here — would be cheaper than finding them one at a time.

**What to do**: redact it. Small, and uncontested.

## Scope

- [ ] Locate the authoritative CPT ID grammar and settle whether `feature` is valid
- [ ] If invalid: scope the rename across all 56 identifiers, and fix `cfs validate` for not catching it
- [ ] If valid: correct the grammar statement so the next reviewer reaches the same conclusion
- [ ] Redact the path in the `semantic_coverage` debug log
- [ ] Consider a lint or convention preventing raw paths/exceptions in log records in this module family

## Context

- #192 — the release PR these were raised on
- #200 — where the original versions of both were addressed

vasylcf:
Closes #228. The last of the four groups from the #192 review.

Both items are comments the reviewer **re-verified and rewrote** rather than resolved, so what they claim now is not what they claimed when raised. Both needed reading fresh.

## 1. The `feature` kind — confirmed a false positive, and the spec now says why

The finding says `cpt-studio-feature-eval-harness` and `cpt-studio-feature-artifact-quality` put `feature` in the kind position, which "the required grammar" restricts to `flow|algo|dod|state|adr`. On re-verification it restated that grammar as `cpt-<project>-<kind>-<domain>-<slug>`.

**Neither matches this repository.** `architecture/specs/traceability.md` is the authority:

| | what the spec says |
|---|---|
| format | `cpt-{hierarchy-prefix}-{kind}-{slug}` — a **variable-length** prefix, not a fixed five-segment form |
| `{kind}` | "actor, cap, fr, nfr, comp, flow, algo, state, req, **etc.**" — an explicitly open list |
| full regex | `cpt-[a-z0-9][a-z0-9-]+` — constrains the kind **not at all** |

No code enforces a closed set either: the marker regex is `(?P<kind>[a-z][a-z0-9-]*)`, the id regex is the one above.

Counted across the repository, the two vocabularies actually in play:

```
code markers   algo 176, flow 90, dod 59, state 17, template 11, req 7,
               scope 4, example 4, test 1, ref 1, actor 1

artifact ids   algo 366, fr 258, actor 226, feature 134, component 127,
               flow 104, dod 90, principle 87, adr 86, constraint 41, ...
```

`flow|algo|dod|state|adr` is **neither list** — it is the top of the first plus one entry from the second. Two vocabularies, two positions, read as one rule. `feature` is the fourth most common artifact kind, in 134 ids, and `cfs validate` passes with 0 errors.

**So: no rename.** But the finding did expose something real — `"etc."` let two readings coexist, which is a defect in the spec even though the code was right. The kind paragraph now states the vocabulary is open, lists the kinds Studio actually uses, and says plainly that code-marker kinds are not the permitted list for artifact ids, with a pointer to #192 so the next reviewer reaches the same conclusion without re-deriving it.

## 2. The path leak moved — and there were three of them

#200 fixed `_relative_posix`'s **return value**. This finding is a different line in the same function:

```python
logger.debug("semantic: %s is outside project_root; using a relative path", path)
```

The raw absolute path — username and directory layout — went into the log instead of the report. A debug record is a narrower audience than a judge prompt, not a different rule.

**Two siblings in the same module did the same thing** and are fixed with it: the duplicate-definition warning logged two absolute artifact paths, the unparseable-file warning logged an absolute code path. Both had `project_root` in scope and now go through `_relative_posix` — which also reads better.

### A second layer the test caught

After the first fix, the unparseable-file test still failed:

```
assert '/tmp/pytest-.../test_an_unparseable_file_warni0' not in logged
  is contained here:
  'path': '/tmp/.../bad.py', 'code': 'marker-begin-no-end', 'location': '/tmp/.../bad.py:1'
```

`errs` is a list of finding dicts, each carrying its own absolute `path` and `location`, so interpolating the structures put back the path the line had just been made to drop. It now logs the error **codes** — the file is already named in the message.

This is the point #228 makes about finding these one at a time: the same leak had three sites and one of them had two layers.

## Verification

Three new tests, each confirmed to fail with the fixes reverted (3 failed / 26 passed). Full suite: **6225 passed, 0 failed.** `cfs validate` 0 errors; spec-coverage gate passes.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that artifact ID kinds use an extensible vocabulary, with examples of currently supported kinds and guidance on interpreting code markers.

* **Bug Fixes**
  * Improved diagnostic messages by replacing absolute file paths with project-relative paths.
  * Simplified unparseable-file warnings to show relevant error codes without exposing full internal details.

* **Tests**
  * Added coverage confirming that log records do not disclose absolute project paths.


skills/studio/scripts/studio/utils/semantic_coverage.py
```diff
@@ -74,7 +81,12 @@ def _relative_posix(path: Path, project_root: Optional[Path]) -> str:
...
-        logger.debug("semantic: %s is outside project_root; using a relative path", path)
+        # The *name*, not the path. The function's first sentence promises no local path
+        # leaks, and #200 made the return value keep that promise -- this line still
+        # interpolated the raw absolute path, so the username and directory layout went
+        # into the log instead of the report. A debug record is a narrower audience than
+        # a judge prompt, not a different rule.
+        logger.debug("semantic: %s is outside project_root; using a relative path", path.name)
     try:
         return Path(os.path.relpath(resolved, root)).as_posix()
     except ValueError:
```
ainetx, line 92: ### Duplicate cpt-id warning can render two different colliding files with the identical bare filename

**Severity:** Minor

**Problem**
_definition_map's new duplicate-definition warning formats both colliding paths via _relative_posix(path, ctx.project_root). For a remote-source artifact (added by _collect_remote_artifacts under cross_repo/resolve_remote_ids, both default True) whose path is outside ctx.project_root, relative_to() raises ValueError and the code falls back to os.path.relpath(); if that itself raises ValueError (Windows cross-drive paths, per the function's own docstring and the pre-existing test_pairings_survive_a_relpath_failure), _relative_posix returns just path.name. Two different same-named files from two different remote sources defining the same cpt id would both collapse to that identical bare filename in the warning.

<details>

**How to reproduce**
1. Configure a workspace with cross_repo=True, resolve_remote_ids=True (defaults). 2. On Windows, have two remote sources on different drives, each containing a file named e.g. `algo.md` (different full paths, different drives). 3. Both files define the same cpt id. 4. Run the semantic coverage pass; _definition_map logs the duplicate warning, calling _relative_posix on both paths against ctx.project_root (on a third drive or otherwise unrelated), triggering the relpath ValueError fallback for both.

**Expected behavior**
The warning should let a reader distinguish which two files actually collided, even when both are outside project_root.

**Actual behavior**
Both colliding paths degrade to the same bare filename (e.g. 'algo.md' and 'algo.md'), making the diagnostic message uninformative for locating the actual conflicting files.

```text
collect_artifacts_to_scan (cross_repo=True) -> remote source A:.../driveX/repoA/algo.md
-> remote source B:.../driveY/repoB/algo.md
_definition_map detects duplicate id -> _relative_posix(A, project_root) -> relative_to fails -> relpath fails (cross-drive) -> 'algo.md'
-> _relative_posix(B, project_root) -> same failure -> 'algo.md'
Warning: "cpt id X defined in both algo.md and algo.md" -- indistinguishable
```

**Impact**
In a genuine cross-repo duplicate-id conflict -- the exact case cross-repo resolution exists to catch -- the diagnostic meant to help a user find and fix the conflict becomes ambiguous, forcing manual investigation across all remote sources to find the real colliding files.

**Suggested correction**
When both relative_to and relpath fail, include a disambiguating source name/tag (e.g. the workspace source name from path_to_source) alongside the bare filename, or fall back to a source-qualified relative path instead of just path.name.

**How to verify**
Add a test simulating two out-of-root paths with different (mocked) source tags that both hit the relpath ValueError branch, and assert the resulting duplicate-warning message still distinguishes the two files.

</details>

ainetx (approved):
Nice cleanup — the debug/warning logs now respect the same project-relative promise `_relative_posix` already made for reports, and the new tests pin that down well. Docs clarification on the `{kind}` vocabulary vs. code marker kinds is a helpful disambiguation too. One minor edge case worth a glance but not blocking:

- **Duplicate cpt-id warning can render two different colliding files with the identical bare filename** — when a colliding path falls outside `project_root` (e.g. a remote-source artifact) and even `os.path.relpath` fails (cross-drive paths), `_relative_posix` falls back to just the filename, so two distinct same-named files from different remote sources could produce an ambiguous warning that looks like a self-collision. ([discussion](https://github.com/constructorfabric/studio/pull/237#discussion_r4037469083))

PR #239, opened 2026-09-18
fix(cf-ux): default the codex pilot to a model accounts still have

vasylcf:
## Problem

`gpt-5.4-mini` has been withdrawn. Every codex case in the prompt pilot errors out:

```
ERROR: {"status":400,"message":"The 'gpt-5.4-mini' model is not supported
when using Codex with a ChatGPT account."}
```

From a clean clone that is **8 of 16 cases red**, for a reason that has nothing to do with the skill under test — and the failure surfaces as a truncated JSON blob inside a promptfoo table cell, so it reads like a harness bug rather than a stale constant.

## Change

Default `CF_UX_CODEX_MODEL` to `gpt-5.6-sol`, the cheapest slug `codex` currently reports for this account type. The full entitled set today: `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-6-astra`, `gpt-5.5`.

The comment now states the slug is **dated, not permanent**, and says where to re-pick from, so the next withdrawal costs a lookup rather than a debugging session. `CF_UX_CODEX_MODEL` is untouched and still overrides.

`DEFAULT_EFFORT="low"` stays valid — `gpt-5.6-sol` lists `low` among its supported reasoning levels.

Two docs naming the same dead default are corrected alongside: `CONTRIBUTING.md` and the header comment in `promptfooconfig.yaml`.

## Result

Measured on this branch — `make test-prompts` went from **0 passed / 16 errors** to **11 passed / 0 failed / 5 errors**, 1m30s, $0.72.

All 8 codex cases now pass. The remaining 5 are claude-side and are genuine pilot findings, not environment: 4 × the `cf` skill was never invoked, 1 × `cf-documenting-review` was invoked directly, bypassing the `cf` router. Those are being handled separately.

## Deliberately out of scope

`skills/studio/scripts/studio/commands/agents.py` maps the cf model tiers onto `gpt-5.4` and `gpt-5.4-mini` too, so the product's own model matrix names withdrawn models. That has a spec and snapshot tests behind it and deserves its own change — flagged here, not touched.


ainetx:
### Production agent model matrix still bakes in the withdrawn gpt-5.4-mini slug for Codex/Cursor+OpenAI

**Severity:** Major

**Problem**
This PR fixes the test pilot's default Codex model in codex_provider.py after discovering gpt-5.4-mini is withdrawn and rejected by ChatGPT-account Codex sessions with the runtime error "The 'gpt-5.4-mini' model is not supported when using Codex with a ChatGPT account." However, skills/studio/scripts/studio/commands/agents.py's _MODEL_MATRIX -- the production resolver that bakes model slugs into real generated agent configs for end users (resolved via _MODEL_MATRIX.get((tool, provider)) at line 992) -- still hard-codes 'gpt-5.4-mini' as the cf:tier:cheap base model for both ('codex','openai') (line 828) and ('cursor','openai') (line 856), with additional 'gpt-5.4' override entries at lines 833-835 and 861-863. The PR's own new comment in codex_provider.py explicitly names this exact matrix and the shared risk ('a withdrawal found here is worth checking there'), showing the authors were aware of the connection, yet the matrix itself was left unfixed.

<details>

**How to reproduce**
1. A user generates a Codex or Cursor+OpenAI agent config with tier cf:tier:cheap (or an analyze/planning override at cf:tier:cheap). 2. The generator resolves the tier via _MODEL_MATRIX to 'gpt-5.4-mini' (or 'gpt-5.4' for the override paths) and bakes it into the emitted agent config. 3. The user runs that agent through Codex CLI on a ChatGPT account. 4. Codex rejects the run with 'The [model] model is not supported when using Codex with a ChatGPT account.'

**Expected behavior**
The production model matrix should track the same withdrawal fact discovered and documented in this PR, resolving cf:tier:cheap (and related overrides) for codex/openai and cursor/openai to a currently-supported slug (e.g. gpt-5.6-sol or another currently-entitled cheap model), just as codex_provider.py's DEFAULT_MODEL was updated.

**Actual behavior**
_MODEL_MATRIX still resolves cf:tier:cheap to the withdrawn 'gpt-5.4-mini' and several overrides to 'gpt-5.4' for both codex/openai and cursor/openai, unchanged by this PR, despite the PR itself proving these slugs are rejected at runtime.

```text
PR fixes test pilot: codex_provider.py DEFAULT_MODEL gpt-5.4-mini -> gpt-5.6-sol
|
v (same withdrawal fact, not propagated)
agents.py _MODEL_MATRIX[('codex','openai')]['base']['cf:tier:cheap'] = 'gpt-5.4-mini' (unchanged)
agents.py _MODEL_MATRIX[('cursor','openai')]['base']['cf:tier:cheap'] = 'gpt-5.4-mini' (unchanged)
|
v
End user generates cheap-tier agent config -> Codex CLI rejects model on ChatGPT account -> same outage class recurs in production
```

**Impact**
End users generating cheap-tier Codex or Cursor+OpenAI agent configs will hit a hard runtime rejection from Codex CLI, recurring the exact outage class this PR fixed in CI, but now affecting real users in production.

**Suggested correction**
Update _MODEL_MATRIX's cf:tier:cheap base and cf:tier:cheap overrides for ('codex','openai') and ('cursor','openai') to a currently-supported model slug, and add tracking/comment noting these two locations must be kept in sync (as codex_provider.py's new comment already flags).

**How to verify**
Confirm _MODEL_MATRIX no longer contains 'gpt-5.4-mini'/'gpt-5.4' for codex/openai and cursor/openai entries, and that a generated agent config for cf:tier:cheap on those tool/provider pairs runs successfully against a real ChatGPT-account Codex session.

</details>

ainetx:
### Makefile still documents the withdrawn gpt-5.4-mini as the pilot's cheap default

**Severity:** Minor

**Problem**
This diff updated the model-default mention in CONTRIBUTING.md and promptfooconfig.yaml but missed the identical mention in Makefile:332, which still reads 'gpt-5.4-mini'.

<details>

**How to reproduce**
1. Read Makefile lines 330-334.
2. Compare against CONTRIBUTING.md:342 and promptfooconfig.yaml:7, both updated in this diff.
3. Makefile still says 'gpt-5.4-mini'.

**Expected behavior**
All three locations describing the pilot's default cheap model should stay in sync.

**Actual behavior**
Makefile's comment is now inconsistent with the actual default and with the other two updated docs.

```text
codex_provider.py DEFAULT_MODEL -> gpt-5.6-sol
CONTRIBUTING.md -> gpt-5.6-sol
promptfooconfig.yaml -> gpt-5.6-sol
Makefile -> gpt-5.4-mini (stale)
```

**Impact**
Contributors reading the Makefile get stale guidance about which model is actually used by default, and may believe (incorrectly) that gpt-5.4-mini still works.

**Suggested correction**
Update Makefile:332 to reference 'gpt-5.6-sol' (or a generic 'see CONTRIBUTING.md' pointer to avoid future duplication).

**How to verify**
Grep the repo for 'gpt-5.4-mini' after the fix and confirm no stale references remain in comments/docs.

</details>

---
_Original location: Makefile:332 -- inline anchoring could not be resolved after 1 attempt(s)._

tests/prompts/cf-ux/providers/codex_provider.py
```diff
@@ -13,7 +13,15 @@
...
-DEFAULT_MODEL = os.environ.get("CF_UX_CODEX_MODEL", "gpt-5.4-mini")
+#: A slug, not a promise: the set a `codex` account is entitled to changes, and
+#: this one is only the cheapest of that set as of 2026-09-18 (`codex` reports
+#: gpt-5.6-sol, -terra, -luna, gpt-6-astra, gpt-5.5). Its predecessor here,
+#: `gpt-5.4-mini`, had been withdrawn, and every codex case in the pilot errored
+#: with `The 'gpt-5.4-mini' model is not supported when using Codex with a
+#: ChatGPT account.` -- half the suite red for a reason that had nothing to do
+#: with the skill under test. When that happens again, re-pick from `codex` and
+#: move the date; `CF_UX_CODEX_MODEL` is the escape hatch meanwhile.
+DEFAULT_MODEL = os.environ.get("CF_UX_CODEX_MODEL", "gpt-5.6-sol")
```
ainetx, line 31: ### No shared source of truth for 'currently entitled Codex model slug' between test pilot and production generator

**Severity:** Minor

**Problem**
The same volatile external fact (which OpenAI model slugs a ChatGPT-plan Codex account currently accepts) is duplicated as separate hardcoded literals in codex_provider.py (test) and agents.py's _MODEL_MATRIX (production), with no shared constant or config tying them together.

<details>

**How to reproduce**
1. grep 'gpt-5.4' across the repo — found in both agents.py and (formerly) codex_provider.py.
2. Observe this PR updated only the test copy, leaving the production copy on the same withdrawn value.

**Expected behavior**
A single source of truth (shared constant/config) for the current cheap Codex model slug, or at minimum a cross-reference comment linking the two locations.

**Actual behavior**
Two independently-maintained literals that have already drifted out of sync as a direct result of this PR's own fix.

```text
external fact (OpenAI model availability) --duplicated into--> codex_provider.py::DEFAULT_MODEL
--duplicated into--> agents.py::_MODEL_MATRIX
(no link between the two; one updated, one not)
```

**Impact**
Future model withdrawals will require remembering to update both locations by hand, and this PR is evidence that's easy to miss.

**Suggested correction**
Extract a shared constant/module (or a small JSON/YAML config) referenced by both codex_provider.py and agents.py for the current cheap-tier OpenAI model slug.

**How to verify**
After refactor, changing the model slug in one place should update both the pilot default and the production matrix.

</details>

tests/prompts/cf-ux/promptfooconfig.yaml
```diff
@@ -4,7 +4,7 @@
 # source (see providers/_sandbox.py). Sandboxes are auto-cleaned on atexit,
 # SIGTERM/SIGINT, and via pid-liveness sweep on next invocation.
 #
-# Cheap-by-default models (Claude Haiku 4.5, gpt-5.4-mini) with low reasoning
+# Cheap-by-default models (Claude Haiku 4.5, gpt-5.6-sol) with low reasoning
```
ainetx, line 7: ### promptfooconfig.yaml's model-slug comment remains an unreferenced duplicate, unlike the Makefile's now-fixed equivalent

**Severity:** Minor

**Problem**
promptfooconfig.yaml line 7 hardcodes 'gpt-5.6-sol' as free-text prose, independent of codex_provider.py's DEFAULT_MODEL constant. This PR demonstrates awareness of exactly this failure mode by fixing the Makefile's identical duplication (removing the literal, explaining 'this comment is the copy that went stale last time'), but did not apply the same treatment to promptfooconfig.yaml, nor does the DEFAULT_MODEL runbook comment or CONTRIBUTING.md mention this yaml file as a location to update on re-pick.

<details>

**How to reproduce**
1. Model gpt-5.6-sol is withdrawn in the future. 2. A maintainer follows codex_provider.py's runbook and updates DEFAULT_MODEL / CF_UX_CODEX_MODEL and CONTRIBUTING.md's table. 3. promptfooconfig.yaml:7 is not mentioned anywhere in the runbook, so it is not updated and now names a stale/withdrawn model.

**Expected behavior**
The yaml comment should either reference the source of truth (e.g. 'see codex_provider.py DEFAULT_MODEL') instead of duplicating the literal, or the update runbook should explicitly list this file as a location to update.

**Actual behavior**
The yaml comment independently duplicates the literal slug with no cross-reference, and no runbook mentions it.

```text
codex_provider.py DEFAULT_MODEL updated on re-pick --> CONTRIBUTING.md table updated --> promptfooconfig.yaml:7 comment NOT updated (not in runbook) --> stale doc
```

**Impact**
Low-severity documentation drift; a contributor reading promptfooconfig.yaml's comment could be misled about the actual current default after a future re-pick, though the functional default lives correctly in codex_provider.py.

**Suggested correction**
Replace the literal slug in promptfooconfig.yaml:7 with a reference to codex_provider.py's DEFAULT_MODEL, or add this file to the re-pick runbook comment in codex_provider.py.

**How to verify**
After a future model re-pick, confirm promptfooconfig.yaml's comment either has no literal to go stale or was updated alongside the other two locations per an explicit runbook step.

</details>

tests/prompts/cf-ux/providers/codex_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of fe5fbce3c330f8078dc37cc16ea09f2ff53fe620 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Direct read of codex_provider.py lines 25-30 (added by this diff) shows an explicit, reasoned cross-reference: it names `studio.commands.agents._MODEL_MATRIX` directly, explains the deliberate architectural reason the two constants are NOT shared (tier-mapping for user configs vs. cheapest-for-test-cost are different concerns; merging them would let a pilot cost decision affect what users' agents run), and explicitly warns 'they do go stale together... a withdrawal found here is worth checking there.' This is a documented, intentional design decision with an advisory cross-reference, not an undocumented/silent duplication — the fact that agents.py wasn't actually fixed in this PR despite that warning is a separate, already-captured issue, not evidence that no cross-reference exists.

</details>

tests/prompts/cf-ux/promptfooconfig.yaml
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): promptfooconfig.yaml line 7 ('Cheap-by-default models (Claude Haiku 4.5, gpt-5.6-sol) with low reasoning...') restates the model slug as free-text prose, independent of codex_provider.py's DEFAULT_MODEL constant. The new remediation guidance (codex_provider.py's comment and CONTRIBUTING.md's updated row) tells a future maintainer to re-pick the model and update CF_UX_CODEX_MODEL/CONTRIBUTING.md, but never mentions this yaml comment.

Problem (now): promptfooconfig.yaml line 7 hardcodes 'gpt-5.6-sol' as free-text prose, independent of codex_provider.py's DEFAULT_MODEL constant. This PR demonstrates awareness of exactly this failure mode by fixing the Makefile's identical duplication (removing the literal, explaining 'this comment is the copy that went stale last time'), but did not apply the same treatment to promptfooconfig.yaml, nor does the DEFAULT_MODEL runbook comment or CONTRIBUTING.md mention this yaml file as a location to update on re-pick.

</details>

vasylcf:
Both findings verified; one taken, one answered.

## Taken — `Makefile:332` (Minor)

Correct, and it was the third copy of the slug. Rather than sync it, `fe5fbce` removes the slug from that comment: its subject is that the pilot spends real tokens on cheap models, not which ones. Fewer copies, one fewer thing to go stale.

## Answered — a shared constant with `agents.py` (Minor)

Not doing this, and the reason is in what the two are for.

`codex_provider.DEFAULT_MODEL` picks **the cheapest thing that can run a test**. `agents._MODEL_MATRIX` maps **a tier a user chose** onto a model. One constant serving both would mean a pilot cost decision changes what users' generated agents run — a coupling that is worse than the duplication it removes.

They do go stale together, so `fe5fbce` says so at the definition, pointing at the other location.

## Escalated, not fixed here — `agents.py` still emits `gpt-5.4-mini` (Major)

The finding is right, and it is the more serious one: generated Codex and Cursor+OpenAI configs at `cf:tier:cheap` name a model a ChatGPT-plan account cannot use. Measured on a real account, the entitled set is `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-6-astra`, `gpt-5.5` — `gpt-5.4-mini` and `gpt-5.4` are both gone.

It is not fixed here because *which* model each tier maps to is a product decision with a spec and snapshot tests behind it (`test_agents_existing_snapshot.py`, `test_agents_model_matrix.py` pin the current values as deliberate), not something to pick while fixing a test default. Widening this PR to cover it would bury a user-facing model change inside a test-harness fix.

Flagged in the PR description before review; tracking it separately so it gets the decision it needs.

tests/prompts/cf-ux/providers/codex_provider.py
```diff
@@ -13,7 +13,22 @@
 CODEX_BIN = "codex"
 CALL_TIMEOUT_S = 850  # under promptfoo worker timeout (900s)
 
-DEFAULT_MODEL = os.environ.get("CF_UX_CODEX_MODEL", "gpt-5.4-mini")
+#: A slug, not a promise: the set a `codex` account is entitled to changes, and
+#: this one is only the cheapest of that set as of 2026-09-18 (`codex` reports
+#: gpt-5.6-sol, -terra, -luna, gpt-6-astra, gpt-5.5). Its predecessor here,
+#: `gpt-5.4-mini`, had been withdrawn, and every codex case in the pilot errored
+#: with `The 'gpt-5.4-mini' model is not supported when using Codex with a
```
ainetx, line 20: ### No fail-fast validation of model availability

_Non-blocking review challenge -- [engineering]_

Rather than letting every codex-backed test case in the suite independently error out with the same 'model not supported' message when the pinned slug is withdrawn, why not have the provider validate DEFAULT_MODEL against the account's entitled set once at startup and raise a single, actionable error?

**Why this is worth asking**
The comment itself describes exactly this failure mode ('half the suite red for a reason that had nothing to do with the skill under test'), and the fix is purely reactive (update the slug, add a longer comment) rather than making the same class of failure cheap to diagnose next time it recurs.

**Industry note**
CLI/SDK wrappers commonly do an early capability/availability check (or catch the specific 'unsupported model' error and re-raise with a short, targeted message) so N parallel workers don't each surface the same root cause as N separate opaque failures.

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725612536

This concern still stands.

<details>
<summary>Why</summary>

The "pushback" is just an automated SonarCloud quality-gate bot comment (Quality Gate passed, 0 new issues) — it contains no argument or claim addressing the finding at all, vague or otherwise. It's boilerplate CI output, not a rebuttal. Separately, `grep` confirms `gpt-5.4-mini` is still present in `skills/studio/scripts/studio/commands/agents.py`, consistent with the finding's claim that the production _MODEL_MATRIX was left unchanged while the PR only updated the test-harness default in codex_provider.py.

</details>

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725612536

Re-verified against the current code (fe5fbce3c330f8078dc37cc16ea09f2ff53fe620) and this is still an issue. The write-up has also been updated to reflect the current code (see the edited comment above).

<details>
<summary>Why</summary>

- **Reviewer:** Confirmed by direct read of skills/studio/scripts/studio/commands/agents.py lines 826-865: _MODEL_MATRIX still hard-codes 'gpt-5.4-mini' as the cf:tier:cheap base model for both ('codex','openai') (line 828) and ('cursor','openai') (line 856), plus 'gpt-5.4' overrides at lines 833-835/861-863, entirely unchanged by this PR. The PR's own new comment in codex_provider.py (lines 16-30) explicitly names this exact matrix, quotes the literal runtime error proving gpt-5.4-mini is rejected for ChatGPT-account Codex usage, and even notes 'a withdrawal found here is worth checking there' — showing the authors were aware of the connection yet left the production-facing resolver unfixed. No plan/roadmap/deferral doc found (checked CONTRIBUTING.md's Tuning and Troubleshooting sections, which only document the test pilot's escape hatch, not a scheduled fix for the production matrix) that would excuse this as intentionally out of scope. This matches the previously accepted finding for this same PR.
- **Verifier:** Confirmed by direct read of skills/studio/scripts/studio/commands/agents.py: _MODEL_MATRIX still hard-codes 'gpt-5.4-mini' as the cf:tier:cheap base model for both ('codex','openai') (line 828) and ('cursor','openai') (line 856), plus 'gpt-5.4' overrides (lines 833-835, 861-863). get_file_diff confirms agents.py was not touched by this PR at all. The PR's own new comment in codex_provider.py (lines 16-30) quotes the literal runtime error proving gpt-5.4-mini is rejected for ChatGPT-account Codex usage, explicitly names _MODEL_MATRIX, and even notes 'a withdrawal found here is worth checking there' -- yet the production resolver (used at line 992 to resolve tiers into real generated agent config model IDs) was left with the same withdrawn slug. No plan/roadmap/deferral doc found (CONTRIBUTING.md's Tuning/Troubleshooting sections only cover the test pilot's escape hatch) that would excuse this as an intentional deferral.

</details>

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725617638

This finding is already closed (fixed-by-new-commit) -- an automated re-check won't reopen it on its own, so flagging this for a human to look at directly rather than reversing it here.

<details>
<summary>Why</summary>

The posted "pushback" is not actually a rebuttal — it's an automated SonarCloud quality-gate comment with no relevance to the model-sync finding, making no checkable claim to evaluate. Separately, even on the merits, the diff and current file content show the finding is stale: Makefile lines 331-338 no longer say 'gpt-5.4-mini' at all — the comment was rewritten to deliberately stop repeating slugs and explains why ("The slugs themselves are deliberately not repeated here... this comment is the copy that went stale last time"). But since the pushback itself offers nothing substantive, there's nothing to agree with here.

</details>

ainetx (changes requested):
Requesting changes: the pilot-side fix is fine, but it doesn't close out the actual production risk, and the PR's own new comment names that risk explicitly without addressing it.

- **Production agent model matrix still bakes in the withdrawn gpt-5.4-mini slug** — `codex_provider.py` is updated to a working default, but `skills/studio/scripts/studio/commands/agents.py`'s `_MODEL_MATRIX` (the resolver that generates real end-user agent configs) still hard-codes `gpt-5.4-mini` for both `('codex','openai')` and `('cursor','openai')`, plus related `gpt-5.4` overrides — meaning any agent generated through this path picks a model that ChatGPT-account Codex sessions will reject at runtime. This needs the same fix applied here before merge. ([comment](https://github.com/constructorfabric/studio/pull/239#issuecomment-5725612536))

vasylcf:
Second round, both taken.

**`promptfooconfig.yaml` was still an unreferenced duplicate** (`051948a`). Fair, and the asymmetry was the point: `fe5fbce` removed the same copy from the Makefile and left this one, in a file the remediation note never mentions — which is exactly how the Makefile's copy went stale. Same treatment rather than a cross-reference: the comment's subject is that the pilot runs cheap models at low effort, not which ones. Also corrected `~3 min` there, which predated several scenarios; the measured run is ~1.5 min.

**"No fail-fast validation of model availability"** — agreed, and it now exists, though not in this PR. #241 adds `tests/prompts/cf-ux/preflight.py`, which `make test-prompts` runs before anything is spent. It reads `codex`'s own model cache — no network, no cost — and refuses with the models the account does have:

```
codex model 'gpt-5.4-mini' is not among the ones this account is entitled to.
  ~/.codex/models_cache.json lists: gpt-5.5, gpt-5.6-luna, gpt-5.6-sol, gpt-5.6-terra, gpt-6-astra
  Pick one and set it:

    CF_UX_CODEX_MODEL=gpt-5.6-sol make test-prompts
```

It stays silent on a missing, unreadable or unfamiliar cache: a preflight that blocks a run on its own ignorance is worse than the 400 it was meant to pre-empt.

The same gap in the **production** generator is #243: `agents.py` emitted the same withdrawn slugs into real Codex and Cursor+OpenAI configs, which is the more serious half of this.

tests/prompts/cf-ux/promptfooconfig.yaml
ainetx: **RESOLVED** -- Confirmed fixed as of 051948a552fd5d5470ff29423e221bbeed6add1d -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Direct read of the current promptfooconfig.yaml lines 7-10 shows it no longer hardcodes any model slug ('Cheap-by-default models at low reasoning effort... The slugs themselves live in providers/ and are deliberately not restated here: this comment was one of the copies that went stale on the last rotation.'). The prior finding described an earlier version of this file that still said 'gpt-5.6-sol' in prose; that has since been fixed identically to how Makefile:332 was fixed, so the third independently-drifting copy no longer exists.

</details>

tests/prompts/cf-ux/providers/codex_provider.py
vasylcf: Accepted, and built -- in two places, because the check is worth having at two different moments.

You are right that the fix in this PR is purely reactive. Bumping the slug and writing a longer comment leaves the next withdrawal to be discovered the same way: N workers each reporting the same root cause as N opaque failures.

**Before the run** -- #241 adds `tests/prompts/cf-ux/preflight.py`, which resolves the account's entitled set from codex's own cache and refuses to start when the pinned slug is not in it, with one message naming the slug and what the account actually has. That is the "single, actionable error" your comment asks for, placed where it costs one check instead of one per case.

**In the product** -- #245 adds `skills/studio/scripts/studio/utils/model_entitlements.py` and wires it into `cf agents`, so a generated agent naming a model the account cannot use is reported at generation time rather than at the point of use. The same withdrawal that produced this PR would now surface as a warning next to the agent it affects.

One deliberate limit, since your "industry note" is about capability checks generally: both read codex's cached listing, and a cache that has entries but lists nothing is treated as *unknown*, not as an empty entitlement set. A check that fails closed on missing local state would ground the suite over a stale cache, which is a worse failure than the one it prevents.

Nothing left to change on this PR, so I am leaving it as it stands; the challenge is answered by #241 and #245 rather than by a third copy of the check here.

ainetx: @vasylcf OK, answer accepted.

vasylcf:
@ainetx — requesting a re-review. Nothing here has changed since your last look; what has changed is that the one thing you asked for now exists elsewhere, so I think the standing `CHANGES_REQUESTED` (18 Sep) no longer reflects the state of the PR.

The review left three threads. Two are resolved. The third — the non-blocking challenge about validating model availability once at startup rather than letting every codex-backed case fail with the same message — you marked *"answer accepted"* earlier today.

That challenge is also implemented, in the two places it belongs:

- **#241** (merged today) adds `tests/prompts/cf-ux/preflight.py`, which resolves the account's entitled set from codex's own cache and refuses to start when the pinned slug is not in it — one message naming the slug and what the account actually has, instead of N workers reporting the same root cause N times.
- **#245** (approved) adds `model_entitlements.py` and wires it into `cf agents`, so the same withdrawal is reported at generation time rather than as a 400 at the point of use.

Worth saying why this one still matters on its own: `gpt-5.4-mini` was written in three independent places, and this is the third. #243 fixes the tier matrix, #245 makes the next withdrawal visible, and this PR fixes the cf-ux eval provider's own default. Leaving it means the eval suite keeps failing on a slug nobody can use, while the other two look fixed.

Happy to rebase onto whatever lands first if you would rather take #243 and #245 ahead of it.

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725612536

Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): This PR fixes the test pilot's default Codex model in codex_provider.py after discovering gpt-5.4-mini is withdrawn and rejected by ChatGPT-account Codex sessions with the runtime error "The 'gpt-5.4-mini' model is not supported when using Codex with a ChatGPT account." However, skills/studio/scripts/studio/commands/agents.py's _MODEL_MATRIX -- the production resolver that bakes model slugs into real generated agent configs for end users (resolved via _MODEL_MATRIX.get((tool, provider)) at line 992) -- still hard-codes 'gpt-5.4-mini' as the cf:tier:cheap base model for both ('codex','openai') (line 828) and ('cursor','openai') (line 856), with additional 'gpt-5.4' override entries at lines 833-835 and 861-863. The PR's own new comment in codex_provider.py explicitly names this exact matrix and the shared risk ('a withdrawal found here is worth checking there'), showing the authors were aware of the connection, yet the matrix itself was left unfixed.

Problem (now): codex_provider.py's new comment documents that gpt-5.4-mini is withdrawn and rejected with the runtime error "The 'gpt-5.4-mini' model is not supported when using Codex with a ChatGPT account." Yet agents.py's _MODEL_MATRIX, the production resolver that bakes model slugs into real generated agent configs for end users, still hard-codes 'gpt-5.4-mini' as the cf:tier:cheap base model for ('codex','openai') and ('cursor','openai'), with 'gpt-5.4' override entries alongside, unchanged by this PR.

</details>

vasylcf:
On the matrix half, which the re-verification still reports. `_MODEL_MATRIX` was fixed in #243, merged to `main` today. I've merged `main` into this branch, so the re-check now sees it. The matrix on this branch holds no `gpt-5.4`/`gpt-5.4-mini` value any more; the one remaining mention in `agents.py` is the comment recording that those slugs were withdrawn. This PR fixes the pilot's default, #243 fixed the production matrix, and #245 adds the warning for the next time a slug is withdrawn.

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725617638

@vasylcf **RESOLVED** -- Acknowledged -- thanks. This finding is already closed (fixed-by-new-commit), and this doesn't change that.

ainetx:
↩️ Replying to https://github.com/constructorfabric/studio/pull/239#issuecomment-5725612536

**RESOLVED** -- Confirmed fixed as of 10209519ac760854e05b80cf61e19e0b18c92f23 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The current _MODEL_MATRIX uses gpt-5.6-luna, gpt-5.6-terra, and gpt-6-astra for both Codex/OpenAI and Cursor/OpenAI; gpt-5.4 and gpt-5.4-mini remain only in explanatory withdrawn-model comments.

</details>

ainetx:
> **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/284

### **MINOR**: Empty Codex model overrides bypass preflight validation

**Severity**: **MINOR**

**Problem**
main() calls check_codex_model(DEFAULT_MODEL) only when DEFAULT_MODEL is truthy, so an exported empty CF_UX_CODEX_MODEL causes preflight to skip model validation and report success even though the provider will invoke Codex with an empty model.

<details>

**How to reproduce**
Preconditions: a usable node and an exported empty CF_UX_CODEX_MODEL. 1. Run the preflight command. 2. Observe that check_node() passes and the model check is skipped. 3. Run the pilot and observe the downstream Codex model error.

**Expected behavior**
An explicitly empty model override should be reported as invalid before the pilot starts.

**Actual behavior**
The falsy DEFAULT_MODEL bypasses check_codex_model entirely, allowing preflight to return success.

```text
CF_UX_CODEX_MODEL='' -> DEFAULT_MODEL='' -> conditional skips check_codex_model -> preflight passes -> codex receives empty model
```

**Impact**
Misconfigured prompt-test runs spend time constructing sandboxes and fail later with an opaque CLI error.

**Suggested correction**
Validate empty overrides explicitly, for example by reporting an invalid empty model before calling check_codex_model, or by removing the truthiness guard while preserving the intended import-failure handling.

**How to verify**
Add a preflight test with CF_UX_CODEX_MODEL set to an empty string and assert a nonzero result containing an actionable configuration message.

</details>

---
_Original location: tests/prompts/cf-ux/preflight.py:223 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx:
> **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/285

### **MINOR**: Kit validation omits registry-only artifact kinds

**Severity**: **MINOR**

**Problem**
_kinds_across_kits() derives known kinds solely from loaded kit constraints, while validate.py's equivalent also unions meta.iter_all_artifacts(); a registry-registered kind absent from every kit's by_kind table can therefore be reported as unknown by validate-kits.

<details>

**How to reproduce**
Preconditions: the artifact registry contains kind CUSTOM, no loaded kit declares CUSTOM in constraints.by_kind, and a kit has [validation.severity.CUSTOM]. 1. Run validate. 2. Observe that the registry kind is accepted. 3. Run validate-kits. 4. Observe an unknown-kind warning for CUSTOM.

**Expected behavior**
Both commands should use the ADR-defined union of kit-declared kinds and registry-registered kinds.

**Actual behavior**
validate-kits compares the severity scope only against _kinds_across_kits(), which contains no registry entries.

```text
registry CUSTOM + kit severity CUSTOM -> validate union includes CUSTOM -> accepted; validate-kits kit-only set omits CUSTOM -> false warning
```

**Impact**
A legitimate cross-kit or registry-backed severity scope is reported as a configuration mistake, reducing trust in validation diagnostics.

**Suggested correction**
Share a common known-kind helper or pass registry metadata into validate-kits and union its iter_all_artifacts() kinds with the kit-declared set.

**How to verify**
Add a fixture with a registry-only kind and a matching validation.severity scope, then assert both validate and validate-kits emit no unknown-kind warning.

</details>

---
_Original location: skills/studio/scripts/studio/commands/validate_kits.py:172 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx:
> **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/286

### **MINOR**: NVM node probing has no aggregate deadline

**Severity**: **MINOR**

**Problem**
_nvm_candidates() probes every candidate in candidates[:NVM_PROBE_LIMIT], and each _node_version() call has its own 20-second timeout; no overall deadline or early exit bounds the cumulative wait.

<details>

**How to reproduce**
Preconditions: the nvm directory contains eight candidate node binaries that each hang until terminated. 1. Run check_node() with PATH's node missing or too old. 2. Let _nvm_candidates() scan the candidates. 3. Observe up to eight sequential 20-second subprocess timeouts before the install remedy is printed.

**Expected behavior**
The preflight should fail fast within a bounded aggregate interval.

**Actual behavior**
The per-process timeout is applied independently, allowing roughly 160 seconds of cumulative probing.

```text
check_node failure -> _nvm_candidates -> node probe 1..8 -> 8 x 20s timeouts -> install remedy
```

**Impact**
A supposedly cheap preflight can stall for minutes before reporting a straightforward environment problem.

**Suggested correction**
Add an aggregate deadline to _nvm_candidates/check_node and stop probing when it expires, or stop after a qualifying candidate when the selection policy permits.

**How to verify**
Use mocked _node_version calls that consume simulated time and assert the candidate scan stops at the aggregate deadline.

</details>

---
_Original location: tests/prompts/cf-ux/preflight.py:99 -- inline anchoring could not be resolved after 1 attempt(s)._

ainetx:
> **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/287

### **MINOR**: Loose version-branch glob can trigger full CI on topic branches

**Severity**: **MINOR**

**Problem**
The push and pull_request filters use v[0-9]*, which matches any branch beginning with v and a digit, including names such as v2-experiment; no path filter or job condition narrows those matches to actual release branches.

<details>

**How to reproduce**
Preconditions: a contributor creates a branch named v2-experiment. 1. Push a commit or open a pull request from that branch. 2. GitHub Actions evaluates the v[0-9]* pattern. 3. Observe that the complete CI workflow is scheduled, including all matrix and analysis jobs.

**Expected behavior**
Only genuine release branches should trigger the release-specific full CI matrix, or the naming convention should be enforced separately.

**Actual behavior**
The raw glob alone admits any matching topic branch and schedules the full workflow.

```text
topic branch v2-experiment -> v[0-9]* matches -> full CI matrix -> unrelated compute spend
```

**Impact**
Accidental branch-name collisions can multiply CI compute and queue usage without providing release-testing value.

**Suggested correction**
Use a stricter release-name glob/branch convention or add a job-level condition/path policy that distinguishes release branches from arbitrary topic branches.

**How to verify**
Exercise representative release and topic branch names against the workflow matcher and confirm only intended release names schedule the matrix.

</details>

---
_Original location: .github/workflows/ci.yml:25 -- inline anchoring could not be resolved after 1 attempt(s)._

Makefile
```diff
@@ -337,9 +337,13 @@ install-proxy: check-pipx
 #                        installed by `make install-proxy`)
```
ainetx: > **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/288

### **MINOR**: Broken pipx pytest executable prevents fallback recovery

**Severity**: **MINOR**

**Problem**
PYTEST selects the pipx path solely because it exists and is executable; check-pytest then tests only that selected path and exits on failure without retrying PATH pytest or python -m pytest.

<details>

**How to reproduce**
Preconditions: ~/.local/bin/pytest is executable but broken, while another working pytest is available on PATH. 1. Invoke make test. 2. PYTEST selects ~/.local/bin/pytest. 3. check-pytest --version fails. 4. The recipe exits instead of trying the working fallback.

**Expected behavior**
A present-but-unrunnable pipx candidate should fall through to a runnable PATH or python -m pytest candidate.

**Actual behavior**
The fallback chain checks only file executability during Make parsing, and the later runtime check hard-fails the selected candidate.

```text
executable stale pipx pytest -> PYTEST selects it -> --version fails -> no retry -> test target aborts
```

**Impact**
A recoverable local environment is rejected and requires manual override or cleanup of the stale pipx binary.

**Suggested correction**
Probe candidates for successful execution when selecting them, or make check-pytest retry the remaining candidates after the selected command fails.

**How to verify**
Create a stale executable pipx shim plus a working PATH pytest and assert make test proceeds using the working fallback.

</details>

ainetx: > **Out of scope for this PR** -- this is a real, evidenced issue, but it's not introduced or worsened by this diff (e.g. it predates it), so it isn't blocking this review. Filed as a standalone issue to track it: https://github.com/constructorfabric/studio/issues/289

### **MINOR**: Pytest discovery runs during every Make invocation

**Severity**: **MINOR**

**Problem**
PYTEST and PYTEST_PARALLEL use top-level:= assignments containing shell commands, so make executes pytest discovery while parsing the Makefile even for unrelated targets such as help or clean.

<details>

**How to reproduce**
Preconditions: invoke any non-test make target. 1. Make parses the top-level assignments. 2. It executes the PYTEST test/command-v probe. 3. It executes pytest --help through PYTEST_PARALLEL. 4. Only afterward does it select the requested target.

**Expected behavior**
Pytest discovery should be deferred until a test-related target requires it.

**Actual behavior**
The shell probes execute unconditionally at parse time for every make invocation.

```text
make any-target -> parse Makefile -> PYTEST shell probe + pytest --help -> execute unrelated target
```

**Impact**
Every make invocation pays avoidable subprocess and potentially Python startup overhead, with the largest cost when pytest is resolved through python -m pytest.

**Suggested correction**
Move pytest discovery and xdist capability probing into test prerequisites/recipes or another lazily evaluated test-only path.

**How to verify**
Instrument subprocess execution while invoking make help and make clean, then confirm no pytest probes occur until a test target is selected.

</details>

ainetx (approved):
Reviewed the model-slug rotation and the accompanying doc/comment updates. The change swaps the withdrawn `gpt-5.4-mini` for `gpt-5.6-sol` and, more importantly, stops restating the actual slug in CONTRIBUTING.md, the Makefile comment, and promptfooconfig.yaml — those copies had gone stale before, and the fix correctly consolidates the source of truth in `codex_provider.py` with a comment explaining why. The added docstring also calls out the deliberate non-sharing with `_MODEL_MATRIX`, which heads off a plausible future "just unify these" refactor. Nothing functional beyond a default-value/env-var change, no tests to break, no security surface here. Looks good to merge.

Makefile
vasylcf: Agreed, and it's mine: this fallback chain came in with #241. The chain checks that the pipx path exists and is executable, and the runnability check only runs later on whatever the chain picked, so a stale pipx pytest never falls through. The fix is to probe each candidate with `--version` inside the chain and take the first that answers. I'll do it together with #289 and #268, because all three are about how the Makefile finds pytest, and #268 is about to pin which version it is. Thanks for filing #288.

vasylcf: Agreed, also from #241. I made these `:=` so the probe would not re-run at every reference, and that moved it to parse time for every target, `help` and `clean` included. A lazily evaluated variable that caches its first expansion, or a stamp file written by `check-pytest`, keeps the one-probe property without running it for unrelated targets. I'll fix it with #288 and #268, as said on the sibling thread. Thanks for filing #289.

PR #240, opened 2026-09-18
feat(cf-ux): report a workflow run without the router as its own state

vasylcf:
## What happened

In a real pilot run, asked to *"Validate the artifact at architecture/decisions/0001-postgres.md and report findings"*, the model invoked `cf-documenting-review` **directly** and never called `cf`. Recorded tool input:

```json
{"skill": "cf-documenting-review", "args": "Validate the artifact at ..."}
```

The harness had two states for this, and neither is true:

- `ran` is false — no gate, no menu, no routing decision. The thing this suite measures never happened.
- `failed` files it with the runs where **nothing** of Studio was reached, which is a materially different finding.

## Change

A third state, `bypassed`. Studio produced the answer, so it is handed to the grader like `ran`. What it is *not* is a clean pass for the router — that shows in `skill_state` and in a `WARNING` naming the workflow.

**The name matcher is untouched.** `cf-documenting-review` still does not equal `cf`, and `_NAME_SEPARATORS` still deliberately does not split on `-`. Only the grading consequence of a `cf-*` call changes.

## The trade, and how it is bounded

Extending recognition from the exact token `cf` to a whole `cf-` prefix widens the net, and `_invoked_names` reports every identifier-shaped string **at any depth** — so a rival skill carrying a `cf-` name in an unrelated field would read as a Studio workflow. That is the loose match the matcher was written to avoid, one prefix wider.

So a bypass requires the **whole call** to name `cf-` identifiers and nothing else:

```python
{"command": "superpowers:brainstorming", "mode": "cf-documenting-review"}
# names {brainstorming, cf-documenting-review} -> not unambiguous -> failed, not bypassed
```

Pinned by `test_a_rival_carrying_a_cf_workflow_name_is_not_a_bypass`.

The evidence bar is unchanged: a bypassing call must come back clean, and come back at all. A workflow that errored, or whose result never arrived, leaves only the model's own prose, and grading that would report on Studio for a run Studio did not produce.

## One existing test changes meaning

`test_a_skill_whose_name_merely_begins_with_cf_is_not_the_cf_skill` asserted `cf-generate` → `failed`. It now asserts `!= "ran"` and `== "bypassed"`. That is the decision this PR implements, stated in the test rather than silently dropped — the guard it was protecting (a `cf-*` name is not the router) is still intact.

## Honest limitation

promptfoo's own tally counts `bypassed` as an ordinary pass — there is no third column in its summary. The bypass is visible in `skill_state`, in `skills_invoked`, and in the log. Documented as such in the README.

## Verified

- `6275 passed, 4 skipped, 15 xfailed` locally; `make spec-coverage` and `make validate` green.
- Replayed against the **actual** recorded tool input from the pilot run: `state: bypassed`, `names: ['cf-documenting-review']`.
- 8 new tests, including the router itself still grading `ran`, an errored workflow and a resultless workflow still refused, and an unrelated skill still `failed`.

## Summary by CodeRabbit

- **New Features**
  - Added a `bypassed` workflow state for direct execution of qualifying workflows without routing.
  - Directly executed workflows are now graded consistently with routed workflows when successful.
  - Metadata now distinguishes bypassed execution and provides clearer candidate-matching details.

- **Documentation**
  - Expanded workflow-state documentation with grading rules, precedence, and requirements for bypass detection.
  - Clarified how successful, failed, absent, and bypassed states are reported.


tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -239,6 +253,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
 
     targeted = [call_id for call_id, found in named.items() if _SKILL_NAME in found]
```
ainetx, line 259: ### A false-positive 'cf' match on one erroring call suppresses bypass detection for a separately succeeding cf-* workflow call

**Severity:** Minor

**Problem**
`_skill_trace` only computes `bypassing` inside the `if not targeted:` branch, where `targeted` is any call whose `_invoked_names` output contains the literal 'cf' — including the accepted false-positive class (any field, any depth) documented earlier in the same file. If one Skill call false-positively matches 'cf' in an unrelated field and errors, while a separate Skill call in the same transcript names only a genuine `cf-*` workflow and succeeds, `targeted` is non-empty so bypass detection never runs, and the function falls through to `failed` because the only targeted (falsely-matched) call errored.

<details>

**How to reproduce**
1. Construct a transcript with two Skill tool_use blocks: call A has an input field containing the literal string 'cf' in some unrelated key (satisfying `_invoked_names`) and its tool_result has `is_error: true`. 2. Call B's input names only `cf-documenting-review` and its tool_result succeeds (`is_error` absent/false). 3. Run `_skill_trace(events, raw)`.

**Expected behavior**
Call B's successful, unambiguous `cf-*` workflow completion should be reported as `bypassed` (graded, with the routing-bypass warning logged).

**Actual behavior**
`targeted` includes call A (false 'cf' match), so the `not targeted` bypass-computation branch never executes; the loop over `targeted` finds only an erroring call and the function returns `failed`, which is excluded from `_GRADED_STATES`, silently dropping both the bypass diagnosis and the gradeable answer.

```text
call A (false 'cf' match, errors) -> targeted=[A] -> bypass check (lines 275-288) skipped
call B (genuine cf-x, succeeds) -> never inspected for bypass
loop over targeted: A errored -> falls through -> return 'failed' (ungraded)
```

**Impact**
Runs where a genuine bypass coincides with an unrelated erroring call that happens to false-positive-match 'cf' are mis-reported as ungraded failures instead of graded bypasses, understating the routing-bypass finding this PR was written to surface and losing a real, scoreable answer from the suite's tally.

**Suggested correction**
Compute `bypassing` independent of whether any call matched 'cf', or restrict the fall-through to `failed` when there is no successful targeted call AND no qualifying bypass candidate exists among the remaining calls.

**How to verify**
Add a test with two Skill calls: one false-positive-matching 'cf' and erroring, one naming only a `cf-*` workflow and succeeding; assert `skill_state == 'bypassed'` rather than `'failed'`.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -239,6 +253,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
+        # depth, so a rival skill carrying `cf-generate` in some unrelated field
+        # offers one -- and the loose match that is an accepted trade for the
+        # exact token `cf` is a much wider net across a whole prefix. Requiring
+        # the call to name nothing else keeps the bypass to calls that are
+        # unambiguously a Studio workflow.
+        bypassing = sorted({
+            name
+            for call_id, found in named.items()
+            if found and results.get(call_id) is False
+            and all(name.startswith(_SKILL_WORKFLOW_PREFIX) for name in found)
```
ainetx: ### A bare 'cf-' identifier (prefix with no workflow suffix) is accepted as a bypassed workflow name

**Severity:** Minor

**Problem**
The bypass matcher at lines 275-281 requires only `name.startswith(_SKILL_WORKFLOW_PREFIX)` where `_SKILL_WORKFLOW_PREFIX = 'cf-'`; it does not require any character after the hyphen. A tool input value that is literally `'cf-'` (permitted by `_NAME_SHAPE`) would pass this check and be reported via `_LOG_ROUTER_BYPASSED` and `skills_invoked` metadata as a legitimate bypassed workflow.

<details>

**How to reproduce**
1. Construct a Skill tool_use whose input contains the literal string 'cf-' as a candidate identifier (any field, matched by `_invoked_names`). 2. Give its tool_result a successful (`is_error` false) status. 3. Ensure no call names exactly 'cf'. 4. Run `_skill_trace`.

**Expected behavior**
A malformed/truncated identifier consisting only of the prefix should not be treated as a named workflow; it should not qualify for `bypassed` classification, or at minimum should be reported distinctly rather than as a real workflow name.

**Actual behavior**
The bare 'cf-' string satisfies `startswith('cf-')`, so it is added to `bypassing`, logged via `_LOG_ROUTER_BYPASSED` as "'cf' was not invoked; the model went straight to 'cf-'", and surfaced in `skills_invoked` metadata as if it were a genuine workflow.

```text
tool input contains literal 'cf-' -> _invoked_names keeps it (matches _NAME_SHAPE, no separator to split) -> startswith('cf-') is True -> bypassing={'cf-'} -> logged/reported as a bypassed workflow named 'cf-'
```

**Impact**
Produces a nonsensical, uninformative diagnostic ('cf-' ran directly) in metadata and logs read by a person triaging a run; low real-world likelihood since real skill identifiers are unlikely to be exactly the bare prefix, but the validation gap is real and unguarded.

**Suggested correction**
Require at least one character after the prefix, e.g. `name.startswith(_SKILL_WORKFLOW_PREFIX) and len(name) > len(_SKILL_WORKFLOW_PREFIX)`, or match against a stricter regex requiring a non-empty suffix.

**How to verify**
Add a test where a Skill call's only candidate name is the literal 'cf-' string and assert it is NOT classified as 'bypassed' (e.g. falls to 'failed' instead).

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -89,6 +89,16 @@
 #: identifier: every false positive of this class is reported, and so are some
 #: perfectly good runs. Narrowing it would need the very knowledge whose absence
 #: created the trade.
+#: A `cf-*` workflow invoked without the router in front of it. Prefixed, not
+#: namespaced, so `_NAME_SEPARATORS` deliberately does not split it and it never
+#: counts as `cf` -- see `_invoked_names`.
+_SKILL_WORKFLOW_PREFIX = "cf-"
```
ainetx: ### _SKILL_WORKFLOW_PREFIX is a bare literal, not derived from _SKILL_NAME

**Severity:** Minor

**Problem**
`_SKILL_WORKFLOW_PREFIX = "cf-"` is written as an independent string literal rather than `f"{_SKILL_NAME}-"`, even though its own comment says it exists conceptually as the workflow-name prefix of `_SKILL_NAME`.

<details>

**How to reproduce**
1. Rename `_SKILL_NAME` to something else (e.g. 'studio').
2. `_SKILL_WORKFLOW_PREFIX` stays 'cf-', silently breaking bypass detection.
3. No test or type error surfaces the mismatch.

**Expected behavior**
`_SKILL_WORKFLOW_PREFIX` should be derived from `_SKILL_NAME` so the two constants cannot drift.

**Actual behavior**
The two constants are hand-maintained literals kept in sync only by a comment.

```text
_SKILL_NAME (literal) -x-> _SKILL_WORKFLOW_PREFIX (separate literal) -- no code-level link
```

**Impact**
A future rename of the router's skill name silently desynchronizes bypass classification with no test failure pointing at the cause.

**Suggested correction**
Define `_SKILL_WORKFLOW_PREFIX = f"{_SKILL_NAME}-"`.

**How to verify**
Change `_SKILL_NAME` and confirm the workflow prefix constant updates automatically without a second edit.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -239,6 +253,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
+        # or came back not at all, leaves only the model's own prose, and grading
+        # that reports on Studio for a run Studio did not produce.
+        # *Every* candidate in the call must be a `cf-` name, not merely one of
+        # them. `_invoked_names` reports every identifier-shaped string at any
+        # depth, so a rival skill carrying `cf-generate` in some unrelated field
+        # offers one -- and the loose match that is an accepted trade for the
+        # exact token `cf` is a much wider net across a whole prefix. Requiring
+        # the call to name nothing else keeps the bypass to calls that are
+        # unambiguously a Studio workflow.
+        bypassing = sorted({
```
ainetx: ### Bypass classification lacks the ambiguity transparency the 'ran' path has for multi-candidate calls

**Severity:** Minor

**Problem**
The `ran` path explicitly surfaces ambiguous multi-name matches via `_LOG_AMBIGUOUS_MATCH` and `other_candidates` metadata. The bypass path (lines 275-281) faces the same 'which field is the real name' uncertainty since `_invoked_names` scans every field at any depth, yet it silently accepts and unions all qualifying cf-* names with no warning or metadata trace of ambiguity.

<details>

**How to reproduce**
1. Construct a Skill call whose input contains two distinct cf-*-prefixed strings, e.g. {'a': 'cf-generate', 'b': 'cf-review'}.
2. Run it through _skill_trace with a successful tool_result.
3. Observe both names land in `bypassing` with no ambiguity warning or other_candidates entry, unlike the ran path's handling of a similar shape.

**Expected behavior**
The bypass path should surface the same kind of ambiguity signal (log + other_candidates-equivalent) when a call names more than one candidate.

**Actual behavior**
Multiple candidate names are silently unioned into the bypass verdict with no diagnostic trace.

```text
ran path: multi-name call -> _LOG_AMBIGUOUS_MATCH + other_candidates
bypass path: multi-name call -> silent union into bypassing, no signal
```

**Impact**
A person auditing bypass results has no signal that a given 'bypassed' verdict rested on multiple candidate names in one call.

**Suggested correction**
Emit a warning/metadata field analogous to _LOG_AMBIGUOUS_MATCH when a bypassing call's `found` set contains more than one name.

**How to verify**
Add a test with two cf-* names in one call's input and assert a warning/metadata field is populated.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -702,3 +722,80 @@ def _dies(cmd, **kwargs):
...
+        out, _seen = run_provider(_stream(
+            _skill_call(skill="cf-documenting-review"),
+            _skill_result(),
+            _result("findings for 0001-postgres.md"),
+        ))
+
+        assert out["output"] == "findings for 0001-postgres.md"
+        assert out["metadata"]["skill_state"] == "bypassed"
+        assert out["metadata"]["skills_invoked"] == ["cf-documenting-review"]
+
```
ainetx, line 751: ### No test exercises the cross-call union in `_bypassing_names` for two simultaneously-successful cf-* calls

**Severity:** Minor

**Problem**
`_bypassing_names` (claude_provider.py:337-342) iterates `named.items()` and unions names from every call_id whose result is `False`. All new bypass tests in TestAWorkflowRunWithoutTheRouter drive either a single qualifying call, or a pair where one call is excluded (it errored), leaving exactly one call contributing to the set each time.

<details>

**How to reproduce**
1. Inspect every test in TestAWorkflowRunWithoutTheRouter. 2. Note test_one_workflow_failing_does_not_hide_another_succeeding and test_a_bypass_survives_an_unrelated_erroring_false_cf_match each have two call_ids, but only one has results.get(call_id) is False. 3. No test has two call_ids both with is_error=False, each naming a distinct cf-* workflow.

**Expected behavior**
A test with two distinct successful cf-* Skill calls in one transcript, asserting the resulting skill_state, skills_invoked ordering/dedup, and the logged names reflect both.

**Actual behavior**
The cross-call aggregation path (multiple call_ids simultaneously qualifying) is untested, so a regression there (e.g. short-circuiting after the first qualifying call, or dedup/order bugs across calls) would not be caught.

```text
two successful cf-* calls -> named.items() loop -> [untested union path] -> bypassing list -> skills_invoked/log
```

**Impact**
A future regression in the multi-call aggregation logic would silently ship undetected by this otherwise thorough test suite.

**Suggested correction**
Add a test with two distinct call_ids, both is_error=False, each naming a different cf-* workflow, asserting skills_invoked contains both names sorted/deduped and the log message lists both.

**How to verify**
Run the new test against the current implementation to confirm it passes, then mutate _bypassing_names (e.g. break after first qualifying call_id) to confirm the test would catch the regression.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -702,3 +722,80 @@ def _dies(cmd, **kwargs):
         assert "unexpected: RuntimeError" in out["error"]
```
ainetx, line 722: ### Single-call multi-candidate bypass path is untested

**Severity:** Minor

**Problem**
The nested `for name in found` iteration inside the bypass comprehension can add multiple names from a single call's `found` set when a Skill input contains two distinct cf-*-prefixed values, but no test in TestAWorkflowRunWithoutTheRouter or elsewhere constructs such an input.

<details>

**How to reproduce**
1. Search test_cf_ux_claude_provider.py for a _skill_call whose input has two distinct cf-* strings.
2. None found; _skill_call only produces {'command': skill}.

**Expected behavior**
A test should drive one Skill call whose input names two distinct cf-* values and assert both land in skills_invoked/log with no truncation.

**Actual behavior**
This sub-path is unexercised by any test.

```text
one call, input={'a':'cf-generate','b':'cf-review'} -> found={'cf-generate','cf-review'} -- inner loop untested
```

**Impact**
A regression specific to multi-name-per-call handling (e.g. accidental truncation to one name) would go unnoticed.

**Suggested correction**
Add a test with a Skill input containing two distinct cf-* values in different fields.

**How to verify**
Run the new test and confirm both names appear in skills_invoked and the log message.

</details>

tests/prompts/cf-ux/README.md
```diff
@@ -120,7 +120,18 @@ or no trace to trust:
 | `result` with a non-success `subtype` | The turn stopped short (`error_max_turns`, `error_during_execution`), so `result` holds a fragment, not an answer — even when the skill did load. An *absent* subtype is not treated this way: an unfamiliar shape should not manufacture failures. |
 | A line that will not parse | Counted in `unparsed_lines` rather than dropped silently, because a lost line can be a lost `tool_result` and the verdict above is read off exactly those. |
 
-Metadata: `skill_state` (`ran` / `failed` / `absent`), `skills_invoked` (the
+`skill_state` has four values:
+
+| State | Meaning | Graded? |
+|---|---|---|
+| `ran` | The `cf` router was invoked and its call came back clean. | yes |
+| `bypassed` | A `cf-*` workflow ran **directly**, the router never did. Studio produced the answer, so it is worth grading — but no gate, menu or routing decision happened, which is what this suite measures. Counted as a pass by promptfoo's own tally, so the metadata and the `WARNING` in the log are where a bypass is actually visible. | yes |
```
ainetx, line 128: ### README doesn't clarify that skill_match_other_candidates is always empty for bypassed runs

**Severity:** Minor

**Problem**
The README explains skill_match_other_candidates right before introducing the bypassed state but never states that bypassed traces never populate this field.

<details>

**How to reproduce**
1. Read README.md lines 98-132.
2. Note the skill_match_other_candidates explanation is immediately followed by the bypassed table row with no cross-reference.
3. In code, `_SkillTrace('bypassed', names, inputs, detail)` omits the other_candidates argument, defaulting to ().

**Expected behavior**
The README should note that other_candidates is always empty for bypassed runs, or that its ambiguity-reporting mechanism doesn't apply there.

**Actual behavior**
A reader debugging a bypassed run could reasonably expect other_candidates data that will never be populated.

```text
README: [other_candidates explained] -> [bypassed row] -- no note that the field never applies to bypassed
```

**Impact**
Wasted debugging time looking for a field that is never populated for this state.

**Suggested correction**
Add a sentence noting other_candidates is always empty/inapplicable for bypassed runs.

**How to verify**
Re-read the README section after the edit and confirm the relationship is explicit.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -239,6 +253,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
```
ainetx, line 258: ### Absent `is_error` on a tool_result is silently treated as confirmed success, now also driving the new 'bypassed' pass grade

**Severity:** Minor

**Problem**
`_skill_trace` builds `results[block.get('tool_use_id')] = bool(block.get('is_error'))` (claude_provider.py:245), which maps a tool_result block that omits the `is_error` key entirely to `False`, identically to an explicit `is_error: false`. This ambiguity previously existed for the `ran` path only; the new `_bypassing_names` helper (lines 316-342) reuses the exact same `results` dict and the identical `results.get(call_id) is False` identity check as its sole confirmation that a cf-* workflow call actually succeeded, extending the weak-evidence pattern to the newly introduced `bypassed` (graded-as-pass) state.

<details>

**How to reproduce**
1. Construct a transcript where a `Skill` call names a `cf-*` workflow (e.g. `cf-documenting-review`).
2. Include a corresponding `tool_result` block for that call's `tool_use_id` that omits the `is_error` field entirely (a truncated/schema-shifted result) rather than setting it explicitly.
3. Run `_skill_trace`/`_invoke` on this transcript.
4. Observe `results.get(call_id)` evaluates to `False`, `_bypassing_names` includes the workflow name, and `skill_state` becomes `bypassed` (graded as a pass) even though there is no confirmed evidence the call succeeded.

**Expected behavior**
A tool_result block that omits `is_error` should not be treated as equivalent to a confirmed clean success; it should be distinguished from an explicit `is_error: false` (e.g. via `block.get('is_error') is False` requiring the key present, or falling to `failed`/unconfirmed when the key is absent).

**Actual behavior**
`bool(block.get('is_error'))` collapses 'key absent' and 'key present and false' into the same `False` value, and this value is reused as the sole success evidence for the new `bypassed` grade in `_bypassing_names`.

```text
tool_result block (no is_error key) -> bool(None) == False -> results[call_id] = False -> _bypassing_names sees results.get(call_id) is False -> workflow call treated as 'ran cleanly' -> skill_state = 'bypassed' -> graded as a pass
```

**Impact**
A truncated or schema-shifted transcript for a directly-invoked cf-* workflow can be silently graded as a genuine Studio pass instead of correctly falling to `failed`/unscored, undermining the suite's core goal of never scoring a run whose success is unconfirmed.

**Suggested correction**
Distinguish an absent `is_error` key from an explicit `false`, e.g. store `None` when the key is missing and require `results.get(call_id) is False` to only match an explicit, present `false`, treating a missing key as unconfirmed (falls through to `failed`).

**How to verify**
Add a test where a `tool_result` block for a cf-* workflow call omits `is_error` entirely (not `_skill_result`'s helper, which always sets it) and assert the run is graded `failed`, not `bypassed`.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -239,6 +253,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
+        # them. `_invoked_names` reports every identifier-shaped string at any
+        # depth, so a rival skill carrying `cf-generate` in some unrelated field
+        # offers one -- and the loose match that is an accepted trade for the
+        # exact token `cf` is a much wider net across a whole prefix. Requiring
+        # the call to name nothing else keeps the bypass to calls that are
+        # unambiguously a Studio workflow.
+        bypassing = sorted({
+            name
+            for call_id, found in named.items()
+            if found and results.get(call_id) is False
```
ainetx: ### Mixed success/failure across multiple distinct cf-* calls in one transcript is untested

**Severity:** Minor

**Problem**
When a transcript has two distinct cf-*-named call_ids where one succeeds and one fails/has no result, the bypass logic correctly excludes the failing call from `bypassing` evidence (via `results.get(call_id) is False`), but this interaction is not exercised by any test, and the failed attempt is only partially visible (its name still appears in skills_invoked, but not in skill_state or detail).

<details>

**How to reproduce**
1. Construct a transcript with two Skill calls, each naming a distinct cf-* value, one followed by a successful tool_result and the other by an is_error:true result. 2. Run _skill_trace. 3. No existing test covers this shape.

**Expected behavior**
A test asserting skill_state=='bypassed' with skills_invoked containing both names but only the successful one in the bypass detail/log, confirming intended behavior explicitly.

**Actual behavior**
This mixed-outcome scenario is currently unverified by any test.

```text
call A (cf-generate, success) + call B (cf-review, error) -> bypassing={cf-generate} but names includes both -- untested
```

**Impact**
A regression that silently drops the failed sibling's visibility entirely (not even in skills_invoked) would not be caught.

**Suggested correction**
Add a test with two distinct cf-* call_ids, one succeeding and one failing, asserting the exact metadata shape.

**How to verify**
New test passes and pins the current partial-visibility behavior.

</details>

tests/prompts/cf-ux/README.md
```diff
@@ -120,7 +120,18 @@ or no trace to trust:
...
+| `absent` | No `Skill` call at all. | no |
```
ainetx, line 129: ### New README state table omits the pre-existing error-mark precedence over 'absent'

**Severity:** Minor

**Problem**
The table added by this diff describes 'absent' as simply 'No Skill call at all', but the code checks _SKILL_ERROR_MARK against raw text before checking for zero calls, so a transcript with no Skill call whose prose happens to contain 'Execute skill: cf' is actually classified 'failed', contradicting a plain reading of the table.

<details>

**How to reproduce**
1. Read claude_provider.py lines 246-252: _SKILL_ERROR_MARK check happens before the empty-calls check. 2. Read README.md's new table: 'absent' row says only 'No Skill call at all.' with no caveat.

**Expected behavior**
The table or surrounding prose should note that the error-mark check takes precedence over the no-calls-implies-absent rule.

**Actual behavior**
The table's plain description could lead a reader to misdiagnose such a transcript as 'absent' rather than 'failed'.

```text
raw text contains 'Execute skill: cf' (no actual Skill call) -> _SKILL_ERROR_MARK matches first -> 'failed', not 'absent' (contrary to naive table reading)
```

**Impact**
Documentation/code mismatch could mislead a maintainer diagnosing test failures around this state.

**Suggested correction**
Add a one-line note to the 'absent' row clarifying the error-mark precedence.

**How to verify**
README updated to mention precedence; existing test_the_real_failure_signature_is_recognised already pins the code behavior.

</details>

vasylcf:
Six findings, all reproduced first. Four were real defects and are fixed in `15a84ba`; two are answered.

## Taken

**1. A false-positive `cf` match suppressed bypass detection.** The sharpest of the six, and correct. The check lived inside `if not targeted:`, and `targeted` accepts the documented false positive — any field, any depth. Reproduced:

```python
call A: {"command": "superpowers:brainstorming", "mode": "cf"}  is_error=True
call B: {"skill": "cf-documenting-review"}                       clean
->  failed        # bypass never checked; the answer dropped too
```

The check now runs whenever no `cf` call came back clean, whatever matched. Same transcript now grades `bypassed` and the answer reaches the grader.

**2. A bare `cf-` was accepted as a workflow.** Reproduced: `skill_state: bypassed`, detail `'cf-' ran directly`. Shaped like an identifier, says nothing. A suffix is required now.

**3. `_SKILL_WORKFLOW_PREFIX` was a second literal.** Now `f"{_SKILL_NAME}-"`, pinned by a test — two hand-maintained constants kept in sync by a comment is one rename away from silently breaking detection.

**4. The bypass path lacked the `ran` path's ambiguity transparency.** Right, and it can happen: my rule requires every candidate to be `cf-*`, and `{"a": "cf-generate", "b": "cf-review"}` satisfies that. It now warns and fills `other_candidates` — with its own message, because the `ran` one reads "matched 'cf'", which is precisely what did not happen.

**5. The untested paths** (multi-call aggregation, single-call multi-candidate, mixed success/failure across `cf-*` calls) now have tests, as does the bare prefix and the derived constant. Five new, 64 in the file.

**6. README.** Added the precedence the state table omitted — an error mark beats everything, no `Skill` call at all beats the rest — and what `skill_match_other_candidates` means for a bypass (populated for `ran` and `bypassed`, empty when one name decided it).

## Answered

**A missing `is_error` read as success, now also feeding `bypassed`.** Not changing this: it is measured behaviour, not an assumption. Against claude-code 2.1.276 a successful `Skill` result carries exactly `{type, tool_use_id, content}` with content `"Launching skill: cf"` and **no `is_error`**, while `Bash` and `Read` results in the same transcript carry it explicitly. Treating the omission as missing evidence graded every successful run as a failure — the whole harness red. Full evidence and the revert are in #229.

The related case that genuinely proves nothing — a targeted call with no `tool_result` at all — is still refused, and has its own test.

## Verified

`6280 passed, 4 skipped, 15 xfailed`; `make spec-coverage` green.

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Direct trace of the current code (claude_provider.py:259-313) shows the bypass computation (`bypassing = _bypassing_names(...)`) is NOT gated by `not targeted` -- that gate was deliberately removed by this diff (see the explanatory comment at lines 273-283: 'Checked here rather than under `not targeted`, which is where it first went'). It runs whenever the `for call_id in targeted` loop fails to return 'ran', including when the only targeted ('cf'-named) call errored. This is exactly what `test_a_bypass_survives_an_unrelated_erroring_false_cf_match` verifies: a call naming 'cf' errors, a separate call naming only 'cf-documenting-review' succeeds, and the result is `skill_state == 'bypassed'` with `output == 'findings'`.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, `_bypassing_names`'s `is_workflow` helper (claude_provider.py:333-335) explicitly requires `len(name) > len(_SKILL_WORKFLOW_PREFIX)` in addition to `startswith`, which excludes the bare 'cf-' string (len 3 is not > len 3). This is directly confirmed by the dedicated test `test_a_bare_prefix_is_not_a_workflow_name`, which asserts a lone 'cf-' skill call yields `skill_state == 'failed'`, not 'bypassed'.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, `_SKILL_WORKFLOW_PREFIX` is not a bare literal: the code reads `_SKILL_WORKFLOW_PREFIX = f"{_SKILL_NAME}-"` (claude_provider.py:95), i.e. it is already derived from `_SKILL_NAME` exactly as the check suggests it should be. A future rename of `_SKILL_NAME` would automatically propagate to the prefix.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, the bypass branch does mirror the `ran` path's transparency: when `bypassing` (the sorted union of qualifying names) has more than one entry, `_skill_trace` logs `_LOG_AMBIGUOUS_BYPASS` and returns `other_candidates=others` on the `_SkillTrace` (claude_provider.py:292-300). This is directly verified by `test_several_workflow_names_in_one_call_are_reported_as_ambiguous`, which asserts `skill_match_other_candidates == ['cf-review']` for a bypassed run with two candidate names.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The `bypassing` set-comprehension (claude_provider.py:275-281) unions and sorts cf-* names across every qualifying call_id in `named.items()`, but no test in TestAWorkflowRunWithoutTheRouter or elsewhere drives two distinct successful cf-* Skill calls in one transcript.

Problem (now): `_bypassing_names` (claude_provider.py:337-342) iterates `named.items()` and unions names from every call_id whose result is `False`. All new bypass tests in TestAWorkflowRunWithoutTheRouter drive either a single qualifying call, or a pair where one call is excluded (it errored), leaving exactly one call contributing to the set each time.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, `test_several_workflow_names_in_one_call_are_reported_as_ambiguous` constructs exactly this scenario -- one Skill call with input `{'a': 'cf-generate', 'b': 'cf-review'}` -- and asserts `skill_state == 'bypassed'` and `skill_match_other_candidates == ['cf-review']`, directly exercising the inner `for name in found` loop for a single call yielding two distinct bypass candidates.

</details>

tests/prompts/cf-ux/README.md
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, the README diff explicitly states this: 'skill_match_other_candidates is populated for ran and for bypassed, and is empty whenever the verdict rested on a single name.' This directly documents the exact behavior the check worried was left unstated, and matches the code (the bypassed `_SkillTrace` does pass a non-empty `other_candidates` when more than one name qualifies).

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `results[...] = bool(block.get('is_error'))` treats an absent is_error key identically to an explicit `false`. This previously-flagged ambiguity for the 'ran' path is reused unchanged by the new bypass branch, so a truncated/schema-shifted tool_result missing is_error would let an unconfirmed direct cf-* workflow call be graded 'bypassed' (a pass) instead of 'failed'.

Problem (now): `_skill_trace` builds `results[tool_use_id] = bool(block.get('is_error'))` (claude_provider.py:245), which maps a missing `is_error` key to `False` exactly like an explicit `is_error: false`. This weakness already existed for the `ran` path; the new `_bypassing_names` function (lines 337-342) reuses the same `results` dict and the same `results.get(call_id) is False` identity check as its sole confirmation that a cf-* workflow call succeeded.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 15a84ba85830556cca31126b79717a000b5f36f2 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to the check's premise, `test_one_workflow_failing_does_not_hide_another_succeeding` directly drives two distinct call_ids (A names 'cf-generate' and errors; B names 'cf-documenting-review' and succeeds) and asserts `skill_state == 'bypassed'` with `skills_invoked == ['cf-documenting-review', 'cf-generate']` -- confirming the successful call correctly produces the bypass verdict while the failed sibling's name still surfaces (via `names`, which is computed from all named calls regardless of success) rather than being silently dropped.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -702,3 +722,140 @@ def _dies(cmd, **kwargs):
...
+            _stream(_skill_call(skill="cf-"), _skill_result(), _result()))
+
+        assert "output" not in out
+        assert out["metadata"]["skill_state"] == "failed"
+
+    def test_several_workflow_names_in_one_call_are_reported_as_ambiguous(self, run_provider):
+        """The same transparency the `ran` path has: a verdict resting on one of
+        several candidates is said out loud."""
+        call = {"type": "assistant", "message": {"content": [
+            {"type": "tool_use", "id": "t1", "name": "Skill", "input": {
```
ainetx, line 836: ### _LOG_AMBIGUOUS_BYPASS warning text is never asserted by any test

**Severity:** Minor

**Problem**
The new _LOG_AMBIGUOUS_BYPASS log call (claude_provider.py ~283-289) has its own %r/%s formatting contract and argument order (name, others, bypassing[0]), mirroring _LOG_AMBIGUOUS_MATCH which IS covered by caplog assertions for the 'ran' path. The corresponding bypass-path test, test_several_workflow_names_in_one_call_are_reported_as_ambiguous, only checks metadata fields and never inspects caplog.

<details>

**How to reproduce**
1. Introduce a bug in _LOG_AMBIGUOUS_BYPASS's argument order or format string, or delete the logger.warning(_LOG_AMBIGUOUS_BYPASS,...) call entirely.
2. Run the full test suite.
3. Observe test_several_workflow_names_in_one_call_are_reported_as_ambiguous still passes because it never reads caplog.

**Expected behavior**
The ambiguous-bypass test should assert on caplog.text (or records) the same way the analogous ran-path ambiguity tests and test_the_bypass_is_logged_by_name do, to pin the log message's content and formatting.

**Actual behavior**
Only metadata assertions are made; the log call's correctness and even its continued existence are unverified.

```text
_bypassing_names finds >1 candidate -> logger.warning(_LOG_AMBIGUOUS_BYPASS,...) fires -> [no test reads caplog] -> regression in log formatting/order/removal goes undetected
```

**Impact**
A future refactor could silently break or remove the ambiguous-bypass diagnostic log without any test failing, reducing on-call triage value.

**Suggested correction**
Add `caplog` fixture usage to test_several_workflow_names_in_one_call_are_reported_as_ambiguous and assert on the expected substrings from _LOG_AMBIGUOUS_BYPASS % (...).

**How to verify**
Re-run the test after adding the caplog assertion; confirm it fails when the log call is removed or its args are reordered.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -702,3 +722,140 @@ def _dies(cmd, **kwargs):
...
+
+    def test_the_bypass_is_logged_by_name(self, run_provider, caplog):
+        """A bypass is graded like a pass, so the log is where a person finds out
+        it happened without diffing metadata."""
+        with caplog.at_level("WARNING", logger=claude_provider.logger.name):
+            run_provider(_stream(
+                _skill_call(skill="cf-documenting-review"),
+                _skill_result(),
+                _result(),
+            ))
```
ainetx, line 799: ### _bypassing_names() helper has no isolated unit test

**Severity:** Minor

**Problem**
_bypassing_names(named, results) was extracted as a standalone top-level function with a detailed docstring specifying several distinct contract rules (every candidate must be cf-*, a bare 'cf-' prefix does not count, and the call's result must be a clean/non-error resolve). No test calls this function directly with hand-built named/results dicts; all coverage comes through the full run_provider -> _skill_trace pipeline.

<details>

**How to reproduce**
1. Grep tests/test_cf_ux_claude_provider.py for `_bypassing_names(` -- no direct invocations found.
2. All bypass-related assertions in TestAWorkflowRunWithoutTheRouter build a full JSON event stream and call run_provider, never the helper directly.

**Expected behavior**
A direct unit test importing claude_provider and calling _bypassing_names(named, results) with hand-constructed dicts to pin its contract independently of the higher-level pipeline.

**Actual behavior**
The helper's contract is only exercised incidentally via end-to-end stream tests, making some edge cases (e.g., a call whose result value is None rather than False) less precisely pinned.

```text
_skill_trace -> _bypassing_names(named, results) -> [no direct test] -> contract pinned only via full pipeline tests
```

**Impact**
Future refactors of _skill_trace's control flow could silently stop exercising some of the helper's edge-case branches without any test noticing.

**Suggested correction**
Add a small test class/function that imports claude_provider._bypassing_names and asserts on its return value directly for the documented edge cases (bare prefix, mixed candidates, missing result).

**How to verify**
Confirm the new direct test fails if _bypassing_names' filtering logic is altered, independent of _skill_trace's surrounding control flow.

</details>

ainetx (approved):
Nice piece of work — the new `bypassed` state is well-motivated (real pilot behavior), the docstrings carry the design intent, and the tests read as genuine specs rather than boilerplate. Nothing here blocks merging; a few small follow-ups worth a glance:

- **Cross-call union in `_bypassing_names` isn't exercised** — every new bypass test drives a single qualifying call or a pair where only one contributes, so the multi-call union path in `_bypassing_names` (claude_provider.py:337-342) is untested. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044306028))
- **Missing `is_error` reads as success for `bypassed` too** — `_skill_trace` treats an absent `is_error` key the same as an explicit `false`; this pre-existing ambiguity now also gates the new bypass grade via the same `results.get(call_id) is False` check. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044310005))
- **README state table glosses over error-mark precedence** — the new table describes `absent` as simply "no Skill call," but the code checks the error mark against raw text before checking for zero calls, so a no-call transcript with `cf` in its prose actually lands as `failed`. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044310284))
- **`_LOG_AMBIGUOUS_BYPASS` text is untested** — the ambiguous-bypass test only checks metadata, unlike its `ran`-path counterpart which asserts on `caplog`. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044723046))
- **`_bypassing_names` has no direct unit test** — its several contract rules (cf-* only, bare prefix excluded, clean result required) are only exercised indirectly through the full `run_provider` pipeline. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044729919))

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -238,11 +257,6 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
         return _SkillTrace("absent", names, inputs, f"no {_SKILL_TOOL} tool call in the transcript")
```
ainetx, line 257: ### Whole-transcript `_SKILL_ERROR_MARK` check preempts `bypassed` detection for a router attempt that errors before a successful direct workflow call

**Severity:** Major

**Problem**
`_skill_trace` in tests/prompts/cf-ux/providers/claude_provider.py checks `_SKILL_ERROR_MARK.search(raw)` unconditionally at the very start (line 251), before evaluating any of the per-call evidence (`named`, `results`) that the rest of the function -- including the new `bypassed` detection via `_bypassing_names` -- relies on. If the raw transcript text contains the literal CLI error signature `<error>Execute skill: cf</error>` anywhere, the whole run is classified `failed`, even if a later, distinct `Skill` call directly and successfully invoked a `cf-*` workflow (e.g. `cf-documenting-review`) and produced a real, scoreable answer. This directly contradicts the design rationale the same PR states just a few lines later for why the bypass check was moved ahead of `not targeted`: to avoid a weaker/coarser signal 'skip the check entirely and lose a real bypass along with a gradeable answer.'

<details>

**How to reproduce**
Build a transcript (via the test helpers) as: (1) `_skill_call("t1", skill="cf")` followed by `_skill_result("t1", is_error=True)`, where the error surface also embeds the literal text `<error>Execute skill: cf</error>` in the stream (as `test_the_real_failure_signature_is_recognised` demonstrates is the real CLI signature), (2) a second call `_skill_call("t2", skill="cf-documenting-review")` with `_skill_result("t2")` (clean), and (3) `_result("findings for 0001-postgres.md")` as the final successful answer. Run this through `_skill_trace`/`call_api`.

**Expected behavior**
Per the PR's own stated intent for `bypassed` ('Studio did the work... worth grading' and 'not `failed`, which would bury a distinct finding among the runs where nothing of Studio was reached at all'), this should be classified `bypassed`, with `skills_invoked == ["cf-documenting-review"]` and the answer graded.

**Actual behavior**
`_SKILL_ERROR_MARK.search(raw)` matches on the literal 'Execute skill: cf' text left over from the router's failed attempt, and `_skill_trace` returns `_SkillTrace("failed",...)` immediately (line 251-255), before `_bypassing_names` is ever called. The answer is discarded to `metadata["unscored_output"]` and `skill_state` is `"failed"` instead of `"bypassed"`.

```text
raw contains '<error>Execute skill: cf</error>' (from failed router attempt)
-> _SKILL_ERROR_MARK.search(raw) matches -> return ("failed",...) [line 251]
-> _bypassing_names(named, results) is NEVER reached, even though
call "t2" = Skill(cf-documenting-review) succeeded cleanly afterward
```

**Impact**
Undermines the exact goal this PR introduces `bypassed` for: a real, gradeable Studio answer reached via a direct workflow call is discarded and misfiled as `failed` whenever the router was attempted and errored first in the same run -- arguably the most realistic bypass scenario (model tries the router, it errors, model falls back to the workflow directly) is the one case guaranteed to be misclassified. This silently suppresses exactly the kind of pilot-run signal the PR's own test class (`TestAWorkflowRunWithoutTheRouter`) was written to surface.

**Suggested correction**
Move the `_SKILL_ERROR_MARK` check to run after (or fold it into) the per-call evidence pipeline, bounding it to calls that are not superseded by a later clean `cf-*` result -- e.g. only treat the error-mark match as authoritative when no call in `named`/`results` independently qualifies as `ran` or `bypassed`, mirroring the reordering already applied to the `targeted`/`not targeted` check.

**How to verify**
Add a test mirroring `test_a_bypass_survives_an_unrelated_erroring_false_cf_match` but using the real `<error>Execute skill: cf</error>` signature from a failed router attempt followed by a successful direct `cf-*` call, and assert `skill_state == "bypassed"` and that the answer is graded (not discarded to `unscored_output`).

</details>

ainetx (changes requested):
Requesting changes — the bypass detection this PR adds is short-circuited by an earlier check, so it can miss the exact case it was built for.

- **Whole-transcript error marker preempts bypass detection** — `_skill_trace` checks `_SKILL_ERROR_MARK.search(raw)` against the *entire* transcript up front, before any of the per-call evidence the new `bypassed` logic relies on. If the literal string `<error>Execute skill: cf</error>` shows up anywhere — even from an unrelated, earlier router attempt — the whole run is marked `failed`, even when a later `Skill` call cleanly invoked a `cf-*` workflow directly and produced a real, gradeable answer. This is the same failure mode the PR's own comments say the bypass check was deliberately ordered to avoid ("skip the check entirely and lose a real bypass along with a gradeable answer"), just tripped by a coarser, whole-transcript signal instead of the per-call one that was fixed. ([discussion](https://github.com/constructorfabric/studio/pull/240#discussion_r4051283555))

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1783efd13c19b91fea213c78aca29799745bb532 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff includes a `TestBypassingNamesDirectly` class (test_cf_ux_claude_provider.py:902-933) that unit-tests `_bypassing_names` directly, including `test_two_clean_calls_are_unioned` which passes two distinct call_ids ({'a': ['cf-generate'], 'b': ['cf-review']}, both results False) and asserts the union `['cf-generate', 'cf-review']` — directly exercising the exact set-comprehension/aggregation logic across multiple simultaneously-successful calls that this check asks about.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `_skill_trace` builds `results[tool_use_id] = bool(block.get('is_error'))` (claude_provider.py:245), which maps a missing `is_error` key to `False` exactly like an explicit `is_error: false`. This weakness already existed for the `ran` path; the new `_bypassing_names` function (lines 337-342) reuses the same `results` dict and the same `results.get(call_id) is False` identity check as its sole confirmation that a cf-* workflow call succeeded.

Problem (now): `_skill_trace` builds `results[block.get('tool_use_id')] = bool(block.get('is_error'))` (claude_provider.py:245), which maps a tool_result block that omits the `is_error` key entirely to `False`, identically to an explicit `is_error: false`. This ambiguity previously existed for the `ran` path only; the new `_bypassing_names` helper (lines 316-342) reuses the exact same `results` dict and the identical `results.get(call_id) is False` identity check as its sole confirmation that a cf-* workflow call actually succeeded, extending the weak-evidence pattern to the newly introduced `bypassed` (graded-as-pass) state.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1783efd13c19b91fea213c78aca29799745bb532 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** test_the_ambiguous_bypass_warning_names_the_reported_one (test_cf_ux_claude_provider.py:886-899) uses caplog to assert 'rests on more than one candidate' (a substring unique to _LOG_AMBIGUOUS_BYPASS) is present and 'matched \'cf\'' (the ran-path message) is absent, directly confirming the new warning fires and is distinguishable from the ran-path message, contradicting the premise that no test asserts on this log text.

</details>

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1783efd13c19b91fea213c78aca29799745bb532 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The diff adds `TestBypassingNamesDirectly` (test_cf_ux_claude_provider.py:902-933), a dedicated unit-test class calling `claude_provider._bypassing_names(named, results)` directly with hand-constructed dicts, covering a clean call, two clean calls unioned, an errored call, a missing-result call, mixed cf-*/non-cf-* names, the bare router name, a bare prefix, and an empty found list -- directly pinning the helper's contract in isolation.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 1783efd13c19b91fea213c78aca29799745bb532 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** This precedence is explicitly deliberate and documented: README lines 132-136 state '<error>Execute skill: cf</error> anywhere in the transcript is failed before any of this is considered -- including before bypassed, so a router that errored and a workflow that then succeeded is a cf failure, not a routing observation.' test_a_router_error_outranks_a_successful_direct_workflow (test_cf_ux_claude_provider.py:863-883) directly pins exactly this scenario and expects skill_state == 'failed' with the workflow name still surfaced in skills_invoked -- confirming the interaction was considered and intentionally decided, not an oversight.

</details>

tests/test_cf_ux_claude_provider.py
```diff
@@ -702,3 +722,212 @@ def _dies(cmd, **kwargs):
...
+            _result("findings"),
+        ))
+
+        assert out["metadata"]["skill_state"] == "bypassed"
+        assert out["metadata"]["skills_invoked"] == ["cf-documenting-review", "cf-generate"]
+
+    def test_the_prefix_is_derived_from_the_router_name(self):
+        """Two hand-maintained literals kept in sync by a comment is a rename
+        away from silently breaking bypass detection."""
+        assert claude_provider._SKILL_WORKFLOW_PREFIX == f"{claude_provider._SKILL_NAME}-"
```
ainetx: ### Test asserts a tautological equality between private module constants instead of exercising bypass-detection behavior

**Severity:** Minor

**Problem**
`test_the_prefix_is_derived_from_the_router_name` (test_cf_ux_claude_provider.py:858-861) asserts `claude_provider._SKILL_WORKFLOW_PREFIX == f'{claude_provider._SKILL_NAME}-'`, reaching directly into two underscore-prefixed module internals. Since the source defines `_SKILL_WORKFLOW_PREFIX = f"{_SKILL_NAME}-"` (claude_provider.py:95), this assertion is tautologically true given the current definition and does not exercise `_skill_trace` or `_bypassing_names` at all. Every other test in the file — including its siblings in `TestAWorkflowRunWithoutTheRouter` — asserts only through `run_provider`'s returned `out`/`metadata` (the module's public surface); the only other private-internal access in the file is `TestBypassingNamesDirectly` which actually calls the helper function to exercise its logic, not merely compares two literals.

<details>

**How to reproduce**
1. Rename `_SKILL_NAME` in claude_provider.py without updating `_SKILL_WORKFLOW_PREFIX`'s derivation logic (or vice versa) in a way that decouples the two.
2. Note that `test_the_prefix_is_derived_from_the_router_name` will catch only a literal-string mismatch between the two constants, not whether bypass detection (`_bypassing_names`, `_skill_trace`) still correctly classifies a renamed router's workflows as `bypassed`.
3. Observe the docstring claims to guard against 'a rename... silently breaking bypass detection', but the test never invokes any bypass-detection code path.

**Expected behavior**
The guarantee that a router rename doesn't silently break bypass detection should be tested through observable behavior (e.g. run a stream where a `cf-*`-prefixed workflow is invoked and assert `skill_state == 'bypassed'`), consistent with how every other behavioral test in the file operates.

**Actual behavior**
The test directly compares two private module-level literals for equality, which is guaranteed true by their very definition in source and exercises no runtime code path.

```text
_SKILL_WORKFLOW_PREFIX = f'{_SKILL_NAME}-' (source) -> test re-evaluates the same f-string -> assert equal -> always true regardless of whether bypass detection actually works correctly
```

**Impact**
Low: creates an informal, unenforced contract between the test suite and module-private internals, and provides a false sense of security that router-rename safety for bypass detection is covered, when it is not actually exercised.

**Suggested correction**
Replace or supplement the test with one that renames/parameterizes the router name (or otherwise drives `_invoke`/`_skill_trace` through the public provider surface) and asserts a `cf-*`-prefixed workflow is still classified `bypassed`.

**How to verify**
Confirm no other test in the file accesses `claude_provider._SKILL_WORKFLOW_PREFIX`/`_SKILL_NAME` directly, and that the replacement test fails if the prefix derivation is broken while the observable classification also breaks.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
```diff
@@ -256,6 +270,39 @@ def _skill_trace(events: list[dict[str, Any]], raw: str) -> _SkillTrace:
...
+    # (no gate, no menu, no routing decision -- the thing this suite measures
+    # never happened), and not `failed`, which would bury a distinct finding
+    # among the runs where nothing of Studio was reached at all.
+    #
+    # Checked here rather than under `not targeted`, which is where it first
+    # went: `targeted` accepts the documented false positive -- any field, any
+    # depth -- so a single unrelated call carrying `cf` somewhere, and erroring,
+    # was enough to skip the check entirely and lose a real bypass along with a
+    # gradeable answer.
+    bypassing = _bypassing_names(named, results)
```
ainetx: ### 'bypassed' verdict can mask a genuine router (`cf`) failure when the error isn't the literal CLI marker text

**Severity:** Major

**Problem**
`_skill_trace` only forces a 'failed' verdict ahead of 'bypassed' when the raw transcript contains the literal string `Execute skill: cf` (via `_SKILL_ERROR_MARK`, checked once against `raw` at the top of the function). Any other error result on the call that named `cf` (e.g. an ordinary `tool_result` with `is_error: true` and different error content) is silently skipped by the `targeted` loop (`continue` at line 262) rather than treated as evidence the router failed. If a separate `cf-*` workflow call in the same transcript then returns cleanly, `_bypassing_names` (called at line 284) finds that clean workflow call and the function returns `bypassed`, whose documented meaning is 'the router never did [run]' — which is false in this scenario, since the router was invoked and did error.

<details>

**How to reproduce**
Construct a transcript with: (1) a `Skill` tool_use call naming `cf`; (2) a `tool_result` for that call with `is_error: true` and content that is any error text NOT containing the literal substring `Execute skill: cf` (e.g. `"tool timed out"` or `"permission denied"`); (3) a second `Skill` tool_use call naming `cf-documenting-review` (or any `cf-` prefixed workflow); (4) a clean `tool_result` for that second call; (5) a final successful `result` event with text output. Run this through `_skill_trace`/`_invoke` as the existing test harness (`run_provider`, `_stream`, `_skill_call`, `_skill_result`) does.

**Expected behavior**
Because the router (`cf`) was actually invoked and its own call errored, `skill_state` should be `failed` (or at minimum should not claim the router 'was never invoked'), consistent with the precedence already enforced for the literal-marker case in `test_a_router_error_outranks_a_successful_direct_workflow`.

**Actual behavior**
`_skill_trace` returns `state='bypassed'` with detail `"'cf' was never invoked; 'cf-documenting-review' ran directly"`, and `_invoke` grades and scores the run as if it were a clean Studio bypass (since `bypassed` is in `_GRADED_STATES`), reporting `skill_state: bypassed` in the metadata — actively misrepresenting that the router never ran when it did run and failed.

```text
cf tool_use --> tool_result(is_error=true, content='timeout') --> _SKILL_ERROR_MARK doesn't match raw --> targeted loop sees results[cf_id] != False --> continue (no 'ran') --> falls through to _bypassing_names() --> finds clean cf-documenting-review call --> returns 'bypassed' (masks the router failure)
```

**Impact**
Undermines the core purpose of this PR, which is to make `skill_state` a trustworthy signal for triaging router regressions. Any router failure whose error text doesn't happen to match the one hardcoded CLI string will be reported as a routing bypass rather than a router failure, hiding real regressions from anyone monitoring `skill_state`/the WARNING log for `cf` failures, and instead surfacing a misleading 'bypass' warning that suggests only a benign routing-skip occurred.

**Suggested correction**
Extend the failure precedence beyond the literal raw-text marker: after building `results`, check whether any call in `targeted` (i.e. any call that named `cf`) has `results.get(call_id) is True` (an observed error), and if so return `failed` before evaluating `_bypassing_names`, mirroring the same precedence already given to `_SKILL_ERROR_MARK`. Alternatively, fold the raw-text marker check into the same per-call evidence model so there is a single source of truth for 'the router call errored'.

**How to verify**
Add a unit test mirroring `test_a_router_error_outranks_a_successful_direct_workflow` but using an ordinary (non-marker) `is_error: true` tool_result for the `cf` call alongside a clean `cf-*` workflow call, and assert `skill_state == 'failed'` (not `bypassed`).

</details>

vasylcf:
Two rounds to answer — the previous push (`1783efd`) went up without a reply, my oversight.

## Round two, taken in `1783efd`

- **The whole-transcript error mark outranks a successful direct workflow.** Correct, and now pinned by a test rather than left to the order of the branches.
- **`_LOG_AMBIGUOUS_BYPASS` had no test asserting its text**, including the part that matters: it must not read "matched 'cf'", which is what the `ran` message says and the one thing that did not happen in a bypass.
- **`_bypassing_names` had no isolated test.** Eight now: the router itself, a bare prefix, an errored call, a resultless call, a call naming something else, an empty candidate list, and the cross-call union.

## Round three, taken in `0795c2a` — and it caught me contradicting myself

**The Major is right.** I argued last round that a router which failed outranks a bypass. That held *only* when the transcript carried the CLI's literal marker text. Without it:

```
cf invoked, came back an error + cf-documenting-review succeeded
  ->  bypassed
  ->  detail: "'cf' was never invoked"
```

The opposite of what happened, stated as fact.

The fix satisfies this finding **and** the earlier one about false positives at the same time, which is what took the thought. Only an **unambiguous** `cf` call suppresses a bypass — one that names `cf` and nothing else. A call naming `cf` *among other candidates* is the documented false positive (any field, any depth), and letting that outrank a bypass was the earlier defect, where one unrelated erroring call hid a real one.

Four shapes, all verified:

| transcript | verdict |
|---|---|
| `cf` errored + workflow succeeded | `failed` — "every 'cf' call came back as an error" |
| false-positive `cf` errored + workflow succeeded | `bypassed` — "no unambiguous 'cf' call" |
| no `cf` at all + workflow succeeded | `bypassed` — "'cf' was never invoked" |
| `cf` clean | `ran` |

The detail no longer claims more than the evidence carries.

**And the test criticism lands.** `test_the_prefix_is_derived_from_the_router_name` asserted `_SKILL_WORKFLOW_PREFIX == f"{_SKILL_NAME}-"` — a restatement of the assignment, proving nothing. Replaced with behaviour: rename the router to `zz`, and `zz-*` becomes the bypass family while `cf-*` stops being one. Reviewing my own additions for the same fault, I found an `or True` in another new test and removed it.

`6295 passed`; spec-coverage green.

tests/test_cf_ux_claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0795c2abec0dca95fad05cc8fc08a52d7414a94f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Reading tests/test_cf_ux_claude_provider.py:858-871 (the actual current content at the target lines), the test 'test_the_prefix_is_derived_from_the_router_name' referenced by this check no longer exists in the diff; it has been replaced by test_a_rename_of_the_router_moves_the_workflow_family_with_it, which only uses monkeypatch.setattr on private constants for test *setup* (simulating a router rename) and asserts exclusively through run_provider's returned metadata (skill_state == 'bypassed'/'failed') -- exactly the observable-behavior pattern the check calls for. The premise (a tautological private-attribute assertion) does not match the current code.

</details>

tests/prompts/cf-ux/providers/claude_provider.py
ainetx: **RESOLVED** -- Confirmed fixed as of 0795c2abec0dca95fad05cc8fc08a52d7414a94f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Traced _skill_trace lines 293-297: 'attempted' captures any targeted call where named[call_id] == [_SKILL_NAME] (unambiguous) AND results.get(call_id) is not False (i.e. errored or missing) -- when attempted is non-empty, bypassing is forced to [] and the code falls through to a 'failed' verdict instead of reaching _bypassing_names. This is directly exercised by test_a_router_call_that_errored_outranks_a_bypass (lines 945-957), which uses an ordinary is_error=True tool_result (content '...', not the literal '<error>Execute skill: cf</error>' marker) alongside a successful cf-documenting-review call, and asserts skill_state == 'failed'. The README (lines 132-137) also documents this precedence explicitly ('so is an unambiguous cf call that came back an error or did not come back'). The check's premise -- that an ordinary is_error result would fall through unguarded to _bypassing_names and be misreported as bypassed -- does not hold against the current code.

</details>

ainetx (approved):
Solid diff — introducing the `bypassed` state cleanly separates "Studio did the work without the router" from both a clean pass and an outright failure, and the added tests cover precedence and ambiguity well. Two minor notes worth a glance, neither blocking:

- **`is_error` absence reads as success, now feeding the `bypassed` grade too** — `_skill_trace`'s `results` dict maps a missing `is_error` key the same as `is_error: false`, and the new `_bypassing_names` helper relies on that same weak signal as its only confirmation a `cf-*` call succeeded. Worth tightening at some point so a malformed/absent field doesn't silently earn a passing grade. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044310005))
- **README's new state table doesn't mention error-mark precedence over `absent`** — the code checks `_SKILL_ERROR_MARK` before checking for zero calls, so a transcript with no `Skill` call but matching error text in its prose classifies as `failed`, not `absent` as the table implies. A one-line caveat would keep the table accurate. ([comment](https://github.com/constructorfabric/studio/pull/240#discussion_r4044310284))

tests/prompts/cf-ux/README.md
vasylcf: Fixed in e93b2ee0.

The precedence was already stated in the prose under the table ("`<error>Execute skill: cf</error>` anywhere in the transcript is `failed` before anything else is considered"), but you are right that the row is what a person reads while diagnosing a transcript, and the row denied it. It now says so itself:

> `absent` -- No `Skill` call at all *and* no error mark in the raw text. The mark is checked first, so a transcript that made no call but whose prose contains `<error>Execute skill: cf</error>` is `failed`, not `absent`.

tests/prompts/cf-ux/providers/claude_provider.py
vasylcf: The mechanism is described correctly, and the suggested fix is the one I shipped on #229 and had to revert -- which is the part worth recording here, because the write-up has now been re-verified three times against a conclusion that measurement contradicts.

`is_error` is not a field that a successful `tool_result` sets to `false`. It is a field that a successful `tool_result` **omits**. Measured against claude-code 2.1.276: a clean `Skill` result carries no `is_error` key at all; the key appears only when the call failed. So `block.get("is_error") is False` -- requiring the key to be present and false -- is true for no successful run in the corpus. I made exactly that change, and it graded every passing run as a failure. The revert and a public correction are on #229.

That inverts the finding's premise. Absence is not weak evidence of success here; absence *is* the success signal, and `bool(...)` is the reading that matches the schema.

What remains true is the residual your reproduction actually depends on: a **truncated or schema-shifted** result, where the key is missing for a different reason. That case is not decided by this line, though -- it is refused before `skill_state` is reached:

- a transcript with no terminal `result` event is not graded at all;
- a `result` whose `subtype` is present and not `success` is an error, not a verdict (`claude_provider.py:486-491`);
- lines that will not parse are counted in `unparsed_lines` (`claude_provider.py:434`) precisely because a lost line can be a lost `tool_result`.

So for a truncated transcript to be graded `bypassed`, it would have to truncate cleanly enough to produce a terminal `result` with `subtype: success` and no unparsed lines, while dropping a field from a block it did emit. I have no example of that shape, and I would rather be shown one than defend against it by breaking every real run.

Two things I am not claiming. The ambiguity is genuinely there -- a schema change at the other end that starts emitting `is_error: false` explicitly, or one that drops it from failures, would be read wrongly and silently. And `bypassed` does widen what it reaches, exactly as you say. What I am saying is that the distinguishing signal does not exist at this layer today, and the fix that assumes it does is measurably worse.

If the schema gains a positive success marker, the tri-state is the right reading and I will take it then.

ainetx: @vasylcf **RESOLVED** -- Agreed.

<details>
<summary>Why</summary>

The repo's own history (pr-229/ and pr-229/, surfaced via search_historical_findings) confirms the human's account: treating an absent `is_error` as success in `_skill_trace` was a deliberate change justified by empirically-observed claude-code 2.1.276 behavior where successful tool_result blocks omit `is_error` entirely and only failures set it. That directly inverts the finding's premise that omission is weak/ambiguous evidence, and corroborates the claim that requiring an explicit `is_error: false` (the finding's suggested fix) would misgrade real passing runs. Separately, the code at claude_provider.py:435-460 and 488-493 shows the specific truncation scenario the finding's repro needs (a schema-shifted tool_result missing `is_error` for a genuinely failed/incomplete run) is already screened out before `skill_state` matters — no terminal `result` event aborts grading entirely, and any present non-`success` subtype is treated as an unscored error — so the narrow case the finding requires to matter isn't demonstrated to survive those guards.

Noted -- recording this as a repo-wide rule for future reviews: "Before flagging ambiguous handling of an absent/optional field as a correctness gap, check whether the observed behavior of the upstream tool/library at the pinned version actually supports the field being omitted only in one specific case (e.g. only on success or only on failure) — a deliberate absence-as-signal convention validated against real output is not the same as weak or missing evidence, and requiring an explicit value instead would misgrade the common case."

</details>

tests/prompts/cf-ux/README.md
ainetx: @vasylcf **RESOLVED** -- Confirmed fixed as of e93b2ee05f7ec8d6035bdf9a42d20ea879e24fd5 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** The new README explicitly states in the 'absent' row that no error mark must exist and that the mark is checked first, so a no-call transcript containing '<error>Execute skill: cf</error>' is classified as 'failed', not 'absent'.
- **Verifier:** The README's 'absent' row (README.md:129, part of this diff) already explicitly states: "No `Skill` call at all — *and* no error mark in the raw text. The mark is checked first, so a transcript that made no call but whose prose contains `<error>Execute skill: cf</error>` is `failed`, not `absent`." This is the exact concrete precedence fact the check demands, stated verbatim in the row's own description, matching the code order in claude_provider.py (lines 251-257, error-mark check before the 'not calls' check) and the pre-existing test `test_the_real_failure_signature_is_recognised` which pins this exact behavior. The prior round's accepted-finding and its stated evidence ('the table describes absent simply as No Skill call at all with no mention of this precedence') is factually contradicted by the actual table row text present in this diff — it does mention the precedence, in the same sentence. The reviewer opinion in this round (PASS) correctly identifies this.

</details>

PR #241, opened 2026-09-18
fix(build): make the prompt pilot and the suite runnable from a clean clone

vasylcf:
Four things that each cost a real run before they were diagnosed. None of them is the skill under test; all of them are the reason it took hours to find that out.

## 1. `node` — existing is not the same as usable

`check-prompt-tests` proved `node` existed. Two ways that is not enough:

- **Too old.** promptfoo's `engines.node` is `>=22.22.0`. An older node fails deep inside npx, talking about the package rather than about node.
- **Not a program at all.** Under nvm's lazy loader, `node` is a **shell function** that sources nvm and re-dispatches. An interactive shell answers `node --version`; `make`, which runs recipes in `sh`, finds nothing on PATH. Measured here: `shutil.which("node") is None` while the shell reports `v20.20.0`. `command -v node` agrees with whichever of the two asked it, which is why the check passed and the run still failed.

New `preflight.py` reports the version it actually found and the path of an installed nvm node that qualifies — as a command to run:

```
no usable `node` program on PATH. promptfoo needs >= 22.22.0.
(Under nvm's lazy loader `node` is a shell function, so an interactive shell
 answers it and `make`, which runs recipes in `sh`, does not.)
nvm has 24.21.0 installed. Put it on PATH for the command:

  PATH=/home/…/.nvm/versions/node/v24.21.0/bin:$PATH make test-prompts
```

## 2. The codex model is checked before the sandboxes are built

A withdrawn slug costs all 8 codex cases, and surfaces as a 400 inside a promptfoo table cell **after** every sandbox has been created. The preflight reads `codex`'s own model cache — no network, no cost — and names what the account does have:

```
codex model 'gpt-5.4-mini' is not among the ones this account is entitled to.
  ~/.codex/models_cache.json lists: gpt-5.5, gpt-5.6-luna, gpt-5.6-sol, gpt-5.6-terra, gpt-6-astra
  Pick one and set it:

    CF_UX_CODEX_MODEL=gpt-5.5 make test-prompts
```

It is **silent** on a missing, unreadable or unfamiliar cache. Ignorance is not evidence, and a preflight that blocks a run on its own is worse than the 400 it was meant to pre-empt. It also never suggests the entries the CLI hides from people (`codex-auto-review`, `gpt-reserve`) — both pinned by tests.

## 3. `make test` accepted exactly one pytest

`check-pytest` required `~/.local/bin/pytest` **and** pipx, so a clean worktree got `ERROR: pytest binary not found` with `make install` as the only advice — for a pytest already on PATH. Now: pipx first (what CI installs), then PATH, then `python3 -m pytest`. Overridable with `make test PYTEST=/path/to/pytest`.

## 4. `pytest-xdist` was a hard gate on an optimisation

Without the plugin the suite did not run **at all**, when it could have run serially. Now `-n 6` when it is present, a note and a slower run when it is not.

## 5. README

The prerequisites lived in comments inside `promptfooconfig.yaml`. They are now where someone looks first, as a table, with the `CF_UX_*` knobs.

## Verified on this machine, where all four bit

- `make test` → **6282 passed, 4 skipped, 15 xfailed**, exit 0 — run *serially*, with no xdist installed. Before this change the same command exited 1 without running a test.
- `make check-prompt-tests` → passes with a qualifying node on PATH and a live model; fails with the exact command to fix it otherwise.
- 14 new tests for the preflight itself, including that it stays silent on every uncertainty.

## Note

The `gpt-5.4-mini` default that example output shows is fixed separately in #239. This PR adds the check that would have caught it.

## Summary by CodeRabbit

- **New Features**
  - Added preflight checks for supported Node.js versions and configured Codex model availability before prompt tests run.
  - Added guidance for resolving missing or incompatible environment requirements.
  - Added configurable model, effort, context, and plugin settings documentation.

- **Bug Fixes**
  - Tests now locate an available pytest installation and run serially when parallel execution is unsupported.
  - Prompt test checks no longer require Node detection before running preflight validation.

- **Tests**
  - Added coverage for Node.js validation, Codex model checks, cache handling, and exit codes.


tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,172 @@
...
+def _version(text: str) -> tuple[int, int, int] | None:
+    found = _VERSION.search(text or "")
+    return (int(found[1]), int(found[2]), int(found[3])) if found else None
+
+
+def _node_version(binary: str = "node") -> tuple[int, int, int] | None:
+    try:
+        done = subprocess.run([binary, "--version"], capture_output=True, text=True,
+                              timeout=20, check=False)
+    except (OSError, subprocess.SubprocessError):
```
ainetx, line 51: ### New except blocks in preflight.py silently swallow exceptions, matching the project's own banned anti-pattern

**Severity:** Minor

**Problem**
_node_version() and check_codex_model() catch broad exception sets (OSError/SubprocessError, OSError/UnicodeDecodeError/JSONDecodeError) and return bare sentinel values (None, []) with no logging, warning, or re-raise -- exactly the pattern scripts/pylint_plugins/silent_exceptions.py's SilentExceptionSwallowedChecker (W9001) is built to flag. The file is not included in PYLINT_TARGETS, so CI's pylint invocation never lints it.

<details>

**How to reproduce**
1. Run `PYTHONPATH=... pylint --load-plugins=scripts.pylint_plugins.silent_exceptions tests/prompts/cf-ux/preflight.py`. 2. Observe W9001 fired on both except blocks. 3. Note `make check-pylint` only targets src/studio_proxy and skills/studio/scripts/studio, so this never runs in CI.

**Expected behavior**
Per the project's own convention, these handlers should log/report the swallowed exception (e.g. via a visible-signal call) or the file should be added to PYLINT_TARGETS so the existing gate catches any regression.

**Actual behavior**
A real unexpected failure (e.g. a permission error reading the cache, a corrupted subprocess pipe) is silently treated the same as 'nothing to report', with no lint coverage to catch or prevent this.

```text
except OSError/...: return None/[] --> caller treats as clean/negative result --> real failure indistinguishable from healthy no-op --> pylint gate doesn't even see this file
```

**Impact**
Debuggability of the preflight tool itself is undermined -- exactly the failure mode (an unexplained silent 'fine' instead of a diagnostic) the tool exists to prevent for node/npx and codex.

**Suggested correction**
Either add a visible-signal (e.g. print to stderr in debug mode) inside these except blocks before returning the sentinel, or add tests/prompts/cf-ux/preflight.py to PYLINT_TARGETS so silent_exceptions.py enforces this going forward.

**How to verify**
Run pylint with the silent_exceptions plugin against preflight.py and confirm W9001 no longer fires, or confirm the file is now part of the CI-enforced PYLINT_TARGETS.

</details>

Makefile
```diff
@@ -18,8 +18,21 @@ PIPX_BIN_DIR ?= $(HOME)/.local/bin
...
+# pipx first, because `make install` puts it there and CI runs that. But a
+# pytest is a pytest: a virtualenv or a distro package runs this suite just as
+# well, and refusing one because it came from somewhere else cost a clean
+# worktree an `ERROR: pytest binary not found` with `make install` as the only
+# advice -- for a tool already on PATH. Whatever answers is checked for the
+# plugins the targets actually use, which is the part that matters.
+PYTEST ?= $(shell test -x "$(PIPX_BIN_DIR)/pytest" && echo "$(PIPX_BIN_DIR)/pytest" \
+                  || command -v pytest 2>/dev/null \
+                  || echo "$(PYTHON) -m pytest")
+PYTEST_PIPX ?= $(PYTEST)
```
ainetx, line 30: ### PYTEST_PIPX can silently diverge from the newly-validated PYTEST variable

**Severity:** Minor

**Problem**
Makefile line 30 sets `PYTEST_PIPX ?= $(PYTEST)`, so PYTEST_PIPX equals the validated/auto-detected PYTEST only when not independently overridden. PYTEST_PIPX was the sole pre-existing public override and is still the variable directly invoked by test/test-verbose/test-quick/test-coverage (lines 157, 170, 175, 180). If a caller sets `make test PYTEST_PIPX=/some/other/pytest`, check-pytest validates `$(PYTEST)` (still auto-detected, unrelated to the override) while the actual test run uses the overridden, unvalidated binary, and PYTEST_PARALLEL is computed from the wrong binary's `--help` output.

<details>

**How to reproduce**
1. Run `make test PYTEST_PIPX=/venv-without-xdist/bin/pytest`. 2. check-pytest validates $(PYTEST), which auto-detects e.g. pipx's pytest (has xdist) -- passes silently. 3. PYTEST_PARALLEL is computed as '-n 6' based on that pipx pytest. 4. `test` target then runs `/venv-without-xdist/bin/pytest tests/ -n 6...` -- the actual invoked binary lacks xdist support and fails on the unknown -n flag, reintroducing exactly the failure class this diff set out to fix.

**Expected behavior**
check-pytest's validation and PYTEST_PARALLEL derivation should be keyed off the same variable actually invoked by the test targets in every override combination, or PYTEST_PIPX should be documented as deprecated/no longer independently overridable.

**Actual behavior**
check-pytest validates $(PYTEST) unconditionally while test targets invoke $(PYTEST_PIPX), which can diverge when set directly.

```text
user sets PYTEST_PIPX=X (bypassing PYTEST) -> check-pytest validates PYTEST (still auto-detected Y) -> passes -> test target runs X with flags computed for Y -> failure
```

**Impact**
Reintroduces the 'pytest binary not found/not runnable' or wrong -n 6 injection failure class this diff's stated purpose was to eliminate, for any CI script or developer habit still using the previously-public override variable.

**Suggested correction**
Either remove/rename PYTEST_PIPX as a public override (keep it purely internal, always derived from PYTEST) or have check-pytest validate whichever of PYTEST/PYTEST_PIPX is actually used downstream, and document the change.

**How to verify**
Run `make test PYTEST_PIPX=/path/to/pytest-without-xdist` and confirm check-pytest either fails appropriately or the test invocation doesn't pass an unsupported -n flag.

</details>

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+"""Preconditions for the cf-ux prompt pilot, checked before anything is spent.
+
+`make check-prompt-tests` already proves the four binaries exist. Existing is
```
ainetx, line 4: ### preflight.py docstring misattributes node's existence check to the Makefile

**Severity:** Minor

**Problem**
The docstring at preflight.py:4 claims `make check-prompt-tests` already proves 'the four binaries exist' before preflight.py runs, but this same diff removed the Makefile's `command -v node` check. The Makefile target now only verifies claude, codex, and cfs (three binaries) before invoking preflight.py, which is the sole place node's existence is checked.

<details>

**How to reproduce**
1. Read Makefile:341-362 (check-prompt-tests target) post-diff. 2. Note only claude/codex/cfs get `command -v` checks. 3. Read preflight.py:4, which still claims all four binaries are already proven to exist.

**Expected behavior**
Docstring should say three binaries are pre-verified by the Makefile, with node's existence-and-version check owned entirely by this script.

**Actual behavior**
Docstring inaccurately implies node's mere existence is already established before this script runs, which could mislead a developer debugging a node-related failure into looking at the Makefile.

```text
Makefile (checks claude/codex/cfs only) --invokes--> preflight.py (docstring wrongly claims Makefile also checked node) --> check_node() actually does full node check
```

**Impact**
Minor documentation inaccuracy; could waste a developer's time looking for a node existence-check in the Makefile that no longer exists.

**Suggested correction**
Update the docstring to say 'three binaries' (claude/codex/cfs) and clarify that node's existence and version are both checked exclusively by this script's check_node().

**How to verify**
Re-read the docstring and confirm the binary count and check-ownership statement matches the current Makefile check-prompt-tests target.

</details>

Makefile
```diff
@@ -360,6 +359,7 @@ check-prompt-tests:
 		echo ""; \
 		exit 1; \
 	}
+	@$(PYTHON) $(PROMPT_TESTS_DIR)/preflight.py
 
 install-prompt-tests: check-prompt-tests
```
ainetx, line 364: ### install-prompt-tests (a package-caching step) now unconditionally requires codex model entitlement

**Severity:** Minor

**Problem**
install-prompt-tests (Makefile:364) depends on check-prompt-tests, whose recipe now ends with an unconditional call to preflight.py (Makefile:362), which includes check_codex_model. A user with a well-formed models_cache.json that legitimately lists models but doesn't include the configured CF_UX_CODEX_MODEL/default slug will fail install-prompt-tests, despite its only stated job being 'Pre-caching promptfoo@... via npx'.

<details>

**How to reproduce**
1. Have a valid ~/.codex/models_cache.json listing models that do not include the currently configured codex model slug. 2. Run `make install-prompt-tests`. 3. Observe it fails via check-prompt-tests -> preflight.py's codex-model check, despite the target's only real job (`npx promptfoo --version`) having nothing to do with codex.

**Expected behavior**
install-prompt-tests should only require what it actually needs (node/npx availability) to cache the promptfoo package, not codex model entitlement.

**Actual behavior**
install-prompt-tests fails on an unrelated codex-model-entitlement issue because it shares the check-prompt-tests prerequisite with test-prompts.

```text
install-prompt-tests --requires--> check-prompt-tests --runs--> preflight.py --includes--> check_codex_model (unrelated to package caching) --> failure blocks npx caching
```

**Impact**
A user just trying to pre-cache the promptfoo npm package can be blocked by, and confused by, an unrelated codex account/model configuration issue.

**Suggested correction**
Split check-prompt-tests into a lighter prerequisite (binary presence only) for install-prompt-tests, or make install-prompt-tests skip/soften the codex-model check since it doesn't invoke codex.

**How to verify**
Run `make install-prompt-tests` with a models_cache.json missing the configured model slug and confirm whether it fails; if intentional, this should be documented.

</details>

vasylcf:
Both CodeRabbit findings reproduced and fixed in `18006c5`.

## 1. A non-numeric `CF_UX_CODEX_CONTEXT` reached the person as a traceback

Correct, and worth stating plainly: importing the provider **runs** it. Its module body reads the `CF_UX_*` knobs, so `int(os.environ.get("CF_UX_CODEX_CONTEXT"))` raises during the import. Only `ImportError` was caught.

Reproduced before the fix:

```
CF_UX_CODEX_CONTEXT=abc python3 tests/prompts/cf-ux/preflight.py
  ...
  ValueError: invalid literal for int() with base 10: 'abc'
```

A traceback out of the one component whose entire job is to replace tracebacks with a sentence. After:

```
ERROR: the prompt pilot cannot run as configured.

  a CF_UX_* setting is not valid: invalid literal for int() with base 10: 'abc'
  (read while importing the codex provider; check CF_UX_CODEX_CONTEXT)
```

Pinned by `test_a_non_numeric_context_is_a_sentence_not_a_traceback`.

## 2. The README overstated what the preflight covers

Also correct. The table listed *"Both CLIs logged in"* among the prerequisites and the line under it said the preflight checks all of this — which it does not, and cannot for free: finding out costs a call, and a preflight that spends money to say a run may proceed is not a preflight.

Reworded to name what it does check (binaries present, node can run promptfoo, the model is one the account has) and to state the consequence explicitly: a logged-out CLI passes here and fails once the run starts.

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,172 @@
...
+        "  (If the cache is stale, `codex` refreshes it -- this check reads it, "
+        "never the network.)",
+    ]
+
+
+def main() -> int:
+    sys.path.insert(0, str(Path(__file__).resolve().parent / "providers"))
+    try:
+        from codex_provider import DEFAULT_MODEL  # noqa: PLC0415 -- optional import
+    except ImportError:
```
ainetx, line 156: ### Codex model entitlement check silently disables itself when codex_provider fails to import

**Severity:** Major

**Problem**
main() imports DEFAULT_MODEL from codex_provider.py, which is where the real default ('gpt-5.4-mini') lives. On ImportError, the fallback `os.environ.get('CF_UX_CODEX_MODEL', '')` loses that default entirely. If the env var is unset, DEFAULT_MODEL becomes empty string, and `check_codex_model(DEFAULT_MODEL) if DEFAULT_MODEL else []` skips the check with zero indication to the user that it was skipped.

<details>

**How to reproduce**
1. Break or move providers/codex_provider.py (or introduce an unrelated ImportError in its dependency chain, e.g. _sandbox.py). 2. Leave CF_UX_CODEX_MODEL unset. 3. Run `make check-prompt-tests` / preflight.py directly. 4. Observe exit code 0 with no warning, even though the codex-model check never ran.

**Expected behavior**
Either print an explicit warning when the codex_provider import fails and the check is consequently skipped, or fail loudly, since silent skipping defeats the exact failure mode this script exists to catch.

**Actual behavior**
The check silently no-ops, and the user proceeds to a real run that can still hit the withdrawn-model 400 this preflight was built to prevent.

```text
codex_provider import fails --> DEFAULT_MODEL='' (no default) --> `if DEFAULT_MODEL else []` --> check_codex_model never runs --> main() returns 0 --> real run later hits 400
```

**Impact**
A broken/moved provider module silently removes the safety net with no signal, defeating the stated purpose of preflight.py for exactly its headline scenario.

**Suggested correction**
On ImportError, print a warning noting the codex-model check is being skipped and why, or hardcode/import the same default fallback used in codex_provider.py.

**How to verify**
Add a test that makes codex_provider unimportable and CF_UX_CODEX_MODEL unset, then assert main() either warns or still performs the check, not silently returns 0.

</details>

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,172 @@
...
+
+    `make` runs its recipes in `sh`, which never sources nvm's shell function,
+    so an nvm user's `node` is whichever version nvm's default symlink points
+    at -- and that is routinely not the one promptfoo needs. Finding the usable
+    one is a directory listing; making the person find it is the half hour this
+    function exists to give back.
+    """
+    root = Path(os.environ.get("NVM_DIR") or Path.home() / ".nvm") / "versions" / "node"
+    found = []
+    for candidate in sorted(root.glob("*/bin/node")) if root.is_dir() else []:
```
ainetx, line 67: ### _nvm_candidates has no cap on subprocess probes, each with a 20s timeout

**Severity:** Minor

**Problem**
_nvm_candidates() (preflight.py:56-71) iterates over every `*/bin/node` found under nvm's versions directory and calls _node_version, a subprocess with up to a 20-second timeout, for each one, with no limit on the number probed nor an overall deadline for check_node().

<details>

**How to reproduce**
1. Have many (e.g. 20+) nvm-installed node versions, or one that hangs on `--version`. 2. Run a node version below NODE_MIN so the fallback path executes. 3. Observe check_node() can take proportionally long (or up to N*20s in a pathological hang case) before printing its report.

**Expected behavior**
A preflight check advertised as fast/'up front' should bound total probing time or candidate count (e.g. stop at first qualifying version scanning newest-first, or wrap the whole loop in an overall timeout).

**Actual behavior**
The loop probes every discovered node binary unconditionally, so worst-case latency scales with however many nvm versions exist on the machine.

```text
check_node() fallback --> _nvm_candidates() --> glob all */bin/node --> subprocess probe each (up to 20s) --> no cap/timeout on total --> slow report on power-user machines
```

**Impact**
On developer machines with many nvm versions or a stale/hanging install, the 'checked up front, cheap' preflight promise can become a multi-minute stall before any expensive step even runs.

**Suggested correction**
Scan newest-first and stop at the first qualifying version, or add an overall time budget/cap on the number of candidates probed.

**How to verify**
Simulate many candidate paths (mocking glob results) and assert _nvm_candidates() stops early or completes within a bounded time.

</details>

Makefile
```diff
@@ -18,8 +18,21 @@ PIPX_BIN_DIR ?= $(HOME)/.local/bin
...
 BOOTSTRAP_STUDIO ?= .bootstrap/.core/skills/studio/scripts/studio.py
 SOURCE_STUDIO ?= skills/studio/scripts/studio.py
-PYTEST_PIPX ?= $(PIPX_BIN_DIR)/pytest
+# pipx first, because `make install` puts it there and CI runs that. But a
+# pytest is a pytest: a virtualenv or a distro package runs this suite just as
+# well, and refusing one because it came from somewhere else cost a clean
+# worktree an `ERROR: pytest binary not found` with `make install` as the only
+# advice -- for a tool already on PATH. Whatever answers is checked for the
+# plugins the targets actually use, which is the part that matters.
+PYTEST ?= $(shell test -x "$(PIPX_BIN_DIR)/pytest" && echo "$(PIPX_BIN_DIR)/pytest" \
```
ainetx, line 27: ### PYTEST is a recursively-expanded Make variable, so its shell probe reruns on every reference

**Severity:** Minor

**Problem**
PYTEST is defined with `?=` and a `$(shell...)` right-hand side, which in GNU Make creates a recursively-expanded (not memoized) variable. Only `:=` would freeze the shell result at parse time; as written, every reference to $(PYTEST) (directly, or transitively via PYTEST_PIPX/PYTEST_PIPX_COV/PYTEST_PARALLEL) re-executes the `test -x` / `command -v` shell pipeline.

<details>

**How to reproduce**
1. Read Makefile:27-35 and note `?=` (not `:=`). 2. Trace references: check-pytest (two uses), PYTEST_PARALLEL's own `$(shell $(PYTEST) --help...)`, and the test/test-verbose/test-quick/test-coverage recipes. 3. Confirm each is a distinct textual reference to a recursively-expanded variable.

**Expected behavior**
PYTEST should be defined with `:=` (simply expanded) so the shell probe runs once per make invocation and is reused consistently.

**Actual behavior**
The shell detection pipeline is re-forked on every reference, multiplying subprocess calls and creating a narrow window where two evaluations within the same recipe could theoretically disagree if PATH/filesystem state changes mid-run.

```text
make test -> check-pytest: evaluates $(PYTEST) (shell fork #1) -> PYTEST_PARALLEL evaluates $(PYTEST) again (shell fork #2) -> test: recipe evaluates $(PYTEST_PIPX)=$(PYTEST) again (shell fork #3)
```

**Impact**
Minor performance overhead from repeated subprocess forks per make invocation, plus a small theoretical consistency risk if resolution could change between evaluations within one run.

**Suggested correction**
Change `PYTEST ?=` to use `:=` (or compute it once via a recursively-safe pattern) so the shell probe runs exactly once and is reused.

**How to verify**
Add `$(info...)` tracing or run under `make --trace` to confirm the shell pipeline executes only once after the fix.

</details>

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,172 @@
...
+        and entry.get("visibility") == "list"
+    })
+    if not slugs or model in slugs:
+        return []
+    return [
+        f"codex model {model!r} is not among the ones this account is entitled to.",
+        f"  {MODELS_CACHE} lists: {', '.join(slugs)}",
+        "  Pick one and set it:",
+        "",
+        f"    CF_UX_CODEX_MODEL={slugs[0]} make test-prompts",
```
ainetx, line 145: ### Suggested replacement codex model is picked alphabetically, ignoring cost/effort tier

**Severity:** Minor

**Problem**
check_codex_model (preflight.py:133-145) builds `slugs = sorted({...})` — an alphabetical sort with no relation to price or capability — and recommends `slugs[0]` verbatim as the CF_UX_CODEX_MODEL value to set, contradicting the Makefile's stated design goal (Makefile:338) of defaulting to cheap models at low effort to bound cost.

<details>

**How to reproduce**
1. Configure a withdrawn codex model. 2. Have a models_cache.json listing an expensive model alphabetically before any cheap one (e.g. 'gpt-5-pro' before 'gpt-5.4-mini' is not alphabetical, but any account entitled to an expensive model whose slug sorts first would trigger this). 3. Run preflight.py and observe the suggested `CF_UX_CODEX_MODEL=` command points at the alphabetically-first slug regardless of cost.

**Expected behavior**
The suggested replacement should prefer a cheap/low-effort-compatible model, or explicitly warn that the user should verify cost before copy-pasting the command.

**Actual behavior**
A user following the printed remedy verbatim could switch to a substantially more expensive model than the suite is designed to run by default.

```text
withdrawn model detected --> slugs sorted alphabetically --> slugs[0] suggested as remedy --> no cost check --> potential expensive model selected
```

**Impact**
Silently undermines the suite's cost-control design goal exactly in the failure scenario (withdrawn cheap default) this check is meant to handle gracefully.

**Suggested correction**
Prefer a slug matching a known cheap/low-cost naming pattern, or explicitly caveat in the printed message that the suggested model's cost/effort should be verified before use.

**How to verify**
Test check_codex_model with a cache containing both cheap and expensive-sounding slugs and confirm the suggested remedy is not simply alphabetical.

</details>

tests/prompts/cf-ux/README.md
```diff
@@ -4,8 +4,29 @@ End-to-end UX tests for the `cf` skill, running real `claude` and `codex` CLIs
...
 
+## Prerequisites
+
+| Needs | Why, and the part that bites |
+|---|---|
+| **Node >= 22.22** | promptfoo's own `engines.node`. An older one fails inside npx talking about the package, not about node. |
+| A node `make` can **spawn** | Under nvm's lazy loader `node` is a *shell function*: an interactive shell answers `node --version` while `make`, which runs recipes in `sh`, finds no program at all. Put a real one on PATH for the command: `PATH="$NVM_DIR/versions/node/v22.22.2/bin:$PATH" make test-prompts`. |
+| `claude`, `codex`, `cfs` on PATH | The pilot drives the real CLIs. `cfs` comes from `make install-proxy`. |
+| Both CLIs **logged in** | Providers inherit only `ANTHROPIC_*`/`CLAUDE_*` and `OPENAI_*`/`CODEX_*`; there is no key to pass in. |
+| A codex model the account **has** | Slugs are withdrawn over time. A stale default costs all 8 codex cases with a 400 in a table cell. |
```
ainetx, line 15: ### README's '8 codex cases' contradicts its own '3 scenarios / 6 cases' baseline table

**Severity:** Minor

**Problem**
README.md line 15 states a stale codex model 'costs all 8 codex cases with a 400 in a table cell,' but the same file's 'Current baseline (pilot)' section (lines 172-180) documents '3 scenarios × 2 providers = 6 cases' with exactly 3 scenario rows, i.e. 3 codex cases, not 8. No cross-reference reconciles the numbers.

<details>

**How to reproduce**
1. Read README.md line 15 ('8 codex cases'). 2. Read README.md lines 172-180 ('3 scenarios x 2 providers = 6 cases', 3 rows). 3. Note no footnote or reference bridges the two numbers.

**Expected behavior**
The Prerequisites table's case count should match the documented baseline (3 codex cases currently, or reference the 'Top 6-10 scenarios' expansion plan explicitly if that's the intended future count).

**Actual behavior**
A reader following the Prerequisites table would expect 8 codex scenarios exist, but the suite documents only 3.

```text
Prerequisites table (line 15: '8 codex cases') <-- inconsistent with --> Current baseline section (lines 172-180: 3 scenarios, 3 codex cases)
```

**Impact**
Confuses readers about actual test coverage; recurring documentation-drift failure mode previously flagged in this exact file.

**Suggested correction**
Update line 15 to reference the correct current case count (3) or explicitly note it refers to a planned future expansion (e.g. 'up to 8 once Top 6-10 scenarios land').

**How to verify**
Re-read both sections after correction and confirm the numbers agree or are explicitly reconciled.

</details>

tests/prompts/cf-ux/README.md
```diff
@@ -30,6 +51,13 @@ the `llm-rubric` asserts; sub-100ms text guards catch skill-load failures.
...
 | `CF_UX_KEEP_SANDBOX=1` | Keep the sandbox after the run; print its path. |
+| `CF_UX_CLAUDE_MODEL` / `CF_UX_CLAUDE_EFFORT` | Override the claude model and reasoning effort. |
+| `CF_UX_CODEX_MODEL` / `CF_UX_CODEX_EFFORT` / `CF_UX_CODEX_CONTEXT` | Override the codex model, reasoning effort and context window. Set the model when the default slug is withdrawn — `codex` lists what the account may use. |
+| `CF_UX_GRADER_MODEL` / `CF_UX_GRADER_EFFORT` | Override the LLM-rubric judge. |
+| `CF_UX_CODEX_DISABLE_PLUGINS` | Comma-separated `name@marketplace` to disable, for isolation debugging only — by default the skill is expected to win against competing plugins. |
+
+Defaults live beside the code they configure: `providers/*.py`. This table says
+they exist and what they are for, not what today's value is.
 
 ## Layout
```
ainetx, line 68: ### README Layout diagram omits the newly-added preflight.py

**Severity:** Minor

**Problem**
The diff adds tests/prompts/cf-ux/preflight.py directly inside tests/prompts/cf-ux/ and documents it extensively (Prerequisites section, Run section), but the pre-existing Layout directory tree (README.md lines 69-78) still lists only promptfooconfig.yaml, providers/*, and README.md — no entry for preflight.py.

<details>

**How to reproduce**
1. Read README.md's '## Layout' section (lines 67-78). 2. Confirm no preflight.py entry. 3. Confirm preflight.py exists in the same directory per the diff.

**Expected behavior**
Layout tree should list preflight.py alongside the other top-level files in tests/prompts/cf-ux/.

**Actual behavior**
Layout tree is incomplete/stale relative to the directory's actual contents post-diff.

```text
diff adds tests/prompts/cf-ux/preflight.py --> README Layout section (unchanged) --> omits preflight.py --> documentation drift
```

**Impact**
Minor documentation inconsistency; a reader relying on Layout for an overview of the directory gets an incomplete picture.

**Suggested correction**
Add a `├── preflight.py` entry (with a short description) to the Layout code block.

**How to verify**
Re-read the Layout section and confirm preflight.py is listed.

</details>

Makefile
ainetx: ### New Makefile PYTEST fallback chain and PYTEST_PARALLEL detection have no test/CI coverage

**Severity:** Minor

**Problem**
Makefile now resolves PYTEST via a three-way fallback (pipx bin dir → `command -v pytest` → `$(PYTHON) -m pytest`) and derives PYTEST_PARALLEL by probing `--help` for `--numprocesses`, but every CI job (test, gates, coverage) runs `make install` first, which always installs pytest via pipx, so only the first branch of the fallback and the xdist-present branch of PYTEST_PARALLEL are ever exercised. No test in the repo references PYTEST, PYTEST_PARALLEL, or check-pytest.

<details>

**How to reproduce**
1. Read.github/workflows/ci.yml: every job invoking `make test`/`make test-gates`/`make test-coverage` runs `make install` immediately before it. 2. Read Makefile:29-37: PYTEST tries pipx path first, only falling through to `command -v pytest` or `$(PYTHON) -m pytest` when the pipx binary is absent -- a state CI never reaches. 3. `grep -r PYTEST tests/` returns no matches, confirming no unit/integration test exercises this shell logic directly.

**Expected behavior**
The fallback chain and the non-fatal xdist detection this diff fixes 'pytest binary not found' hard failures for should have some automated check (unit test invoking the Makefile logic in isolation, or a CI job that runs `make test` in an environment without pipx-installed pytest) guarding against regressions.

**Actual behavior**
No test or CI job ever runs `make test`/`check-pytest` without pipx-provided pytest already present, so the PATH-lookup fallback, the python -m pytest fallback, and the pytest-xdist absent-path NOTE are logic that could regress (e.g. reintroducing a hard fail) without any CI signal.

```text
CI job -> make install (pipx pytest) -> make test -> PYTEST resolves via first branch only -> fallback branches (PATH lookup, python -m pytest) and PYTEST_PARALLEL absent-xdist path: never executed by any job or test
```

**Impact**
A future edit that breaks the PATH-lookup or python -m pytest fallback, or reintroduces the old hard-fail behavior for missing pytest-xdist, would pass CI silently and only surface for a developer running from a non-pipx environment -- exactly the scenario this diff set out to fix.

**Suggested correction**
Add a CI job or Makefile-focused test that invokes `make check-pytest`/`make test` with PIPX_BIN_DIR pointed at a directory without pytest, forcing the PATH-lookup/python-m-pytest fallback and the absent-xdist NOTE path to execute and assert on their outcomes.

**How to verify**
Confirm the new test/CI job fails if the fallback ordering in Makefile:29-31 or the PYTEST_PARALLEL detection in Makefile:37 is reverted to the old hard-fail behavior.

</details>

tests/test_cf_ux_preflight.py
```diff
@@ -0,0 +1,138 @@
...
+_PILOT = Path(__file__).resolve().parents[1] / "tests" / "prompts" / "cf-ux"
+if str(_PILOT) not in sys.path:
+    sys.path.insert(0, str(_PILOT))
+
+preflight = pytest.importorskip("preflight")
+
+
+class TestTheNodeCheck:
+    def test_a_new_enough_node_says_nothing(self, monkeypatch):
+        monkeypatch.setattr(preflight, "_node_version", lambda *a: (22, 22, 0))
```
ainetx, line 29: ### check_node()'s real subprocess/regex parsing in _node_version() is never exercised by any test

**Severity:** Minor

**Problem**
_node_version() (subprocess.run against a `node` binary with a 20s timeout, then regex-parses stdout via _VERSION) is the actual detection logic replacing the Makefile's removed `command -v node` gate. Every test in TestTheNodeCheck monkeypatches _node_version directly with a lambda returning a pre-fabricated tuple. TestTheNvmScan does invoke the real _nvm_candidates() (confirming its glob/filter/sort/cap logic), but it too stubs _node_version rather than calling it for real.

<details>

**How to reproduce**
1. grep '_node_version' tests/test_cf_ux_preflight.py -- every hit is a monkeypatch.setattr call. 2. Read preflight.py:41-51 -- the real function does subprocess.run(['node','--version'],...) then regex-parses. 3. No test anywhere spawns a real or fake executable and lets _node_version's subprocess call and regex run unmocked.

**Expected behavior**
At least one test should exercise the real _node_version() (e.g. via a fake `node` script placed on PATH or passed as `binary=`) to confirm the regex correctly parses realistic `node --version` output (e.g. 'v22.22.0'), handles a non-zero exit code, and handles a timeout/OSError.

**Actual behavior**
A regression in the _VERSION regex, the subprocess invocation, or the timeout/exit-code handling in _node_version() would pass the full test suite silently, since check_node()'s only tested surface is the wrapper logic around already-parsed tuples.

```text
check_node() -> _node_version() [subprocess.run + regex parse, UNTESTED for real] -> compared against NODE_MIN -> problem message
tests -> monkeypatch _node_version() with canned tuple -> only wrapper/message logic tested
```

**Impact**
Node-version detection is the sole mechanism now guarding against the exact npx-failure and nvm shell-function traps this file was built to prevent; an unnoticed regression in its real parsing logic would silently defeat that protection with no test failure to catch it.

**Suggested correction**
Add a test that writes a small fake `node` script (e.g. `#!/bin/sh\necho v22.22.0`) to tmp_path, calls the real preflight._node_version(str(fake_node_path)), and asserts the parsed tuple; similarly test a non-zero-exit fake script.

**How to verify**
Confirm the new test fails if _VERSION's regex or the subprocess.run call in _node_version() is altered/broken.

</details>

tests/prompts/cf-ux/preflight.py
ainetx, line 67: ### _nvm_candidates() filesystem traversal is unguarded, can crash preflight.py with a raw traceback

**Severity:** Minor

**Problem**
root.glob("*/bin/node") is forced to iterate via sorted() with no try/except, while every other IO operation in this file (subprocess calls, cache reads) is defensively wrapped. A permission-denied entry or broken symlink loop under $NVM_DIR/versions/node raises an uncaught OSError.

<details>

**How to reproduce**
1. Create $NVM_DIR/versions/node/badperm as a directory with no read/execute permission. 2. Run `make check-prompt-tests` (or `python preflight.py`) with an old/missing node so check_node() calls _nvm_candidates(). 3. glob() raises PermissionError, propagating as an unhandled traceback instead of a diagnostic line.

**Expected behavior**
The traversal should be guarded the same way _node_version()'s subprocess call is, converting an OSError into an empty candidate list (or a reported problem line) rather than crashing.

**Actual behavior**
An OSError during glob() propagates uncaught through check_node() and main(), producing a raw Python traceback -- exactly the failure mode this preflight tool exists to prevent.

```text
check_node() -> _nvm_candidates() -> root.glob() raises OSError -> no try/except anywhere in the call chain -> uncaught traceback out of main()
```

**Impact**
On developer machines with unusual or permission-restricted nvm installs, the preflight itself crashes with an unhelpful traceback instead of reporting a clear remedy, undermining the tool's stated purpose.

**Suggested correction**
Wrap the glob/iteration in _nvm_candidates() in a try/except (OSError) and treat a traversal failure as 'no candidates found' (or surface it as one of the reported problem lines).

**How to verify**
Add a test where root.glob is monkeypatched to raise OSError and assert check_node() still returns a clean problem list instead of raising.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): `_node_version()`'s `except (OSError, subprocess.SubprocessError): return None` and `check_codex_model()`'s `except (OSError, UnicodeDecodeError, json.JSONDecodeError): return []` both collapse distinct failure causes (permission errors, disk errors, corrupted vs. absent files) into an indistinguishable sentinel with no logging/diagnostic emitted, matching the project's own silent_exceptions pylint plugin's flagged anti-pattern (bare sentinel return, no raise, no visible-signal call).

Problem (now): _node_version() and check_codex_model() catch broad exception sets (OSError/SubprocessError, OSError/UnicodeDecodeError/JSONDecodeError) and return bare sentinel values (None, []) with no logging, warning, or re-raise -- exactly the pattern scripts/pylint_plugins/silent_exceptions.py's SilentExceptionSwallowedChecker (W9001) is built to flag. The file is not included in PYLINT_TARGETS, so CI's pylint invocation never lints it.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The module docstring says 'make check-prompt-tests already proves the four binaries exist' before describing what preflight.py itself checks. But this diff removed the Makefile's `command -v node` check from check-prompt-tests, leaving only claude/codex/cfs (three binaries) checked via `command -v` before preflight.py runs; node's presence is now solely established inside preflight.py's own check_node().

Problem (now): The docstring at preflight.py:4 claims `make check-prompt-tests` already proves 'the four binaries exist' before preflight.py runs, but this same diff removed the Makefile's `command -v node` check. The Makefile target now only verifies claude, codex, and cfs (three binaries) before invoking preflight.py, which is the sole place node's existence is checked.

</details>

Makefile
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): install-prompt-tests depends on check-prompt-tests, which now runs preflight.py's check_codex_model(). If ~/.codex/models_cache.json is present and well-formed but doesn't list the configured model, preflight.py exits 1, and since the Makefile recipe line has no `-` prefix, make aborts check-prompt-tests, which in turn blocks install-prompt-tests.

Problem (now): install-prompt-tests (Makefile:364) depends on check-prompt-tests, whose recipe now ends with an unconditional call to preflight.py (Makefile:362), which includes check_codex_model. A user with a well-formed models_cache.json that legitimately lists models but doesn't include the configured CF_UX_CODEX_MODEL/default slug will fail install-prompt-tests, despite its only stated job being 'Pre-caching promptfoo@... via npx'.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): main()'s `try: from codex_provider import DEFAULT_MODEL except ImportError: DEFAULT_MODEL = os.environ.get('CF_UX_CODEX_MODEL', '')` produces an empty string when the env var is unset, and `check_codex_model(DEFAULT_MODEL) if DEFAULT_MODEL else []` then skips the check entirely with zero diagnostic output.

Problem (now): main() imports DEFAULT_MODEL from codex_provider.py, which is where the real default ('gpt-5.4-mini') lives. On ImportError, the fallback `os.environ.get('CF_UX_CODEX_MODEL', '')` loses that default entirely. If the env var is unset, DEFAULT_MODEL becomes empty string, and `check_codex_model(DEFAULT_MODEL) if DEFAULT_MODEL else []` skips the check with zero indication to the user that it was skipped.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): _nvm_candidates() iterates every matched nvm node binary and probes each with a subprocess call (up to 20s timeout apiece) with no limit on how many are probed and no overall timeout wrapping the whole scan.

Problem (now): _nvm_candidates() (preflight.py:56-71) iterates over every `*/bin/node` found under nvm's versions directory and calls _node_version, a subprocess with up to a 20-second timeout, for each one, with no limit on the number probed nor an overall deadline for check_node().

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): check_codex_model recommends `slugs[0]` from an alphabetically sorted set of entitled model slugs as the value to set via CF_UX_CODEX_MODEL, with no relation between alphabetical order and a model's cost or effort tier.

Problem (now): check_codex_model (preflight.py:133-145) builds `slugs = sorted({...})` — an alphabetical sort with no relation to price or capability — and recommends `slugs[0]` verbatim as the CF_UX_CODEX_MODEL value to set, contradicting the Makefile's stated design goal (Makefile:338) of defaulting to cheap models at low effort to bound cost.

</details>

tests/prompts/cf-ux/README.md
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): README.md:15 claims a stale codex model 'costs all 8 codex cases,' but the same file's Current baseline section documents exactly 3 scenarios × 2 providers = 6 cases total (3 codex cases), with no reconciling reference to a larger separate audit.

Problem (now): README.md line 15 states a stale codex model 'costs all 8 codex cases with a 400 in a table cell,' but the same file's 'Current baseline (pilot)' section (lines 172-180) documents '3 scenarios × 2 providers = 6 cases' with exactly 3 scenario rows, i.e. 3 codex cases, not 8. No cross-reference reconciles the numbers.

</details>

tests/prompts/cf-ux/README.md
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): This diff adds tests/prompts/cf-ux/preflight.py directly under tests/prompts/cf-ux/, but the pre-existing '## Layout' section's directory tree (README.md:64-73) still lists only promptfooconfig.yaml, providers/*, and README.md, omitting the new top-level file that the same diff repeatedly references elsewhere (Prerequisites and Run sections).

Problem (now): The diff adds tests/prompts/cf-ux/preflight.py directly inside tests/prompts/cf-ux/ and documents it extensively (Prerequisites section, Run section), but the pre-existing Layout directory tree (README.md lines 69-78) still lists only promptfooconfig.yaml, providers/*, and README.md — no entry for preflight.py.

</details>

tests/test_cf_ux_preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): Every test in TestTheNodeCheck replaces preflight._node_version and preflight._nvm_candidates with monkeypatched lambdas/list returning canned tuples, so the real subprocess.run + regex parsing in _node_version() and the real glob/filter/sort logic in _nvm_candidates() are never actually executed by the test suite.

Problem (now): _node_version() (subprocess.run + regex parse) and _nvm_candidates() (Path.glob + version-sort over $NVM_DIR/versions/node) are the actual detection logic replacing the Makefile's removed `command -v node` gate, but every test in TestTheNodeCheck replaces both functions with lambdas/list before calling check_node(), so only the wrapper/message-formatting logic around pre-fabricated tuples is tested.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): _nvm_candidates() calls `sorted(root.glob("*/bin/node"))` with no try/except around it, unlike _node_version()'s subprocess call which is explicitly wrapped in `except (OSError, subprocess.SubprocessError)`. A permission-denied entry or broken symlink loop under $NVM_DIR/versions/node during traversal could raise an OSError that propagates uncaught.

Problem (now): root.glob("*/bin/node") is forced to iterate via sorted() with no try/except, while every other IO operation in this file (subprocess calls, cache reads) is defensively wrapped. A permission-denied entry or broken symlink loop under $NVM_DIR/versions/node raises an uncaught OSError.

</details>

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,185 @@
...
+        # replace exactly that with a sentence.
+        return _report([f"a CF_UX_* setting is not valid: {exc}",
+                        "  (read while importing the codex provider; check "
+                        "CF_UX_CODEX_CONTEXT)"])
+
+    return _report(check_node()
+                   + (check_codex_model(DEFAULT_MODEL) if DEFAULT_MODEL else []))
+
+
+def _report(problems: list[str]) -> int:
```
ainetx, line 226: ### _report() helper's exact formatting/exit-code contract has no direct unit test

**Severity:** Minor

**Problem**
grep confirms zero direct calls to `preflight._report(...)` anywhere in tests/test_cf_ux_preflight.py. All coverage of `_report()` (banner text, blank-line spacing, the `if not line.startswith(' ')` indentation branch, and the empty-list->0 / non-empty->1 short-circuit) is incidental, reached only via `main()` with `check_node`/`check_codex_model` monkeypatched to non-empty lists, and only asserts a substring appears in captured stderr.

<details>

**How to reproduce**
1. Regress `_report()`'s empty-list short-circuit (e.g. change `if not problems: return 0` to always print/return 1), or break the indentation condition. 2. Run `pytest tests/test_cf_ux_preflight.py`. 3. All existing tests still pass because none of them call `_report([])` or pin the exact banner/indentation output.

**Expected behavior**
A test that calls `preflight._report([])` directly and asserts return 0 with no stderr, plus `preflight._report(["some problem"])` asserting the exact banner text, blank-line spacing, indentation behavior, and return 1.

**Actual behavior**
Only indirect, substring-level coverage via `main()`; `_report`'s own contract is unpinned.

```text
_report([]) -> [untested] -> should return 0, no output
_report([...]) -> [only tested via main()] -> banner/indent/blank-line format unpinned
```

**Impact**
A regression in _report()'s own formatting logic or its empty-list short-circuit would go undetected by the current test suite.

**Suggested correction**
Add a small test class calling preflight._report(...) directly for both the empty and non-empty cases, asserting exact stderr content and exit codes.

**How to verify**
Introduce a deliberate regression in _report() (e.g. remove the empty-list short-circuit) and confirm the new direct test fails while it currently passes.

</details>

ainetx (changes requested):
Requesting changes: this diff regresses the codex model entitlement check into a silent no-op in a common configuration, undoing the safety the new preflight is supposed to add.

- **Codex model entitlement check silently disables itself when `codex_provider` fails to import** — `main()` falls back to `os.environ.get('CF_UX_CODEX_MODEL', '')` on `ImportError`, losing the real default (`gpt-5.4-mini`) from `codex_provider.DEFAULT_MODEL`. With the env var unset, `DEFAULT_MODEL` is `''` and `check_codex_model(...) if DEFAULT_MODEL else []` skips the check entirely — with no warning that it didn't run. This is exactly the failure mode the preflight was built to catch (withdrawn model → 400 after every sandbox is built), reappearing through the one code path most likely to hit it. ([discussion](https://github.com/constructorfabric/studio/pull/241#discussion_r4044606330))

vasylcf:
Second round: one Major and eight smaller findings, all reproduced. `5f80d8f`.

## Major — the preflight reported a success it never established

Correct, and the worst kind of bug for this component. Reproduced:

```
provider unimportable + CF_UX_CODEX_MODEL unset  ->  exit 0
```

`DEFAULT_MODEL` fell back to an unset env var, the guard read `""` as "nothing to check", and the preflight exited clean. A component whose entire job is to say whether a run may proceed was saying yes on a check that never ran.

An unimportable provider is now reported outright — the pilot cannot run without it either way. The env override still works and still lets a run proceed; pinned by two tests.

## Taken

- **`PYTEST` was recursively expanded.** Its `$(shell)` probe re-ran at every reference, and most targets reference it. `:=` now, same for `PYTEST_PARALLEL`.
- **`PYTEST_PIPX` could diverge from the validated variable.** Both the check and the parallel probe now consult `PYTEST_PIPX` — the one the recipes actually call — so a user overriding it gets the checked one.
- **`install-prompt-tests` required a live account to cache an npm package.** It now runs the node half only, via a new `--node-only`.
- **`_nvm_candidates` was unguarded and uncapped.** It globbed the whole directory and probed every entry, each a process with a 20s timeout; a long-lived nvm directory has dozens. Capped at the newest 8, and a traversal that fails returns "no candidates" instead of a traceback.
- **The suggested model was alphabetical**, which put `gpt-5.5` — the previous generation — at the head. It now follows `codex`'s own `priority` ordering, so the suggestion matches what the CLI itself would offer.
- **The `except` blocks returned without a word.** They log at debug now. "No version" is still the answer the caller acts on, but the reason belongs somewhere.
- **The docstring credited the Makefile with a node check** this PR had moved into the preflight.
- **README:** the layout diagram omitted `preflight.py`, and the baseline said "3 scenarios × 2 providers = 6 cases, ~3-4 min" where the config has 8 scenarios and the measured run was 16 cases in ~1.5 min. That line predated several scenarios; corrected with the measurement date.

## Verified

`6289 passed, 4 skipped, 15 xfailed`; `make spec-coverage` green. Six new tests cover the Major, the `--node-only` path, the priority ordering, and both `_nvm_candidates` guards.

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Verifier:** Traced the actual astroid-based plugin logic (scripts/pylint_plugins/silent_exceptions.py) against the two new except blocks: each handler's body is [Expr(logger.debug(...)), Return(sentinel)]. `_is_passive_statement` on the logger.debug() Expr statement requires either a silent statement, a passive assignment, a visible signal, or a call whose name is in `_PASSIVE_CALL_NAMES={append,extend,write,writelines}` -- "debug" matches none of these, so `_is_passive_statement` returns False for that statement, which makes `all(_is_passive_statement(...) for stmt in handler.body)` False, so `visit_try` never calls `add_message` -- W9001 would NOT actually fire on this exact pattern (verified by manually replaying the plugin's own test cases, e.g. test_flags_list_append_then_silent_return vs test_allows_local_warn_wrapper_then_fallback, which behave exactly as this trace predicts). This is corroborated empirically: `skills/studio/scripts/studio` IS in PYLINT_TARGETS and this plugin IS registered via pyproject.toml load-plugins, yet the identical `logger.debug(...); return <sentinel>` pattern is used dozens of times throughout that actively-linted directory (decision_log.py, change_summary.py, kit.py, eval.py, okf.py, doc_index.py, whatsnew.py) without CI failing on it. Per the general lesson about checking sibling/pre-existing code paths before attributing a defect to new code, this is an established, pylint-clean, repo-wide convention, not a genuine violation of the silent_exceptions plugin contract.

</details>

Makefile
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Reading the current Makefile shows check-pytest validates $(PYTEST_PIPX) directly (line 104: `$(PYTEST_PIPX) --version`) and PYTEST_PARALLEL is derived from `$(shell $(PYTEST_PIPX) --help...)` (line 37) -- the exact same variable that test/test-verbose/test-quick/test-coverage invoke, not $(PYTEST). Since both the validation and the invocation reference $(PYTEST_PIPX) consistently within a single make run (command-line overrides of PYTEST_PIPX apply uniformly to all references), there is no split-brain: whatever PYTEST_PIPX resolves to (default via PYTEST's auto-detection, or an explicit override) is exactly what's checked and exactly what's run.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read preflight.py's actual docstring (lines 2-18): it says 'make check-prompt-tests proves that claude, codex and cfs exist' (three binaries explicitly named, no 'four') and immediately clarifies '(node it checks here too, for the reason in check_node)' -- i.e. it already explicitly attributes node's check to this script, not to the Makefile. The premise of this check (that the docstring claims 'four binaries' were already proven) does not match the current text.

</details>

Makefile
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read the current Makefile (lines 366-373): install-prompt-tests no longer depends on check-prompt-tests at all (`install-prompt-tests:` with no prerequisite) and its recipe explicitly runs `preflight.py --node-only`, which (per preflight.py main(), line 184-185) returns immediately after check_node() alone, never touching check_codex_model. A comment explicitly documents this as deliberate: 'Deliberately not check-prompt-tests: this caches an npm package, and making that wait on a live codex model entitlement meant a download could not be prepared offline...' This fully satisfies the check's expected evidence.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read preflight.py main() (lines 187-212): on ImportError, if CF_UX_CODEX_MODEL is unset, it explicitly returns `_report([..."the codex provider will not import"...])`, printing an error and exiting 1 -- not a silent skip. There is also a dedicated ValueError handler for bad CF_UX_* env vars, and tests_cf_ux_preflight.py's TestWhatTheCheckCannotDoItSays and TestAMisconfiguredEnvVar classes directly exercise both of these branches through main(). The check's premise (silent, untested skip) does not match the current code.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read preflight.py lines 37-40 and 88: NVM_PROBE_LIMIT = 8 is defined and used as `for candidate in candidates[:NVM_PROBE_LIMIT]`, explicitly capping the number of subprocess probes at 8 (worst case ~160s with the 20s timeout), directly contradicting the check's premise of 'no cap on the number of subprocess probes'. TestTheNvmScan.test_the_number_of_probes_is_capped also verifies this cap with 40 fake versions.

</details>

Makefile
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read the current Makefile line 29 and the diff directly: `PYTEST:= $(shell...)` uses `:=` (simply-expanded/immediate), not `?=`/recursive expansion as the check claims. The value is computed once at parse time; PYTEST_PARALLEL (line 37) is also `:=`. Subsequent references to $(PYTEST)/$(PYTEST_PIPX)/$(PYTEST_PARALLEL) are cheap text substitutions of already-computed strings, not repeated shell forks. The Makefile's own comment explicitly explains this choice ('`:=`, not `?=`: a recursively-expanded variable re-runs its $(shell) at every reference...'), directly refuting the check's premise.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read preflight.py lines 155-166: the code explicitly sorts by `(priority, slug)` via `ranked = sorted(listed, key=lambda e: (e.get('priority') if isinstance(e.get('priority'), int) else 10**6, e['slug']))`, not a plain alphabetical sort, with a comment explicitly stating 'Ordered by the CLI's own priority, not alphabetically. Alphabetical put gpt-5.5 -- the previous generation -- at the head of the suggestion, which is a worse default...' `slugs[0]` therefore reflects priority order, not alphabetical order, contradicting the check's premise. A dedicated test (TestTheSuggestedModel) confirms this.

</details>

tests/prompts/cf-ux/README.md
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read the current README.md 'Current baseline (pilot)' section (line 175): it already states '8 scenarios × 2 providers = 16 cases, ~1.5 min wall-clock at the configured concurrency (measured 2026-09-18; the table below predates several scenarios)' -- matching the '8 codex cases' figure in the Prerequisites table, with an explicit reconciling note explaining why the table below only shows 3 scenario rows. The premised contradiction with a stale '3 scenarios/6 cases' baseline does not exist in the current text.

</details>

tests/prompts/cf-ux/README.md
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read the current README.md Layout section (lines 69-79): it already includes `├── preflight.py — node version + codex model entitlement, pre-run` in the directory tree, directly contradicting the check's premise that this file is omitted.

</details>

Makefile
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): The new PYTEST resolution logic (pipx path -> PATH lookup -> `$(PYTHON) -m pytest` fallback) and PYTEST_PARALLEL's non-fatal pytest-xdist detection are pure Make/shell logic with no accompanying automated test. CI's `make install && make test` job only exercises the pipx-installed-pytest branch, since `make install` always installs pytest via pipx.

Problem (now): Makefile now resolves PYTEST via a three-way fallback (pipx bin dir → `command -v pytest` → `$(PYTHON) -m pytest`) and derives PYTEST_PARALLEL by probing `--help` for `--numprocesses`, but every CI job (test, gates, coverage) runs `make install` first, which always installs pytest via pipx, so only the first branch of the fallback and the xdist-present branch of PYTEST_PARALLEL are ever exercised. No test in the repo references PYTEST, PYTEST_PARALLEL, or check-pytest.

</details>

tests/test_cf_ux_preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): _node_version() (subprocess.run + regex parse) and _nvm_candidates() (Path.glob + version-sort over $NVM_DIR/versions/node) are the actual detection logic replacing the Makefile's removed `command -v node` gate, but every test in TestTheNodeCheck replaces both functions with lambdas/list before calling check_node(), so only the wrapper/message-formatting logic around pre-fabricated tuples is tested.

Problem (now): _node_version() (subprocess.run against a `node` binary with a 20s timeout, then regex-parses stdout via _VERSION) is the actual detection logic replacing the Makefile's removed `command -v node` gate. Every test in TestTheNodeCheck monkeypatches _node_version directly with a lambda returning a pre-fabricated tuple. TestTheNvmScan does invoke the real _nvm_candidates() (confirming its glob/filter/sort/cap logic), but it too stubs _node_version rather than calling it for real.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of 5f80d8fc44100948256f9d8bafd599d98709dadc -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Verifier:** Read the actual current code at tests/prompts/cf-ux/preflight.py lines 68-92: `_nvm_candidates()` already wraps `sorted(root.glob("*/bin/node"), reverse=True)` in `try:... except OSError as exc: logger.debug(...); return []`, exactly mirroring the guard pattern in `_node_version()`. The prior reviewer/verifier claims of 'no try/except' appear to have inspected only lines 56-71 (the check's own stated target range), which cuts off before the except clause at line 83-85 -- the function is fully guarded when read in its entirety. This is also directly exercised by tests/test_cf_ux_preflight.py::TestTheNvmScan::test_an_unreadable_nvm_directory_is_not_a_traceback, which monkeypatches Path.glob to raise OSError and asserts `_nvm_candidates() == []` (no traceback). The finding does not hold against the current code.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): _report() in tests/prompts/cf-ux/preflight.py was factored out to be shared by two call sites in main() (the ValueError branch and the check_node/check_codex_model branch), but no test in tests/test_cf_ux_preflight.py calls preflight._report(...) directly. All coverage of it is incidental, reached only through main() with monkeypatched check_node/check_codex_model, and only asserts that a given substring appears somewhere in captured stderr.

Problem (now): grep confirms zero direct calls to `preflight._report(...)` anywhere in tests/test_cf_ux_preflight.py. All coverage of `_report()` (banner text, blank-line spacing, the `if not line.startswith(' ')` indentation branch, and the empty-list->0 / non-empty->1 short-circuit) is incidental, reached only via `main()` with `check_node`/`check_codex_model` monkeypatched to non-empty lists, and only asserts a substring appears in captured stderr.

</details>

tests/prompts/cf-ux/preflight.py
```diff
@@ -0,0 +1,228 @@
...
+    at -- and that is routinely not the one promptfoo needs. Finding the usable
+    one is a directory listing; making the person find it is the half hour this
+    function exists to give back.
+    """
+    root = Path(os.environ.get("NVM_DIR") or Path.home() / ".nvm") / "versions" / "node"
+    try:
+        # A traversal, so it can fail the way traversals do: a permission-denied
+        # directory, a broken symlink, a path that is not one. None of that is
+        # worth a traceback out of a preflight -- it means "no candidate here".
+        candidates = sorted(root.glob("*/bin/node"), reverse=True) if root.is_dir() else []
```
ainetx: ### _nvm_candidates() sorts version directories lexicographically before capping the probe count, breaking the 'newest first' guarantee

**Severity:** Major

**Problem**
`candidates = sorted(root.glob("*/bin/node"), reverse=True)` in _nvm_candidates() sorts Path objects by their string path components, not by parsed numeric version. String 'v9.x.x' compares greater than 'v22.x.x' or 'v30.x.x' because '9' > '2'/'3' as the first differing character. The list is sliced to the first NVM_PROBE_LIMIT (8) entries before _node_version() ever parses a real (major,minor,patch) tuple, so truncation can permanently discard the genuinely newest, qualifying node install from consideration.

<details>

**How to reproduce**
1. Install nvm node versions v6, v7, v8, v9, v18, v19, v20, v21, v22 (9 versions > NVM_PROBE_LIMIT=8). 2. Run check_node() with the PATH node too old. 3. Reverse-lexicographic string order puts v9...v6 ahead of v22...v18, so the [:8] slice can miss v22 entirely, causing the tool to recommend an older version or report no candidate even though a newer qualifying one is installed.

**Expected behavior**
_nvm_candidates() should probe/report the numerically newest installed node first, as its own docstring states ('Probed newest first').

**Actual behavior**
Candidates are ordered by raw path string, not numeric version, before the probe-limit cap is applied, so the numerically newest install can be excluded from the probed subset entirely.

```text
glob -> ['v6..','v7..','v8..','v9..','v18..','v19..','v20..','v21..','v22..'] -> sorted(reverse=True) as strings -> ['v9..','v8..','v7..','v6..','v22..','v21..','v20..','v19..','v18..'] -> [:8] wastes probe slots on v6-v9 ahead of v18-v22; with more low-digit versions present, v22 itself can fall outside the top 8
```

**Impact**
In long-lived nvm environments (explicitly anticipated by the function's own comment: 'accumulates dozens of versions'), the preflight can silently recommend a stale/wrong node path or report 'no usable node' even when a good one is installed, undermining the purpose of the remediation message.

**Suggested correction**
Sort using a parsed version key, e.g. `candidates = sorted(root.glob("*/bin/node"), key=lambda p: _node_version(str(p)) or (0,0,0), reverse=True)`, or parse the directory name into an int tuple for the sort key before applying NVM_PROBE_LIMIT.

**How to verify**
Add a test with nvm directories mixing single- and double-digit majors (e.g. v9, v22, v30) exceeding NVM_PROBE_LIMIT, and assert the numerically newest qualifying version is the one returned/suggested.

</details>

vasylcf:
Round three, `aefb171`. The Major is a defect the previous round's own fix introduced.

## Major — the probe cap could discard exactly what it was meant to find

Correct. I capped `_nvm_candidates` at the newest 8 and sorted the paths as **text**, which puts `v9.0.0` above `v24.21.0`:

```
['v9.0.0', 'v8.0.0', 'v24.21.0', 'v22.22.2', 'v20.20.0']
```

On a machine with many installed versions, every node new enough to qualify could fall outside the window — and the preflight would report that none was installed, which is the exact failure it exists to prevent, with a confident message attached.

Ordered by the version parsed from the directory name now. Directories carrying no version (nvm aliases, a named install) sort **last** rather than being dropped: still worth a probe if there is room, never ahead of a real one. Three tests: the cap, the ordering with a cap of 2, and the alias case.

## Taken — the two test gaps

- **`_report`** is what every path returns through, so its exit code and "problems reach stderr, output stays clean" contract is pinned directly instead of inferred from whichever caller happened to be under test.
- **`_node_version`'s own parsing** was monkeypatched away everywhere. Four tests now exercise it: a real `v22.22.2` string, a non-zero exit, output with no version in it, and a binary that will not run.

## Already addressed

The two CodeRabbit comments re-anchored at `preflight.py:189` and `README.md:19` are the ones fixed in `18006c5` — the `ValueError` on a non-numeric `CF_UX_CODEX_CONTEXT`, and the README claiming the preflight checks every prerequisite when it cannot verify a login. Both still hold; the line numbers just moved.

**Makefile fallback chain has no test coverage** — not taking this one. Testing `$(shell)` resolution inside a Makefile means invoking `make` from pytest with a synthetic `PATH` and `PIPX_BIN_DIR`, which tests the harness more than the code. The chain is three ordered probes with an explicit override (`make test PYTEST=…`), and its failure mode is loud and immediate. Happy to add it if you want the coverage anyway.

`6297 passed`; spec-coverage green.

tests/test_cf_ux_preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of aefb1712b3407d6c8fd399771ce185c2e599330f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Contrary to prior rounds, the diff's tests/test_cf_ux_preflight.py includes a `TestTheVersionParsingItself` class (verified by reading the file) that calls the real, unmocked `preflight._node_version()` while only stubbing the low-level `subprocess.run`, directly exercising its real control flow and the real `_VERSION` regex parsing across success, non-zero-exit, malformed-output, and OSError cases -- this is exactly the missing coverage the check asked for.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of aefb1712b3407d6c8fd399771ce185c2e599330f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Current diff/file has a dedicated `class TestTheReportContract` (tests/test_cf_ux_preflight.py:274-287) with direct calls `preflight._report([])` (expects 0, stdout and stderr both empty) and `preflight._report(["first thing", "second thing"])` (expects 1, both strings in stderr, stdout empty) -- this directly pins the exit-code contract and stderr-routing independent of main(), contradicting the prior rounds' claim that no direct call to _report() exists; grep confirms two direct call sites at lines 280 and 286.

</details>

tests/prompts/cf-ux/preflight.py
ainetx: **RESOLVED** -- Confirmed fixed as of aefb1712b3407d6c8fd399771ce185c2e599330f -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Read the actual current code (lines 87-96): `candidates = sorted(entries, key=newest_first, reverse=True)` where `newest_first` explicitly calls `_version(path.parent.parent.name)` to parse the vX.Y.Z directory name into a numeric (major, minor, patch) tuple before sorting -- this is NOT a lexicographic string sort as the check claims. The code's own comment states exactly why this key function exists ('sorting paths as text puts v9.0.0 above v24.21.0'), and test_the_newest_is_probed_first_not_the_alphabetically_last explicitly exercises a mixed single/double-digit-major scenario (v8,v9,v20,v22,v24) and asserts the correct newest-first probe order, directly covering this scenario. The prior 'accepted-finding' verdicts appear to have misread the sort call (mistaking it for a keyless `sorted(root.glob(...), reverse=True)`), which does not match the code actually present in this diff.

</details>

ainetx (approved):
Solid change: the Makefile's pytest resolution is more forgiving of non-pipx installs (venv, distro package, etc.) without losing the CI-relied-upon pipx path, and the fallback logic reads cleanly with good rationale in the comments. The new `preflight.py` for the cf-ux prompt pilot is well-scoped and thoroughly tested (node version detection, nvm shell-function gotcha, codex model entitlement checks, and graceful handling of unreadable/missing caches). Docs updates in the README are clear and match the new behavior.

One minor note worth a glance:

- **Makefile PYTEST/PYTEST_PARALLEL fallback paths are untested in CI** — since every CI job runs `make install` first, only the pipx-present branch of the `PYTEST` fallback and the xdist-present branch of `PYTEST_PARALLEL` ever actually execute; the `command -v pytest` and `$(PYTHON) -m pytest` fallbacks, and the no-xdist serial path, have no coverage. Not blocking, but worth keeping in mind if those paths ever need to change. ([comment](https://github.com/constructorfabric/studio/pull/241#discussion_r4044637189))

ainetx (approved):
Approving — this is a solid, well-reasoned hardening of the pytest resolution path plus a genuinely useful preflight for the prompt pilot (node version and codex model entitlement checks are well-tested and correctly err on the side of silence when uncertain). One minor item worth a glance before it's forgotten:

- **Makefile PYTEST fallback chain has no test/CI coverage** — the new three-way `PYTEST` resolution (pipx → `command -v pytest` → `$(PYTHON) -m pytest`) and the `PYTEST_PARALLEL` xdist probe are only ever exercised on their first/happy-path branches in CI, since every job runs `make install` first. Nothing in the repo asserts on `PYTEST`, `PYTEST_PARALLEL`, or `check-pytest` directly, so a regression in the fallback or non-xdist path wouldn't be caught. ([comment](https://github.com/constructorfabric/studio/pull/241#discussion_r4044637189))

PR #223, opened 2026-09-16
test(pdsl): pin why a gate's declared type is not a lintable judgement

SanjeevSolanki:
Part of #219. **This PR withdraws the rule it originally proposed.** It now contains no production code: `skills/` is byte-identical to `main`, and two tests remain.

## Why this is safe

Stated first, because the change is a removal.

- **Nothing regressed and nothing shipped.** `PDSL704` and `AUTO_PROCEEDING_GATES` were introduced by *this* pull request and never merged. Removing them takes nothing away from any released version, any kit, or any user.
- **No runtime behaviour changes, in either direction.** Nothing resolves a gate from its declared type today — `architecture/specs/PDSL.md` says so, and the three shipped paths that do auto-resolve read no declaration. Registered or unregistered, Studio behaved identically; with the rule gone it still does.
- **Nothing depends on it.** No workflow, module, command or other test reads `AUTO_PROCEEDING_GATES`. Verified by search across `skills/`, `src/` and `tests/`.
- **It cannot loosen a check that existed before.** The rule this removes was stricter than `main`, not part of it. `main`'s behaviour for a `TYPE: confirmation` declaration is unchanged: the vocabulary lint from #153 still validates the token and still rejects near-miss spellings.
- **The full local gate battery passes** on this branch: `validate`, `self-check`, `check-versions`, `validate-kits`, `test-gates`, `test`, `test-coverage`, `pylint`, `vulture-ci`, `spec-coverage` — 6401 tests.

**What is given up, stated rather than implied:** there is now no authoring-time check on a `confirmation` label. A wrong label is caught by review and nothing else. That is a deliberate trade, argued below.

## What was wrong with the rule

The registry keyed on a bare `MENU` name. The validator sees one source at a time, has no project root, and the same file has a different path here and inside an installed kit — so a name cannot be scoped to a file. Registering one reviewed gate therefore authorises **every** gate sharing that name. Reproduced:

```
registry = {"ApprovalGate"}
skills/reviewed.md        MENU ApprovalGate  TYPE: confirmation  -> accepted
skills/somewhere-else.md  MENU ApprovalGate  TYPE: confirmation  -> accepted
                          (its only option is "delete everything")
```

`TerminalStates` is declared twice in this tree today, so name collisions are not hypothetical.

[...]

Issue #219, opened 2026-09-16
Gate risk types are lintable but undeclared: 1 of ~106 menus carries a TYPE, including none of the ten that carry 71% of the surface


SanjeevSolanki:
The PDSL validator understands a `TYPE:` declaration on a `MENU` and lints it (#152, merged). Almost nothing declares one: there is a single `TYPE:` in the tree, in `skills/studio/modules/plan-compile.md`.

That leaves the vocabulary shipped and unused, and it blocks the work that depends on it — a gate's declared type is the ceiling on how it may be resolved, so anything reading that type finds nothing to read.

## Which gates, and why these

A static walk of the 47 workflows counts, for each menu, how many workflows can reach it. Ten menus carry **71%** of the total reachability (338 of 473):

| # | menu | reachable from |
|---|---|---|
| 1 | `NextActionsMenu` | 46 of 47 |
| 2 | `SubAgentFallbackRequest` | 44 |
| 3 | `SubAgentFallbackLimitRequest` | 44 |
| 4 | `SubAgentApprovalRequest` | 44 |
| 5 | `DebuggerMenu` | 39 |
| 6 | `SimpleModeBraveNewWorldChoice` | 39 |
| 7 | `SimpleModeChoice` | 39 |
| 8 | `BlockedNextActionsMenu` | 23 |
| 9 | `ReviewFindingsNavigation` | 10 |
| 10 | `ReviewFixScope` | 10 |

Typing these ten covers most of the surface; the tail can follow behind the lint rather than all at once.

## What this is not

**Not a mechanical pass.** Choosing between `confirmation`, `decision` and `blocking` is a judgement about whether the answer changes the work product, and it is made deliberately by a person at labelling time so that it is reviewable — rather than by a runtime heuristic that is invisible and different every run.

Several of these ten are visibly **not** `confirmation`: `ReviewFindingsNavigation` and `ReviewFixScope` appear by name in the never-auto-answer invariants at `skills/studio/modules/brave-new-world-eligibility.md:33-43`, as do debugger prompts. Those are `blocking`, and the lint should already refuse anything weaker. `NextActionsMenu` is the interesting one: nothing in its own definition says which option is right, so no label makes it quiet — it can only ever be answered from an approved plan.

## Acceptance

- each of the ten carries a declared, reviewed `TYPE:`
- behaviour is provably unchanged — declaring a type resolves nothing on its own
- the declared-stop count from the static walk does not rise

## Note on expected benefit

[...]

PR #246, opened 2026-09-21
feat(pdsl): declare the four gates the never-auto-answer invariants cover

SanjeevSolanki:
Part of #219. Four of the ten most-reachable menus now declare `TYPE: blocking`, taking the typed surface from **1 of ~106 to 5**.

## Why this is safe

- **No behaviour changes, in any mode.** Nothing resolves a gate from its declared type today — `architecture/specs/PDSL.md` says so, and the three shipped paths that do auto-resolve read no declaration. Verified by search: outside `utils/pdsl.py`, which validates the token, nothing in `skills/studio/scripts/` reads a declared type at all.
- **It cannot loosen anything.** `blocking` is the most restrictive of the three types — "never auto-proceeds in any mode". Every one of these four is declared at that level, so the change moves in the safe direction by construction, and there is no direction in which it could relax an existing check.
- **It adds no rule.** No lint, no gate, no CI check. Four declarations and two tests.
- **Full local battery passes**: `validate`, `self-check`, `check-versions`, `validate-kits`, `test-gates`, `test`, `test-coverage`, `pylint`, `vulture-ci`, `spec-coverage` — 6401 tests.

**What this does not achieve, stated rather than implied:** it reduces no questions. All four are labelled *must always ask*. Typing them correctly records that they keep asking. The stop-reduction this story is after has to come from the other six, and four of those are waiting on a ruling (below).

## The four, and why

| menu | reach | why `blocking` |
|---|---|---|
| `ReviewFindingsNavigation` | 10 | named outright in the never-auto-answer invariants (`brave-new-world-eligibility.md:33-43`) |
| `ReviewFixScope` | 10 | named outright in the same list |
| `DebuggerMenu` | 39 | the invariants name *"debugger prompts, breakpoint controls, step/continue approvals, debug-gate prompts"* — this menu is all four |
| `BlockedNextActionsMenu` | 23 | the invariants name *"missing prerequisites"*, and its `override` option exists to bypass the gates that are missing |

## How they are pinned

The lint that would have refused a weaker declaration was withdrawn as unsound (#223), so between an author's label and the behaviour it authorises there is review and these two tests.

- `test_every_menu_the_invariants_name_is_declared_blocking` **derives** the named menus from the invariant list itself, so a menu added to that list later is covered without anyone remembering this test exists.
- `test_the_gates_covered_by_an_invariant_category_are_declared_blocking` carries the other two as literals with the reasoning in its docstring. Those matches are a judgement, and a judgement recorded as a literal is honest where one dressed up as a derivation is not.

Mutation-checked individually: weakening any of the four to `confirmation` fails them.

## What is deliberately not in this PR

The remaining six of the ten. Four turn on one question that is a person's to answer — **whether dispatching a sub-agent is the "external delegation" the invariants forbid.** Those three sub-agent gates are 132 of the ten's 338 reachability, so that single answer decides most of what typing buys; if it goes the restrictive way, typing the ten quiets almost nothing. `NextActionsMenu` and `SimpleModeChoice` have proposed readings in #219 and are held with them.

Labelling any of the six provisionally would be exactly the trusted-rather-than-reviewed failure the withdrawn lint was meant to catch, and there is now no lint.

## A finding worth flagging against the sizing estimate

Reading all ten against the invariants gives, by reachability: **35% `blocking`, 13% `decision`, 50% resting on the open ruling.** The sizing for this work credits typed gates with 22% of the expected reduction. That looks like the optimistic end rather than the expected one, and it is cheaper to say now than to discover it when the stop-count metric is measured against it. Detail in #219.

## Summary by CodeRabbit

- **Updates**
  - Debugger, review, and blocked-action menus are now explicitly treated as blocking menus, requiring a deliberate selection before proceeding.
  - Sub-agent fallback menus are now explicitly treated as confirmation menus.

- **Reliability**
  - Added safeguards to ensure blocking menus retain their intended behavior and confirmation menus do not offer session-wide options.
  - Added automated coverage to verify menu types remain correctly declared across supported scenarios.


tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,94 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+        for path in sorted((REPO_ROOT / folder).rglob("*.md")):
+            menu = None
+            for line in path.read_text(encoding="utf-8-sig", errors="replace").splitlines():
+                header = re.match(r"^\s*MENU\s+([A-Za-z][\w-]*)", line)
+                if header:
+                    menu = header.group(1)
+                    continue
+                declared = re.match(r"^\s*TYPE:\s*(\S+)\s*$", line)
+                if declared and menu:
+                    types[menu] = declared.group(1)
```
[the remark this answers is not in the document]
SanjeevSolanki: Correct, and reproduced before fixing:

```
MENU Alpha
TITLE: first
OPTIONS:
  1 go -> CONTINUE X

UNIT Somewhere
DO:
  EMIT "x"
TYPE: confirmation        <- belongs to no menu

_declared_types() -> {'Alpha': 'confirmation'}
```

The scan kept a menu current to the end of the file, so a `TYPE:` the validator would itself reject as `PDSL702` was read as the preceding menu's declaration. That matters more here than in an ordinary helper: this scan backs the only enforcement these labels have, since the lint that would have refused a weaker declaration was withdrawn as unsound (#223). A misattributed type means the guard is checking the wrong menu.

The region now closes the way the validator closes it — at the first sub-header that is not `TITLE`/`TYPE`/`SHAPE`, at a `UNIT`, and at a fence boundary.

**Pinning it took three cases, and the reason is worth recording.** The corpus has no stray `TYPE:`, so removing the termination entirely leaves every corpus answer identical — only a crafted source can tell the two apart, which meant lifting the scan out of the tree walk so it takes text. Then mutating each half in turn showed the first fixture was not enough either: it contained both a section header and a `UNIT`, so whichever rule fired first made the other deletable with everything still green. There is now one case per rule — after `OPTIONS:`, after a `UNIT`, after a fence close — and dropping either half fails a test that only that half covers.

ainetx (changes requested):
Requesting changes — this closes off part of the enforcement the diff itself is trying to establish, so the gate ends up weaker than the tests claim it is.

- **Invariant-to-menu guard silently drops stale/renamed menu references instead of failing closed** [Major] — In `test_every_menu_the_invariants_name_is_declared_blocking`, `named` only keeps CamelCase tokens from the `- NEVER auto-answer …` invariant lines that already resolve to a declared or existing menu name. If a menu referenced by an invariant gets renamed, removed, or the invariant text drifts with a typo, that token is quietly filtered out before `wrong` is built — and since the test only asserts `named` is non-empty overall, one stale reference among several valid ones goes completely unnoticed. That's exactly the failure mode this test set out to catch, so it needs to fail closed (e.g., flag or fail on invariant-named tokens that don't resolve to any live menu) rather than silently drop them.

tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,163 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+
+def _declared_types() -> dict:
+    """``menu name -> declared TYPE`` across the authored tree.
+
+    Read from the source rather than through the validator, because the question here is
+    what an author wrote, not whether it parses -- the validator has its own tests for that.
+    """
+    types: dict = {}
+    for folder in ("workflows", "skills"):
+        for path in sorted((REPO_ROOT / folder).rglob("*.md")):
```
ainetx, line 2959: ### Renamed/removed menu silently drops out of the never-auto-answer blocking guard

**Severity:** Major

**Problem**
In test_every_menu_the_invariants_name_is_declared_blocking, `named` is computed as the intersection of CamelCase tokens found in NEVER invariant text with menus that currently exist (`declared` or `_menu_names()`). If a menu named by the invariants (e.g. ReviewFindingsNavigation) is later renamed or removed while the invariant prose still refers to the old name, that token simply vanishes from `named` instead of failing.

<details>

**How to reproduce**
1. Rename `ReviewFindingsNavigation` to some new menu name in fix-approval.md without updating brave-new-world-eligibility.md's NEVER line that still says 'ReviewFindingsNavigation'. 2. Run the test suite. 3. `named` no longer contains the old token (it's absent from declared/_menu_names()), so `wrong` is never populated for it, and the test passes even though the invariant text now references a non-existent menu and the new menu's TYPE was never checked.

**Expected behavior**
An invariant-named menu token that can't be resolved to a currently-declared/existing menu should cause the test to fail (or otherwise surface as an unchecked reference), not be silently dropped.

**Actual behavior**
The token is filtered out by the `if word in declared or word in _menu_names()` clause before the blocking check runs, so a stale/renamed reference is invisible to the guard.

```text
invariant text names OldMenu -> OldMenu renamed/removed -> token not in declared/_menu_names() -> filtered out of `named` -> `wrong` check never runs for it -> test passes despite lost enforcement
```

**Impact**
The only mechanical enforcement that a security-relevant, never-auto-answer menu stays declared `blocking` can be silently defeated by an ordinary rename, with no test failure to flag it.

**Suggested correction**
Instead of filtering candidate tokens down to only those matching existing menus, extract all CamelCase-shaped candidates from NEVER lines and separately assert that every one which looks like a menu reference resolves to a currently existing declared menu (fail if not), then check it is 'blocking'.

**How to verify**
Rename an invariant-named menu without updating the invariant text and confirm the test suite fails instead of passing silently.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,163 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+    in the tree (GH #223) — but where one names a menu outright, the label is mechanical and
+    is pinned here.
+
+    This is the only enforcement there is. The lint that would have refused a weaker
+    declaration was withdrawn as unsound, so between an author's label and the behaviour it
+    authorises there is review and this test. Part of GH #219.
+    """
+    invariants = (REPO_ROOT / "skills/studio/modules/brave-new-world-eligibility.md").read_text(
+        encoding="utf-8")
+    never = [line.strip()[2:] for line in invariants.splitlines()
```
ainetx: ### Invariant-name extraction reads only the first physical line of a NEVER bullet, missing PDSL-legal continuation lines

**Severity:** Minor

**Problem**
`never = [line.strip()[2:] for line in invariants.splitlines() if line.strip().startswith("- NEVER")]` only captures lines that themselves start with '- NEVER'. PDSL.md documents continuation lines as a legal construct under any list item, including INVARIANTS bullets.

<details>

**How to reproduce**
1. Wrap a long NEVER bullet in brave-new-world-eligibility.md onto a continuation line, moving a named menu (e.g. 'ReviewFixScope') onto that second physical line. 2. Run the test suite. 3. The continuation line is never captured by `never`, so the menu name never enters `named`, and its TYPE is never checked even if it were downgraded from `blocking`.

**Expected behavior**
The extraction should gather the full logical text of each NEVER item, including any continuation lines, before scanning for named menus.

**Actual behavior**
Only the first physical line of each bullet is scanned; anything wrapped to a following line is dropped from the enforced text with no failure.

```text
NEVER bullet wraps onto continuation line -> extraction reads only first physical line -> named menu on continuation line never captured -> never checked for TYPE: blocking
```

**Impact**
A future, currently-unused-but-spec-legal formatting edit (wrapping a long NEVER bullet) can silently remove a named menu from the only mechanical enforcement of the never-auto-answer blocking requirement.

**Suggested correction**
Join each NEVER item's continuation lines (lines indented at or beyond the bullet's own indent that don't start a new top-level bullet) into the bullet's text before scanning for menu-name tokens.

**How to verify**
Wrap a NEVER bullet naming a live menu onto a continuation line and confirm the guard still catches a mislabeled TYPE on that menu.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,163 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+
+
+def _declared_types() -> dict:
+    """``menu name -> declared TYPE`` across the authored tree.
+
+    Read from the source rather than through the validator, because the question here is
+    what an author wrote, not whether it parses -- the validator has its own tests for that.
+    """
+    types: dict = {}
+    for folder in ("workflows", "skills"):
```
ainetx: ### New _declared_types()/_menu_names() helpers scan only 2 of the 4 authoritative prompt roots

**Severity:** Major

**Problem**
PROMPT_ROOTS (tests/test_pdsl_keywords.py:20-25) authoritatively defines four prompt roots: skills, workflows, requirements, architecture -- and requirements/ and architecture/ both contain real, authored MENU blocks. The new `_declared_types()` and `_menu_names()` helpers hardcode `for folder in ("workflows", "skills")`, omitting requirements/ and architecture/ entirely, unlike the file's existing `_prompt_files()`-based scanner used for the same purpose elsewhere in this same file.

<details>

**How to reproduce**
1. Add a menu under requirements/ or architecture/ (e.g. a new gate in requirements/*.md). 2. Name that menu outright in brave-new-world-eligibility.md's NEVER list. 3. Declare that menu with TYPE anything other than blocking (or leave it undeclared). 4. Run test_every_menu_the_invariants_name_is_declared_blocking -- it will not detect the mismatch because `_declared_types()`/`_menu_names()` never scanned that root, so the menu name never enters the `named` set being checked.

**Expected behavior**
The helper roots should match the file's authoritative PROMPT_ROOTS (or otherwise explicitly and correctly justify a narrower corpus), so a menu declared under any supported prompt root can reach the invariant/type comparison.

**Actual behavior**
The new helpers hardcode only workflows/ and skills/, silently excluding requirements/ and architecture/, both of which are real prompt roots containing authored MENU blocks.

```text
PROMPT_ROOTS (4 roots) --> _prompt_files() [correct, all 4] --> _menu_declaration_scan (existing, synced)
\
--> _declared_types()/_menu_names() [new, only 2 of 4] --> misses menus in requirements/, architecture/
```

**Impact**
A future NEVER-invariant naming a menu declared under requirements/ or architecture/ with the wrong (or no) TYPE would pass this guard silently, defeating 'the only enforcement there is' for the never-auto-answer blocking invariant on that menu.

**Suggested correction**
Change the folder loops in `_declared_types()` and `_menu_names()` to iterate the same set of roots as `PROMPT_ROOTS` (or reuse `_prompt_files()`/`PROMPT_ROOTS` directly) instead of the hardcoded ("workflows", "skills") tuple.

**How to verify**
Add a MENU under requirements/ or architecture/, declare it non-blocking, name it outright in the NEVER list, and confirm test_every_menu_the_invariants_name_is_declared_blocking now fails as expected once the roots are corrected.

</details>

tests/test_pdsl_keywords.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): In test_every_menu_the_invariants_name_is_declared_blocking, `named` is computed as the set of CamelCase-looking tokens extracted from the never-auto-answer invariant lines, filtered down to only those tokens that already appear as a key in `declared` (menus with a declared TYPE) or as a value in `_menu_names()` (all menu names currently in the tree). Any invariant token that does NOT match a currently-existing menu name — because the menu was renamed, deleted, or the invariant text has a typo — is silently excluded from `named` rather than being reported. The subsequent `wrong = {name: declared.get(name, '(undeclared)') for name in named if declared.get(name) != 'blocking'}` check can only ever inspect tokens that already resolved to a live menu, so it structurally cannot catch this class of drift.

Problem (now): In test_every_menu_the_invariants_name_is_declared_blocking, `named` is computed as the intersection of CamelCase tokens found in NEVER invariant text with menus that currently exist (`declared` or `_menu_names()`). If a menu named by the invariants (e.g. ReviewFindingsNavigation) is later renamed or removed while the invariant prose still refers to the old name, that token simply vanishes from `named` instead of failing.

</details>

tests/test_pdsl_keywords.py
ainetx: Re-verified against the current code -- this write-up has been updated.

<details>
<summary>Why</summary>

Problem (was): PDSL.md documents that continuation lines/nested bullets are valid under a list item, but `never = [line.strip()[2:] for line in invariants.splitlines() if line.strip().startswith('- NEVER')]` only captures the first physical line of each NEVER entry.

Problem (now): `never = [line.strip()[2:] for line in invariants.splitlines() if line.strip().startswith("- NEVER")]` only captures lines that themselves start with '- NEVER'. PDSL.md documents continuation lines as a legal construct under any list item, including INVARIANTS bullets.

</details>

tests/test_pdsl_keywords.py
ainetx: Re-verified against the current code -- this write-up has been updated. Severity: Minor -> Major.

<details>
<summary>Why</summary>

Problem (was): _declared_types() and _menu_names() (test_pdsl_keywords.py:2897, 2950) only walk ("workflows", "skills"), while this same test file defines PROMPT_ROOTS as (skills, workflows, requirements, architecture) and confirmed MENU blocks exist in requirements/ and architecture/ files.

Problem (now): PROMPT_ROOTS (tests/test_pdsl_keywords.py:20-25) authoritatively defines four prompt roots: skills, workflows, requirements, architecture -- and requirements/ and architecture/ both contain real, authored MENU blocks. The new `_declared_types()` and `_menu_names()` helpers hardcode `for folder in ("workflows", "skills")`, omitting requirements/ and architecture/ entirely, unlike the file's existing `_prompt_files()`-based scanner used for the same purpose elsewhere in this same file.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,236 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+        # `TITLE`/`TYPE`/`SHAPE`, and at anything that opens a new block or ends the fence
+        # -- the same rule the validator applies.
+        if stripped.startswith("UNIT ") or stripped.startswith("```"):
+            menu = None
+            continue
+        section = re.match(r"^\s*([A-Z][A-Z_]*):\s*$", line)
+        if section and section.group(1) not in {"TITLE", "TYPE", "SHAPE"}:
+            menu = None
+            continue
+        declared = re.match(r"^\s*TYPE:\s*(\S+)\s*$", line)
```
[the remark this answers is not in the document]
SanjeevSolanki: Correct, and reproduced:

```
MENU Alpha
TITLE: a
    TYPE: blocking      <- indented deeper than the menu's sub-headers
OPTIONS: …

_types_in(...) -> {'Alpha': 'blocking'}
```

The validator reads that as continuation text and sees no declaration, so this scan could report a type nothing else agrees exists — and the guard built on it would pass while the gate was, to the validator, undeclared. That is the worst shape for this particular helper, since it is the only enforcement these labels have.

The indent is learned per menu from its first sub-header now, which keeps the indented style `DebuggerMenu` uses working (`  TITLE:` / `  TYPE:` — same indent, accepted) while rejecting a nested one.

Two fixtures, and both are needed: the nested case, and an all-indented menu that must still read. No corpus file is written either way, so measuring the real tree cannot tell the rule from its absence — dropping it left all 65 tests green until the crafted cases existed.

tests/test_pdsl_keywords.py
SanjeevSolanki: Correct, and fixed — a rename now fails loudly instead of shrinking the set being checked.

```
AssertionError: the never-auto-answer invariants name ['ReviewFixScope'], which match
no MENU in the tree. Either the menu was renamed or removed and the invariant needs
updating, or the name belongs in _NOT_MENU_NAMES because it was never a menu.
```

The previous version kept only candidates it could resolve, so a renamed menu simply left the check with one fewer name and nothing to say. Verified by renaming `ReviewFixScope` in the tree, which is the message above.

Non-menu tokens are now listed explicitly in `_NOT_MENU_NAMES` (`ID`, from "finding-ID capture", plus the PDSL keywords) rather than filtered by whether they happen to resolve — so the filter cannot absorb a real rename.

tests/test_pdsl_keywords.py
SanjeevSolanki: Correct, and fixed by reading each bullet whole rather than line by line — a wrapped invariant lost everything after its first physical line, including any menu named there.

None wrap today, which is why this had to be fixed by reading rather than confirmed by measuring: the corpus cannot tell the two versions apart. Continuations are folded in when the line is indented and is neither a new bullet nor a fence; 13 bullets are read.

tests/test_pdsl_keywords.py
SanjeevSolanki: Correct, and widening the scan turned up more than the eight missing declarations.

All four authored roots are scanned now. Worth recording what that revealed, because it changes a number rather than just a code path: the two extra `TYPE:` declarations it can now see are in `architecture/specs/PDSL.md` — **specification examples**, illustrations of the language rather than gates any workflow reaches.

```
shipped surface (workflows+skills) : 106 menus,  7 typed
specification examples            :   8 menus,  2 typed
```

So a naive total reads 9 typed and would overstate the reachable surface, which is the figure #219 and the story's sizing both use. The extra roots are scanned for **existence**, so the rename check above is complete, and never counted as progress — the comment on `AUTHORED_ROOTS` says so, since the distinction is invisible from the code alone.

tests/test_pdsl_keywords.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 0e35d910436583fbd40f9d8ae95fd2e50e137f5b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Current code at tests/test_pdsl_keywords.py:3082-3094 computes `missing = sorted(candidates - known)` and asserts it is empty before filtering, so an invariant token naming a menu that no longer resolves to any live menu in `_menu_names()` now fails loudly with an explicit message ('name is renamed/removed... or belongs in _NOT_MENU_NAMES') rather than being silently dropped from `named`; non-menu tokens are explicitly enumerated in `_NOT_MENU_NAMES` instead of being filtered by resolvability, closing the fail-open gap identified in prior rounds.
- **Verifier:** The current code computes candidates - known via `missing = sorted(candidates - known)` and asserts it's empty BEFORE building `named`, so any invariant-named token that no longer resolves to a live menu in `_menu_names()` (renamed, removed, or typo'd) now fails the test loudly with an explicit diagnostic instead of being silently filtered out; `named` is then simply `sorted(candidates)`, which by that point is guaranteed to be a subset of `known`, so the previous fail-open filtering gap (fixing candidates to only what already resolved) is closed. Verified by reading tests/test_pdsl_keywords.py's current test_every_menu_the_invariants_name_is_declared_blocking body in the diff.

</details>

tests/test_pdsl_keywords.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 0e35d910436583fbd40f9d8ae95fd2e50e137f5b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Current code at tests/test_pdsl_keywords.py:3030-3056 defines `_never_bullets()`, which accumulates indented continuation lines onto the current bullet (`current += " " + stripped`) rather than reading only the first physical line of each `- NEVER` entry, so a menu name wrapped onto a continuation line would still be captured; this directly replaces the single-line extraction flagged in prior rounds.
- **Verifier:** `_never_bullets()` in the diff accumulates indented continuation lines onto the current bullet (`current += " " + stripped`) rather than only reading the first physical line starting with '- NEVER', explicitly designed to fold continuation text into the bullet (per its own docstring citing this exact prior gap and 'Raised in review'); this directly replaces the single-line extraction flagged in earlier rounds, so a menu name on a legal PDSL continuation line would now be captured.

</details>

tests/test_pdsl_keywords.py
ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 0e35d910436583fbd40f9d8ae95fd2e50e137f5b -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Current code at tests/test_pdsl_keywords.py:2948,2958,3107 defines `AUTHORED_ROOTS = ("workflows", "skills", "requirements", "architecture")` and both `_declared_types()` and `_menu_names()` iterate over `AUTHORED_ROOTS`, matching this file's own authoritative `PROMPT_ROOTS` (lines 20-25) exactly; the two-root omission flagged in prior rounds is no longer present in this code.
- **Verifier:** `AUTHORED_ROOTS = ("workflows", "skills", "requirements", "architecture")` now matches this file's authoritative PROMPT_ROOTS exactly, and both `_declared_types()` and `_menu_names()` (the two helpers named in the check target) iterate over `AUTHORED_ROOTS`, not a hardcoded two-root tuple; the accompanying comment on AUTHORED_ROOTS even documents the requirements/architecture menu counts found once the scan was widened, confirming the fix is real and intentional, not accidental.

</details>

tests/test_pdsl_keywords.py
```diff
@@ -2885,3 +2885,324 @@ def test_holding_a_phase_back_is_announced_not_silent() -> None:
...
+        assert declared.get(menu) == "blocking", (
+            f"{menu} is declared {declared.get(menu, '(nothing)')}, but the "
+            "never-auto-answer invariants cover it by subject"
+        )
+
+
+def _menu_bodies() -> dict:
+    """``menu name -> the option lines of its body`` across the authored tree."""
+    bodies: dict = {}
+    for folder in ("workflows", "skills"):
```
ainetx: ### _menu_bodies() scans a narrower root set than _declared_types(), letting a future confirmation-typed gate outside workflows/skills evade the session-wide-option guard

**Severity:** Minor

**Problem**
test_no_auto_proceeding_gate_offers_a_session_wide_option() derives `declared` from _declared_types(), which walks AUTHORED_ROOTS = (workflows, skills, requirements, architecture), but derives `bodies` from _menu_bodies(), which only walks (workflows, skills) -- contradicting its own docstring 'across the authored tree'. Any menu declared TYPE: confirmation under requirements/ or architecture/ would have bodies.get(menu, []) silently return [], so the session-scoped-option scan finds nothing for it no matter what its OPTIONS body actually contains.

<details>

**How to reproduce**
1. Add a MENU under architecture/ or requirements/ with `TYPE: confirmation` and an OPTIONS line containing `SET SOME_VAR = approve-session` (a session-scoped preference). 2. Run test_no_auto_proceeding_gate_offers_a_session_wide_option(). 3. declared[menu] == 'confirmation' but bodies.get(menu, []) == [] since _menu_bodies() never visited that folder, so `session` is empty and the menu never lands in `offenders`. 4. The test passes green despite the exact violation it exists to catch.

**Expected behavior**
_menu_bodies() should scan the same root set as _declared_types()/_menu_names() (AUTHORED_ROOTS), or the test should otherwise ensure every menu key it inspects for `kind == confirmation` has its body actually collected, so the guard cannot pass merely because a menu's body was never gathered.

**Actual behavior**
_menu_bodies() is hardcoded to ("workflows", "skills") while the menus it is queried against via `declared` can come from any of the four AUTHORED_ROOTS, creating a silent false-negative for the two roots it omits.

```text
_declared_types() [4 roots] --> declared{menu:type}
_menu_bodies() [2 roots] --> bodies{menu:[lines]}
test iterates declared.items() --> bodies.get(menu, []) --> [] for requirements/architecture menus --> session=[] --> never in offenders --> assert passes even with a real violation
```

**Impact**
The sole enforcement mechanism for 'no confirmation-declared gate may carry a session-scoped option' has a blind spot for any such gate authored under requirements/ or architecture/; currently unexploited (verified no such menu exists there today), but the guard would stay green if one were added, defeating its stated purpose (GH #219).

**Suggested correction**
Change `for folder in ("workflows", "skills"):` in _menu_bodies() to `for folder in AUTHORED_ROOTS:` (or otherwise keep the two root sets in sync), so any confirmation-typed menu anywhere in the authored tree has its body collected and checked.

**How to verify**
Add a crafted menu with TYPE: confirmation and a session-scoped SET under requirements/ or architecture/ in a test fixture (or temporarily point AUTHORED_ROOTS's extra roots at a temp dir) and confirm the test now fails before the fix and passes/correctly flags after aligning the root sets.

</details>

SanjeevSolanki: Correct — and worse than a future risk, the guard was already inapplicable to part of what it checks.

`_declared_types()` walked four authored roots and `_menu_bodies()` walked two, so a menu declared `confirmation` under `requirements/` or `architecture/` had `bodies.get(menu, [])` return empty, and the session-wide-option guard found nothing for it whatever its options actually said. It passed by having no subject, which reads as coverage.

`_menu_bodies()` now walks the same `AUTHORED_ROOTS`, and the two scans are pinned together rather than left to drift again: `test_the_two_corpus_scans_cover_the_same_tree` asserts every menu with a declared type has a body for the guards to read. It is written as a relationship between the two scans rather than a fixed number, so it stays true as the corpus grows.

**Verified by mutation:** narrowing `_menu_bodies()` back to two roots fails the new test and names the menus that lost their body.

**Safe:** widening the scan can only give an existing guard more to check; no guard was relaxed, and no declared type changed. Head is now `1825089f`.

ainetx: @SanjeevSolanki **RESOLVED** -- Confirmed fixed as of 1825089fe792136edeac99365659fd40c76fbc82 -- independent re-verification no longer reproduces this issue.

<details>
<summary>Why</summary>

- **Reviewer:** Verified directly in the current file: _menu_bodies() (line 3146) now iterates the same AUTHORED_ROOTS as _declared_types() (line 2958), and a new test test_the_two_corpus_scans_cover_the_same_tree() (line 3168) pins the two scans together, asserting every menu with a declared type has a body -- so the gap described (bodies scanning only workflows/skills while declared scans all four roots) has already been fixed in this diff, matching the posted discussion thread's description of the fix at head 1825089f.
- **Verifier:** Verified directly in current code: _menu_bodies() (line 3146) now iterates AUTHORED_ROOTS (workflows, skills, requirements, architecture) -- the same four roots as _declared_types() (line 2958) and _menu_names() -- rather than the previously-hardcoded (workflows, skills). A new test, test_the_two_corpus_scans_cover_the_same_tree() (line 3168), asserts every menu key in `declared` also has an entry in `bodies`, which would fail if the two scans' root sets ever diverge again. The finding accurately describes a bug that existed in an earlier state of this PR (per the prior reviewer opinions and the discussion thread confirming head 1825089f), but the diff/code as currently presented already contains the fix plus a pinning regression test, so there is no live defect to report.

</details>

SanjeevSolanki:
**Re-review requested — head is `1825089f`.** The changes-requested review of 21 Sep was against `60cd5ae4`; **5 findings have been fixed since then**.

**The condition from that review is met.** You flagged that the invariant-to-menu guard filtered out CamelCase tokens it could not resolve, so a renamed or mistyped menu reference vanished silently — the exact failure the test existed to catch. It now **fails closed**: unresolved names are collected and asserted on by name, with a message saying whether the menu moved or the token belongs in the not-a-menu list. The non-empty check stayed, because "some names resolve" and "no name is stale" are different claims.

Since then, one more from you and one I found: the two corpus scans walked different root sets, so a `confirmation` gate declared outside `workflows/` or `skills/` had an empty body as far as the session-wide-option guard was concerned — it passed by having no subject. Both scans now walk the same roots, and a test asserts every menu with a declared type has a body for the guards to read.

**Why it is safe:** test-side only, and each change widens what a guard examines rather than relaxing it. No declared type changed, and the four typed gates are unchanged from the version you reviewed.

23/23 green, no conflicts.

ainetx (approved):
Reviewed the new `TYPE:` declarations on the four menus (`DebuggerMenu`, `ReviewFindingsNavigation`, `ReviewFixScope`, `BlockedNextActionsMenu` as `blocking`; `SubAgentFallbackRequest`/`SubAgentFallbackLimitRequest` as `confirmation`) alongside the supporting test changes. The tightened `DECLARED_TYPE_READ_RE` correctly avoids matching a menu's own `TYPE:` line as evidence of a declaration read elsewhere, and the new tests around `_types_in`, invariant-name resolution, session-wide-option guarding, and cross-scan tree coverage are well-targeted and exercise real edge cases (region termination, indentation, continuation lines) with crafted fixtures rather than relying on incidental corpus shape. The baseline removals line up exactly with the newly typed menus. No correctness, security, or coverage concerns found — nothing blocking here.

ainetx (approved):
Reviewed the new `TYPE:` declarations on the debugger, review, blocked-next-actions, and sub-agent fallback menus, along with the corresponding test suite additions. The test coverage is thorough — it exercises menu-region parsing edge cases (stray types, nested indentation, multi-header files), cross-checks the two corpus scans walk the same authored roots, and pins the `blocking`/`confirmation` labels to the invariants that require them, including the session-wide-option guard for `confirmation` gates. Logic is sound and well-justified throughout, with no correctness, security, or test-gap concerns. Nothing blocking here — approved.

